PR #29, opened 2021-03-15
MPDX-6947 : Design new contact row

gyasistory:
[MPDX-6947](https://jira.cru.org/browse/MPDX-6947)


OzzieOrca (dismissed):
Thanks for making all these changes! I've got a bunch more... Looks like you're getting more into the business logic of stuff which is cool.

Maybe try and not introduce too much else. This PR is already big and we can work on specific pieces of it or delegate some work after it's merged.

src/components/Contacts/ContactRow.tsx
```diff
@@ -37,55 +36,35 @@ interface Props {
...
+  const contactHasBirthday = (): boolean =>
+    contact.people.nodes.some((person) =>
+      Interval.after(DateTime.fromJSDate(new Date()).startOf('day'), {
+        days: 5,
+      }).contains(
+        DateTime.fromISO(
+          `${new Date().getFullYear}-${person.birthdayMonth}-${
+            person.birthdayDay
+          }`,
+        ),
```
OzzieOrca, line 49: Luxon has `DateTime.fromObject(...)` for this. Took me a bit to find. Also I missed that month and day were separate fields (I still had an ISO string I was playing with in an example) and than the year needs to be ignored in the comparison. Thanks for fixing my last suggestion to account for those :)
```suggestion
      DateTime.fromObject({ month: person.birthdayMonth, day: person.birthdayDay }),
```

src/components/Contacts/ContactRow.tsx
```diff
@@ -37,55 +36,35 @@ interface Props {
...
+  const contactHasAnniversary = (): boolean =>
+    contact.people.nodes.some((person) =>
+      Interval.after(DateTime.fromJSDate(new Date()).startOf('day'), {
+        days: 5,
+      }).contains(
+        DateTime.fromISO(
+          `${new Date().getFullYear}-${person.anniversaryMonth}-${
+            person.anniversaryDay
+          }`,
+        ),
```
OzzieOrca, line 62:
```suggestion
        DateTime.fromObject({ month: person.anniversaryMonth, day: person.anniversaryDay }),
```

src/components/Contacts/GiftStatus/GiftStatus.tsx
```diff
@@ -12,6 +12,20 @@ export enum GiftStatusEnum {
   Late,
   Hidden,
 }
+export const GiftIsLateStatus = (
```
OzzieOrca, line 15: What do you think of calling this in the `GiftStatus` component instead of exporting it and making the parent call it? You'd have to change the GiftStatusProps to take the date instead of the status enum... I think it'd be cleaner and keep the implementation isolated in this component.

OzzieOrca (changes requested):
Just a couple more date things and one more idea with GiftStatus

OzzieOrca (approved):

PR #26, opened 2021-03-10
Add GraphQL automock tooling

OzzieOrca:
- Add graphql-ergonomock
- Add helper function to make graphql-ergonomock support fragments and generated types
- Add ContactRow test
- Add contacts page test

PR #34, opened 2021-03-18
MPDX-6929: Contacts Screen Layout

reldredge71:
Layout the ContactsScreen with Filters, Header, and Table.  Each of these will be fleshed out later.


src/components/Contacts/ContactsTable/ContactsTable.tsx
```diff
@@ -0,0 +1,59 @@
...
+  );
+
+  const renderError = () => (
+    <Box data-testID="ErrorText" bgcolor={colors.red[600]}>
+      Error: {error.toString()}
+    </Box>
+  );
+
+  return (
+    <Box height="100%">
```
reldredge71, line 42: I'm having trouble getting the table to stretch and fill the rest of the window space.  `height="100%"` is ultimately not what I want -- it will make it the same height as the window, but since there is a header above this component, the bottom of this component extends beyond the view.  I've been trying to find the right combination of CSS and MaterialUI props to get the right effect, but no luck so far.  My React Native experience with flex boxes is not working here!

What I want:
- Scrolling disabled for the overall screen
- ContactsHeader should have fixed height
- ContactsTable should fill remaining available space
- ContactsTable should allow for scrolling inside itself, if the list of contacts is long

pages/accountLists/[accountListId]/contacts.page.tsx
```diff
@@ -30,39 +34,15 @@ const ContactsPage: React.FC = () => {
...
-          ) : !data?.contacts?.nodes ? (
-            <p>No data</p>
-          ) : (
-            data.contacts.nodes.map((contact) => (
-              <ContactRow key={contact.id} contact={contact} />
-            ))
-          )}
-        </div>
-      </div>
+      <Contacts
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/contacts.page.tsx:37]: Some thoughts about how to structure components here:
1. I think the `ContactsPage` component should be the one to split the page into 2 columns.
2. We can make `ContactsPage` store the selected filters state (future work) and pass those between the left filters pane and the right contacts list.
3. We should make the left filters component and the right contact list contain the GraphQL queries. I don't think there's any reason to have them at the top level and pass everything down.

src/components/Contacts/ContactFilters/ContactFilters.tsx
```diff
@@ -0,0 +1,44 @@
...
+  loading,
+  error,
+  loadFilters,
+}: Props) => {
+  return (
+    <div style={{ backgroundColor: colors.amber[600] }}>
+      <h2>Filters</h2>
+      <button data-testID="LoadFiltersButton" onClick={() => loadFilters()}>
+        Load Filters
+      </button>
```
OzzieOrca, line 27: Just adding context, I think the filters panel is collapsible and will be collapsed by default. So I was thinking we'd make it lazy load once it's open. I just added a button for now.

src/components/Contacts/ContactFilters/ContactFilters.tsx
```diff
@@ -0,0 +1,44 @@
...
+      {error && <p data-testID="ErrorText">Error: {error.toString()}</p>}
```
OzzieOrca, line 28: I think you can use [`getByText`](https://testing-library.com/docs/queries/bytext) here.
You should be able to use an `exact: false` match or a regex to match the beginning of the string https://testing-library.com/docs/queries/about#textmatch

src/components/Contacts/Contacts.tsx
```diff
@@ -0,0 +1,54 @@
...
+          loadFilters={loadContactFilters}
+        />
+      </Box>
+      <Box flex={4} flexDirection={'column'}>
+        <ContactsHeader />
+        <ContactsTable
+          contacts={contactsData?.contacts.nodes}
+          loading={contactsLoading}
+          error={contactsError}
+        />{' '}
```
OzzieOrca, line 48: I assume Prettier added this
```suggestion
        />
```

src/components/Contacts/Contacts.tsx
```diff
@@ -0,0 +1,54 @@
...
+      </Box>
+    </Box>
+  );
+};
+
+export default Contacts;
```
OzzieOrca, line 54: Can we use named exports?
```suggestion
```
The exception is pages where Next.js is looking for the default export.

src/components/Contacts/ContactsHeader/ContactsHeader.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+  title: 'ContactHeaders',
+};
+
+export const Default = (): ReactElement => {
+  return <ContactHeaders />;
+};
+
+Default.story = {
+  parameters: {
+    chromatic: { delay: 1000 },
```
OzzieOrca, line 14: We should eventually use the graphql mock provider here. I don't think it should do a real query. Not quite sure how to get it past the loading state. We'll have to do some testing.

We should spend some time making the testing wrappers consistent. They are wildly different between tests.

src/components/Contacts/ContactsTable/ContactsTable.tsx
OzzieOrca: I think we should make things sticky instead of nested scroll bars and trying to size things correctly to the screen. That way we just use the main body scroll which is less buggy.

Idk what Material has but something involving `position: sticky;` CSS or absolute positioning.

The AppBar supports it https://github.com/CruGlobal/mpdx-react/commit/c66ae9f7ef2abf9a5922943fd3958aa89689f086

OzzieOrca (changes requested):
Thanks for setting this all up! I have a few thoughts/things I saw.

I can probably go back a take a better look at the tests. We might be able to use some of the automock stuff.

reldredge71:
@OzzieOrca I think a sticky header makes sense.  I'm still thinking through how to properly set up the contents of the table.  When there's data for the contacts we don't have to worry so much -- we just have the rows appear and we can scroll the whole length of the table. For the Loading and Empty states, though, I want to have some kind of text and/or loading indicator centered in the space between the header and the bottom of the window.  Somehow I have to do that calculation.

src/components/Contacts/ContactsHeader/ContactsHeader.stories.tsx
reldredge71: Yeah I would like to use the graphql mock provider, I'm not sure how to best use it yet.  It sounds like you were having trouble making tests that required loading data also.  I agree that this is something we should work on.

OzzieOrca: It works but you can't customize the mock data. https://github.com/CruGlobal/mpdx-react/pull/26/files#diff-3862096e13eaf47c09b8236fea6ae30abac092cae98e4d99c977fc11fea47c68R28-R41

I'll try and figure that out soon.

OzzieOrca (commented):
I don't think that necessarily needs needs to be full height. We could try and set up a full height layout but probably just putting tons of padding around whatever error/loading/empty icon/message would work. Or `50vh`-`70vh` could work but might be weird on mobile as the height shrinks a lot.

Looks like you have some tests to update. This looks cleaner though :)

pages/accountLists/[accountListId]/contacts.page.tsx
```diff
@@ -2,67 +2,29 @@ import React from 'react';
...
-import { useContactFiltersLazyQuery } from './ContactFilters.generated';
+import { Box } from '@material-ui/core';
+import { ContactFilters } from '../../../src/components/Contacts/ContactFilters/ContactFilters';
+import { ContactsTable } from '../../../src/components/Contacts/ContactsTable/ContactsTable';
 
 const ContactsPage: React.FC = () => {
   const { t } = useTranslation();
   const {
     query: { accountListId },
   } = useRouter();
```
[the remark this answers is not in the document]
Comment (TheNoodleMoose) [reply pages/accountLists/[accountListId]/contacts.page.tsx:13]: I think a hook would be nice, would also help with that breadcrumb logic we wanted to remove as well.

reldredge71:
I can't get the header to be sticky.  It's behaving strangely.  Using `AppBar` puts it at the very top of the document, right where the existing app bar already is.  Using any variation of the "sticky" property doesn't seem to do anything.  I'm going to move on from that and work on fixing the tests.

OzzieOrca:
It's one of the harder CSS things to do. The parent element where you want it to be sticky too probably needs `position: relative`. I can play with it sometime after you merge this.

src/components/Contacts/ContactFilters/ContactFilters.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+};
+
+const accountListId = '111';
+
+export const Default = (): ReactElement => {
+  return <ContactFilters accountListId={accountListId} />;
+};
+
+export const Loading = (): ReactElement => {
+  return <ContactFilters accountListId={accountListId} />;
```
reldredge71, line 15: @OzzieOrca Is there a way I can mock the GraphQL results for Storybook to display this component in different states?  Same for the ContactsTable.

TheNoodleMoose: @reldredge71  I believe you can wrap it in a ```<MockedProvider>``` and then pass your mocks you want to it, Here's an example: https://github.com/CruGlobal/mpdx-react/blob/811726cf039830f5449da17790f2ec28f6c6aa7b/src/components/Task/Drawer/ContactList/ContactList.stories.tsx#L15-L18

And if you wanted to do a loading state you would do something like this:
https://github.com/CruGlobal/mpdx-react/blob/811726cf039830f5449da17790f2ec28f6c6aa7b/src/components/Task/Drawer/ContactList/ContactList.stories.tsx#L29-L34

reldredge71: @TheNoodleMoose thanks!  Another question: I want to display the filters panel before and after the "load filters" button is pushed, because the loading state doesn't appear until after the button triggers the lazy query.  Is there a way to simulate a button push in order to display the loading state in Storybook?  (Maybe I can leave that alone for now and let @frett do it when he work on the filters)

frett: I'm fine with you leaving that for me if you want

TheNoodleMoose: Hmm, I'm not sure as far as storybook how to do that or if it's possible(yet). Maybe @OzzieOrca would know. But maybe a unit test would be better for that scenario? You have [this test](https://github.com/CruGlobal/mpdx-react/pull/34/files/e9416cccab9c79483b5dcdeea01d100c860f6386#diff-2f3a5e662d909bbf6e4ed9179d38312e1b7e8f288c861584499ca26a61d1fec4R26) which I believe covers the test case. But at the same time, it would be nice to visually see it happen.

reldredge71: Yeah I would like to visually show the before, during, and after loading states for the design team's approval, hopefully we find a way to do that.  So far it looks like Storybook will give us outputs for the provided inputs, but triggering things inside the component is different.

TheNoodleMoose: Yeah, I tried looking at the docs but didn't see anything about mocking data and waiting for that data to be provided to the component until something is interacted with.

src/components/Contacts/ContactFilters/ContactFilters.mock.tsx
```diff
@@ -0,0 +1,59 @@
...
+        query: ContactFiltersDocument,
+        variables: {
+          accountListId,
+        },
+      },
+      result: {
+        data: contactFiltersQuery,
+      },
+    },
+  ];
```
OzzieOrca, line 29: Interesting combination of automocks and manual mocks :) I like it although I hope we can keep the normal usage a little less verbose.

`<GqlMockedProvider>` should work in place of `<MockedProvider>`. Currently you can't customize the mocked data (the bug I've been meaning to look at more). But I'm pretty sure customizing the data works with `gqlMock(...)`. This could be a good way to get around that issue or create more complex mocks that match variables or return different values for different variables or return specific errors.

If I can't figure out my bug, this could be a cool way to implement it ourselves :)

src/components/Contacts/ContactFilters/ContactFilters.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+  return (
+    <MockedProvider mocks={ContactFiltersQueryEmptyMocks(accountListId)}>
+      <ContactFilters accountListId={accountListId} />
+    </MockedProvider>
+  );
+};
+
+Default.story = {
+  parameters: {
+    chromatic: { delay: 1000 },
```
OzzieOrca, line 42: This still needed? You've got several of them.

src/components/Contacts/ContactFilters/ContactFilters.tsx
```diff
@@ -0,0 +1,35 @@
...
+interface Props {
+  accountListId: string;
+}
+
+export const ContactFilters: React.FC<Props> = ({ accountListId }: Props) => {
+  const [
+    loadContactFilters,
+    { data, loading, error },
+  ] = useContactFiltersLazyQuery({
+    variables: { accountListId: accountListId as string },
```
OzzieOrca, line 14: This is already a string right?
```suggestion
    variables: { accountListId: accountListId },
```

src/components/Contacts/ContactFilters/ContactFilters.mock.tsx
```diff
@@ -0,0 +1,59 @@
...
+};
+
+export const ContactFiltersQueryLoadingMocks = (
+  accountListId: string,
+): MockedResponse[] => {
+  return [
+    {
+      ...ContactFiltersQueryDefaultMocks(accountListId)[0],
+      delay: 100931731455,
```
OzzieOrca, line 56: What is this? You've got several of them.

src/components/Contacts/ContactsTable/ContactsTable.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+    const mocks = {
+      Contacts: {
+        contacts: {
+          nodes: [],
+        },
+      },
+    };
+
+    const { queryByTestId, queryByText } = render(
+      <GqlMockedProvider<ContactsQuery> mocks={mocks}>
```
OzzieOrca, line 50: Huh, looks like mocks worked for you here. I was never specifying the top level query name... Looks like you have `Contacts` at the top here. That sounds like my automock bug.

Maybe you've solved it? 🎉 If so it was user error 🤦

src/components/Contacts/ContactsTable/ContactsTable.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { ContactsHeader } from '../ContactsHeader/ContactsHeader';
+import { useContactsQuery } from '../../../../pages/accountLists/[accountListId]/Contacts.generated';
+
+interface Props {
+  accountListId: string;
+}
+
+export const ContactsTable: React.FC<Props> = ({ accountListId }: Props) => {
+  const { data, loading, error } = useContactsQuery({
+    variables: { accountListId: accountListId as string },
```
OzzieOrca, line 20:
```suggestion
    variables: { accountListId: accountListId },
```

OzzieOrca (changes requested):
Looks good! Couple more small things.

Seems like you're using several different strategies for mocking GraphQL data. Maybe we could move towards `<GqlMockedProvider>` since it looks like you got that working. We can use `gqlMock(...)` for fragments. And use `gqlMock(...)` with Apollo's `<MockedProvider>` if we need the data to change depending on variables or need to test errors.

You're the one who's been writing this code, do you see any cases where a specific strategy makes the most sense? We can keep iterating on our testing helpers.

src/components/Contacts/ContactsTable/ContactsTable.test.tsx
TheNoodleMoose: Yep, this seems to fix it! I just fixed this test: https://github.com/CruGlobal/mpdx-react/blob/811726cf039830f5449da17790f2ec28f6c6aa7b/pages/accountLists/%5BaccountListId%5D/contacts.test.tsx#L38-L40
by adding the top level query name!

src/components/Contacts/ContactFilters/ContactFilters.mock.tsx
reldredge71: I didn't know how to create a mock loading state, but I found this in one of the other files: https://github.com/CruGlobal/mpdx-react/blob/main/src/components/Dashboard/ThisWeek/WeeklyActivity/WeeklyActivity.mock.ts#L110.  From what I can tell, it delays the results from actually returning, so when you test with Jest or Storybook it will still render in a loading state.  There's probably a better way to do that though.

src/components/Contacts/ContactFilters/ContactFilters.stories.tsx
reldredge71: We probably don't need them... again, I found them in another Storybook file, so I thought it should be there for a reason.

src/components/Contacts/ContactFilters/ContactFilters.mock.tsx
reldredge71: Yeah my only reason for passing `accountListId` into the mocks is that the component itself requires you to pass `accountListId` into it, so I want the mock data to be consistent with the data being put into the component.  If the mocks were completely automated but we hardcode the `accountListId` that is passed into the component, there's a chance that they might not be equal to each other.  So I wanted the test or stories file to determine what the `accountlistId` should be, and pass it in here.

I can try replacing `MockedProvider` with `GqlMockedProvider` throughout this PR.

OzzieOrca: If you want to assert that it was queried with certain values you can use:
https://github.com/SurveyMonkey/graphql-ergonomock#:~:text=Finally,%20you%20can%20spy%20on%20executed%20operations%20via%20the%20onCall%20prop.

src/components/Contacts/ContactFilters/ContactFilters.stories.tsx
OzzieOrca: If you're using some sort of mocked provider, hopefully you don't need to delay it. There could still be issues though...

src/components/Contacts/ContactFilters/ContactFilters.mock.tsx
OzzieOrca: Is this part of Apollo's MockedProvider API? I haven't played with it... Are you having trouble getting it out of the loading state or keeping it in the loading state?

reldredge71: Yeah I'm having trouble keeping it in a loading state... I hope there's a better way than this

OzzieOrca (approved):
The tests look a lot cleaner, thanks :)

PR #35, opened 2021-03-18
Add Tasks menu item back

OzzieOrca:


TheNoodleMoose (approved):
Thanks for adding this back after I ripped everything out haha

OzzieOrca:
I think it's missing from Figma for some reason. I don't think they have permission to drop one of the 2 main features of the app haha :)

PR #40, opened 2021-03-22
MPDX-6944-Add-Delete-Button-Task-Detail-Drawer

TheNoodleMoose:
@OzzieOrca and I discussed this on slack, but in order to remove the task from the apollo cache after deleting it(to avoid having to query all tasks again), I had to pass down the ```filter``` and the ```rowsPerPage``` in order to correctly query the data from the cache. 

I haven't added stories or tests yet for this new logic as I wanted to get others' opinions first on this implementation and see if anyone has any suggestions for improvement.

### DIFF
--- a/pages/accountLists/[accountListId]/contacts.test.tsx
+++ b/pages/accountLists/[accountListId]/contacts.test.tsx
@@ -37,6 +37,6 @@ it('should render list of people', async () => {
   );
   // TODO: Fix custom mocks so we don't have to rely on finding first and checking if it matches a hardcoded random value
   expect((await findAllByRole('row'))[0]).toHaveTextContent(
-    'Stomach BrainCoffee Elephant Bee',
+    'Cup SandwichStar God Baby',
   );
 });
--- a/pages/accountLists/[accountListId]/contacts.test.tsx
+++ b/pages/accountLists/[accountListId]/contacts.test.tsx
@@ -28,15 +28,14 @@ it('should render list of people', async () => {
     <TestRouter router={router}>
       <GqlMockedProvider<ContactsQuery>
         mocks={{
-          contacts: { nodes: [{ name }] },
+          Contacts: {
+            contacts: { nodes: [{ name }] },
+          },
         }}
       >
         <Contacts />
       </GqlMockedProvider>
     </TestRouter>,
   );
-  // TODO: Fix custom mocks so we don't have to rely on finding first and checking if it matches a hardcoded random value
-  expect((await findAllByRole('row'))[0]).toHaveTextContent(
-    'Cup SandwichStar God Baby',
-  );
+  expect((await findAllByRole('row'))[0]).toHaveTextContent(name);
 });
--- a/src/components/Task/Drawer/Form/Form.stories.tsx
+++ b/src/components/Task/Drawer/Form/Form.stories.tsx
@@ -19,6 +19,17 @@ export default {
   title: 'Task/Drawer/Form',
 };
 
+const mockFilter = {
+  userIds: [],
+  tags: [],
+  contactIds: [],
+  activityType: [],
+  completed: null,
+  startAt: null,
+  before: null,
+  after: null,
+};
+
 export const Default = (): ReactElement => {
   return (
     <MockedProvider
@@ -28,7 +39,12 @@ export const Default = (): ReactElement => {
       ]}
       addTypename={false}
     >
-      <TaskDrawerForm accountListId="abc" onClose={(): void => {}} />
+      <TaskDrawerForm
+        accountListId="abc"
+        filter={mockFilter}
+        rowsPerPage={100}
+        onClose={(): void => {}}
+      />
     </MockedProvider>
   );
 };
@@ -36,7 +52,12 @@ export const Default = (): ReactElement => {
 export const Loading = (): ReactElement => {
   return (
     <MockedProvider mocks={[]} addTypename={false}>
-      <TaskDrawerForm accountListId="abc" onClose={(): void => {}} />
+      <TaskDrawerForm
+        accountListId="abc"
+        filter={mockFilter}
+        rowsPerPage={100}
+        onClose={(): void => {}}
+      />
     </MockedProvider>
   );
 };
@@ -88,6 +109,8 @@ export const Persisted = (): ReactElement => {
     >
       <TaskDrawerForm
         accountListId="abc"
+        filter={mockFilter}
+        rowsPerPage={100}
         task={task}
         onClose={(): void => {}}
       />
--- a/src/components/Task/Drawer/Form/Form.stories.tsx
+++ b/src/components/Task/Drawer/Form/Form.stories.tsx
@@ -12,8 +12,10 @@ import {
   getDataForTaskDrawerMock,
   createTaskMutationMock,
   updateTaskMutationMock,
+  deleteTaskMutationMock,
 } from './Form.mock';
 import TaskDrawerForm from '.';
+import { getTasksForTaskListMock } from '../../List/List.mock';
 
 export default {
   title: 'Task/Drawer/Form',
@@ -103,7 +105,9 @@ export const Persisted = (): ReactElement => {
     <MockedProvider
       mocks={[
         getDataForTaskDrawerMock(),
+        getTasksForTaskListMock(),
         { ...updateTaskMutationMock(), delay: 500 },
+        { ...deleteTaskMutationMock(), delay: 1000 },
       ]}
       addTypename={false}
     >
--- a/src/components/Task/Drawer/Form/Form.stories.tsx
+++ b/src/components/Task/Drawer/Form/Form.stories.tsx
@@ -1,6 +1,7 @@
 import React, { ReactElement } from 'react';
 import { MockedProvider } from '@apollo/client/testing';
 import { DateTime } from 'luxon';
+import { InMemoryCache } from '@apollo/client';
 import {
   ActivityTypeEnum,
   Contact,
@@ -8,6 +9,8 @@ import {
   NotificationTypeEnum,
   Task,
 } from '../../../../../graphql/types.generated';
+import { getTasksForTaskListMock } from '../../List/List.mock';
+import { GetTasksForTaskListDocument } from '../../List/TaskList.generated';
 import {
   getDataForTaskDrawerMock,
   createTaskMutationMock,
@@ -15,7 +18,6 @@ import {
   deleteTaskMutationMock,
 } from './Form.mock';
 import TaskDrawerForm from '.';
-import { getTasksForTaskListMock } from '../../List/List.mock';
 
 export default {
   title: 'Task/Drawer/Form',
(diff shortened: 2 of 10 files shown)


pages/accountLists/[accountListId]/contacts.test.tsx
```diff
@@ -37,6 +37,6 @@ it('should render list of people', async () => {
   );
   // TODO: Fix custom mocks so we don't have to rely on finding first and checking if it matches a hardcoded random value
   expect((await findAllByRole('row'))[0]).toHaveTextContent(
-    'Stomach BrainCoffee Elephant Bee',
+    'Cup SandwichStar God Baby',
```
Comment (TheNoodleMoose) [inline pages/accountLists/[accountListId]/contacts.test.tsx:40]: Another unfortunate name haha

src/components/Task/Drawer/Form/Form.test.tsx
```diff
@@ -54,14 +70,13 @@ describe('TaskDrawerForm', () => {
...
+              filter={mockFilter}
+              rowsPerPage={100}
               onClose={onClose}
               task={{
                 activityType: null,
                 contacts: {
                   nodes: [],
-                  pageInfo: { hasNextPage: false, hasPreviousPage: false },
-                  totalCount: 0,
-                  totalPageCount: 0,
```
TheNoodleMoose, line 64: @OzzieOrca, I'm not quite sure why typescript was yelling at me to remove this. I stashed all my changes and the error went away, so it has to be something I added. Just not sure what.

src/components/Task/Drawer/Form/Form.tsx
```diff
@@ -550,6 +640,32 @@ const TaskDrawerForm = ({
                 </Button>
               </Grid>
             </Grid>
+            <Dialog
```
TheNoodleMoose, line 643: I tried to mimic the current design and flow of deleting a task from mpdx.org. I think the ```Dialog``` component from MaterialUI works pretty well for this kind of confirmation popup.

src/components/Task/Drawer/Form/Form.stories.tsx
```diff
@@ -19,6 +19,17 @@ export default {
   title: 'Task/Drawer/Form',
 };
 
+const mockFilter = {
```
TheNoodleMoose, line 22: Also not sure what's the best way to approach where this mockFIlter lives. Should I just put it in every file that needs it, or should I create a separate file for it and just import that into a test or storybook if it needs it? I will also have different variations of the filter(ex. ```emptyMockFilter```, ```mockFilter```, etc).

OzzieOrca: 🤷‍♂️ You could export it.. Idk where would be best. Maybe the test?

src/components/Task/Drawer/Form/Form.test.tsx
OzzieOrca: I assume this is affected by the GraphQL query and the codegen for it. Maybe these fields aren't requested.

src/components/Task/Drawer/Form/Form.tsx
```diff
@@ -58,6 +71,13 @@ const useStyles = makeStyles((theme: Theme) => ({
   title: {
     flexGrow: 1,
   },
+  removeButton: {
+    backgroundColor: '#d9534f',
+    color: '#fff',
+    '&:hover': {
+      backgroundColor: '#C9302F',
+    },
```
OzzieOrca, line 79: Are these theme colors?

src/components/Task/Drawer/Form/Form.tsx
```diff
@@ -191,6 +218,49 @@ const TaskDrawerForm = ({
...
+              },
+            });
+          },
+        });
+        enqueueSnackbar(t('Task deleted successfully'), { variant: 'success' });
+        handleRemoveDialog(false);
+        onClose();
+      }
+    } catch (error) {
+      enqueueSnackbar(error.message, { variant: 'error' });
```
OzzieOrca, line 260: If we want this error in Rollbar, we should probably rethrow the error.

src/components/Task/Drawer/Form/Form.tsx
```diff
@@ -527,12 +597,32 @@ const TaskDrawerForm = ({
...
+                <Button
+                  size="large"
+                  variant="contained"
+                  className={classes.removeButton}
+                  onClick={() => handleRemoveDialog(true)}
+                >
+                  {deleting ? (
+                    <>
+                      <CircularProgress color="primary" size={20} />
+                      &nbsp;
```
OzzieOrca, line 610: This intentional?

src/components/Task/Drawer/Form/Form.tsx
```diff
@@ -550,6 +640,32 @@ const TaskDrawerForm = ({
                 </Button>
               </Grid>
             </Grid>
+            <Dialog
+              disableBackdropClick
+              disableEscapeKeyDown
```
OzzieOrca, line 645: Is there a good reason for these? They always seem rude to disable on a user.

src/components/Task/Drawer/Form/Form.tsx
```diff
@@ -550,6 +640,32 @@ const TaskDrawerForm = ({
...
+              open={removeDialogOpen}
+              aria-labelledby="remove task confirmation"
```
OzzieOrca, line 647: This should probably be translated?

src/components/Task/List/List.tsx
```diff
@@ -434,7 +434,11 @@ const TaskList = ({ initialFilter }: Props): ReactElement => {
       setCurrentPage(newPage);
     },
     onRowClick: (_rowData, rowMeta) => {
-      openTaskDrawer({ taskId: data.tasks.nodes[rowMeta.dataIndex].id });
+      openTaskDrawer({
+        taskId: data.tasks.nodes[rowMeta.dataIndex].id,
```
OzzieOrca, line 438: This doesn't work? (I didn't look at any docs...)
```suggestion
        taskId: rowData.id,
```

OzzieOrca (changes requested):
Looks good, just a few smaller things.

src/components/Task/Drawer/Form/Form.tsx
TheNoodleMoose: ```#fff``` is one so I switched to using it from the theme instead of hard coding it here. ```#d9534f``` was the color from the delete button on mpdx.org that I just copied over, but I'm going to instead use the error color from the theme. I'm also going to add ```#C9302F``` to the theme as the dark version of the error color.

src/components/Task/Drawer/Form/Form.tsx
TheNoodleMoose: I think I added it, but it's probably not needed. I'll delete it.

src/components/Task/Drawer/Form/Form.tsx
TheNoodleMoose: Yeah, you're probably right. I'll remove these. I had noticed currently on mpdx.org that if I have the confirmation modal open to delete a task, and click outside of the modal it will just delete the task without the user confirming. Not sure if that's intentional or not... but I think that's why I went ahead and prevented that here(except here it just calls ```onClose``` and nothing else).

src/components/Task/Drawer/Form/Form.tsx
TheNoodleMoose: Yep, will add.

src/components/Task/List/List.tsx
TheNoodleMoose: It looks like ```rowData``` is just an array of strings

OzzieOrca: Huh, weird 🤷‍♂️

pages/accountLists/[accountListId]/contacts.test.tsx
```diff
@@ -28,15 +28,14 @@ it('should render list of people', async () => {
...
       >
         <Contacts />
       </GqlMockedProvider>
     </TestRouter>,
   );
-  // TODO: Fix custom mocks so we don't have to rely on finding first and checking if it matches a hardcoded random value
-  expect((await findAllByRole('row'))[0]).toHaveTextContent(
-    'Cup SandwichStar God Baby',
-  );
+  expect((await findAllByRole('row'))[0]).toHaveTextContent(name);
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/contacts.test.tsx:40]: 🎉 Thanks for fixing this 😄

src/components/Task/Drawer/Form/Form.tsx
OzzieOrca: To be clear I was commenting on the `&nbsp;` line only, not the progress indicator. Looks like you removed the whole indicator...

OzzieOrca (approved):

src/components/Task/Drawer/Form/Form.tsx
TheNoodleMoose: Ahhhh... gotcha haha. I was a bit confused as the dialog component is focused on the screen so the user doesn't really see this loading indicator but I'll add it back. I think I just copied the ```&nbsp;``` that was in the other loading indicator on the form. Does this need to be there? https://github.com/CruGlobal/mpdx-react/blob/811726cf039830f5449da17790f2ec28f6c6aa7b/src/components/Task/Drawer/Form/Form.tsx#L543-L547

OzzieOrca: Oh weird that it's hidden. Maybe there's a better place for it or it should show in the modal before it closes.

`&nbsp;` (non breaking space) is just a space that won't collapse into other spaces. Seems odd that you'd need a space character around a loading indicator. Usually it's a hack to get extra spacing or a way to HTML encode things that need the spacing.

TheNoodleMoose: It's not necessarily hidden, just not what the user is focused on.
Screen Shot 2021-03-25 at 9 23 19 AM

 I'm going to move it into the modal and switch out the text for the loading indicator while it's loading.

src/components/Task/Drawer/Form/Form.stories.tsx
```diff
@@ -103,7 +105,9 @@ export const Persisted = (): ReactElement => {
     <MockedProvider
       mocks={[
         getDataForTaskDrawerMock(),
+        getTasksForTaskListMock(),
         { ...updateTaskMutationMock(), delay: 500 },
+        { ...deleteTaskMutationMock(), delay: 1000 },
```
TheNoodleMoose, line 110: @OzzieOrca So the mutation works in this storybook, but I ran into an error as it's trying to update the cache after the mutation. Mainly because the cache does not have the data from the ```getTasksForTaskListQuery```. I can add the mock here, but since it doesn't get called in the file, it's not in the cache.
Screen Shot 2021-03-25 at 10 14 10 AM

Any ideas on how I could, I guess, force it to run the query but just in this storybook? Or can I have some fallback logic in the update function to query for that data if there is nothing in the cache?

TheNoodleMoose: Actually, I figured it out! 🙂

src/components/Task/Drawer/Form/Form.tsx
```diff
@@ -609,7 +610,7 @@ const TaskDrawerForm = ({
                   className={classes.removeButton}
                   onClick={() => handleRemoveDialog(true)}
                 >
-                  <DeleteIcon titleAccess={t('Remove')} />
+                  <DeleteIcon />
```
TheNoodleMoose, line 613: I removed this ```titleAccess``` as it was making the accessibility name of the button ```remove remove```

OzzieOrca (approved):

PR #41, opened 2021-03-23
MPDX-6952 : Contact view status widget

gyasistory:
https://jira.cru.org/browse/MPDX-6952

### DIFF
--- a/src/components/Contacts/ContactView/StatusSelector/StatusSelector.tsx
+++ b/src/components/Contacts/ContactView/StatusSelector/StatusSelector.tsx
@@ -1,27 +1,77 @@
 import { Box, InputLabel, MenuItem, Select } from '@material-ui/core';
-import React, { ReactElement, ReactNode } from 'react';
+import { makeStyles, Theme } from '@material-ui/core/styles';
+import { ErrorOutline } from '@material-ui/icons';
+import React from 'react';
 import { useTranslation } from 'react-i18next';
 
+const useStyles = makeStyles((theme: Theme) => ({
+  statusLabel: {
+    padding: theme.spacing(0),
+    fontFamily: theme.typography.fontFamily,
+    color: theme.palette.text.primary,
+    fontSize: '12px',
+    letterSpacing: '0.4px',
+    margin: '4px 0px',
+  },
+  statusMenuItem: {
+    fontFamily: theme.typography.fontFamily,
+    fontSize: '16px',
+    background: theme.palette.common.white,
+  },
+  statusIcon: {
+    color: theme.palette.secondary.main,
+    marginRight: theme.spacing(1),
+    width: '20px',
+    height: '20px',
+  },
+  statusSelector: {
+    fontFamily: theme.typography.fontFamily,
+    fontSize: '16px',
+    background: theme.palette.common.white,
+    paddingLeft: '16px',
+    paddingRight: '16px',
+    paddingTop: '8px',
+    paddingBottom: '8px',
+  },
+}));
 interface StatusSelectorProps {
   statuses: string[];
   selectedStatus: string;
+  saveContactStatus: (status: string) => void;
 }
 
 export const StatusSelector: React.FC<StatusSelectorProps> = ({
   statuses,
   selectedStatus,
+  saveContactStatus,
 }) => {
   const { t } = useTranslation();
+  const classes = useStyles();
   const [contactStatus, setContactStatus] = React.useState(selectedStatus);
   const handleChange = (event: React.ChangeEvent<{ value: unknown }>) => {
     setContactStatus(event.target.value as string);
+    saveContactStatus(event.target.value as string);
   };
   return (
     <Box>
-      <InputLabel>{t('Status')}</InputLabel>
-      <Select id="contactStatus" value={contactStatus} onChange={handleChange}>
+      <InputLabel className={classes.statusLabel}>{t('Status')}</InputLabel>
+      <Select
+        className={classes.statusSelector}
+        id="contactStatus"
+        value={contactStatus}
+        onChange={handleChange}
+      >
         {statuses.map(function (status, _index) {
-          return <MenuItem value={status}>{status}</MenuItem>;
+          return (
+            <MenuItem
+              className={classes.statusMenuItem}
+              value={status}
+              key={status}
+            >
+              <ErrorOutline className={classes.statusIcon} />
+              {status}
+            </MenuItem>
+          );
         })}
       </Select>
     </Box>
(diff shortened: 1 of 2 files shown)


j2trumpet:
It looks like this is using a lot of the same CSS as [MPDX-6952 : Frequency Contact Selector](https://github.com/CruGlobal/mpdx-react/pull/43). Is there a way to make those two components run off the same styles?

src/components/Contacts/ContactView/StatusSelector/StatusSelector.tsx
```diff
@@ -0,0 +1,79 @@
...
+    fontSize: '16px',
+    background: theme.palette.common.white,
+    paddingLeft: '16px',
+    paddingRight: '16px',
+    paddingTop: '8px',
+    paddingBottom: '8px',
+  },
+}));
+interface StatusSelectorProps {
+  statuses: string[];
```
OzzieOrca, line 38: I created https://jira.cru.org/browse/MPDX-6958 to load these. It seemed like a good first custom hook/GraphQL ticket for you so I assigned it to you. Hopefully this component could call it like `const { statuses } = useApiConstants();`. I'd be happy to work with you on it if you need more direction. You can create a `.graphql` file with the query and then run `yarn gql` which should generate a `useLoadConstantsQuery()` hook for you that you can call in this new custom hook.

I'm fine if you want to merge this PR as is first though and come back to this.

Seems like a cleaner way than making something else load them and pass them in.

src/components/Contacts/ContactView/StatusSelector/StatusSelector.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { makeStyles, Theme } from '@material-ui/core/styles';
+import { ErrorOutline } from '@material-ui/icons';
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+
+const useStyles = makeStyles((theme: Theme) => ({
+  statusLabel: {
+    padding: theme.spacing(0),
+    fontFamily: theme.typography.fontFamily,
+    color: theme.palette.text.primary,
```
OzzieOrca, line 11: Are any of these CSS properties already the defaults? I think it's better to not specify them repeatedly unless you need to overwrite styles elsewhere.

src/components/Contacts/ContactView/StatusSelector/StatusSelector.tsx
```diff
@@ -0,0 +1,79 @@
...
+    height: '20px',
+  },
+  statusSelector: {
+    fontFamily: theme.typography.fontFamily,
+    fontSize: '16px',
+    background: theme.palette.common.white,
+    paddingLeft: '16px',
+    paddingRight: '16px',
+    paddingTop: '8px',
+    paddingBottom: '8px',
```
OzzieOrca, line 34: Is `theme.spacing(x)` useful for any of the above spacing values?

src/components/Contacts/ContactView/StatusSelector/StatusSelector.tsx
```diff
@@ -0,0 +1,79 @@
...
+  },
+}));
+interface StatusSelectorProps {
+  statuses: string[];
+  selectedStatus: string;
+  saveContactStatus: (status: string) => void;
```
OzzieOrca, line 40:
```suggestion
  updateContactStatus: (status: string) => void;
```
might be a better name for this since it isn't actually doing any saving yet. 🤷‍♂️  We'll have to figure out if we want this component to fire off a mutation or make the parent component do it but we can figure that out later.

src/components/Contacts/ContactView/StatusSelector/StatusSelector.tsx
```diff
@@ -0,0 +1,79 @@
...
+  statuses,
+  selectedStatus,
+  saveContactStatus,
+}) => {
+  const { t } = useTranslation();
+  const classes = useStyles();
+  const [contactStatus, setContactStatus] = React.useState(selectedStatus);
+  const handleChange = (event: React.ChangeEvent<{ value: unknown }>) => {
+    setContactStatus(event.target.value as string);
+    saveContactStatus(event.target.value as string);
```
OzzieOrca, line 53:
```suggestion
  const handleChange = (event: React.ChangeEvent<{ value: string }>) => {
    setContactStatus(event.target.value);
    saveContactStatus(event.target.value);
```
should type this function better.

src/components/Contacts/ContactView/StatusSelector/StatusSelector.tsx
```diff
@@ -0,0 +1,79 @@
...
+}
+
+export const StatusSelector: React.FC<StatusSelectorProps> = ({
+  statuses,
+  selectedStatus,
+  saveContactStatus,
+}) => {
+  const { t } = useTranslation();
+  const classes = useStyles();
+  const [contactStatus, setContactStatus] = React.useState(selectedStatus);
```
OzzieOrca, line 50:
```suggestion
  const [contactStatus, setContactStatus] = useState(selectedStatus);
```
I usually like to import this by name but it doesn't really matter.
`import React, { useState } from 'react';`

src/components/Contacts/ContactView/StatusSelector/StatusSelector.tsx
```diff
@@ -0,0 +1,79 @@
...
+            <MenuItem
+              className={classes.statusMenuItem}
+              value={status}
+              key={status}
+            >
+              <ErrorOutline className={classes.statusIcon} />
+              {status}
+            </MenuItem>
+          );
+        })}
```
OzzieOrca, line 75: Arrow functions! 🎉 (also you don't need the `_index` if you aren't using it)
```suggestion
        {statuses.map((status) => (
          <MenuItem
            className={classes.statusMenuItem}
            value={status}
            key={status}
          >
            <ErrorOutline className={classes.statusIcon} />
            {status}
          </MenuItem>
        ))}
```

OzzieOrca (changes requested):
Looks good :) Few trivial things and then some thoughts about loading the statuses from the API.

OzzieOrca:
To @j2trumpet's point maybe we could create a shared component that handles styling a select that could be customized by these specific components. You could hand the shared component the possible options and an icon or something...

OzzieOrca:
Looks like the design may have changed on you 😞 
https://www.figma.com/file/zvqkcHImucOoPyXrYNqrkn/MPDX-Update-Material-UI?node-id=631%3A324
https://www.figma.com/file/zvqkcHImucOoPyXrYNqrkn/MPDX-Update-Material-UI?node-id=631%3A508

At least it's labeled `Truth` now 😄

PR #42, opened 2021-03-24
MPDX-6920: Fix Storybook Renders

reldredge71:
Fixed the Storybook tests that were producing errors.  Also added error messages to the components and wrote tests for them.

### DIFF
--- a/src/components/Task/Drawer/CommentList/CommentList.tsx
+++ b/src/components/Task/Drawer/CommentList/CommentList.tsx
@@ -51,11 +51,11 @@ const TaskDrawerCommentList = ({
 
   const ref = useRef(null);
 
+  const nodes = data?.task?.comments?.nodes;
+
   useEffect(() => {
     ref.current.scrollIntoView({ behaviour: 'smooth' });
-  }, [data?.task?.comments?.nodes]);
-
-  const nodes = data?.task.comments.nodes;
+  }, [nodes]);
 
   return (
     <>
--- a/src/components/Task/Drawer/CommentList/CommentList.tsx
+++ b/src/components/Task/Drawer/CommentList/CommentList.tsx
@@ -58,6 +58,8 @@ const TaskDrawerCommentList = ({
     ref.current.scrollIntoView({ behaviour: 'smooth' });
   }, [data?.task?.comments?.nodes]);
 
+  const nodes = data?.task.comments.nodes;
+
   return (
     <>
       <Box m={2}>
@@ -70,7 +72,7 @@ const TaskDrawerCommentList = ({
           </Box>
         ) : (
           <>
-            {data?.task.comments.nodes?.length === 0 && (
+            {nodes?.length === 0 && (
               <Card data-testid="TaskDrawerCommentListEmpty">
                 <CardContent className={classes.cardContent}>
                   <img
@@ -82,8 +84,8 @@ const TaskDrawerCommentList = ({
                 </CardContent>
               </Card>
             )}
-            {data?.task.comments.nodes?.length > 0 &&
-              data.task.comments.nodes?.reduce((result, comment) => {
+            {nodes?.length > 0 &&
+              nodes.reduce((result, comment, index) => {
                 return [
                   ...result,
                   <Box
@@ -93,7 +95,7 @@ const TaskDrawerCommentList = ({
                     <TaskDrawerCommentListItem
                       comment={comment}
                       reverse={comment.me}
-                      nextComment={data.task.comments?.nodes[result.length + 1]}
+                      nextComment={nodes[index + 1]}
                     />
                   </Box>,
                 ];
--- a/src/components/Task/Drawer/CommentList/CommentList.tsx
+++ b/src/components/Task/Drawer/CommentList/CommentList.tsx
@@ -39,18 +39,15 @@ const TaskDrawerCommentList = ({
   const { t } = useTranslation();
   const { enqueueSnackbar } = useSnackbar();
 
-  const { data, loading, error } = useGetCommentsForTaskDrawerCommentListQuery({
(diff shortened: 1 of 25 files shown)


src/components/Task/Drawer/CommentList/CommentList.tsx
```diff
@@ -53,7 +53,9 @@ const TaskDrawerCommentList = ({
   return (
     <>
       <Box m={2}>
-        {loading ? (
+        {error ? (
+          <p>{error.toString()}</p>
```
reldredge71, line 57: This is very bare bones, not sure how we want to render error messages

OzzieOrca: Christian was using the snackbar. https://github.com/CruGlobal/mpdx-react/pull/40/files#diff-c2a85f1c4f401282f7b2bffdfc7e006980106f80113c4e0668b8d798ea616643R260

Seems ok. We haven't discussed any best practices but I assume Tatai was the one who brought that in. Not sure how that would work during render... Maybe it needs a useEffect (or a custom hook to watch it and call the snackbar). Or we could make an error icon component with a message that could be rendered in the body of the list.

OzzieOrca (approved):
Thanks :) Looks good.

src/components/Task/Drawer/CommentList/CommentList.tsx
```diff
@@ -44,6 +46,12 @@ const TaskDrawerCommentList = ({
     },
   });
 
+  useEffect(() => {
```
OzzieOrca, line 49: I may have misled you with the `useEffect` suggestion. It's fine but do you think `onError` https://www.apollographql.com/docs/react/api/react/hooks/#options would be better? Idk how easy that is to test. Hopefully it's easy.

src/components/Task/Drawer/CommentList/CommentList.tsx
```diff
@@ -76,8 +82,8 @@ const TaskDrawerCommentList = ({
                 </CardContent>
               </Card>
             )}
-            {data.task.comments.nodes.length > 0 &&
-              data.task.comments.nodes.reduce((result, comment) => {
+            {data?.task.comments.nodes?.length > 0 &&
+              data.task.comments.nodes?.reduce((result, comment) => {
```
OzzieOrca, line 86: This is defined if the last condition was true
```suggestion
              data.task.comments.nodes.reduce((result, comment) => {
```

src/components/Task/Drawer/CommentList/CommentList.tsx
```diff
@@ -74,8 +82,8 @@ const TaskDrawerCommentList = ({
...
```
OzzieOrca, line 86:
```suggestion
              data.task.comments.nodes.reduce((result, comment, index, nodes) => {
```
See comment below

src/components/Task/Drawer/CommentList/CommentList.tsx
```diff
@@ -85,7 +93,7 @@ const TaskDrawerCommentList = ({
                     <TaskDrawerCommentListItem
                       comment={comment}
                       reverse={comment.me}
-                      nextComment={data.task.comments.nodes[result.length + 1]}
+                      nextComment={data.task.comments?.nodes[result.length + 1]}
```
OzzieOrca, line 96: You didn't write this but it seems a little odd. I guess it's fine, I was just trying to think if there was a better way to do it. You could maybe make two arrays, shift the second array left one (by removing the first element), zip them together so you've got pairs of the comment and the next comment and then map but I think that's worse 😄 

Does this help make it slightly simpler? 🤷‍♂️ 
```suggestion
                      nextComment={nodes[index + 1]}
```

OzzieOrca (changes requested):
Thanks, looks good. Just a couple small things.

src/components/Task/Drawer/CommentList/CommentList.tsx
reldredge71: Yeah I might have been overly cautious here, but you're right.

OzzieOrca (approved):

OzzieOrca:
Just dumping the Vercel logs here. Idk why the rest of the checks are hanging. Hopefully fixing the conflicts will get it going again.
Screen Shot 2021-03-25 at 2 29 25 PM

reldredge71:
@OzzieOrca I was trying to replace all instances of `MockedProvider` with `GqlMockedProvider`.  However, I was having a lot of issues with the latter.  I couldn't tell what was going on -- the Stories kept producing errors saying something was null when I clearly defined it in the mocks.  `MockedProvider` works perfectly though.  Maybe we need to fix the implementation of `GqlMockedProvider`?

Also I would eventually like to have a way to create a Story for a loading state, without having to add "delay" to the mocks... I haven't found another way to do that yet.  Maybe we can make that a part of `GqlMockedProvider` somehow?

OzzieOrca (approved):

OzzieOrca:
Sorry it's been giving you issues... We can definitely iterate on it. The implementation right now is pass everything down to Ergonomock, add the schema prop, and improve the typings of the mocks. https://github.com/CruGlobal/mpdx-react/blob/7e58576276ece2685e54c26a038334a603e5258d/__tests__/util/graphqlMocking.tsx#L31-L37 I haven't written many tests with it yet...

You have an example of a test that was giving you trouble? Theoretically it should work for both tests and stories... I haven't looked into the delays with stories yet...

reldredge71:
@OzzieOrca I'm not sure what the exact issue was... I've had tests fail just with using `<GqlMockedProvider<QUERYTYPE>>`.  I think in my ContactsTable PR you were talking about how using a top level query made a difference.... maybe the queries I was using in this PR were not as high level as `ContactsQuery`? https://github.com/CruGlobal/mpdx-react/pull/34#discussion_r600671250

PR #43, opened 2021-03-24
MPDX-6952 : Frequency Contact Selector 

gyasistory:
https://jira.cru.org/browse/MPDX-6952


j2trumpet (approved):
Assuming I'm looking at the right part of the design, I think the CSS in this PR looks good.

OzzieOrca:
I glanced at this. I can look more thoroughly at this if you want but it would probably have a lot of the same comments as your other PR and the design change probably affects it too.

PR #46, opened 2021-03-29
MPDX-6930 Contact filters UI

frett:


frett:
probably could use some tests around the different filter types

gyasistory (changes requested):
At first, glace looks good just need to resolve the merge conflict.

src/components/Contacts/ContactFilters/ContactFilters.tsx
```diff
@@ -1,35 +1,211 @@
...
+}) => {
+  const theme = useTheme();
+  const { t } = useTranslation();
+
+  const { data, loading, error } = useContactFiltersQuery({
     variables: { accountListId },
   });
 
+  const [selectedGroup, showGroup] = useState<null | ContactFilterGroup>(null);
+  const [selectedFilters, setSelectedFilters] = useState({});
```
OzzieOrca, line 62:
```suggestion
  const [selectedFilters, setSelectedFilters] = useState<{
    [name: string]: unknown;
  }>({});
```

I'm not sure how to type `value`... I just stuck `unknown` here for now. It seems like some combination of string, string array, and boolean. Maybe that's good enough for now. Or maybe there's some way to set the type for each kind of filter.

I guess that can wait till we wire this up to the contacts query as that probably will have a specific format.

src/components/Contacts/ContactFilters/ContactFilters.tsx
```diff
@@ -1,35 +1,211 @@
...
+      });
+    else
+      setSelectedFilters((prev) => {
+        const updated = { ...prev };
+        delete updated[name];
+        return updated;
+      });
+  };
+
+  const getSelectedFilters = (group: ContactFilterGroup) =>
```
OzzieOrca, line 78: Using updated type from `Filter.ts`:
```suggestion
  const getSelectedFilters = (group: FilterGroup) =>
```

src/components/Contacts/ContactFilters/ContactFilters.tsx
```diff
@@ -1,35 +1,211 @@
...
+    group.filters.filter((value) => selectedFilters[value.name]);
+  const isGroupVisible = (group: ContactFilterGroup) =>
```
OzzieOrca, line 80: Using updated type from `Filter.ts`:
```suggestion
  const isGroupVisible = (group: FilterGroup) =>
```

src/components/Contacts/ContactFilters/ContactFilters.tsx
```diff
@@ -1,35 +1,211 @@
...
+              ) : data?.contactFilters?.length == 0 ? (
+                <ListItem data-testid="NoFiltersState">
+                  <ListItemText
+                    primary={t('No Contact Filters Found')}
+                    primaryTypographyProps={{ variant: 'subtitle1' }}
+                  />
+                </ListItem>
+              ) : (
+                <>
+                  {data?.contactFilters?.map((group: ContactFilterGroup) => (
```
OzzieOrca, line 144: TS can infer this better than manually updating it since `data` has a type
```suggestion
                  {data?.contactFilters?.map((group) => (
```

src/components/Contacts/ContactFilters/ContactFilters.tsx
```diff
@@ -1,35 +1,211 @@
...
+                        <ListItemText
+                          primary={group.title}
+                          primaryTypographyProps={{ variant: 'subtitle1' }}
+                        />
+                        <ArrowForwardIos fontSize="small" color="disabled" />
+                      </ListItem>
+                    </Collapse>
+                  ))}
+                  {data?.contactFilters?.some(
+                    (g: ContactFilterGroup) => !isGroupVisible(g),
```
OzzieOrca, line 160:
```suggestion
                    (g) => !isGroupVisible(g),
```

src/components/Contacts/ContactFilters/ContactFilters.tsx
```diff
@@ -1,35 +1,211 @@
...
+                  ) == true ? (
```
OzzieOrca, line 161:
```suggestion
                  ) ? (
```

src/components/Contacts/ContactFilters/ContactFilters.tsx
```diff
@@ -1,35 +1,211 @@
...
+                style={{ verticalAlign: 'middle' }}
+              >
+                {selectedGroup?.title}
+              </Typography>
+            </FilterHeader>
+            <FilterList dense>
+              {selectedGroup?.filters?.map((filter) => (
+                <FilterListItem
+                  key={filter.id}
+                  filter={filter as Filter}
```
OzzieOrca, line 200: We should be able to make the child components use the updated types from `Filter.ts`:
```suggestion
                  filter={filter}
```

src/components/Shared/Filters/FilterListItemDateRange.tsx
```diff
@@ -0,0 +1,72 @@
...
+  value,
+  onUpdate,
+}) => {
+  const { t } = useTranslation();
+
+  const [startDate, endDate] = ((value: string | null) =>
+    value?.split('..', 2)?.map((value) => DateTime.fromISO(value) || null) ?? [
+      null,
+      null,
+    ])(value);
```
OzzieOrca, line 25: Would using intervals work for this? Looks like we might have to replace `..` with a `/`. https://moment.github.io/luxon/docs/class/src/interval.js~Interval.html#static-method-fromISO.

Also it's odd that this is a IIFE.

src/components/Shared/Filters/FilterListItemMultiselect.tsx
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export const FilterListItemMultiselect: React.FC<Props> = ({
+  filter,
+  selected,
+  onUpdate,
+}: Props) => {
+  const isChecked = (value: string) =>
+    isArray(selected)
+      ? selected.findIndex((it) => it == value) != -1
```
OzzieOrca, line 24: Doesn't appear that you care about the index so https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some would be better.
```suggestion
      ? selected.some((it) => it === value)
```

src/components/Shared/Filters/FilterListItemSelect.tsx
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export const FilterListItemSelect: React.FC<Props> = ({
+  filter,
+  value,
+  onUpdate,
+}: Props) => {
+  return (
+    <ListItem>
+      <FormControl style={{ flex: '1 1 auto' }}>
```
OzzieOrca, line 24:
```suggestion
      <FormControl style={{ flex: 'auto' }}>
```
https://developer.mozilla.org/en-US/docs/Web/CSS/flex#values
> This is equivalent to setting "flex: 1 1 auto".

OzzieOrca (changes requested):

src/components/Contacts/ContactFilters/ContactFilters.tsx
```diff
@@ -1,35 +1,211 @@
...
-      ) : (
-        <ul data-testID="FiltersList">
-          {data.contactFilters.map(({ id, name }) => (
-            <li key={id}>{name}</li>
-          ))}
-        </ul>
-      )}
-    </div>
+    <Box {...boxProps}>
+      <div style={{ overflow: 'hidden' }}>
```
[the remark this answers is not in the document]
frett: this `overflow` attribute is to isolate the horizontal slide animation within the filters div. otherwise the UI slides across the entire screen.

Honestly the current horizontal slide animation needs work, it has positioning issues while it's animating. it goes fast enough that you may not notice it unless you are specifically looking for it

src/components/Contacts/ContactFilters/ContactFilters.tsx
```diff
@@ -1,35 +1,211 @@
...
+        <Slide
+          in={selectedGroup == null}
+          direction="right"
+          appear={false}
+          mountOnEnter
+          unmountOnExit
+        >
+          <div>
```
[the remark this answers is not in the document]
frett: using a fragment here causes a null error because the Slide animation depends on having a single child to write values to

src/components/Shared/Filters/Filter.ts
```diff
@@ -0,0 +1,10 @@
+export type Filter = {
+  title: string;
+  type: string;
+  options: FilterOption[];
+};
+
+export type FilterOption = {
+  id: string | null;
+  name: string;
+};
```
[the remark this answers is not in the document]
frett: I started creating these as a generic class to share between ContactFilters and the future TaskFilters. I'm assuming they would have 2 different GraphQL types, but I wanted to be able to re-use components/types between the 2.

src/components/Contacts/ContactFilters/ContactFilters.tsx
frett: yeah, I was very loose on the filter type for now due to not having an actual API defined, I'm sure some filter value cleanup will be needed once we start using this for filtering queries

src/components/Shared/Filters/Filter.ts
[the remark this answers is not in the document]
OzzieOrca: I was thinking about that a little. It would be great if we could share GraphQL types too but that probably doesn't need to be addressed immediately. The `contactFilters` and `taskFilters` could both return a `FilterGroup` type if we can get all the fields to match up.

If we get that far, we could make a GraphQL fragment for all the fields needed by this filter component and use that fragment for the types here and use that fragment in whatever query is loading the filters.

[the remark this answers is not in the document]
frett: I'll have to take a look at GraphQL fragments, that might be a better immediate solution that will push us towards sharing this logic as the task and contact filter APIs are implemented

pages/api/graphql-rest.page.ts
```diff
@@ -80,20 +86,47 @@ class MpdxRestApi extends RESTDataSource {
...
+      return {
+        id: parent,
+        title: parent,
+        alwaysVisible: false,
+        filters: [],
+      };
+    };
+
+    const response = [];
+    data.forEach(
```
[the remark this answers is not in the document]
frett: I can take a look at using `.reduce` to do this transform, but this is also the temporary REST -> GraphQL glue API until the MPDX API directly supports return contact filters.

src/components/Contacts/ContactFilters/ContactFilters.tsx
```diff
@@ -1,35 +1,211 @@
...
+  ...boxProps
+}) => {
+  const theme = useTheme();
+  const { t } = useTranslation();
+
+  const { data, loading, error } = useContactFiltersQuery({
     variables: { accountListId },
   });
 
+  const [selectedGroup, showGroup] = useState<null | ContactFilterGroup>(null);
```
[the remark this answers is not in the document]
frett: I'm not used to have a different `undefined` and `null` concept, I used `null` everywhere because that is what I'm used to from java/kotlin

pages/api/graphql-rest.page.ts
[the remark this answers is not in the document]
OzzieOrca: Ya, may not be worth the effort. It's fine as is. I just like reduce since it forces you to do things in a more immutable way.

src/components/Contacts/ContactFilters/ContactFilters.tsx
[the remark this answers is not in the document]
OzzieOrca: Ya it's weird. Usually when things are not specified, the value is `undefined` so it's usually less verbose to use that. Also TypeScript optional types (`interface example { optional?: string }`) work with `undefined`. Using `null` is slightly more verbose. So generally I like using `undefined` if I can. But everything optional coming from GraphQL will use `null` though...

It generally doesn't matter in conditions where you aren't matching against a specific value though as both null and undefined are falsy values.

src/components/Contacts/ContactFilters/ContactFilters.tsx
frett: removing the type there causes a different not assignable type error on the calls to `isGroupVisible(group)` and `showGroup(group)`

src/components/Contacts/ContactFilters/ContactFilters.tsx
frett: this causes a not assignable type error for `isGroupVisible(g)`

frett: nm, this goes away once I use the `FilterGroup = ContactFiltersQuery['contactFilters']` type alias

src/components/Contacts/ContactFilters/ContactFilters.tsx
frett: this did get addressed once I setup the `FilterGroup` alias of the GraphQL type

OzzieOrca (approved):

PR #48, opened 2021-03-31
MPDX-6975 : Contact details   contact details tab base

gyasistory:
https://jira.cru.org/browse/MPDX-6975

This is just the base PR for the Contact Detail Tab.  After approval, I will have a separate PR completing each section.


src/components/Contacts/ContactDetail/ContactDetailTab/ContactDetailTab.stories.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export default {
+  title: 'Contacts/Tab/ContactDetailTab',
+  component: ContactDetailTab,
+};
+
+export const Default: Story = () => {
+  const contact = gqlMock<ContactDetailQuery>(ContactDetailDocument, {
+    mocks: {},
+  });
```
OzzieOrca, line 18:
```suggestion
  const contact = gqlMock<ContactDetailQuery, ContactDetailQueryVariables>(
    ContactDetailDocument,
    {
      variables: { accountListId: 'account-list-1', contactId: 'contact-1' },
    },
  );
```

src/theme.ts
```diff
@@ -83,6 +83,12 @@ const theme = createMuiTheme({
         body: {
           backgroundColor: '#05699b',
         },
+        h6: {
+          fontSize: '16px',
+          color: '#c4c4c4',
+          fontWeight: 600,
+          letterSpacing: '0.15px',
+        },
```
OzzieOrca, line 91: I think we should use the Typography component. https://material-ui.com/api/typography/ Maybe we could setup these styles in the theme if they don't match the defaults.

@j2trumpet you have any thoughts/want to help with this?

src/components/Contacts/ContactDetail/ContactDetailTab/ContactDetailTab.tsx
```diff
@@ -0,0 +1,87 @@
...
+            <Create className={classes.contactDetailHeadingIcon} />
+          </Box>
+        </Box>
+        <hr />
+        <Box
+          id="contact-detail-mailing"
+          className={classes.contactDetailContainer}
+        >
+          <Box className={classes.contactDetailHeadingContainer}>
+            <h6 className={classes.contactDetailHeadingText}>Mailing</h6>
```
TheNoodleMoose, line 68: This probably needs translation:

```suggestion
            <h6 className={classes.contactDetailHeadingText}>{t('Mailing')}</h6>
```

src/components/Contacts/ContactDetail/ContactDetailTab/ContactDetailTab.tsx
```diff
@@ -0,0 +1,87 @@
...
+}
+
+export const ContactDetailTab: React.FC<ContactDetailTabProps> = ({
+  contact,
+}) => {
+  const classes = useStyles();
+  const calledContact = contact.contact;
+
+  return (
+    <ThemeProvider theme={theme}>
```
OzzieOrca, line 44: This is provided globally for the app and Storybook. At this point tests may still need to include it.
https://github.com/CruGlobal/mpdx-react/blob/main/pages/_app.page.tsx#L85
https://github.com/CruGlobal/mpdx-react/blob/main/.storybook/preview.js#L31

src/components/Contacts/ContactDetail/ContactDetailTab/ContactDetailTab.tsx
```diff
@@ -0,0 +1,87 @@
...
+            <Create className={classes.contactDetailHeadingIcon} />
+          </Box>
+        </Box>
+        <hr />
+        <Box
+          id="contact-detail-other"
+          className={classes.contactDetailContainer}
+        >
+          <Box className={classes.contactDetailHeadingContainer}>
+            <h6 className={classes.contactDetailHeadingText}>Other</h6>
```
TheNoodleMoose, line 78: Translation here as well:
```suggestion
            <h6 className={classes.contactDetailHeadingText}>{t('Other')}</h6>
```

src/components/Contacts/ContactDetail/ContactDetailTab/ContactDetailTab.tsx
```diff
@@ -0,0 +1,87 @@
...
+  makeStyles,
+  Theme,
+  ThemeProvider,
+} from '@material-ui/core';
+import { Create } from '@material-ui/icons';
+import React from 'react';
+import theme from '../../../../theme';
+import { ContactDetailQuery } from '../ContactDetail.generated';
+
+const useStyles = makeStyles((theme: Theme) => ({
```
OzzieOrca, line 13: I would like to move toward the `styled` API as I think it's easier/cleaner to reuse on multiple components and styles are tied directly to a specific child component instead of shared across the whole component.
https://cru-main.slack.com/archives/C01P00MTREU/p1616802400006300

src/components/Contacts/ContactDetail/ContactDetailTab/ContactDetailTab.tsx
```diff
@@ -0,0 +1,87 @@
...
+          id="contact-detail-other"
+          className={classes.contactDetailContainer}
+        >
+          <Box className={classes.contactDetailHeadingContainer}>
+            <h6 className={classes.contactDetailHeadingText}>Other</h6>
+            <Create className={classes.contactDetailHeadingIcon} />
+          </Box>
+        </Box>
+        <hr />
+        <Button>DELETE CONTACT</Button>
```
TheNoodleMoose, line 83: And here too, although I'm not sure if it would be better to have it all caps, or lower case and then use something like ```.toUpperCase()``` to make it all caps.

```suggestion
        <Button>{t('DELETE CONTACT')}</Button>
```
or maybe
```suggestion
        <Button>{t('Delete Contact').toUpperCase()}</Button>
```

OzzieOrca, line 83: This needs to be translated.

Also stylistic capitalization is better to do with CSS's text-transform. It makes sure the translation also gets capitalized.

OzzieOrca (dismissed):

src/theme.ts
```diff
@@ -83,6 +83,12 @@ const theme = createMuiTheme({
         body: {
           backgroundColor: '#05699b',
         },
+        h6: {
```
TheNoodleMoose, line 86: Is this just a one-off style, or have you noticed a consistent pattern where we should override this in the theme? If we are using this everywhere then go for it. But if it's only in the ContactDetailsTab, maybe we should just make a custom style/class there?

gyasistory: @TheNoodleMoose this is the Heading 6 style in the Figma doc.

src/components/Contacts/ContactDetail/ContactDetailTab/ContactDetailTab.stories.tsx
gyasistory: Thanks, @OzzieOrca that is exactly what I was missing.

src/theme.ts
gyasistory: I created this to use the style that was created in Figma which I think is used in multiple places.  @j2trumpet I would love to know your thoughts on this.

MPDX_Update_Material_UI

src/components/Contacts/ContactDetail/ContactDetailTab/ContactDetailTab.tsx
gyasistory: Thanks, I forgot to remove this.

src/components/Contacts/ContactDetail/ContactDetailTab/ContactDetailTab.tsx
OzzieOrca: `.toUpperCase()` is how you do it in React Native since it doesn't support `text-transform: uppercase`. On web we can use CSS.

src/theme.ts
frett: using the typography component you do this:
```
<Typography variant="h6">Text Content</Typography>
```

it will then automatically pull the h6 styles from the MUI theme

frett: I don't know if this is or isn't the right spot to specify that, I would assume that it might inherit some of these settings automatically from the base MUI theme.

src/theme.ts
frett: I agree with @OzzieOrca , if we want to use the preset typography that MUI provides us, this is the wrong place to specify the "h6" styles

src/components/Contacts/ContactDetail/ContactDetailTab/ContactDetailTab.tsx
TheNoodleMoose: Yep, you're right. We've been doing it the RN way so long I forgot it could be done with CSS!

src/components/Contacts/ContactDetail/ContactDetailTab/ContactDetailTab.tsx
```diff
@@ -0,0 +1,98 @@
...
+  marginLeft: 'auto',
+  marginRight: 'auto',
+}));
+
+const ContactDetailHeadingText = styled(Typography)(() => ({
+  flexGrow: 5,
+}));
+
+interface ContactDetailTabProps {
+  contact: ContactDetailQuery;
```
OzzieOrca, line 41: What component is going to load this data? I feel like this component probably should. It could just take a `contactId` as a prop and query the GraphQL API.

gyasistory: @OzzieOrca I was looking at @reldredge71  PR for the whole contact Detail Page and it looks like his ContactDetail fragment could replace this Query.  What do you think?

OzzieOrca: Maybe @reldredge71  has thoughts based on how he's implementing the tabs.

It appears that tabs aren't rendered unless they are active so normal querying should work fine. https://github.com/mui-org/material-ui/blob/next/packages/material-ui-lab/src/TabPanel/TabPanel.js#L35

I think the contacts panel should have a query for the header/summary section and a query for each tab. Each of those components can be responsible for running their respective queries. I don't think there's any reason for these components to share data or move the query up to any parent components.

Since the tabs don't all load at once I don't think it makes sense to query this all at once. If it did we could use fragments but I'm not convinced we need one big query.

Guess there might be a few fields that would be loaded twice between the header and the contact details tab. I think it's fine to just fetch them twice, especially since the contact details isn't the default tab.

gyasistory: @OzzieOrca sounds good I will make the changes.

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.tsx
```diff
@@ -0,0 +1,103 @@
...
+  paddingTop: '4px',
+  margin: 0,
+  alignContent: 'start',
+}));
+
+const ContactDetailHeadingContainer = styled(Box)(() => ({
+  display: 'flex',
+  flexDirection: 'row',
+  alignContent: 'center',
+  alignItems: 'center',
```
TheNoodleMoose, line 18: Do you need both ```alignContent``` and ```alignItems``` here?

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.tsx
```diff
@@ -0,0 +1,103 @@
...
+const ContactDetailSectionContainer = styled(Box)(({ theme }) => ({
+  padding: theme.spacing(0),
+  paddingTop: '4px',
+  margin: 0,
+  alignContent: 'start',
+}));
+
+const ContactDetailHeadingContainer = styled(Box)(() => ({
+  display: 'flex',
+  flexDirection: 'row',
```
TheNoodleMoose, line 16: I don't think you need this, as the automatic direction when using ```flex``` is ```row```

TheNoodleMoose (changes requested):

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.test.tsx
```diff
@@ -0,0 +1,28 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { GqlMockedProvider } from '../../../../../__tests__/util/graphqlMocking';
+import { useContactDetailsTabQuery } from './ContactDetailsTab.generated';
+
+describe('ContactDetailTab', () => {
+  it('test query', async () => {
```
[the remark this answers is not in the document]
gyasistory: This was the way that @OzzieOrca suggested.  I am good with using either way to test.  I think the benefit with going this way is that you can auto generate the objects. which can be difficult to emulate if you don't know the layout of the data.

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.tsx
```diff
@@ -96,8 +98,8 @@ export const ContactDetailsTab: React.FC<ContactDetailTabProps> = ({
       </ContactDetailSectionContainer>
       <Divider />
       <ContactDeleteButton variant="outlined" color="default">
-        {t('DELETE CONTACT').toLocaleUpperCase()}
+        {t('delete contact').toLocaleUpperCase()}
```
TheNoodleMoose, line 101: Looks like you still have the ```.toLocaleUpperCase()```. Or did the ```textTransform: 'uppercase'``` not work?

TheNoodleMoose (approved):

PR #49, opened 2021-03-31
MPDX-6961: Improve theme

j2trumpet:
Added Cru and MPDX colors to the theme. Any values I just straight up deleted were either invalid, unnecessary, or duplicates from the default theme values. I still want to go through the files we've already got and swap out color values and references, but I wanted to get some eyes on what I've done so far first. I made each color its own key, thinking that the `light`, `dark`, and `contrastText` values would be automatically created, but those don't seem to be working at the moment.

### DIFF
--- a/src/components/Loading/Loading.tsx
+++ b/src/components/Loading/Loading.tsx
@@ -12,10 +12,10 @@ const useStyles = makeStyles((_theme: Theme) => ({
     marginTop: '-28px',
   },
   fab: {
-    backgroundColor: '#fff',
+    backgroundColor: _theme.palette.common.white,
     cursor: 'default',
     '&:hover': {
-      backgroundColor: '#fff',
+      backgroundColor: _theme.palette.common.white,
     },
   },
 }));
--- a/src/components/Loading/Loading.tsx
+++ b/src/components/Loading/Loading.tsx
@@ -3,7 +3,7 @@ import { makeStyles, Theme, Fab, CircularProgress } from '@material-ui/core';
 import { useRouter } from 'next/router';
 import { motion, AnimatePresence } from 'framer-motion';
 
-const useStyles = makeStyles((_theme: Theme) => ({
+const useStyles = makeStyles((theme: Theme) => ({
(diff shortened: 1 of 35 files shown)


src/theme.ts
```diff
@@ -2,36 +2,85 @@ import { createMuiTheme } from '@material-ui/core/styles';
...
+}
+
 const theme = createMuiTheme({
   typography: {
     fontFamily: "'Source Sans Pro', sans-serif",
   },
   palette: {
-    common: {
-      black: '#000',
-      white: '#fff',
```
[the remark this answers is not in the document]
OzzieOrca: Nvm the default theme solves these issues :)

OzzieOrca (approved):

OzzieOrca (approved):

src/components/Loading/Loading.tsx
```diff
@@ -12,10 +12,10 @@ const useStyles = makeStyles((_theme: Theme) => ({
     marginTop: '-28px',
   },
   fab: {
-    backgroundColor: '#fff',
+    backgroundColor: _theme.palette.common.white,
```
OzzieOrca, line 15: Can you rename `_theme` to `theme`? Before it was an unused variable which eslint would complain about. So someone prefixed it with an `_` to bypass that lint rule. Sometimes that is useful if you don't need a specific argument of a function or item of an array but want to access something that comes after it.

Now that you are using it again, we don't need to prefix it.

OzzieOrca (approved):
Looks good to me :)

You've got a couple of lint errors about unused vars that should be really easy to fix.

A bunch of tests are failing because they don't have the theme provider. I guess we need to go through and add the provider like we did for the Footer component. Let me know if you want help with figuring out that workflow or fixing up those tests.

src/components/Loading/Loading.tsx
j2trumpet: Will do. I didn't know why the underscore was there. I was just working with what was already provided.

PR #50, opened 2021-04-01
MPDX-6959: Setup ContactDetails Screen

reldredge71:
This is a bare bones setup of the ContactDetails panel that pops up when a user clicks on a contact row.


pages/accountLists/[accountListId]/contacts.page.tsx
```diff
@@ -1,28 +1,48 @@
...
 import { ContactsTable } from '../../../src/components/Contacts/ContactsTable/ContactsTable';
+import { ContactDetails } from '../../../src/components/Contacts/ContactDetails/ContactDetails';
 
 const ContactsPage: React.FC = () => {
   const { t } = useTranslation();
   const {
     query: { accountListId },
   } = useRouter();
 
+  const [contactDetailsId, setContactDetailsId] = useState<string | null>(null);
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/contacts.page.tsx:16]: I think using `undefined` as the default value is cleaner :)
```suggestion
  const [contactDetailsId, setContactDetailsId] = useState<string>();
```

pages/accountLists/[accountListId]/contacts.page.tsx
```diff
@@ -1,28 +1,48 @@
...
+  const [contactDetailsHidden, setContactDetailsHidden] = useState<boolean>(
+    true,
+  );
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/contacts.page.tsx:19]: Do we need to keep `contactDetailsId` around when it's hidden? Or could we just clear `contactDetailsId` instead and use a single state value for both of these usages?

The current version takes 2 clicks to switch from one contact to the next since it's toggling the hidden state every time.

pages/accountLists/[accountListId]/contacts.page.tsx
```diff
@@ -1,28 +1,48 @@
...
+  };
+
   return (
     <>
       <Head>
         <title>MPDX | {t('Contacts')}</title>
       </Head>
-      <Box height="100vh" display="flex">
-        <Box flex={1}>
+      <Box height="100vh" display="flex" overflow="hidden">
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/contacts.page.tsx:31]: Do we want the filters to scroll instead? Guess there's still work to do here later to make them sticky.
```suggestion
      <Box height="100vh" display="flex" overflow-y="scroll">
```

src/components/Contacts/ContactDetails/ContactDetails.graphql
```diff
@@ -0,0 +1,59 @@
+fragment ContactDetails on Contact {
+  id
+  ...ContactRow
```
OzzieOrca, line 3: What's your thought process here?

src/components/Contacts/ContactDetails/ContactDetails.graphql
```diff
@@ -0,0 +1,59 @@
...
+  avatar
+  churchName
```
OzzieOrca, line 5: What are you loading data for here? This looks like a lot of fields that might not be used everywhere. The header? Header and the first tab (Tasks)? I think the contact details tab should be lazy loaded. (To be clear I mean the contact details tab in the contact details page. Maybe we should name those 2 components differently somehow.)

Gyasi wrote a query for the tasks tab already https://github.com/CruGlobal/mpdx-react/pull/47

src/components/Contacts/ContactDetails/ContactDetails.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const [
+    loadContactDetails,
+    { data, loading },
+  ] = useGetContactDetailsLazyQuery();
+
+  useEffect(() => {
+    if (contactId != null) {
+      loadContactDetails({ variables: { accountListId, contactId } });
+    }
+  }, [contactId]);
```
OzzieOrca, line 27: I think the header should load data immediately. Depending on how the tabs work, they should lazy load if they render immediately but should load data immediately if they aren't rendered until navigated to.

`contactId` shouldn't be nullable but if was needed to be, `skip` could be used in this case. Tatai was using this pattern of lazy load with a `useEffect` instead of a `skip` in some places. I'm working on changing those in my ts-strict PR.

src/components/Contacts/ContactDetails/ContactDetails.tsx
```diff
@@ -0,0 +1,43 @@
+import { AppBar, Box, Tab, Tabs } from '@material-ui/core';
+import React, { useEffect } from 'react';
+import { useTranslation } from 'react-i18next';
+import { useGetContactDetailsLazyQuery } from './ContactDetails.generated';
+import { ContactDetailsHeader } from './ContactDetailsHeader/ContactDetailsHeader';
+
+interface Props {
+  accountListId: string;
+  contactId: string | null;
```
OzzieOrca, line 9: I don't think this should be nullable. The parent page should render this component if there is no contact to show.

src/components/Contacts/ContactRow.stories.tsx
```diff
@@ -18,7 +18,7 @@ export default {
 export const Default: Story = () => {
   const contact = gqlMock<ContactRowFragment>(ContactRowFragmentDoc);
 
-  return <ContactRow contact={contact} />;
+  return <ContactRow contact={contact} onContactSelected={() => {}} />;
```
OzzieOrca, line 21: This could be wired up to a story action. Not important though...

OzzieOrca (changes requested):
Looking good :) I think my biggest question is how you are thinking of firing off GraphQL queries. Does the main contact details component need to make any queries? Could the contact header and the tabs all load their own data? Since the default tab is Tasks (I'm assuming since it's the first tab), there might not be a whole lot of shared data between the header and the tasks tab.

pages/accountLists/[accountListId]/contacts.page.tsx
Comment (reldredge71) [reply pages/accountLists/[accountListId]/contacts.page.tsx:19]: Yeah that could be nice.  Hiding the panel could be controlled by whether or not `contactDetailsId` is defined or not.

pages/accountLists/[accountListId]/contacts.page.tsx
Comment (reldredge71) [reply pages/accountLists/[accountListId]/contacts.page.tsx:31]: I can try that... It seemed to work fine with `overflow="hidden"`, but I haven't tested it with a lot of screen sizes and scenarios.  We can keep trying things to get the experience that we want.

src/components/Contacts/ContactDetails/ContactDetails.graphql
reldredge71: I was creating this query to load the data needed for the contact details screen, at the time I didn't see that there were many tabs off of this screen, I was just going off of what I saw in Figma.  So basically this data is for the Contact Details header as well as the tab labeled "Contact Details".  I decided to keep all of it because I thought "Contact Details" would be the first tab.  Maybe it would be good to keep this simpler and only load data for the header for now.

I decided to expand the ContactRow here because pretty much all of the properties on the ContactRow are also needed for this query as well.

src/components/Contacts/ContactDetails/ContactDetails.graphql
reldredge71: I didn't realize Gyasi made a query for the Tasks Tab, We could probably create separate data for each tab and make the query only when the tab is opened.

src/components/Contacts/ContactDetails/ContactDetails.tsx
reldredge71: Yeah the way I'm handling the drawer opening/closing is that the the ContactDetails is rendered along with the whole Contact screen, but is controlled by `hidden`.  When the user first launches the screen, there is no contactId passed to ContactDetails, and I don't want to call the query until we have a contactId to query by, so a lazy query seemed like the best idea.  Later on, the user clicks on a row, the contactId is passed to the ContactDetails that is already present, the `useEffect` detects a change of value and launches the query.  It worked pretty well when I was testing it, opening and closing the ContactDetails for different rows.  But I'm not too attached to this implementation, if you know of a better way I'm open to trying it.

src/components/Contacts/ContactRow.stories.tsx
reldredge71: Yeah maybe, although the ContactRow doesn't change at all when the row is selected, so I don't think wiring this to a story action would change how this appears in Storybook.

src/components/Contacts/ContactDetails/ContactDetails.tsx
OzzieOrca: I think it would be better to not render the details panel at all when it's not shown instead of using hidden to control it. Then you can just query as normal in the child component when it mounts.

src/components/Contacts/ContactRow.stories.tsx
TheNoodleMoose: I think the story action is less focused on how it appears in the storybook and more so on if logic is called when expected. Wiring it up to the action could test that the callback function is fired when expected inside the component.

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.tsx
```diff
@@ -1,21 +1,29 @@
...
+  const { data, loading } = useGetContactDetailsHeaderQuery({
+    variables: { accountListId, contactId },
+  });
+
   return (
     <Box>
       {/*TODO: Build Header*/}
-      {loading ? <p>loading</p> : <p role="contactName">{contact?.name}</p>}
+      {loading ? (
+        <p>loading</p>
```
TheNoodleMoose, line 23: I assume this loading text is just placeholder? I've been using the ```Skeleton``` component for loading states and it visually works very well. Here's an example: https://github.com/CruGlobal/mpdx-react/blob/54d75ad1c29cc1a2f6670de4200be234d2b503f8/src/components/Dashboard/DonationHistories/DonationHistories.tsx#L172-L176

src/components/Contacts/ContactRow.tsx
```diff
@@ -31,22 +31,30 @@ const useStyles = makeStyles((theme: Theme) => ({
...
           height: '72px',
+          width: '100%',
           display: 'flex',
           flexDirection: 'row',
           justifyContent: 'space-evenly',
           alignItems: 'center',
           alignContent: 'center',
         }}
+        onClick={onClick}
       >
```
TheNoodleMoose, line 58: What do you think about moving these styles out and using the ```styled``` API instead?
```
const  ContactRowButton = styled(Box)(({}) => ({
      height: '72px',
      width: '100%',
      display: 'flex',
      flexDirection: 'row',
      justifyContent: 'space-evenly',
      alignItems: 'center',
      alignContent: 'center',
}))

```

and then you would just have

```suggestion
<ContactRowButton
    role="rowButton"
    onClick={onClick}
>
```

I think the fewer inline styles the better as it makes the component tree a little bit more readable.

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+          contactId={contactId}
+        />
+      </GqlMockedProvider>,
+    );
+
+    await waitFor(async () =>
+      expect((await findAllByRole('contactName'))[0]).toBeInTheDocument(),
+    );
+
+    expect(queryAllByText('loading')).toEqual([]);
```
TheNoodleMoose, line 38: Could you also do?
```suggestion
    expect(queryByText('loading')).toBeNull()
```

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { GetContactDetailsHeaderQuery } from './ContactDetailsHeader.generated';
+
+export default {
+  title: 'Contacts/ContactDetails/Header',
+};
+
+const accountListId = 'abc';
+const contactId = 'contact-1';
+
+export const Default = (): ReactElement => {
```
TheNoodleMoose, line 15: Should we have a storybook for a loading state?

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.tsx
```diff
@@ -0,0 +1,29 @@
...
+  accountListId: string;
+  contactId: string;
+}
+
+export const ContactDetailsHeader: React.FC<Props> = ({
+  accountListId,
+  contactId,
+}: Props) => {
+  const { data, loading } = useGetContactDetailsHeaderQuery({
+    variables: { accountListId, contactId },
```
TheNoodleMoose, line 16: What should we do about errors for this component? It would be great if we could have a consistent pattern for handling and displaying errors. I've used ``` const { enqueueSnackbar } = useSnackbar();``` and it seems to work really well but not sure if everyone else is on board with it.

src/components/Contacts/ContactDetails/ContactDetails.tsx
```diff
@@ -0,0 +1,35 @@
+import { AppBar, Box, Tab, Tabs } from '@material-ui/core';
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { ContactDetailsHeader } from './ContactDetailsHeader/ContactDetailsHeader';
+
+interface Props {
+  accountListId: string;
+  contactId: string;
+  onClose: () => void;
```
TheNoodleMoose, line 9: Is this ```onClose``` prop needed at the moment? It doesn't appear to be used in this component. Or will it be used in the future?

src/components/Contacts/ContactDetails/ContactDetails.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+          onClose={onClose}
+        />
+      </GqlMockedProvider>,
+    );
+
+    await waitFor(async () =>
+      expect((await findAllByRole('contactName'))[0]).toBeInTheDocument(),
+    );
+
+    expect(queryAllByText('loading')).toEqual([]);
```
TheNoodleMoose, line 45: A similar question about this.

pages/accountLists/[accountListId]/contacts.page.tsx
```diff
@@ -1,29 +1,44 @@
...
-        <Box flex={4}>
-          <ContactsTable accountListId={accountListId as string} />
+        <Box flex={1}>
+          <ContactsTable
+            accountListId={accountListId as string}
+            onContactSelected={setContactDetailsId}
+          />
         </Box>
+        {contactDetailsId ? (
+          <Box flex={1} hidden={!contactDetailsId}>
```
Comment (TheNoodleMoose) [inline pages/accountLists/[accountListId]/contacts.page.tsx:34]: Do you need the ```hidden``` if it doesn't get rendered unless ```contactDetailsId``` is true?

pages/accountLists/[accountListId]/contacts.page.tsx
```diff
@@ -1,29 +1,44 @@
...
-        <Box flex={1}>
+      <Box height="100vh" display="flex" overflow-y="scroll">
+        <Box width="20vw">
           <ContactFilters accountListId={accountListId as string} />
         </Box>
-        <Box flex={4}>
-          <ContactsTable accountListId={accountListId as string} />
+        <Box flex={1}>
+          <ContactsTable
+            accountListId={accountListId as string}
```
Comment (TheNoodleMoose) [inline pages/accountLists/[accountListId]/contacts.page.tsx:29]: What is ```accountListId``` typed as?

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.tsx
frett: Material UI does provide a Indeterminate Progress Indicator that might be useful to use for loading UIs.

https://material-ui.com/components/progress/#circular-indeterminate

reldredge71: Thanks!  I might use Skeleton for now, since we've used that in other places already.  But yes, this is just a placeholder, I'm actually already working on the header in another branch.  Maybe I can leave this as it is for now, and create a proper loading state in that other branch?

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.stories.tsx
reldredge71: In a previous PR, I found out how to mock a loading state using the `MockedProvider`, but I haven't found a good way to do it with `GQLMockedProvider` yet.  I need some way to keep the query from finishing in order for Storybook to keep the loading state visible.

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.tsx
reldredge71: I'm good with Snackbar, we should discuss if we're good with keeping that or if we should find something else.

src/components/Contacts/ContactDetails/ContactDetails.tsx
reldredge71: Yeah, I will build out the ContactDetailsHeader in another PR.  I added this anticipating that it will be used soon -- there needs to be a way to close the details panel after it's opened.

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.stories.tsx
TheNoodleMoose: Daniel and I had a discussion earlier today and we agreed we could not find a good way to mock the loading state using ```GQLMockProvider``` and instead decided on using ```MockProvider``` and a delay to mock loading states. Here how I handled it in the NewsletterMenu story: https://github.com/CruGlobal/mpdx-react/pull/51/files#diff-8f40b0ad90892d047b8f9edf7b387e7ac1c7af607e31b1572f52b43b3316e2adR37

pages/accountLists/[accountListId]/contacts.page.tsx
Comment (reldredge71) [reply pages/accountLists/[accountListId]/contacts.page.tsx:34]: Oh, I forgot to remove that from the previous way I implemented this.

pages/accountLists/[accountListId]/contacts.page.tsx
Comment (reldredge71) [reply pages/accountLists/[accountListId]/contacts.page.tsx:29]: It says `Namespace<string>`

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.tsx
TheNoodleMoose: Doing it in another branch is fine with me 🙂

src/components/Contacts/ContactRow.stories.tsx
TheNoodleMoose: But like Scotty mentioned, not that important

src/components/Contacts/ContactRow.tsx
```diff
@@ -8,13 +8,29 @@ import { GiftStatus } from './GiftStatus/GiftStatus';
 import { StarContactIcon } from './StarContactIcon/StarContactIcon';
 
 const useStyles = makeStyles((theme: Theme) => ({
+  contactRowButton: {
```
TheNoodleMoose, line 11: Thanks for cleaning this up 🙂 Do you think in your next pr we should migrate this away from ```makeStyles``` in favor of the ```styled``` API?

pages/accountLists/[accountListId]/contacts.page.tsx
```diff
@@ -22,7 +22,7 @@ const ContactsPage: React.FC = () => {
       </Head>
       <Box height="100vh" display="flex" overflow-y="scroll">
         <Box width="20vw">
-          <ContactFilters accountListId={accountListId as string} />
+          <ContactFilters accountListId={accountListId.toString()} />
```
Comment (TheNoodleMoose) [inline pages/accountLists/[accountListId]/contacts.page.tsx:25]: It's strange that this is the only place where ```accountListId``` isn't typed as a string. Not sure which method of typing this I prefer... what made you decide in favor ```.toString()``` if I may ask?

Comment (reldredge71) [reply pages/accountLists/[accountListId]/contacts.page.tsx:25]: Here, accountListId is coming from the router, and I guess the type of accountListId from that source is `Namespace<string>`.  Here's another page in the app that is fetching accountListId from the router, and is converted from `Namespace` to `string`: https://github.com/CruGlobal/mpdx-react/blob/main/pages/accountLists/%5BaccountListId%5D/tasks.page.tsx#L77

I really don't know what the best way to handle this is, I thought this might be a more natural way to convert it than forcefully casting it.

src/components/Contacts/ContactRow.tsx
reldredge71: I wasn't sure if `styled` would work for the Material UI icon in this component... would it work?  If so then I have no problem moving to `styled`.

TheNoodleMoose: Hmm, I haven't specifically tested it on a material UI icon, but I would say it should work 🙂

TheNoodleMoose:
Looks like vercel is having some issues... https://www.vercel-status.com/

gyasistory:
@reldredge71 if you merge in the main branch you can add the ContactDetailTab to your tabs.

TheNoodleMoose:
@reldredge71 It looks like the ```.toString()``` is throwing an error on the build. Although the ```.toString()``` makes more sense, I think casting the type via ```as string``` is safer.

Screen Shot 2021-04-07 at 11 45 43 AM

TheNoodleMoose (approved):

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.tsx
OzzieOrca: Probably a merged PR isn't the best place for this discussion but sticking it here anyway...

I think we could handle errors globally with an Apollo link like https://github.com/CruGlobal/missionhub-react-native/blob/develop/src/apolloClient.ts#L27-L57.

Since we're using the snackbars which just display over everything and we aren't displaying anything at the component level, we probably don't need to make the individual components do anything.

Guess we should make a Jira ticket if we want to go this route.

Or maybe design has thoughts... The UX could be better if we put the error message right where it occurred. But it's probably more work to do that on a per component basis.

What should the component do if it gets an error and no data? Try and stay in the loading state? Render nothing? I guess we'll already have to be handling any fields that aren't defined.

pages/accountLists/[accountListId]/contacts.page.tsx
Comment (OzzieOrca) [reply pages/accountLists/[accountListId]/contacts.page.tsx]: VSCode says it's a `string | string[] | undefined` for me. It's coming from the `querystring` module.
```ts
interface ParsedUrlQuery extends NodeJS.Dict<string | string[]> { }
```
Seems like casting to a string makes sense since I don't see any way of telling the router what the types of its variables are. Maybe we should make a hook for this.

PR #51, opened 2021-04-05
MPDX-6943-Add-Newsletter-Dropdown-Functionality

TheNoodleMoose:
So this ticket adds functionality to the newsletter menu, but to a certain degree. This menu has three dropdown options:
1. Log Newsletter
2. Export Email
3. Export Physical

Only "Export Email" has full functionality and design in this PR, and I plan to create separate tickets/prs to address the other two. I did this as I didn't want the pr to get too large. "Log Newsletter" and "Export Physical" are just the barebones for those dialogs and tests.  Let me know what you think and if you have any questions 🙂

### DIFF
--- a/pages/api/graphql-rest.page.ts
+++ b/pages/api/graphql-rest.page.ts
@@ -12,6 +12,26 @@ import { Resolvers } from './graphql-rest.page.generated';
 const typeDefs = gql`
   type Query {
     contactFilters(accountListId: ID!): [ContactFilter!]!
+    taskAnalytics(accountListId: ID!): TaskAnalytics!
+  }
+
+  type TaskAnalytics {
+    id: ID!
+    type: String!
+    createdAt: ISO8601DateTime!
+    lastElectronicNewsletterCompletedAt: ISO8601DateTime
+    lastPhysicalNewsletterCompletedAt: ISO8601DateTime
+    tasksOverdueOrDueTodayCounts: [OverdueOrDueTodayTaskAnalytic!]!
+    totalTasksDueCount: Int!
+    updatedAt: ISO8601DateTime!
+    updatedInDbAt: ISO8601DateTime!
+  }
+
+  scalar ISO8601DateTime
+
+  type OverdueOrDueTodayTaskAnalytic {
+    label: String!
+    count: Int!
   }
 
   type ContactFilter {
@@ -38,6 +58,9 @@ const resolvers: Resolvers = {
   Query: {
     contactFilters: async (_source, { accountListId }, { dataSources }) => {
       return dataSources.mpdxRestApi.getContactFilters(accountListId);
+    },
+    taskAnalytics: async (_source, { accountListId }, { dataSources }) => {
+      return dataSources.mpdxRestApi.getTaskAnalytics(accountListId);
     },
   },
 };
@@ -94,6 +117,35 @@ class MpdxRestApi extends RESTDataSource {
             : [default_selection],
       }),
     );
+  }
+
+  async getTaskAnalytics(accountListId) {
+    const { data } = await this.get(
+      `tasks/analytics?filter[account_list_id]=${accountListId}`,
+    );
+
+    const {
+      attributes: {
+        created_at,
+        last_electronic_newsletter_completed_at,
+        last_physical_newsletter_completed_at,
+        tasks_overdue_or_due_today_counts,
+        total_tasks_due_count,
+        updated_at,
+        updated_in_db_at,
+      },
+    } = data;
+
+    return {
+      ...data,
+      createdAt: created_at,
+      lastElectronicNewsletterCompletedAt: last_electronic_newsletter_completed_at,
+      lastPhysicalNewsletterCompletedAt: last_physical_newsletter_completed_at,
+      tasksOverdueOrDueTodayCounts: tasks_overdue_or_due_today_counts,
+      totalTasksDueCount: total_tasks_due_count,
+      updatedAt: updated_at,
+      updatedInDbAt: updated_in_db_at,
+    };
   }
 }
 
--- a/src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.stories.tsx
+++ b/src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.stories.tsx
@@ -3,11 +3,13 @@ import { GraphQLError } from 'graphql';
 import React, { ReactElement } from 'react';
 import { GqlMockedProvider } from '../../../../../__tests__/util/graphqlMocking';
 import {
-  GetEmailNewsletterContactsDocument,
-  GetEmailNewsletterContactsQuery,
   GetTaskAnalyticsDocument,
   GetTaskAnalyticsQuery,
-} from '../NewsletterMenu.generated';
+} from './NewsletterMenu.generated';
+import {
+  GetEmailNewsletterContactsQuery,
+  GetEmailNewsletterContactsDocument,
+} from './MenuItems/ExportEmail/GetNewsletterContacts.generated';
 import NewsletterMenu from './NewsletterMenu';
 
 export default {
--- a/src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.stories.tsx
+++ b/src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.stories.tsx
@@ -1,0 +1,104 @@
+import React, { ReactElement } from 'react';
+import { GqlMockedProvider } from '../../../../../__tests__/util/graphqlMocking';
+import {
+  GetEmailNewsletterContactsQuery,
+  GetTaskAnalyticsQuery,
+} from '../NewsletterMenu.generated';
+import NewsletterMenu from './NewsletterMenu';
+
+export default {
+  title: 'Dashboard/ThisWeek/NewsletterMenu',
+};
+
+const accountListId = '111';
+
+export const Default = (): ReactElement => {
+  const mocks = {
+    GetTaskAnalytics: {
+      taskAnalytics: {
+        lastElectronicNewsletterCompletedAt: '2020-10-27T16:20:06Z',
+        lastPhysicalNewsletterCompletedAt: '2020-11-11T19:42:03Z',
+      },
+    },
+  };
+  return (
+    <GqlMockedProvider<GetTaskAnalyticsQuery & GetEmailNewsletterContactsQuery>
+      mocks={mocks}
+    >
+      <NewsletterMenu accountListId={accountListId} />
+    </GqlMockedProvider>
+  );
+};
+
+export const Loading = (): ReactElement => {
+  const mocks = {
+    GetEmailNewsletterContacts: {
+      contacts: {
+        nodes: [],
+      },
+    },
+
+    GetTaskAnalytics: {
+      taskAnalytics: {
+        lastElectronicNewsletterCompletedAt: undefined,
+        lastPhysicalNewsletterCompletedAt: undefined,
+      },
+    },
+  };
+  return (
+    <GqlMockedProvider<GetTaskAnalyticsQuery & GetEmailNewsletterContactsQuery>
+      mocks={mocks}
+    >
+      <NewsletterMenu accountListId={accountListId} />
+    </GqlMockedProvider>
+  );
+};
+
+export const Empty = (): ReactElement => {
+  const mocks = {
+    GetEmailNewsletterContacts: {
+      contacts: {
+        nodes: [],
+      },
+    },
+
+    GetTaskAnalytics: {
+      taskAnalytics: {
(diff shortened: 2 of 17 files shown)


pages/api/graphql-rest.page.ts
```diff
@@ -12,6 +12,26 @@ import { Resolvers } from './graphql-rest.page.generated';
...
+    createdAt: ISO8601DateTime!
+    lastElectronicNewsletterCompletedAt: ISO8601DateTime
+    lastPhysicalNewsletterCompletedAt: ISO8601DateTime
+    tasksOverdueOrDueTodayCounts: [OverdueOrDueTodayTaskAnalytic!]!
+    totalTasksDueCount: Int!
+    updatedAt: ISO8601DateTime!
+    updatedInDbAt: ISO8601DateTime!
+  }
+
+  scalar ISO8601DateTime
```
TheNoodleMoose, line 30: Was super happy to figure this out. I created a custom scalar here and then use it up in my `TaskAnalytics` type. On codegen, https://github.com/CruGlobal/mpdx-react/blob/7554b164b815d4fc08d60695e13d2c7cdea68524/codegen-server.yml#L17-L18 sees this scalar and will generate the correct type for it.

pages/api/graphql-rest.page.ts
```diff
@@ -39,6 +59,9 @@ const resolvers: Resolvers = {
     contactFilters: async (_source, { accountListId }, { dataSources }) => {
       return dataSources.mpdxRestApi.getContactFilters(accountListId);
     },
+    taskAnalytics: async (_source, { accountListId }, { dataSources }) => {
+      return dataSources.mpdxRestApi.getTaskAnalytics(accountListId);
+    },
```
TheNoodleMoose, line 64: Once this is merged in, I'd also like to take a crack at trying to modularize this and break these queries and types up into different files.

src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+};
+
+const accountListId = '111';
+
+export const Default = (): ReactElement => {
+  const mocks = {
+    GetTaskAnalytics: {
+      taskAnalytics: {
+        lastElectronicNewsletterCompletedAt: '2020-10-27T16:20:06Z',
+        lastPhysicalNewsletterCompletedAt: '2020-11-11T19:42:03Z',
```
TheNoodleMoose, line 20: For some reason, ergonomock was always returning null, even though the type should be a ```string```. I think it may have something to do with it being a custom scalar, but not too sure. To fix it, I just had to mock the specific dates I wanted.

src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+    );
+
+    userEvent.click(queryByTestId('NewsletterMenuButton'));
+
+    expect(queryByText('Log Newsletter')).toBeInTheDocument();
+    expect(queryByText('Export Email')).toBeInTheDocument();
+    expect(queryByText('Export Physical')).toBeInTheDocument();
+  });
+
+  describe('Log Newsletter', () => {
```
TheNoodleMoose, line 45: As I mentioned, I created super barebones tests for the two menu options that I didn't fully fledge out in this pr.

src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+      expect(queryByTestId('emailList')).toHaveValue(`${email1},${email2}`);
+    });
+  });
+
+  it('should copy email list to clipboard', async () => {
+    Object.assign(navigator, {
+      clipboard: {
+        writeText: () => {},
+      },
+    });
```
TheNoodleMoose, line 115: I had to do some googling to figure out how to test copying text to the clipboard. This is what I based this on: https://stackoverflow.com/questions/62351935/how-to-mock-navigator-clipboard-writetext-in-jest

src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.tsx
```diff
@@ -0,0 +1,303 @@
...
+import {
+  useGetTaskAnalyticsQuery,
+  useGetEmailNewsletterContactsLazyQuery,
+} from '../NewsletterMenu.generated';
+
+interface Props {
+  accountListId: string;
+}
+
+const CloseButton = styled(IconButton)(({ theme }) => ({
```
TheNoodleMoose, line 30: I really prefer the ```styled``` API over what we were doing before. I may be biased though as it's very similar to what I always use in https://styled-components.com/ or https://emotion.sh/docs/@emotion/styled 🙂

src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.tsx
```diff
@@ -0,0 +1,303 @@
...
+    ) {
+      return physicalDate.toLocaleString();
+    } else {
+      return electronicDate < physicalDate
+        ? physicalDate.toLocaleString()
+        : electronicDate.toLocaleString();
+    }
+  };
+  // TODO: Finish creating dialog for "Log Newsletter" and "Export Physical"
+  const renderDialog = () => {
```
TheNoodleMoose, line 153: Also very important feedback I'd love to hear, how do you feel about how I approached rendering these dialogs? Could it be done better? I didn't want to create a separate component for each of these menu items so this seemed like the best way to handle it. Let me know what you guys think 🙂

pages/api/graphql-rest.page.ts
frett: just a heads up that I've been modifying this file as well to tweak the data that is returned for contact filters

TheNoodleMoose: Okay, thanks for the heads up. I'll wait then until it's in a good place to be modified.

src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.tsx
```diff
@@ -0,0 +1,303 @@
...
+
+interface Props {
+  accountListId: string;
+}
+
+const CloseButton = styled(IconButton)(({ theme }) => ({
+  position: 'absolute',
+  right: theme.spacing(1),
+  top: theme.spacing(1),
+  color: theme.palette.grey[500],
```
j2trumpet, line 34: @TheNoodleMoose I'd change this gray tone to `theme.palette.text.primary`. I'm not sure where it's defined, but the light gray tone used for the background circle on hover should be `#EBECEC`. That light gray shade isn't available in the theme yet, but if you hard code it, I'll fix it as I'm working on the theme overhaul.

src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.tsx
```diff
@@ -0,0 +1,303 @@
...
+}));
+
+const NewsletterTextContainer = styled(ListItemText)(({}) => ({
+  display: 'flex',
+  flexDirection: 'column',
+  alignItems: 'flex-start',
```
j2trumpet, line 40: @TheNoodleMoose I don't think you need these three style rules. Turning them off in local testing and adding `textAlign: 'left'` achieves the same look.

src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.stories.tsx
```diff
@@ -0,0 +1,111 @@
+import { MockedProvider, MockedResponse } from '@apollo/client/testing';
```
gyasistory, line 1:
```suggestion
import { MockedProvider } from '@apollo/client/testing';
```

gyasistory (changes requested):

src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.test.tsx
```diff
@@ -197,6 +225,9 @@ describe('NewsletterMenu', () => {
                   },
                 },
               },
+              {
+                primaryPerson: null,
+              },
```
TheNoodleMoose, line 230: I added this to show that the logic will filter out a value from the email list string if the person does not have an ```email```.

TheNoodleMoose:
I've decided to go ahead and move the different menu items into a new folder and be contained to their own components.

src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.tsx
```diff
@@ -0,0 +1,304 @@
...
+  });
+
+  const [
+    loadEmailNewsletterContacts,
+    { data: newsletterContactsData, loading: contactsLoading },
+  ] = useGetEmailNewsletterContactsLazyQuery({
+    variables: { accountListId },
+  });
+
+  useEffect(() => {
```
OzzieOrca, line 82: Use `useMemo` for this instead of `useState` and `useEffect`.

src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.tsx
```diff
@@ -0,0 +1,304 @@
...
+    if (newsletterContactsData?.contacts.nodes.length > 0) {
+      changeEmailList(
+        newsletterContactsData.contacts.nodes.reduce(
```
OzzieOrca, line 85: a `.map` to get an array of emails and a `.join(',')` might be cleaner than the reduce you have. Might have to throw a `.filter(Boolean)` in there if you need null checks. 🤷‍♂️

src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.tsx
```diff
@@ -0,0 +1,304 @@
...
+
+const NewsletterMenuDialogTitle = styled(DialogTitle)(({}) => ({
+  textTransform: 'uppercase',
+  textAlign: 'center',
+}));
+
+const NewsletterMenu = ({ accountListId }: Props): ReactElement<Props> => {
+  const { t } = useTranslation();
+
+  const [anchorEl, setAnchorEl] = useState(null);
```
OzzieOrca, line 56: I think using undefined is cleaner `useState<SomeNonNullType>()`. I've been cleaning up some of these cuz TS strict wants a type here.

src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.tsx
OzzieOrca: Instead of several different switches, might be better to group the title, content, action etc into an array of objects... 🤷‍♂️

src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.tsx
TheNoodleMoose: Yep, I like that better 🙂

src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.tsx
TheNoodleMoose: Good idea, been a while since I've used ```usedMemo```!

src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.tsx
```diff
@@ -150,105 +90,41 @@ const NewsletterMenu = ({ accountListId }: Props): ReactElement<Props> => {
...
-            </Button>
-          );
-        case 2:
-          return null;
-      }
-    };
     return (
       <Dialog
         open={newsletterMenuDialogOpen}
         aria-labelledby={t('')}
```
OzzieOrca, line 123: This seems bad...

OzzieOrca (approved):

src/components/Dashboard/ThisWeek/NewsletterMenu/NewsletterMenu.tsx
TheNoodleMoose: Oh I missed that, thanks 👍

OzzieOrca (approved):

TheNoodleMoose:
@OzzieOrca Looks like there was a type change on the API. Should I merge this in and we can address the type issue in a separate pr or should I wait until that fix is merged in? The type is unrelated to my work here.

OzzieOrca:
It's unrelated, let's just fix it in a new PR.

PR #52, opened 2021-04-05
Mpdx 6975 add tag chips to contact detail tab

gyasistory:
### Contact Detail Tab - Tags

https://jira.cru.org/browse/MPDX-6975

Add Contact Tags to the Contact Detail Tab

_Relies on_ 
- [x] #48


src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+};
+
+export const Default = (): ReactElement => {
+  return (
+    <Box m={2}>
+      <MuiThemeProvider theme={theme}>
+        <CssBaseline />
+        <ContactTags
+          contactId="contact_id"
+          contactTags={array('contactTags', ['help', 'something'])}
```
TheNoodleMoose, line 19: Good idea to use the array addon here, it's neat to be able to edit the values dynamically.

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
```diff
@@ -0,0 +1,58 @@
...
+    console.info('You clicked the Delete tag for tag' + tag + '' + contactId);
+  };
+  const handleAddTag = () => {
+    // TODO: add mutation to update tag for contact
+    console.info('You clicked add contact Tag to contact' + contactId);
+  };
+
+  return (
+    <ContactTagsContainer>
+      {contactTags.map((tag, _index) => (
```
TheNoodleMoose, line 48: You don't need the ```_index``` if you're not using it. The only question would be, is there a chance there could be two tags with the same value? If so maybe using the ```index``` would be better.

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <ContactTagChip key={tag} label={tag} onDelete={handleTagDelete} />
+      ))}
+      <Link color="textPrimary" onClick={handleAddTag}>
```
TheNoodleMoose, line 51: Looking at the design in figma, I think the color needs to be ```#9C9FA1``` while this is currently ```#383F43```.

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.stories.tsx
```diff
@@ -17,9 +19,15 @@ const contactId = '222';
 
 export const Default = (): ReactElement => {
   return (
-    <GqlMockedProvider<ContactDetailsTabQuery>>
-      <ContactDetailsTab accountListId={accountListId} contactId={contactId} />
-    </GqlMockedProvider>
+    <MuiThemeProvider theme={theme}>
```
TheNoodleMoose, line 22: I haven't done this yet, but is this needed to provide the theme to the child components in storybook? What happens if you don't provide this, does it just use the default theme values?

TheNoodleMoose (dismissed):
Looks good, just a few minor things

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.stories.tsx
gyasistory: @TheNoodleMoose yes if you don't use it you will not get the theme that we create.

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
gyasistory: @TheNoodleMoose logically the tag should not be repeated that wouldn't make since.  I think if we handle duplication on the creation of new tags that should be enough.  Also this is logic that we are going to have to comeback to since we don't have the mutation call for contact set up yet.

TheNoodleMoose: I went to mpdx.org and tested trying to add two of the same tag. It allows you to submit it but it doesn't duplicate the tag so we should be good.

TheNoodleMoose (approved):

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.stories.tsx
frett: just a follow up to this, this is already defined in `.storybook/preview.js`, there shouldn't be a need to specify it for each story.

PR #53, opened 2021-04-05
MPDX-6975 : Other contact detail tab section

gyasistory:
### Contact Details Tab - Other Section

Relies on
- [x] #48

### DIFF
--- a/src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.graphql
+++ b/src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.graphql
@@ -45,5 +45,6 @@ query ContactDetailsTab($accountListId: ID!, $contactId: ID!) {
     sendNewsletter
     greeting
     #TODO add Magazine
+    ...ContactOther
   }
 }
--- a/src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.graphql
+++ b/src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.graphql
@@ -1,6 +1,7 @@
 query ContactDetailsTab($accountListId: ID!, $contactId: ID!) {
   contact(accountListId: $accountListId, id: $contactId) {
     id
+    locale
     name
     primaryPerson {
       id
@@ -13,6 +14,7 @@ query ContactDetailsTab($accountListId: ID!, $contactId: ID!) {
       country
     }
     tagList
+    churchName
     envelopeGreeting
     people {
       nodes {
--- a/src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.graphql
+++ b/src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.graphql
@@ -1,7 +1,6 @@
 query ContactDetailsTab($accountListId: ID!, $contactId: ID!) {
   contact(accountListId: $accountListId, id: $contactId) {
     id
-    locale
     name
     primaryPerson {
       id
@@ -14,7 +13,6 @@ query ContactDetailsTab($accountListId: ID!, $contactId: ID!) {
       country
     }
     tagList
-    churchName
     envelopeGreeting
     people {
       nodes {
@@ -35,8 +33,6 @@ query ContactDetailsTab($accountListId: ID!, $contactId: ID!) {
         birthdayYear
       }
     }
-    preferredContactMethod
-    timezone
     user {
       firstName
       lastName
--- a/src/components/Contacts/ContactDetails/ContactDetailsTab/Other/ContactDetailsOther.stories.tsx
+++ b/src/components/Contacts/ContactDetails/ContactDetailsTab/Other/ContactDetailsOther.stories.tsx
@@ -1,0 +1,15 @@
+import { Box } from '@material-ui/core';
+import React, { ReactElement } from 'react';
+import { ContactDetailsOther } from './ContactDetailsOther';
+
+export default {
+  title: 'Contacts/Tab/ContactDetailsTab/Other',
+};
+
+export const Default = (): ReactElement => {
+  return (
+    <Box m={2}>
+      <ContactDetailsOther contact={null} />
+    </Box>
+  );
+};
--- a/src/components/Contacts/ContactDetails/ContactDetailsTab/Other/ContactDetailsOther.stories.tsx
+++ b/src/components/Contacts/ContactDetails/ContactDetailsTab/Other/ContactDetailsOther.stories.tsx
@@ -5,6 +5,7 @@ import theme from '../../../../../theme';
 import {
   ContactDetailsTabDocument,
   ContactDetailsTabQuery,
+  ContactDetailsTabQueryVariables,
 } from '../ContactDetailsTab.generated';
 import { ContactDetailsOther } from './ContactDetailsOther';
 
@@ -17,12 +18,15 @@ const accountListId = '111';
 const contactId = '222';
 
 export const Default = (): ReactElement => {
-  const mock = gqlMock<ContactDetailsTabQuery>(ContactDetailsTabDocument, {
-    variables: {
-      accountListId: accountListId,
-      contactId: contactId,
+  const mock = gqlMock<ContactDetailsTabQuery, ContactDetailsTabQueryVariables>(
+    ContactDetailsTabDocument,
+    {
+      variables: {
+        accountListId: accountListId,
+        contactId: contactId,
+      },
     },
-  });
+  );
   return (
     <Box m={2}>
       <MuiThemeProvider theme={theme}>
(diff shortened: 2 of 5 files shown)


src/components/Contacts/ContactDetails/ContactDetailsTab/Other/ContactDetailsOther.stories.tsx
```diff
@@ -0,0 +1,34 @@
...
+  component: ContactDetailsOther,
+};
+
+const accountListId = '111';
+const contactId = '222';
+
+export const Default = (): ReactElement => {
+  const mock = gqlMock<ContactDetailsTabQuery>(ContactDetailsTabDocument, {
+    variables: {
+      accountListId: accountListId,
```
gyasistory, line 22: Not sure why this is failing. the check.

TheNoodleMoose: Looked into this and turns out the default type for the variables is ```never```. See https://github.com/CruGlobal/mpdx-react/blob/9c2102b7d4a45216fd074408b09aaf31815254ae/__tests__/util/graphqlMocking.tsx#L133
So to fix this, I just included the generated types for the query variables.

src/components/Contacts/ContactDetails/ContactDetailsTab/Other/ContactDetailsOther.tsx
```diff
@@ -0,0 +1,86 @@
+import React from 'react';
+import { Box, Link, styled, Typography } from '@material-ui/core';
+import { useTranslation } from 'react-i18next';
+import { ContactDetailsTabQuery } from '../ContactDetailsTab.generated';
+
+const ContactOtherContainer = styled(Box)(({ theme }) => ({
+  marginLeft: theme.spacing(5),
+  margin: theme.spacing(1),
```
TheNoodleMoose, line 8: You could maybe consolidate this down do:
```suggestion
  margin: theme.spacing(1, 1, 1, 5)
```
but I'm not sure if that's really any better...

src/components/Contacts/ContactDetails/ContactDetailsTab/Other/ContactDetailsOther.tsx
```diff
@@ -0,0 +1,86 @@
...
+}));
+
+const ContactOtherTextContainer = styled(Box)(({ theme }) => ({
+  display: 'flex',
+  flexDirection: 'row',
```
TheNoodleMoose, line 13: Shouldn't need this ```flexDirection```

src/components/Contacts/ContactDetails/ContactDetailsTab/Other/ContactDetailsOther.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const { t } = useTranslation();
+  return (
+    <Box>
+      <Typography variant="h6">{t('Other')}</Typography>
+      <ContactOtherContainer>
+        <ContactOtherTextContainer>
+          <ContactOtherTextLabel variant="subtitle1">
+            {t('Assignee')}
+          </ContactOtherTextLabel>
+          <Typography variant="subtitle1">{contact.contact.name}</Typography>
```
TheNoodleMoose, line 39: Could probably do some destructuring here as well.
```ts
const {
    contact: {
      name,
      preferredContactMethod,
      locale,
      timezone,
      churchName,
      website,
    },
} = data;
```

src/components/Contacts/ContactDetails/ContactDetailsTab/Other/ContactDetailsOther.tsx
```diff
@@ -0,0 +1,86 @@
...
+}));
+
+const ContactOtherTextLabel = styled(Typography)(({ theme }) => ({
+  color: theme.palette.text.hint,
+  marginRight: '5px',
+}));
+
+interface ContactDetailsOtherProp {
+  //TODO: convert to ContactDetailTabQuery
+  contact: ContactDetailsTabQuery;
```
TheNoodleMoose, line 24: Is this comment still relevant?

src/components/Contacts/ContactDetails/ContactDetailsTab/Other/ContactDetailsOther.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+      variables: {
+        accountListId: accountListId,
+        contactId: contactId,
+      },
+    },
+  );
+  return (
+    <Box m={2}>
+      <MuiThemeProvider theme={theme}>
+        <CssBaseline />
```
TheNoodleMoose, line 33: I think you can remove these.

TheNoodleMoose (changes requested):

src/components/Contacts/ContactDetails/ContactDetailsTab/Other/ContactDetailsOther.tsx
```diff
@@ -0,0 +1,79 @@
...
+        </ContactOtherTextContainer>
+        <ContactOtherTextContainer>
+          <ContactOtherTextLabel variant="subtitle1">
+            {t('Church')}
+          </ContactOtherTextLabel>
+          <Typography variant="subtitle1">{churchName}</Typography>
+        </ContactOtherTextContainer>
+        <ContactOtherTextContainer>
+          <Link
+            href={website.startsWith('http') ? website : 'http://' + website}
```
TheNoodleMoose, line 69: One final template literal 🙂 
```suggestion
            href={website.startsWith('http') ? website : `http://${website}`}
```

TheNoodleMoose (changes requested):

src/components/Contacts/ContactDetails/ContactDetailsTab/Other/ContactDetailsOther.tsx
```diff
@@ -0,0 +1,79 @@
...
+        </ContactOtherTextContainer>
+        <ContactOtherTextContainer>
+          <ContactOtherTextLabel variant="subtitle1">
+            {t('Church')}
+          </ContactOtherTextLabel>
+          <Typography variant="subtitle1">{churchName}</Typography>
+        </ContactOtherTextContainer>
+        <ContactOtherTextContainer>
+          <Link
+            href={website.startsWith('http') ? website : `http://${website}`}
```
OzzieOrca, line 69: This seems a little odd that we just redirect out to random websites. Guess they should be entered but the current user though. Also a https redirect would be better but idk if that's practical.

I almost feel better about dumping the text on the page and letting the user copy/paste it into their url bar if they really want to visit the site. I feel (with no evidence) that this is a rarely used feature.

src/components/Contacts/ContactDetails/ContactDetailsTab/Other/ContactDetailsOther.tsx
```diff
@@ -0,0 +1,79 @@
...
+  marginLeft: theme.spacing(2),
+}));
+
+const ContactOtherTextLabel = styled(Typography)(({ theme }) => ({
+  color: theme.palette.text.hint,
+  marginRight: '5px',
+}));
+
+interface ContactDetailsOtherProp {
+  contact: ContactDetailsTabQuery;
```
OzzieOrca, line 21: This would be better typed if you made a fragment for just the fields needed by this component and used that fragment in the parent query. Then you could just use the generated type for that fragment.

I think it's better for each component to specify the fields they need. But this is a small component so it could be ok 🤷‍♂️

OzzieOrca (commented):
I'm not certain that either of the two comments need to be fixed, just wanted to get your thoughts before approving.

src/components/Contacts/ContactDetails/ContactDetailsTab/Other/ContactDetailsOther.tsx
gyasistory: @OzzieOrca I can go either way on this one.   This maybe something that Scott can answer later.  I will go ahead and remove it for now and we can add it back later.

OzzieOrca: Sounds good. I didn't go and look at what the current behavior is.

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.graphql
```diff
@@ -1,6 +1,7 @@
 query ContactDetailsTab($accountListId: ID!, $contactId: ID!) {
   contact(accountListId: $accountListId, id: $contactId) {
     id
+    locale
```
OzzieOrca, line 4: Are `locale` and `churchName` still needed in this query? If they are only needed for the other section, I think they should only be in the fragment. But if something else needs them, keep them.

OzzieOrca (changes requested):

OzzieOrca (approved):

PR #54, opened 2021-04-05
fix form List issue

gyasistory:
This is to fix a Lint issue that I keep recieving


TheNoodleMoose (approved):
Thanks, for fixing this. I guess it wasn't caught before because it's just a warning and not an error.

PR #55, opened 2021-04-06
MPDX-6975 : Contact details mailing section

gyasistory:
### Contact Details Tab - Mailing Section

Relies on
- [x] #48


src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/ContactDetailsTabMailing.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const Default = (): ReactElement => {
+  const mock = gqlMock<ContactDetailsTabQuery, ContactDetailsTabQueryVariables>(
+    ContactDetailsTabDocument,
+    { variables: { accountListId: '111', contactId: '222' } },
+  );
+  return (
+    <Box m={2}>
+      <MuiThemeProvider theme={theme}>
+        <CssBaseline />
```
frett, line 24: this should already be applied to all stories via: https://github.com/CruGlobal/mpdx-react/blob/main/.storybook/preview.js#L31-L42

gyasistory: @frett I don't think it is working because in the code below if I remove the ThemeProvider I don't get the correct style.

https://github.com/CruGlobal/mpdx-react/blob/27b7990c748ebb04e3258a35fb00cdad203944d8/src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.stories.tsx#L29

frett: I added a box with a background color of red to `preview.js` and it changed the background color of the preview area to red. so it is definitely processing it, not sure if something else is messing with it

frett: I bet it needs to be a `MuiThemeProvider` and not just a `ThemeProvider`

gyasistory: OK, I will remove the theme not sure why my test environment was not showing the theme at first but it currently is working.

src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/ContactDetailsTabMailing.tsx
```diff
@@ -0,0 +1,98 @@
...
+          <Typography variant="subtitle1">{data.contact.name}</Typography>
+          <Typography variant="subtitle1">
+            {data.contact.primaryAddress.street}
+          </Typography>
+          <Typography variant="subtitle1">
+            {data.contact.primaryAddress.city +
+              ', ' +
+              data.contact.primaryAddress.state +
+              ' ' +
+              data.contact.primaryAddress.postalCode}
```
TheNoodleMoose, line 63: Do we need a loading state for this? Or does this not get rendered until the data is available? Also, you could use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) here. A lot better than having concatting the string together.
```suggestion
{`${data.contact.primaryAddress.city}, ${data.contact.primaryAddress.state}, ${data.contact.primaryAddress.postalCode}`}
```
You could also do some [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#object_destructuring) so you don't have to dive so deep into the object to get the value you want for the string. You would probably need something like this.

```
  const {
    contact: {
      primaryAddress: { street, city, state, postalCode },
    },
  } = data;
```
And then all you would have for the string is
```
{`${city}, ${state}, ${postalCode}`}
```

src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/ContactDetailsTabMailing.tsx
```diff
@@ -0,0 +1,98 @@
...
+            <Typography variant="subtitle1">
+              {data.contact.sendNewsletter}
+            </Typography>
+          </ContactDetailsMailingLabelTextContainer>
+          {/* Magizine Section */}
+          <ContactDetailsMailingLabelTextContainer>
+            <ContactDetailsMailingLabel variant="subtitle1">
+              {t('Magazine')}
+            </ContactDetailsMailingLabel>
+            <Typography variant="subtitle1">Not Implemented</Typography>
```
TheNoodleMoose, line 92: I assume this will be implemented in a future pr? Is the data not available yet?

src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/ContactDetailsTabMailing.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { useTranslation } from 'react-i18next';
+import { ContactDetailsTabQuery } from '../ContactDetailsTab.generated';
+
+const ContactDetailsMailingTitle = styled(Typography)(({ theme }) => ({
+  marginTop: theme.spacing(2),
+}));
+
+const ContactDetailsMailingMainContainer = styled(Box)(({ theme }) => ({
+  display: 'flex',
+  flexDirection: 'row',
```
TheNoodleMoose, line 13: Shouldn't need this ```flexDirection```

src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/ContactDetailsTabMailing.tsx
```diff
@@ -0,0 +1,98 @@
...
+  flexGrow: 1,
+}));
+
+const ContactDetailsMailingTexContainer = styled(Box)(({}) => ({
+  flexGrow: 8,
+}));
+
+const ContactDetailsMailingLabelTextContainer = styled(Box)(({}) => ({
+  display: 'flex',
+  flexDirection: 'row',
```
TheNoodleMoose, line 30: Or this one.

src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/ContactDetailsTabMailing.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+const ContactDetailsMailingIcon = styled(LocationOn)(({ theme }) => ({
+  margin: theme.spacing(1),
+  width: '14px',
+  height: '20px',
+  flexGrow: 1,
+}));
+
+const ContactDetailsMailingTexContainer = styled(Box)(({}) => ({
+  flexGrow: 8,
```
TheNoodleMoose, line 25: This feels like a really large number for a flexGrow.. why was 8 the magic number? Also I think you can add the flex properties directly as a prop on the ```Box``` components.

```jsx
<ContactDetailsMailingTextContainer flex flexGrow="8" />
```

BUT! At the same time, that could get very messy especially with these large component trees. So I think having these styled-components would be the better option, but let me know what you think.

src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/ContactDetailsTabMailing.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+  ContactDetailsTabQuery,
+  ContactDetailsTabQueryVariables,
+} from '../ContactDetailsTab.generated';
+import { ContactDetailsTabMailing } from './ContactDetailsTabMailing';
+
+export default {
+  title: 'Contacts/Tab/ContactDetailsTab/Mailing',
+};
+
+export const Default = (): ReactElement => {
```
TheNoodleMoose, line 15: Like my question below, will this ever need a loading indicator? If so we could probably add a story here for loading.

TheNoodleMoose (dismissed):

src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/ContactDetailsTabMailing.tsx
gyasistory: @TheNoodleMoose yes the data is not currently available in graphql.

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.test.tsx
```diff
@@ -22,7 +22,7 @@ describe('ContactDetailTab', () => {
       }
     `);
     expect(result.current.data.contact.name).toMatchInlineSnapshot(
-      `"Ship Star"`,
+      `"Elephant Film Stomach"`,
```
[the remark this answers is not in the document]
gyasistory: @TheNoodleMoose I can switch to toMatchSnapshot this will create a snapshot file, instead of writing it inline.  I will make that change and let me know if you like that way better.

src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/ContactDetailsTabMailing.tsx
gyasistory: @TheNoodleMoose it didn't make a big deal on web but when I testing on a mobile screen 8 was the magic number.

src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/ContactDetailsTabMailing.stories.tsx
gyasistory: @TheNoodleMoose I don't think it will need a loading because it is passing in the query results from the ContactDetailsTab.  That view already has a loading screen.

src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/ContactDetailsTabMailing.tsx
```diff
@@ -0,0 +1,92 @@
...
+          {/* Newsletter Section */}
+          <ContactDetailsMailingLabelTextContainer>
+            <ContactDetailsMailingLabel variant="subtitle1">
+              {t('Newsletter')}
+            </ContactDetailsMailingLabel>
+            <Typography variant="subtitle1">
+              {data.contact.sendNewsletter}
+            </Typography>
+          </ContactDetailsMailingLabelTextContainer>
+          {/* Magizine Section */}
```
TheNoodleMoose, line 81: Not a huge deal since it's just a comment but 
```suggestion
          {/* Magazine Section */}
```

src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/ContactDetailsTabMailing.tsx
```diff
@@ -0,0 +1,92 @@
...
+      <ContactDetailsMailingMainContainer>
+        <ContactDetailsMailingIcon />
+        <ContactDetailsMailingTexContainer>
+          {/* Address Section */}
+          <Typography variant="subtitle1">{data.contact.name}</Typography>
+          <Typography variant="subtitle1">
+            {data.contact.primaryAddress.street}
+          </Typography>
+          <Typography variant="subtitle1">
+            {`${data.contact.primaryAddress.city}, ${data.contact.primaryAddress.state} ${data.contact.primaryAddress.postalCode}`}
```
TheNoodleMoose, line 57: What's your opinion on using the object destructuring for all this data? I think it could really help clean up some of this so you don't have to do ```data.contact``` for everything. Also is ```primaryAddress``` a nullable field? Can a contact not have a primary address?

TheNoodleMoose (changes requested):

src/components/Contacts/ContactDetails/ContactDetailsTab/__snapshots__/ContactDetailsTab.test.tsx.snap
```diff
@@ -0,0 +1,10 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`ContactDetailTab test query 1`] = `
+Object {
+  "accountListId": "accountList-id",
+  "contactId": "contact-id",
+}
+`;
+
+exports[`ContactDetailTab test query 2`] = `"Elephant Film Stomach"`;
```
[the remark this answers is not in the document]
gyasistory: @TheNoodleMoose NP I kind of like the other way as well.

src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/ContactDetailsTabMailing.tsx
gyasistory: @TheNoodleMoose  I really like I am going to switch to it here.  I wish Kotlin had something like this.

src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/ContactDetailsTabMailing.tsx
```diff
@@ -0,0 +1,94 @@
...
+      </ContactDetailsMailingTitle>
+      <ContactDetailsMailingMainContainer>
+        <ContactDetailsMailingIcon />
+        <ContactDetailsMailingTexContainer>
+          {/* Address Section */}
+          <Typography variant="subtitle1">{name}</Typography>
+          <Typography variant="subtitle1">{street}</Typography>
+          <Typography variant="subtitle1">
+            {`${city}, ${state} ${postalCode}`}
+          </Typography>
```
TheNoodleMoose, line 62: So ```primaryAddress``` can be null, so we should probably have a check for that.
```suggestion
 {primaryAddress !== null ? (
 <Typography variant="subtitle1">{primaryAddress?.street}</Typography>
          <Typography variant="subtitle1">
            {`${primaryAddress?.city}, ${primaryAddress?.state} ${primaryAddress?.postalCode}`}
          </Typography>
 ) : null}
```

not sure what design wants to do if they don't have a primary address, maybe instead of rendering ```null``` we render ```N/A``` or something to that extent?

TheNoodleMoose (approved):

TheNoodleMoose:
@OzzieOrca Pinging here as I accidentally approved this but auto-merge was on. What's your opinion of how we should address my last comment here?

src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/ContactDetailsTabMailing.tsx
OzzieOrca: I've been working on enabling TS strict, so that should catch this eventually if it doesn't get resolved. Maybe design has thoughts on a null state. I could see either a sentence about no addresses or removing the whole section as viable options. Is this supposed to display more than one address?

TheNoodleMoose: I believe it's just the one address

PR #57, opened 2021-04-08
MPDX-6970: Contact Details Header

reldredge71:
Filling out the content in the Contact Details Header


src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.tsx
```diff
@@ -1,29 +1,131 @@
...
-    <Box>
-      {/*TODO: Build Header*/}
-      {loading ? (
-        <p>loading</p>
-      ) : (
-        <p role="contactName">{data?.contact.name}</p>
-      )}
+    <Box style={{ padding: 24 }}>
+      <HeaderBar>
+        <ContactAvatar />
```
gyasistory, line 84:
```suggestion
        <ContactAvatar src={contact.avatar} />
```

This will load the avatar of the contact if it is available.

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.tsx
```diff
@@ -1,29 +1,131 @@
...
+                marginLeft: 18,
+                width: 240,
+                fontSize: 24,
+              }}
+            />
+          )}
+        </HeaderBarContactWrap>
+        <HeaderBarButtonsWrap>
+          <ButtonWrap>
+            <StarButtonIcon />
```
gyasistory, line 109:
```suggestion
            <StarContactIcon hasStar={true} />
```
The star functionality is already in place and can be reused.  Currently it is colored but we can change that.

gyasistory (changes requested):
A few minor suggestions also once your merge it with main you should get rid of compile error.

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.tsx
reldredge71: Yeah I was avoiding it because it was colored, but yeah we can change that later

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.test.tsx
```diff
@@ -18,11 +18,11 @@ describe('ContactDetails', () => {
       </GqlMockedProvider>,
     );
 
-    await waitFor(() => expect(getByText('loading')).toBeInTheDocument());
+    expect((await findAllByRole('ContactName'))[0]).toBeInTheDocument();
```
OzzieOrca, line 21: Sorta seems like a loading state should be able to find this in the document...

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.tsx
```diff
@@ -1,29 +1,127 @@
...
 
 interface Props {
   accountListId: string;
   contactId: string;
 }
 
-export const ContactDetailsHeader: React.FC<Props> = ({
+const HeaderBar = styled(Box)(({}) => ({
+  display: 'flex',
+  paddingBottom: 24,
```
OzzieOrca, line 22: There's a lot of magic numbers in here. Would the `theme.spacing(...)` function be useful here?

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.tsx
```diff
@@ -1,29 +1,127 @@
...
+}));
+const HeaderSectionWrap = styled(Box)(({}) => ({
+  display: 'flex',
+}));
+
+export const ContactDetailsHeader = ({
   accountListId,
   contactId,
-}: Props) => {
+}: Props): ReactElement => {
```
OzzieOrca, line 69: Use `export const ContactDetailsHeader: React.FC<Props>` instead.

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.tsx
```diff
@@ -1,29 +1,127 @@
...
   const { data, loading } = useGetContactDetailsHeaderQuery({
     variables: { accountListId, contactId },
   });
+  const { t } = useTranslation();
+
+  const { contact } = data || {};
```
OzzieOrca, line 75: Do the types survive this? I feel like they haven't in the past. Also enabling ts strict mode might complain. https://github.com/CruGlobal/mpdx-react/blob/378a455a9685f5d276dad23f00f141b1bfa98f25/tsconfig.json#L7

I'm working on enabling that globally still.

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeaderSection/ContactHeaderAddressSection.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+import React, { ReactElement } from 'react';
+import { useTranslation } from 'react-i18next';
+import theme from '../../../../../theme';
+import { ContactDetailsHeaderFragment } from '../ContactDetailsHeader.generated';
+import { ContactHeaderSection } from './ContactHeaderSection';
+
+interface Props {
+  loading: boolean;
+  contact?: ContactDetailsHeaderFragment;
```
OzzieOrca, line 13: You could have a separate fragment for each of these sub components. That way each component specifies the fields it needs directly. Might be easier to maintain in the future.

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeaderSection/ContactHeaderEmailSection.tsx
```diff
@@ -0,0 +1,49 @@
...
+  marginLeft: 18,
+  width: 200,
+  fontSize: 16,
+}));
+
+export const ContactHeaderEmailSection = ({
+  loading,
+  contact,
+}: Props): ReactElement => {
+  let content: ReactElement = null;
```
OzzieOrca, line 32: I'd prefer to avoid `let` if we can help it and use immutable patterns.

OzzieOrca (changes requested):

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.test.tsx
reldredge71: Hmm yeah now that I think about this, this is not a good test for the loading state.  When it's loading, almost all the content should be `Skeleton` components.  Should I search for items with the Skeleton type to tell if it is a loading state?

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.tsx
LizzyRandall: Or grid might be useful for arranging, that has a spacing option

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.tsx
reldredge71: I think they do... but if it's safer to not unpack it this way, I ca do something else

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.test.tsx
OzzieOrca: I think that sounds ok. Or at least assert that the data isn't in the document. But that doesn't test much... What you've got looks good :)

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.tsx
OzzieOrca: I'd need to test it with strict mode. Maybe the types survive...

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactHeaderSection/ContactHeaderAddressSection.tsx
```diff
@@ -0,0 +1,71 @@
...
+  fontSize: 16,
+}));
+
+export const ContactHeaderAddressSection = ({
+  loading,
+  contact,
+}: Props): ReactElement => {
+  const { t } = useTranslation();
+
+  //let content: ReactElement = null;
```
OzzieOrca, line 35:
```suggestion
```

OzzieOrca (approved):

OzzieOrca (approved):

PR #59, opened 2021-04-09
Mpdx 6975 add mutation for updating contact tags

gyasistory:
This PR is to add delete functionality to Contact Chip.  I used the mutation to update the chip

### DIFF
--- a/src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
+++ b/src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
@@ -1,6 +1,7 @@
 import { Box, Chip, Link, styled, Typography } from '@material-ui/core';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
+import { useUpdateContactTagsMutation } from './ContactTab.generated';
 
 const ContactTagsContainer = styled(Box)(({ theme }) => ({
   display: 'flex',
@@ -21,11 +22,13 @@ const ContactTagAddTagText = styled(Typography)(({ theme }) => ({
 }));
 
 interface ContactTagsProps {
+  accountListId: string;
   contactId: string;
   contactTags: string[];
 }
 
 export const ContactTags: React.FC<ContactTagsProps> = ({
+  accountListId,
   contactId,
   contactTags,
 }) => {
@@ -33,8 +36,16 @@ export const ContactTags: React.FC<ContactTagsProps> = ({
   const handleTagDelete = (tag: string) => {
     const index = contactTags.indexOf(tag);
     if (index > -1) {
-      contactTags = contactTags.splice(index, 1);
-      // TODO: add mutation to update tag for contact
+      const updatedTags = contactTags.splice(index, 1);
+
+      const [, { data }] = useUpdateContactTagsMutation({
+        variables: {
+          accountList: accountListId,
+          contactId: contactId,
+          tagList: updatedTags,
+        },
+      });
+      contactTags = data.updateContact.contact.tagList;
     }
     console.info('You clicked the Delete tag for tag' + tag + '' + contactId);
   };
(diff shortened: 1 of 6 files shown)


reldredge71 (approved):

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
```diff
@@ -21,22 +25,37 @@ const ContactTagAddTagText = styled(Typography)(({ theme }) => ({
...
 export const ContactTags: React.FC<ContactTagsProps> = ({
+  accountListId,
   contactId,
-  contactTags,
 }) => {
   const { t } = useTranslation();
+  const { data, loading } = useGetContactTagsQuery({
+    variables: {
+      accountListID: accountListId, // value for 'accountListID'
+      contactId: contactId, // value for 'contactId'
```
OzzieOrca, line 40: I don't think these comments add any value.

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
```diff
@@ -21,22 +25,37 @@ const ContactTagAddTagText = styled(Typography)(({ theme }) => ({
...
+  });
+
   const handleTagDelete = (tag: string) => {
-    const index = contactTags.indexOf(tag);
-    if (index > -1) {
-      contactTags = contactTags.splice(index, 1);
-      // TODO: add mutation to update tag for contact
-    }
-    console.info('You clicked the Delete tag for tag' + tag + '' + contactId);
+    console.info('You clicked the Delete tag for tag ' + tag + ' ' + contactId);
```
OzzieOrca, line 45: Probably don't need to keep this.

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
```diff
@@ -21,22 +25,37 @@ const ContactTagAddTagText = styled(Typography)(({ theme }) => ({
...
+    const list = data.contact.tagList;
+    const updatedTags: string[] = [];
+    list.forEach((element) => {
+      element != tag ? updatedTags.push(tag) : null;
+    });
```
OzzieOrca, line 50: `.filter` would be better here. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
```diff
@@ -21,22 +25,37 @@ const ContactTagAddTagText = styled(Typography)(({ theme }) => ({
...
+
+    const [] = useUpdateContactTagsMutation({
+      variables: {
+        accountList: accountListId,
+        contactId: contactId,
+        tagList: updatedTags,
+      },
+    });
```
OzzieOrca, line 58: Hooks need to be at the top level of a react component. You can't put them inside other functions. This hook returns a function that you can use to call it when desired.
I would use the hook like this:
```ts
const [updateTags, { data }] = useUpdateContactTagsMutation({
  variables: {
    accountList: accountListId,
    contactId: contactId,
  },
});
```
and then later call the `updateTags` function to execute the mutation when needed
```ts
updateTags({
  variables: {
    tagList: updatedTags,
  },
});
```
You can set variables either in the hook or when you call the mutation. In this case you don't know the tags until you're inside this function so you can wait to set that until it's executed.

https://www.apollographql.com/docs/react/data/mutations/#executing-a-mutation

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
```diff
@@ -21,22 +25,37 @@ const ContactTagAddTagText = styled(Typography)(({ theme }) => ({
...
-  contactTags: string[];
 }
 
 export const ContactTags: React.FC<ContactTagsProps> = ({
+  accountListId,
   contactId,
-  contactTags,
 }) => {
   const { t } = useTranslation();
+  const { data, loading } = useGetContactTagsQuery({
```
OzzieOrca, line 37: Could the parent component load these? You could pass it a fragment.

OzzieOrca (changes requested):

PR #67, opened 2021-04-19
MPDX-6943-Export-Physical-Newsletter-Functionality

TheNoodleMoose:
This PR focuses on finishing up the Export Physical menu item. I discussed with Scotty via Slack, but I was able to get downloading the CSV/XLSX file working but for some reason, it's taking waaay too long. On stage.mpdx.org the file download is almost instantaneous, but locally it takes about 30 seconds to start. Hopefully, we can figure out why before this is merged in.
Screen Shot 2021-04-19 at 4 17 47 PM


pages/api/graphql-rest.page.ts
```diff
@@ -15,6 +15,18 @@ const typeDefs = gql`
     taskAnalytics(accountListId: ID!): TaskAnalytics!
   }
 
+  type Mutation {
+    exportContacts(input: ExportContactsInput!): String!
+  }
+
+  input ExportContactsInput {
+    format: String!
```
OzzieOrca, line 23: Can we make this an enum?

pages/api/graphql-rest.page.ts
```diff
@@ -15,6 +15,18 @@ const typeDefs = gql`
...
+    mailing: Boolean!
```
OzzieOrca, line 24: Can you add a description? https://www.apollographql.com/docs/apollo-server/schema/schema/#documentation-strings

It's not immediately clear to me what this filter is doing. Is it only including mailing contacts when true? excluding mailing contacts when false?

A name change is another option if you like that any better

pages/api/graphql-rest.page.ts
```diff
@@ -15,6 +15,18 @@ const typeDefs = gql`
...
+    labelType: String
+    sort: String
```
OzzieOrca, line 26: Could either of these be enums?

pages/api/graphql-rest.page.ts
```diff
@@ -99,6 +132,35 @@ class MpdxRestApi extends RESTDataSource {
...
+      newsletter: string;
+      status: string;
+    },
+    labelType?: string | null,
+    sort?: string | null,
+  ) {
+    let pathAddition = '';
+    if (mailing) {
+      pathAddition = '/mailing';
+    }
```
OzzieOrca, line 149:
```suggestion
    const pathAddition = mailing ? '/mailing' : '';
```
I prefer the immutable syntax. Trivial though.

pages/api/graphql-rest.page.ts
```diff
@@ -99,6 +132,35 @@ class MpdxRestApi extends RESTDataSource {
...
+        params: {
+          filter,
+          type: labelType,
+          sort,
+        },
+        type: 'export_logs',
+      },
+    });
+
+    return `${process.env.REST_API_URL}contacts/exports${pathAddition}/${data.id}.${format}?access_token=`;
```
OzzieOrca, line 161: Maybe let the frontend tack on the `access_token=`? Just seems a little odd to return part of a query param.

src/components/Dashboard/ThisWeek/NewsletterMenu/MenuItems/ExportPhysical/ExportPhysical.tsx
```diff
@@ -6,16 +6,25 @@ import {
...
 interface Props {
   accountListId: string;
   handleClose: () => void;
 }
 
+enum ExportFormatEnum {
+  CSV = 'csv',
+  XLSX = 'xlsx',
+}
+
```
OzzieOrca, line 27: Ooh ya can you move this to the server? The client can get it from codegen.

src/components/Dashboard/ThisWeek/NewsletterMenu/MenuItems/ExportPhysical/ExportPhysical.tsx
```diff
@@ -31,8 +40,36 @@ const CloseButton = styled(IconButton)(({ theme }) => ({
...
-const ExportPhysical = ({ handleClose }: Props): ReactElement<Props> => {
+const LabelButton = styled(Button)(({ theme }) => ({
+  width: '90%',
+  margin: theme.spacing(1, 0),
+}));
+
+const ExportPhysical = ({
+  handleClose,
+  accountListId,
+}: Props): ReactElement<Props> => {
```
OzzieOrca, line 51:
```suggestion
const ExportPhysical: React.FC<Props> = ({
  handleClose,
  accountListId,
}) => {
```

I think the React.FC handles the children case better and you don't have to specify a return type. Also the JSX you are returning doesn't take any props so `ReactElement<Props>` isn't accurate.

src/components/Dashboard/ThisWeek/NewsletterMenu/MenuItems/ExportPhysical/ExportPhysical.tsx
```diff
@@ -31,8 +40,36 @@ const CloseButton = styled(IconButton)(({ theme }) => ({
...
   const { t } = useTranslation();
+  const [session] = useSession();
+
+  const [createExportedContacts] = useCreateExportedContactsMutation({
+    onCompleted: (data) => {
+      window.location.replace(`${data.exportContacts}${session?.user.token}`);
```
OzzieOrca, line 57: Would this be better in a new window? Or does it behave fine?

OzzieOrca (changes requested):

OzzieOrca:
Looks like you're filtering the requests incorrectly somehow. I'm clicking Export Physical > CSV for mail merge. The existing version of the CSV export returns 17 contacts but this implementation returns 2260. Somehow you're not filtering right and is probably the cause of your performance differences.

In the existing app the request looks like:
Screen Shot 2021-04-19 at 6 09 53 PM
The GraphQL one looks like
Screen Shot 2021-04-19 at 6 10 21 PM
with a rest body of
```json
{
  "data": {
    "params": {
      "filter": {
        "account_list_id": "3c051f91-7806-4ce3-9132-314eaba7c5f4",
        "newsletter": "address",
        "status": "active"
      }
    },
    "type": "export_logs"
  }
}
```

I think the rest body needs to go `data.attributes.params.filter`. Looks like you're missing the `attributes` key. It really should throw some errors but it doesn't...

pages/api/graphql-rest.page.ts
TheNoodleMoose: Possibly but none of the examples in the angular codebase ever use these fields for the export so I'm not exactly sure what the values are. The docs don't have anything about them either: https://developers.mpdx.org/api/v2/contacts/exports

src/components/Dashboard/ThisWeek/NewsletterMenu/MenuItems/ExportPhysical/ExportPhysical.tsx
TheNoodleMoose: Ah yep, will update!

src/components/Dashboard/ThisWeek/NewsletterMenu/MenuItems/ExportPhysical/ExportPhysical.tsx
TheNoodleMoose: Yeah it works fine, visually it doesn't do anything except start the download

TheNoodleMoose:
@OzzieOrca Yep updated the data object with ```attributes``` and the download is MUCH quicker haha. Thanks for catching that!

src/components/Dashboard/ThisWeek/NewsletterMenu/MenuItems/ExportPhysical/ExportPhysical.tsx
```diff
@@ -31,8 +36,38 @@ const CloseButton = styled(IconButton)(({ theme }) => ({
...
   const { t } = useTranslation();
+  const [session] = useSession();
+
+  const [createExportedContacts] = useCreateExportedContactsMutation({
+    onCompleted: (data) => {
+      window.location.replace(
+        `${data.exportContacts}?access_token=${session?.user.token}`,
+      );
+      handleClose();
+    },
```
OzzieOrca, line 57: You can't get this from the return value of `await createExportedContacts`? Seems like it would be better to group that code instead of having a callback elsewhere.

OzzieOrca (approved):

src/components/Dashboard/ThisWeek/NewsletterMenu/MenuItems/ExportPhysical/ExportPhysical.tsx
TheNoodleMoose: Yeah looking at it now, that's probably better. I think I had issues originally when trying but it seems to work fine now. I'll use this way instead!

pages/api/graphql-rest.page.ts
TheNoodleMoose: Actually, I figured this out while working on the last bit of the exporting logic. I've turned these into Enums.

pages/api/graphql-rest.page.ts
```diff
@@ -103,8 +118,8 @@ const resolvers: Resolvers = {
         mailing,
         format,
         filter,
-        labelType,
-        sort,
+        labelType ?? undefined,
+        sort ?? undefined,
```
TheNoodleMoose, line 122: I had to add this here as I was getting a TS error about it being ```null```, even though I'm passing ```undefined```...

src/components/Dashboard/ThisWeek/NewsletterMenu/MenuItems/ExportPhysical/ExportPhysical.test.tsx
```diff
@@ -41,6 +52,121 @@ describe('ExportPhysical', () => {
...
+      // Have to ignore TS complaining about deleting window.location because location is not optional
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      delete window.location;
+      window.location = { ...location, replace: jest.fn() };
+    });
+
+    afterAll(() => {
+      window.location = location;
+    });
```
TheNoodleMoose, line 71: I looked everywhere to find a way to mock ```window.location.replace``` and this was the only one that worked. However, a change in TypeScript 4 causes a TS issue with the ```delete window.location```. Let me know how you feel about this, or if you have any suggestions.

src/components/Dashboard/ThisWeek/NewsletterMenu/MenuItems/ExportPhysical/ExportPhysical.test.tsx
```diff
@@ -41,6 +52,121 @@ describe('ExportPhysical', () => {
...
+      await waitFor(() =>
+        expect(window.location.replace).toHaveBeenCalledWith(
+          `${url}?access_token=someToken1234`,
+        ),
+      );
+    });
+  });
+
+  describe('Exporting Contacts Mutation', () => {
+    // it('Exports Contacts - PDF of Mail Merged Labels', async () => {
```
TheNoodleMoose, line 98: @OzzieOrca I was talking with @gyasistory and he mentioned a way of testing that you suggested, [example here](https://github.com/CruGlobal/mpdx-react/blob/d3f052ff6aa0007fe0bb97cb513da110b49456b7/src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.test.tsx#L7). So I tried to implement this for the mutation. However, I'm getting a ```timed out in waitfornextupdate after 1000ms``` error from the ```waitForNextUpdate```. Any ideas why? I tried looking around but the suggestions I found didn't seem to work. Also, do we want to even do this way of testing with mutations? Would love to get your opinion on which test format would be better. Or maybe even having both would be fine 🤷‍♂️

src/components/Dashboard/ThisWeek/NewsletterMenu/MenuItems/ExportPhysical/ExportPhysical.tsx
```diff
@@ -41,105 +48,175 @@ const LabelButton = styled(Button)(({ theme }) => ({
...
+  width: '100%',
+}));
+
 const ExportPhysical: React.FC<Props> = ({
   handleClose,
   accountListId,
 }: Props) => {
   const { t } = useTranslation();
   const [session] = useSession();
+  const [isExportingPdf, changeIsExportingPdf] = useState(false);
```
TheNoodleMoose, line 66: If the user clicks ```PDF of Mail Merged Labels``` a new menu pops up to select the different filters they wish to apply to the export. Instead of creating a whole new component, I figured using some conditional logic inside this component would be simple enough. Let me know what you think though.

pages/api/graphql-rest.page.ts
TheNoodleMoose: Here's what I get when I remove the ```?? undefined```
Screen Shot 2021-04-21 at 4 43 52 PM
Does the ```Maybe``` mean it could be ```null``` as well?

OzzieOrca: Ya codegen made `Maybe` and it means it can be null.

TheNoodleMoose: So would it be better to leave it how it is, or remove the ```?? undefined``` and just add ``` | null``` to the type below?

src/components/Dashboard/ThisWeek/NewsletterMenu/MenuItems/ExportPhysical/ExportPhysical.test.tsx
OzzieOrca: https://stackoverflow.com/a/56999581/665224 is a possibility.

I also saw mention of using `global` instead of `window`. Not quite sure what there is to be gained from that.

With this solution, how come `delete window.location` is required? Does overwriting it throw errors?

TheNoodleMoose: I think I tried that solution with no success, but I'll give it another shot. Overwriting doesn't throw errors, but it doesn't work. It appears you have to delete the existing one set by jsdom and then set it again but with your mock.

src/components/Dashboard/ThisWeek/NewsletterMenu/MenuItems/ExportPhysical/ExportPhysical.test.tsx
OzzieOrca: Umm I don't think we need to test generated code. That should be the responsibility of the Apollo library and codegen libraries. We should test by interacting with our component as a user would by clicking a button or something.

I'm trying to figure out the context of what I suggested...

It's probably not working cuz it's a mutation and you never called the mutation function before calling `await waitForNextUpdate();` so there were no changes.

src/components/Dashboard/ThisWeek/NewsletterMenu/MenuItems/ExportPhysical/ExportPhysical.tsx
OzzieOrca: That's fine. Could be worth making some sort of modal component that would let you pass these sections in as props:
```ts
<Modal
      title="Something"
      body={
        <SomeJSX />
      }
      footer={
        <SomeJSX />
      } />
```
I'm not sure what's best. This is okay.

pages/api/graphql-rest.page.ts
```diff
@@ -22,21 +26,32 @@ const typeDefs = gql`
 
   input ExportContactsInput {
     """
-    Enum value to determine the file format of the exported contacts(Either csv or xlsx)
+    Enum value to determine the file format of the exported contacts(Either csv,xlsx, or pdf)
```
OzzieOrca:
```suggestion
    Enum value to determine the file format of the exported contacts (Either csv, xlsx, or pdf)
```

pages/api/graphql-rest.page.ts
OzzieOrca: GraphQL seems to operate having `null` as the only nullish value (it doesn't have undefined like JS). Maybe it would be best to use `labelType: string | null,` on the server side instead of `labelType?: string,` (even though that's more concise) so you can just keep the `null` instead of having to convert to/from `undefined`

OzzieOrca (changes requested):

src/components/Dashboard/ThisWeek/NewsletterMenu/MenuItems/ExportPhysical/ExportPhysical.test.tsx
```diff
@@ -70,15 +97,96 @@ describe('ExportPhysical', () => {
...
+          <ExportPhysical
+            accountListId={accountListId}
+            handleClose={handleClose}
+          />
+        </GqlMockedProvider>,
+      );
+
+      userEvent.click(getByText('PDF of Mail Merged Labels'));
+      //  material ui selects do not use actual Select elements in the DOM
+      //  reference https://stackoverflow.com/a/61491607
```
TheNoodleMoose, line 142: Having to test the selects this way is a bit annoying but glad I was able to find this solution.

pages/api/graphql-rest.page.ts
```diff
@@ -167,8 +167,8 @@ class MpdxRestApi extends RESTDataSource {
       newsletter: string;
       status: string;
     },
-    labelType?: ExportLabelTypeEnum,
-    sort?: ExportSortEnum,
+    labelType?: ExportLabelTypeEnum | null,
+    sort?: ExportSortEnum | null,
```
TheNoodleMoose, line 171: So up in the typeDef, I have these set to be Nullable fields, and that seems to add the types ```Maybe<ExportLabelTypeEnum> | undefined```. So I add the ```?``` to the argument to resolve the undefined type. I then added the ```null``` to resolve the ```Maybe```? I'm partially confused why it's generating ```Maybe``` and ```undefined``` for the type. Maybe I'm missing something. On the client-side I'm still passing either ```ExportLabelTypeEnum``` or undefined for this.

OzzieOrca: Oh weird.

Maybe we want this? https://www.graphql-code-generator.com/docs/generated-config/typescript#avoidoptionals Idk why this hasn't been a problem on the client side.

I also saw https://www.graphql-code-generator.com/docs/generated-config/typescript#maybevalue. Would be sorta nice to just use `undefined` everywhere but that just changes the typings, not what Apollo returns when GraphQL returns a `null` value.

TheNoodleMoose: I added the ```avoidOptionals: true``` to the config but it didn't appear to change anything. Maybe it's something we could address in the eventual PR to improve the file structure/modularization of the REST -> GraphQL logic?

OzzieOrca: Which codegen config did you add it to? The server one?

Ya we can deal with it later.

OzzieOrca (approved):

pages/api/graphql-rest.page.ts
TheNoodleMoose: Yeah, the server one.

PR #69, opened 2021-04-21
MPDX-6990: Contacts Table Header bar

reldredge71:
Building out the header bar for the contacts table.  The UI is pretty much finished, minus the search bar and the actions dropdown.  Needs to be connected to the filter panel for certain functions.


src/components/Contacts/CheckBox/CheckBox.tsx
```diff
@@ -0,0 +1,37 @@
...
+  onClick?: () => void;
+}
+
+const StyledCheckBox = styled(Checkbox)(({ checked }: CheckboxProps) => ({
+  width: '18px',
+  height: '18px',
+  color: checked ? theme.palette.primary.dark : theme.palette.secondary.dark,
+}));
+
+export const CheckBox: React.FC<Props> = ({
```
LizzyRandall, line 22: Is naming wrapper like components the same as Material UI a good practice? Just curious, I've tried to make the names different but haven't ran into needing the same name.

src/components/Contacts/ContactsTable/ContactsTable.tsx
```diff
@@ -14,11 +14,17 @@ import { useContactsQuery } from '../../../../pages/accountLists/[accountListId]
...
+  filterPanelOpen: boolean;
+  toggleFilterPanel: () => void;
 }
 
 export const ContactsTable: React.FC<Props> = ({
   accountListId,
   onContactSelected,
+  activeFilters,
+  filterPanelOpen,
+  toggleFilterPanel,
```
LizzyRandall, line 27: Is this the standard you guys have for deconstructing props in the project?

src/components/Contacts/StarContactIcon/StarContactIcon.tsx
```diff
@@ -1,26 +1,25 @@
...
-    height: '24px',
-    color: theme.palette.secondary.main,
-  },
-}));
-
 interface Props {
   hasStar: boolean;
 }
 
+const StarFilled = styled(StarIcon)(({}) => ({
```
LizzyRandall, line 11: Are you guys using styled option instead of makeStyles now?

LizzyRandall (commented):
I had a few questions mainly around standards.

src/components/Contacts/CheckBox/CheckBox.tsx
OzzieOrca: Idk that I have a strong preference. ES6 imports make it clear where the component is imported from. But I could still see there being some ambiguity until you look up the import.

src/components/Contacts/CheckBox/CheckBox.tsx
```diff
@@ -0,0 +1,37 @@
...
+  state = CheckBoxState.unchecked,
+  onClick,
+}) => {
+  const checked = state != CheckBoxState.unchecked;
```
OzzieOrca, line 26:
```suggestion
  const checked = state !== CheckBoxState.unchecked;
```
We should find this eslint rule and enable it.

src/components/Contacts/ContactsHeader/ContactsHeader.tsx
```diff
@@ -1,10 +1,192 @@
-import { Box, colors } from '@material-ui/core';
-import React from 'react';
+import { Box, styled, IconButton, Hidden } from '@material-ui/core';
+import { FilterList, FormatListBulleted, ViewColumn } from '@material-ui/icons';
+import { ToggleButton, ToggleButtonProps } from '@material-ui/lab';
+import React, { useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import theme from '../../../theme';
```
OzzieOrca, line 6: I think you should be pulling `theme` from the styled context argument instead.

src/components/Contacts/ContactsHeader/ContactsHeader.tsx
```diff
@@ -1,10 +1,192 @@
...
+    activeFilters: boolean;
+    panelOpen: boolean;
+  }) => ({
+    display: 'inline-block',
+    width: 48,
+    height: 48,
+    borderRadius: 24,
+    margin: theme.spacing(1),
+    backgroundColor: activeFilters
+      ? '#FFCF07'
```
OzzieOrca, line 45: Let's get this in the theme. I'll try and get John's PR merged today which has it https://github.com/CruGlobal/mpdx-react/pull/49/files#diff-0e21fb1c9519a397050defc2148a2d1b966a0982cc4cf196e1e6b007f8d095e5R7

src/components/Contacts/ContactsTable/ContactsTable.tsx
OzzieOrca: Ya I think we usually try and use destructuring vs having `props.` everywhere.

src/components/Contacts/StarContactIcon/StarContactIcon.tsx
OzzieOrca: Ya I think we prefer this syntax. Ties styles to a specific component directly and makes it clear when something is no longer used and can be deleted.

OzzieOrca (changes requested):

src/components/Contacts/CheckBox/CheckBox.tsx
reldredge71: Technically, Material UI has `Checkbox`, this is `CheckBox`... just kidding I'll change it, it can still get confusing.

OzzieOrca (approved):

PR #70, opened 2021-04-21
Set SITE_URL and NEXTAUTH_URL based on branch first

OzzieOrca:
- Fixes issues with VERCEL_URL being the Vercel deployment url instead of the domain name

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -37,13 +37,15 @@ module.exports = withPlugins([
   withGraphql,
   withBundleAnalyzer,
   {
+    serverRuntimeConfig: {
+      NEXTAUTH_URL: siteUrl,
+    },
     env: {
       JWT_SECRET: process.env.JWT_SECRET || 'development-key',
       API_URL: process.env.API_URL || 'https://api.stage.mpdx.org/graphql',
       REST_API_URL:
         process.env.REST_API_URL || 'https://api.stage.mpdx.org/api/v2/',
       SITE_URL: siteUrl,
-      NEXTAUTH_URL: siteUrl,
       CLIENT_ID: process.env.CLIENT_ID || '4027334344069527005',
       CLIENT_SECRET: process.env.CLIENT_SECRET,
       BEACON_TOKEN:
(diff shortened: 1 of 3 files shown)


next.config.js
```diff
@@ -38,18 +37,22 @@ module.exports = withPlugins([
...
-      SITE_URL: SiteUrl,
+      SITE_URL: siteUrl,
       CLIENT_ID: process.env.CLIENT_ID || '4027334344069527005',
       CLIENT_SECRET: process.env.CLIENT_SECRET,
       BEACON_TOKEN:
         process.env.BEACON_TOKEN || '01b4f5f0-7fff-492a-b5ec-d536f3657d10',
     },
     // Force .page prefix on page files (ex. index.page.tsx) so generated files can be included in /pages directory without Next.js throwing build errors
     pageExtensions: ['page.tsx', 'page.ts', 'page.jsx', 'page.js'],
+    productionBrowserSourceMaps: true,
```
TheNoodleMoose, line 56: Was this needed to get these changes to work or was it just something that should have been set to ```true``` originally?

pages/login.page.tsx
```diff
@@ -18,11 +18,7 @@ const IndexPage = (): ReactElement => (
       subtitle="MPDX is fundraising software from Cru that helps you grow and maintain your ministry
 partners in a quick and easy way."
     >
-      <Button
-        size="large"
-        variant="contained"
-        onClick={() => signin('thekey', { callbackUrl: process.env.SITE_URL })}
-      >
+      <Button size="large" variant="contained" onClick={() => signin('thekey')}>
```
TheNoodleMoose, line 21: The ```callbackUrl``` isn't needed anymore because the ```serverRunTimeConfig.NEXTAUTH_URL``` now contains the URL to navigate to after auth?

TheNoodleMoose (approved):

next.config.js
OzzieOrca: No I just wanted to debug on prod and couldn't. Unrelated to this PR but thought would be nice to enable.

pages/login.page.tsx
OzzieOrca: Ya it just uses the current browser url I think. Not sure why Tatai set it originally.

PR #71, opened 2021-04-24
Add eslint-plugin-jsx-a11y

OzzieOrca:
I think you guys have been abusing the HTML `role` attribute haha. There are only a limited number of role values that should be used:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques#roles

If you are reaching for this in tests, looks like using the accessible name along with the role: `getByRole(expectedRole, { name: 'The name' })` is a better way to do it followed by `getByText`/`getByAltText`/other ones for specific use cases and `getByTestId` if you can't get anything else to work.
https://testing-library.com/docs/queries/about#priority

If you've got something like an icon, we should probably set some sort of an alt text on it. https://material-ui.com/api/svg-icon/#svgicon-api has a `titleAccess` prop.


src/components/HandoffLink/HandoffLink.test.tsx
```diff
@@ -28,6 +28,7 @@ describe('HandoffLink', () => {
         }}
       >
         <HandoffLink path="/contacts">
+          {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
```
TheNoodleMoose, line 31: Could you just define this once at the top of the file to disable it for the whole file?

TheNoodleMoose (approved):
@OzzieOrca Looks good, thanks for adding this and fixing the lint issues that followed 🙂 Do you have a preferred query you would think would be best to use? I know they have priority but wasn't sure if you had a preference over just throwing ```data-testid```'s everywhere vs trying to add ```titleAccess``` where we can. I think accessibility is something we should keep trying to improve! 🙂

src/components/HandoffLink/HandoffLink.test.tsx
OzzieOrca: In general, I like having the rules enabled for the rest of the file and like having the line specific ones go away when the line is deleted. If they are up top, they usually last forever.

But this specific file is a test. Maybe we don't need to enforce a11y rules on test files. Also I expect this file specifically to be deleted at the end of the rewrite.

OzzieOrca:
I think the priority listed in the docs is a good one. It forces you to add accessibility features to elements that don't have an easy text selector and makes you think about accessibility more which I like. There's a reason test id is last since it doesn't help the accessibility story. But if things already seem accessible and it's hard/frustrating to query things without it, I think using test id sparingly is fine.

PR #73, opened 2021-04-26
Add TabPanels to ContactDetails

OzzieOrca:


src/components/Contacts/ContactDetails/ContactDetails.tsx
```diff
@@ -50,11 +52,11 @@ export const ContactDetails: React.FC<Props> = ({
 }: Props) => {
   const { t } = useTranslation();
 
-  const [selectedTabIndex, setSelectedTabIndex] = React.useState(0);
+  const [selectedTabIndex, setSelectedTabIndex] = React.useState('0');
```
OzzieOrca, line 55: Seemed like some of the typings for the tab components were expecting strings so I changed everything to a string.

LizzyRandall:
Looks good to me. Do you know if you guys will have more tabs in the future? If so, maybe an enum would help then? I just think of Siebel we get more and more dumped on an account with different entities (notes, attachments, events, donations, activities... it's like 15 or so entities!)

reldredge71:
This looks good to me!  I agree that an enum might be nice to have, I don't think we'll be adding to this tab bar too much, but I think enums are more precise than an index value.  I'm fine either way.

@OzzieOrca Do you know what's up with the eslint warning in the change log?

reldredge71 (approved):

OzzieOrca:
Sure we can do an enum :)

PR #74, opened 2021-04-28
Mpdx 6974 Contact referral tab

gyasistory:


gyasistory:
I currently don't have the `createdAt` attribute implemented because it causes the storyboard to fail as the mock data doesn't create it properly.

gyasistory:
> I currently don't have the `createdAt` attribute implemented because it causes the storyboard to fail as the mock data doesn't create it properly.

I was able to add the createdAt into this PR.  It just will not display on the Storybook and you will have to look at the site to view it.

src/components/Contacts/ContactDetails/ContactReferralTab/ContactReferralTab.test.tsx
```diff
@@ -1,6 +1,12 @@
+/* eslint-disable react/react-in-jsx-scope */
```
OzzieOrca, line 1: Let's just disable this rule. It used to be needed but importing `React` is no longer needed. https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

You can add `'react/react-in-jsx-scope': 'off',` to the `rules` section of https://github.com/CruGlobal/mpdx-react/blob/main/.eslintrc.js#L18-L33 instead of disabling this one file.

src/components/Contacts/ContactDetails/ContactReferralTab/ContactReferralTab.tsx
```diff
@@ -0,0 +1,86 @@
...
+        <TableContainer component={Paper}>
+          <Table>
+            <TableHead>
+              <TableRow>
+                <TableCell>{t('Name')}</TableCell>
+                <TableCell>{t('Referral Date')}</TableCell>
+              </TableRow>
+            </TableHead>
+            <TableBody>
+              {(data?.contact as Contact)?.contactReferralsByMe.nodes.length >
```
OzzieOrca, line 65: This cast essentially just suppresses the error that there's no null check. Making sure we've actually got a definite value going into the `> 0` comparison is better.
```suggestion
              {data?.contact && data.contact.contactReferralsByMe.nodes.length >
```

OzzieOrca (dismissed):

src/components/Contacts/ContactDetails/ContactReferralTab/ContactReferralTab.tsx
```diff
@@ -0,0 +1,86 @@
...
+              0 ? (
+                data?.contact.contactReferralsByMe.nodes.map((referral) => (
+                  <TableRow key={referral.id}>
+                    <TableCell>{referral.referredTo.name}</TableCell>
+                    <TableCell>
+                      {DateTime.fromISO(referral.createdAt).toLocaleString()}
+                    </TableCell>
+                  </TableRow>
+                ))
```
TheNoodleMoose, line 74: Not a necessary change but you could destructure here.
```suggestion
                data?.contact.contactReferralsByMe.nodes.map(({id, referredTo, createdAt}) => (
                  <TableRow key={id}>
                    <TableCell>{referredTo.name}</TableCell>
                    <TableCell>
                      {DateTime.fromISO(createdAt).toLocaleString()}
                    </TableCell>
                  </TableRow>
                ))
```

src/components/Contacts/ContactDetails/ContactReferralTab/ContactReferralTab.tsx
```diff
@@ -0,0 +1,86 @@
...
+interface ContactReferralTabProps {
+  accountListId: string;
+  contactId: string;
+}
+
+export const ContactReferralTab: React.FC<ContactReferralTabProps> = ({
+  accountListId,
+  contactId,
+}) => {
+  const { data, loading } = useContactReferralTabQuery({
```
TheNoodleMoose, line 38: Should we have something for errors?

OzzieOrca: I'll make a Jira ticket. I think we can handle it globally.

src/components/Contacts/ContactDetails/ContactReferralTab/ContactReferralTab.test.tsx
OzzieOrca: Or continuing to import it works too I guess. We can disable it in another PR.

OzzieOrca (approved):

PR #76, opened 2021-04-30
MPDX-6972: Contact Tasks Tab

reldredge71:


src/components/Contacts/ContactDetails/ContactTaskRow/ContactTaskRow.tsx
```diff
@@ -0,0 +1,156 @@
...
+  color: theme.palette.text.primary,
+}));
+
+const TaskDescription = styled(Typography)(({ theme }) => ({
+  fontSize: 14,
+  color: theme.palette.text.primary,
+  marginLeft: theme.spacing(0.5),
+}));
+
+const Spacer = styled(Box)(({}) => ({ flex: 1 }));
```
OzzieOrca, line 36: Idk if this is the best pattern. You can do stuff like this instead:
```html
<div style="display: flex; justify-content: space-between;">
  <div>Left stuff</div>
  <div>Right stuff</div>
</div>
```
(not suggesting inline styles, just writing quick code)

Or maybe CSS Grid could be better here since it is more like a table.

src/components/Contacts/ContactDetails/ContactTaskRow/ContactTaskRow.test.tsx
```diff
@@ -0,0 +1,279 @@
...
+      },
+    });
+
+    const { queryByText } = render(
+      <MuiThemeProvider theme={theme}>
+        <ContactTaskRow task={task} />
+      </MuiThemeProvider>,
+    );
+
+    expect(queryByText(task.subject)).toBeInTheDocument();
```
OzzieOrca, line 33: I this is better because `findBy*` throws errors that show you more info about why it wasn't found instead of just null and `toBeVisible` to make sure it's actually on screen.
```suggestion
    expect(findByText(task.subject)).toBeVisible();
```
But what you have is ok too.

src/components/Contacts/ContactDetails/ContactTaskRow/TaskCompleteButton/TaskCompleteButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+  onClick: () => void;
+}
+
+export const TaskCompleteButton: React.FC<TaskCompleteButtonProps> = ({
+  isComplete,
+  onClick,
+}) => {
+  return (
+    <TaskRowWrap isComplete={isComplete} onClick={() => onClick()}>
+      <TaskCheckIcon data-testid="checkIcon" isComplete={isComplete} />
```
OzzieOrca, line 38:
```suggestion
      <TaskCheckIcon titleAccess={t('Check Icon')} isComplete={isComplete} />
```
would be better. You can do `findByRole('img', { name: 'Check Icon' })` to get at it.

src/components/Contacts/ContactDetails/ContactTasksTab/ContactTasksTab.tsx
```diff
@@ -0,0 +1,138 @@
...
+const TaskButton = styled(Button)(({ theme }) => ({
+  margin: theme.spacing(1),
+}));
+
+const AddTaskButtonIcon = styled(Add)(({ theme }) => ({
+  width: 20,
+  height: 20,
+  fontWeight: 600,
+  margin: theme.spacing(0.5),
+  color: '#2196F3',
```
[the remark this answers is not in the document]
reldredge71: I didn't see it in the `theme.ts` file.... but now that I look at it, it might be a default Google color, let me see if I can access it.

src/components/Contacts/ContactDetails/ContactTaskRow/TaskCommentsButton/TaskCommentsButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+  onClick: () => void;
+}
+
+export const TaskCommentsButton: React.FC<TaskCommentsButtonProps> = ({
+  isComplete,
+  numberOfComments = 0,
+  onClick,
+}) => {
+  return (
+    <TaskRowWrap onClick={() => onClick()}>
```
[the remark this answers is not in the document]
reldredge71: I was doing this because the `onClick` property expects a function with `Event` as a parameter...  and I wanted to avoid adding that to the callback function because it doesn't use it.  But I agree that this is not as clean to look at.

LizzyRandall (commented):
Looks good to me, the findByTestId does seem like a better name for the variable

src/components/Contacts/ContactDetails/ContactDetails.test.tsx
```diff
@@ -11,19 +13,21 @@ const onClose = jest.fn();
 
 describe('ContactDetails', () => {
   it('should show loading state', async () => {
-    const { getByTestId } = render(
+    const { queryByTestId } = render(
```
OzzieOrca, line 16: I think I misled you. 😞  Sorry... I meant `getBy` but I said use `findBy` instead of `queryBy` https://testing-library.com/docs/queries/about#:~:text=Summary%20Table has:

Type of Query | 0 Matches | 1 Match | >1 Matches | Retry (Async/Await)
-- | -- | -- | -- | --
Single Element |   |   |   |  
getBy... | Throw error | Return element | Throw error | No
queryBy... | Return null | Return element | Throw error | No
findBy... | Throw error | Return element | Throw error | Yes
Multiple Elements |   |   |   |  
getAllBy... | Throw | Return array | Return array | No
queryAllBy... | Return [] | Return array | Return array | No
findAllBy... | Throw | Return array | Return array | Yes

I think the takeaway from that is use `queryBy` when it might not be on the page, use `findBy` when it might not be on the page immediately, and `getBy` otherwise. I think we should use `getBy` by default unless more features are needed.

They all work, so it's not a big deal as long as you get the test working but it's probably best to use them as they were designed.

OzzieOrca (approved):
If you want to take another pass at using different queries, go for it but if you're done playing with the tests here, what've you've got is fine since they are passing.

src/components/Contacts/ContactDetails/ContactDetails.test.tsx
reldredge71: Gotcha, thank you for that clarification!  I can go through and use `getBy` instead, I'd rather stay consistent to our standards.  I'd also rather it not be asynchronous unless it needs to be.

OzzieOrca (approved):

PR #77, opened 2021-04-30
MPDX-6973 contact donation tab base

gyasistory:
https://jira.cru.org/browse/MPDX-6973

This is the Outline for the Contact Donations Tab

### DIFF
--- a/src/components/Contacts/ContactDetails/ContactDontationsTab/ConatctDonationsTab.test.tsx
+++ b/src/components/Contacts/ContactDetails/ContactDontationsTab/ConatctDonationsTab.test.tsx
@@ -1,0 +1,31 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { GqlMockedProvider } from '../../../../../__tests__/util/graphqlMocking';
+import { useGetContactDonationsQuery } from './ContactDonationsTab.generated';
+
+const accountListId = 'account-list-1';
+const contactId = 'contact-id-1';
+
+describe('ContactDonationsTab', () => {
+  it('test query', async () => {
+    const { result, waitForNextUpdate } = renderHook(
+      () =>
+        useGetContactDonationsQuery({
+          variables: {
+            accountListId: accountListId,
+            contactId: contactId,
+          },
+        }),
+      { wrapper: GqlMockedProvider },
+    );
+    await waitForNextUpdate();
+    expect(result.current.variables).toMatchInlineSnapshot(`
+      Object {
+        "accountListId": "account-list-1",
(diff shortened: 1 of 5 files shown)


src/components/Contacts/ContactDetails/ContactDontationsTab/ConatctDonationsTab.test.tsx
```diff
@@ -0,0 +1,31 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { GqlMockedProvider } from '../../../../../__tests__/util/graphqlMocking';
+import { useGetContactDonationsQuery } from './ContactDonationsTab.generated';
+
+const accountListId = 'account-list-1';
+const contactId = 'contact-id-1';
+
+describe('ContactDonationsTab', () => {
+  it('test query', async () => {
```
OzzieOrca, line 9: What are you testing here? Seems like you're just testing generated code instead of any app code. Good to know that it works but idk that it really helps build confidence that our app is working correctly since we didn't write this code.

src/components/Contacts/ContactDetails/ContactDontationsTab/ContactDonationsTab.tsx
```diff
@@ -0,0 +1,123 @@
...
+  PartnershipInfo = 'Partnership Info',
+}
+
+export const ContactDonationsTab: React.FC<ContactDontationsProp> = ({
+  accountListId,
+  contactId,
+}) => {
+  const { data, loading } = useGetContactDonationsQuery({
+    variables: { accountListId: accountListId, contactId: contactId },
+  });
```
OzzieOrca, line 56: Do you think we should just load the graph data with this query and make the sub tabs load their own data? Maybe it's all about the same and makes sense to share.

OzzieOrca (approved):

src/components/Contacts/ContactDetails/ContactDontationsTab/ContactDonationsTab.tsx
gyasistory: @OzzieOrca I was thinking this call would be for all of the data.  Since it is just one instance of contact.

src/components/Contacts/ContactDetails/ContactDontationsTab/ConatctDonationsTab.test.tsx
gyasistory: @OzzieOrca I am only testing the query here because the view doesn't contain much logic in the view.  Once I work on the other view I will ad more view-specific tests in them.

PR #78, opened 2021-05-04
MPDX-6973 : Contact donations tab graph

gyasistory:
https://jira.cru.org/browse/MPDX-6973

This is just for creating the graph for this section


src/components/Contacts/ContactDetails/ContactDontationsTab/DonationsGraph/DonationsGraph.tsx
```diff
@@ -0,0 +1,109 @@
...
+  });
+
+  const donationsForLastYear = donations?.nodes.filter((donation) => {
+    return (
+      DateTime.fromISO(donation.donationDate).toJSDate() <
+        DateTime.local().minus({ year: 1 }).toJSDate() ||
+      DateTime.fromISO(donation.donationDate).toJSDate() >
+        DateTime.local().minus({ year: 2 }).toJSDate()
+    );
+  });
```
OzzieOrca, line 54: This can often be something that's worth adding filters for server side. @soberstadt you have any thoughts?

You could do something like
```graphql
  contact(...){
    ...
    donationsForThisYear: donations(interval: '<ISO8601 Time Interval representing this year>') {
      nodes {
        ...ContactDonation
      }
    }
    donationsForLastYear: donations(interval: '<ISO8601 Time Interval representing last year>') {
      nodes {
        ...ContactDonation
      }
    }
  }
```

Also we may run into pagination issues here as only 25 nodes are returned by default and the API will only return a max of 50 nodes at a time if `first: 50` is specified. I haven't really thought about loading chart data with GraphQL before.

src/components/Contacts/ContactDetails/ContactDontationsTab/DonationsGraph/DonationsGraph.tsx
```diff
@@ -0,0 +1,109 @@
...
+}
+
+export const DonationsGraph: React.FC<DonationsGraphProps> = ({
+  donations,
+}) => {
+  const { t } = useTranslation();
+  const donationsForThisYear = donations?.nodes.filter((donation) => {
+    return (
+      DateTime.fromISO(donation.donationDate).toJSDate() >
+      DateTime.local().minus({ year: 1 }).toJSDate()
```
OzzieOrca, line 43: I don't think  you have to convert to JS date first but I could be wrong
```suggestion
      DateTime.fromISO(donation.donationDate) >
      DateTime.local().minus({ year: 1 })
```

Also you could create a Luxon Interval and call https://moment.github.io/luxon/docs/class/src/interval.js~Interval.html#instance-method-contains

src/components/Contacts/ContactDetails/ContactDontationsTab/DonationsGraph/DonationsGraph.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const donationsForLastYear = donations?.nodes.filter((donation) => {
+    return (
+      DateTime.fromISO(donation.donationDate).toJSDate() <
+        DateTime.local().minus({ year: 1 }).toJSDate() ||
+      DateTime.fromISO(donation.donationDate).toJSDate() >
+        DateTime.local().minus({ year: 2 }).toJSDate()
+    );
+  });
+
+  let convertedCurrency = '';
```
OzzieOrca, line 56: Looks like you are just pulling out the last donation in the array from either `donationsForThisYear` or `donationsForLastYear` and looking up the currency. Maybe that doesn't need to be in the loop since you just throw all the intermediate values away. Maybe there's a better way to look this up for a contact...

src/components/Contacts/ContactDetails/ContactDontationsTab/DonationsGraph/DonationsGraph.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  const contactDonationsMap = [...Array(12)].map((x, i) => {
+    const mapDate = DateTime.now().minus({ month: i });
```
OzzieOrca, line 59: The graph is reverse sorted because you are counting backwards here.

src/components/Contacts/ContactDetails/ContactDontationsTab/DonationsGraph/DonationsGraph.tsx
```diff
@@ -0,0 +1,109 @@
...
+      }
+      convertedCurrency = lastDonation.amount.convertedCurrency;
+    });
+
+    const data: { month: string; lastYear: number; thisYear: number } = {
+      month: mapDate.monthShort,
+      lastYear: lastYearCurrencyConvertedTotal,
+      thisYear: thisYearCurrencyConvertedTotal,
+    };
+    return data;
```
OzzieOrca, line 80:
```suggestion
   return {
      month: mapDate.monthShort,
      lastYear: lastYearCurrencyConvertedTotal,
      thisYear: thisYearCurrencyConvertedTotal,
    };
```

src/components/Contacts/ContactDetails/ContactDontationsTab/DonationsGraph/DonationsGraph.tsx
```diff
@@ -0,0 +1,109 @@
...
+    );
+  });
+
+  let convertedCurrency = '';
+
+  const contactDonationsMap = [...Array(12)].map((x, i) => {
+    const mapDate = DateTime.now().minus({ month: i });
+    let thisYearCurrencyConvertedTotal = 0;
+    let lastYearCurrencyConvertedTotal = 0;
+    donationsForThisYear?.forEach((thisDonation) => {
```
OzzieOrca, line 62: You could use a `reduce` instead of a `forEach`.

OzzieOrca (changes requested):

OzzieOrca (approved):

PR #81, opened 2021-05-07
MPDX-7011: Contact SearchBar

reldredge71:
Connect the SearchBox component to the query for the Contacts list


src/components/common/SearchBox/SearchBox.tsx
```diff
@@ -56,7 +54,6 @@ export const SearchBox: React.FC<SearchBoxProps> = ({
       startAdornment={
         <Icon path={mdiAccountSearch} title="Search contacts" size="24px" />
       }
-      value={searchTerm}
       placeholder={placeholder ?? t('Search')}
       onChange={(event) => {
         onChange(event.target.value);
```
TheNoodleMoose, line 59: @reldredge71 @OzzieOrca  Do you think we should debounce this ```onChange```? I could just see someone typing in a last name like ```Huffman``` and it refetching for each letter. With a debounce, we could limit the number of calls made.

reldredge71: @TheNoodleMoose yes that's a good idea.  I'm pretty sure what I have now will make an API call with each character added/taken away.

src/components/Contacts/ContactDetails/ContactTasksTab/ContactTasksTab.tsx
```diff
@@ -90,10 +84,14 @@ export const ContactTasksTab: React.FC<ContactTasksTabProps> = ({
...
   contactId,
 }) => {
-  const { data, loading } = useContactTasksTabQuery({
+  const { data, loading, refetch } = useContactTasksTabQuery({
     variables: { accountListId, contactId },
   });
 
+  const handleSearchTermChanged = (searchTerm: string) => {
+    refetch({ accountListId, contactId, searchTerm });
+  };
```
OzzieOrca, line 93: You shouldn't need to call `refetch`. The `useQuery` hook should see that its variables have changed  on rerender and refetch for you. It must use the `useEffect` dependency array under the hood or some sort of change detection.

src/components/Contacts/ContactsTable/ContactsTable.tsx
```diff
@@ -26,10 +26,14 @@ export const ContactsTable: React.FC<Props> = ({
...
   toggleFilterPanel,
 }: Props) => {
-  const { data, loading, error } = useContactsQuery({
+  const { data, loading, error, refetch } = useContactsQuery({
     variables: { accountListId },
   });
 
+  const handleSearchTermChanged = (searchTerm: string) => {
+    refetch({ accountListId, searchTerm });
+  };
```
OzzieOrca, line 35: Same here

src/components/common/SearchBox/SearchBox.tsx
OzzieOrca: Ya that's a great idea. This is an expensive query to load a bunch of contacts every time a character changes.

https://github.com/nkbt/react-debounce-input looks kinda nice after having scrolled through https://stackoverflow.com/questions/23123138/perform-debounce-in-react-js. A lot of ideas there could work but they all seem to have a bunch of code to add to each component. A lot of them seem to depend on lodash.debounce which is fine.

I didn't research this very long, maybe there are better ideas out there. I didn't see an easy way to debounce this on the Apollo side.

OzzieOrca (changes requested):

src/components/Contacts/ContactDetails/ContactTasksTab/ContactTasksTab.tsx
reldredge71: Okay, should I put `searchTerm` inside `useState` and use this method to update the state?

src/components/Contacts/ContactsHeader/ContactsHeader.test.tsx
```diff
@@ -195,12 +195,13 @@ describe('ContactFilters', () => {
...
-
     const textbox = getByRole('textbox');
 
     userEvent.type(textbox, searchText);
 
     expect(toggleFilterPanel).not.toHaveBeenCalled();
-    expect(onSearchTermChanged).toHaveBeenCalledWith(searchText);
+    setTimeout(() => {
+      expect(onSearchTermChanged).toHaveBeenCalledWith(searchText);
+    }, 500);
```
OzzieOrca, line 205: Idk if this is the right way to test it. The test will complete successfully before the timeout is fired so that expect is probably not called until after the test finishes.

src/components/Contacts/ContactsTable/ContactsTable.tsx
```diff
@@ -26,12 +26,14 @@ export const ContactsTable: React.FC<Props> = ({
   filterPanelOpen,
   toggleFilterPanel,
 }: Props) => {
-  const { data, loading, error, refetch } = useContactsQuery({
-    variables: { accountListId },
+  const [searchTerm, setSearchTerm] = useState<string | undefined>(undefined);
```
OzzieOrca, line 29: I think TS is happy with this which is slightly shorter. But same thing 🤷‍♂️ 
```suggestion
  const [searchTerm, setSearchTerm] = useState<string>();
```

src/components/Contacts/ContactsTable/ContactsTable.tsx
```diff
@@ -26,12 +26,14 @@ export const ContactsTable: React.FC<Props> = ({
...
+
+  const { data, loading, error } = useContactsQuery({
+    variables: { accountListId, searchTerm },
   });
 
   const handleSearchTermChanged = (searchTerm: string) => {
-    refetch({ accountListId, searchTerm });
+    setSearchTerm(searchTerm);
```
OzzieOrca: Can we just call `setSearchTerm` directly?

OzzieOrca (approved):

src/components/Contacts/ContactsHeader/ContactsHeader.test.tsx
reldredge71: gotcha... I'll try to see if there's a better way to wait for the debounce to finish before testing this.

src/components/Contacts/ContactsHeader/ContactsHeader.test.tsx
```diff
@@ -200,8 +200,9 @@ describe('ContactsHeader', () => {
     userEvent.type(textbox, searchText);
 
     expect(toggleFilterPanel).not.toHaveBeenCalled();
-    setTimeout(() => {
-      expect(onSearchTermChanged).toHaveBeenCalledWith(searchText);
-    }, 500);
+
+    await new Promise((resolve) => setTimeout(resolve, 500));
+
+    expect(onSearchTermChanged).toHaveBeenCalledWith(searchText);
```
OzzieOrca, line 206: This is better and actually runs the expect but I think using Jest's fake timers would be better so the test doesn't have to wait half a second doing nothing.
https://jestjs.io/docs/timer-mocks#advance-timers-by-time

reldredge71: Hmm from what I can tell, Jest fake timers don't work with lodash/debounce.  I was looking at this comment thread (https://github.com/facebook/jest/issues/3465), it looks like the options are to either use real timers along with `setTimeout` https://github.com/facebook/jest/issues/3465#issuecomment-335160011 or mocking lodash https://github.com/facebook/jest/issues/3465#issuecomment-351186130.

OzzieOrca: Ah good to know. This is fine.

OzzieOrca (approved):

PR #83, opened 2021-05-11
Modularize-GraphQL-Schema

TheNoodleMoose:


codegen-server.yml
```diff
@@ -4,10 +4,7 @@ hooks:
     - prettier --write
 generates:
   ./pages/api/graphql-rest.page.generated.ts:
-    schema:
-      - './pages/api/Schema/ContactFilters/contactFilters.graphql'
-      - './pages/api/Schema/ExportContacts/exportContacts.graphql'
-      - './pages/api/Schema/TaskAnalytics/taskAnalytics.graphql'
+    schema: ./pages/api/Schema/**/*.graphql
```
TheNoodleMoose, line 7: I think the ``` ''``` was causing issues with this. I believe you can use them for specific file paths, but when trying to use them with wildcards it seems to break things.

pages/api/Schema/ContactFilters/datahandler.ts
```diff
@@ -0,0 +1,67 @@
+import {
```
TheNoodleMoose, line 1: I was debating if this should live in its own separate file called ```dataHandler.ts``` (let me know if you can think of a better name), or if it should live in ```resolvers.ts```. I'm leaning more towards a separate file to help separate concerns but let me know what you guys think.

pages/api/graphql-rest.page.ts
```diff
@@ -11,17 +10,10 @@ import {
...
-import ExportContactsResolvers from './Schema/ExportContacts/resolvers';
-import ContactFiltersTypeDefs from './Schema/ContactFilters/typeDefs';
-import TaskAnalyticsTypeDefs from './Schema/TaskAnalytics/typeDefs';
-import ExportContactsTypeDefs from './Schema/ExportContacts/typeDefs';
+import { ContactFilterOption } from './graphql-rest.page.generated';
+import schema from './Schema';
+import getTaskAnalytics from './Schema/TaskAnalytics/dataHandler';
+import getContactFilters from './Schema/ContactFilters/datahandler';
 
 class MpdxRestApi extends RESTDataSource {
```
TheNoodleMoose, line 18: Also thinking of moving this out and into its own file. Let me know what you guys think.

pages/api/Schema/index.ts
```diff
@@ -0,0 +1,15 @@
+import { buildFederatedSchema } from '@apollo/federation';
+import ContactFiltersResolvers from './ContactFilters/resolvers';
+import ContactFiltersTypeDefs from './ContactFilters/typeDefs';
+import ExportContactsResolvers from './ExportContacts/resolvers';
+import ExportContactsTypeDefs from './ExportContacts/typeDefs';
+import TaskAnalyticsResolvers from './TaskAnalytics/resolvers';
+import TaskAnalyticsTypeDefs from './TaskAnalytics/typeDefs';
+
+const schema = buildFederatedSchema([
```
TheNoodleMoose, line 9: I thought it would be a good idea to have this in one place that we can just export.

pages/api/Schema/index.ts
```diff
@@ -0,0 +1,15 @@
+import { buildFederatedSchema } from '@apollo/federation';
+import ContactFiltersResolvers from './ContactFilters/resolvers';
+import ContactFiltersTypeDefs from './ContactFilters/typeDefs';
+import ExportContactsResolvers from './ExportContacts/resolvers';
+import ExportContactsTypeDefs from './ExportContacts/typeDefs';
+import TaskAnalyticsResolvers from './TaskAnalytics/resolvers';
+import TaskAnalyticsTypeDefs from './TaskAnalytics/typeDefs';
```
TheNoodleMoose, line 7: Also considering creating a ```index.ts``` for each folder that would export the typeDefs.ts, resolvers.ts, and the dataHandler.ts.

reldredge71: Yeah I think that would be a good idea... That might clean up the imports here.

pages/api/Schema/ContactFilters/datahandler.ts
reldredge71: Yeah I think separate is good... it looks like graphql-rest.page.ts uses this file, it's probably easier to import this into more places by keeping it separate from resolvers.ts.

pages/api/Schema/ContactFilters/datahandler.ts
```diff
@@ -0,0 +1,67 @@
...
+        const group = createFilterGroup(title);
+        response.push(group);
+        group.filters.push(filter);
+      }
+    },
+  );
+  return response;
+};
+
+export default getContactFilters;
```
OzzieOrca, line 67: Can we use named exports? I think they are easier to auto import and you can't mess up the spelling.

pages/api/Schema/ContactFilters/typeDefs.ts
```diff
@@ -0,0 +1,34 @@
+import { gql } from 'graphql-tag';
```
OzzieOrca, line 1: This file seems like a duplicate of the `.graphql` file. Is it needed? Can codegen make it? Or should we use this instead of the `.graphql` file?

OzzieOrca (changes requested):

pages/api/Schema/ContactFilters/typeDefs.ts
TheNoodleMoose: So we need ```.graphql``` file for codegen and then the ```typeDefs``` files are what is actually used for the logic to generate the schema. I can try again to use the ```.graphql``` file but I think I had issues with it if I remember correctly.

TheNoodleMoose: Actually, it does work! My error was probably before I got the wildcards working.

pages/api/Schema/index.ts
TheNoodleMoose: @OzzieOrca @reldredge71 Thoughts on this now that we no longer need the typeDefs file and we're using named exports? Could help to clean up imports, but then it's also just an extra file.

pages/api/Schema/index.ts
```diff
@@ -1,10 +1,10 @@
 import { buildFederatedSchema } from '@apollo/federation';
-import ContactFiltersResolvers from './ContactFilters/resolvers';
-import ContactFiltersTypeDefs from './ContactFilters/typeDefs';
-import ExportContactsResolvers from './ExportContacts/resolvers';
-import ExportContactsTypeDefs from './ExportContacts/typeDefs';
-import TaskAnalyticsResolvers from './TaskAnalytics/resolvers';
-import TaskAnalyticsTypeDefs from './TaskAnalytics/typeDefs';
+import { ContactFiltersResolvers } from './ContactFilters/resolvers';
+import ContactFiltersTypeDefs from './ContactFilters/contactFilters.graphql';
```
TheNoodleMoose, line 3: Only bad thing is, ```.graphql``` files don't seem to be able to be auto imported in. Had to manually type out the file name. But I guess that's better than an additional file.

OzzieOrca: Hmm. You could use the ts files with the typeDefs instead of the `.graphql` files if that's better. Codegen should be able to read from ts files with `gql` template literals.

TheNoodleMoose: I feel like codegen wasn't working with just the typedef ts files, but once again that could be a different issue I was experiencing. I'll give it another try and see what works.

TheNoodleMoose: @OzzieOrca Tried switching to just a typeDefs ts file and deleting the ```.graphql``` file. But once I did that it doesn't generate the types.

OzzieOrca: Hmm. What file pattern was in `schema` for codegen?

TheNoodleMoose: I changed both ```codegen-server.yml``` and ```codegen.yml``` to use ```schema: ./pages/api/Schema/**/*.ts``` instead of ```schema: ./pages/api/Schema/**/*.graphql```, but no luck.

OzzieOrca: `schema: ./pages/api/Schema/**/typeDefs.ts` seems to work if you want to go that route. I'm still not sure which is the better format. `.graphql` files are nice.

I thought I might just be able to generate the gql tag version with https://www.graphql-code-generator.com/docs/plugins/typescript-document-nodes but it didn't output anything new other than the gql import. Maybe it only works with operations...

TheNoodleMoose:
@OzzieOrca Does this look good to merge?

pages/api/Schema/index.ts
reldredge71: Yeah I think you're good now, the named exports helps.

OzzieOrca (approved):

reldredge71 (approved):

PR #87, opened 2021-05-17
M6977 : Donations tab   donations tab

gyasistory:


src/components/Contacts/ContactDetails/ContactDontationsTab/ContactDonationsList/ContactDonationsList.tsx
```diff
@@ -0,0 +1,138 @@
...
+      first: range,
+    },
+    notifyOnNetworkStatusChange: true,
+  });
+
+  const { t } = useTranslation();
+
+  return (
+    <Box>
+      {loading || networkStatus === 3 ? (
```
OzzieOrca, line 56: If you add `notifyOnNetworkStatusChange` from my other comment globally, you can just use `loading` here instead.

OzzieOrca, line 56: I think you want to show the loading indicator and the data at the same time because during `fetchMore` and `refetch` you can have both data that you don't want to unrender and a loading state. Maybe show the loading indicator below the content or something so it looks like more is loading.

src/components/Contacts/ContactDetails/ContactDontationsTab/ContactDonationsList/ContactDonationsList.tsx
```diff
@@ -0,0 +1,138 @@
...
+      ) : (
+        <>
+          <Table>
+            <TableHead>
+              <TableCell>{t('Date')}</TableCell>
+              <TableCell>{t('Amount')}</TableCell>
+              <TableCell>{t('Converted Amount')}</TableCell>
+            </TableHead>
+            {data?.contact !== null &&
+            data?.contact.donations.nodes !== undefined ? (
```
OzzieOrca, line 71: Idk if this is quite right but you should be able to simplify this check.
```suggestion
            {data?.contact.donations.nodes ? (
```

src/components/Contacts/ContactDetails/ContactDontationsTab/ContactDonationsList/ContactDonationsList.tsx
```diff
@@ -0,0 +1,138 @@
...
+                  </TableCell>
+                </TableRow>
+              ))
+            ) : (
+              <></>
+            )}
+          </Table>
+          {!loading &&
+          data?.contact.donations.pageInfo !== undefined &&
+          data?.contact.donations.pageInfo.hasNextPage === true ? (
```
OzzieOrca, line 97:
```suggestion
          data?.contact.donations.pageInfo.hasNextPage ? (
```
Add a `?.` if something is still potentially undefined here.

src/components/Contacts/ContactDetails/ContactDontationsTab/ContactDonationsList/ContactDonationsList.tsx
```diff
@@ -0,0 +1,138 @@
...
+                            ...prev.contact.donations.nodes,
+                            ...fetchMoreResult.contact.donations.nodes,
+                          ],
+                          pageInfo: fetchMoreResult.contact.donations.pageInfo,
+                          totalCount:
+                            fetchMoreResult.contact.donations.totalCount,
+                          totalPageCount:
+                            fetchMoreResult.contact.donations.totalPageCount,
+                        },
+                      },
```
OzzieOrca, line 123: You are probably only loading one root key in the query here but spreading every level is safer if more was added to the query in the future. Also by spreading `fetchMoreResult.contact.donations`, you don't need to manually set all the pagination metadata.
```suggestion
                      ...prev
                      ...fetchMoreResult,
                      contact: {
                        ...prev.contact,
                        ...fetchMoreResult.contact,
                        donations: {
                          ...prev.contact.donations,
                          ...fetchMoreResult.contact.donations,
                          nodes: [
                            ...prev.contact.donations.nodes,
                            ...fetchMoreResult.contact.donations.nodes,
                          ],
                        },
                      },
```

OzzieOrca (changes requested):

src/components/Contacts/ContactDetails/ContactDontationsTab/ContactDonationsList/ContactDonationsList.tsx
```diff
@@ -0,0 +1,135 @@
...
+  width: '100%',
+  height: '24px',
+  margin: theme.spacing(2, 0),
+}));
+
+export const ContactDonationsList: React.FC<ContactDonationsListProp> = ({
+  accountListId,
+  contactId,
+}) => {
+  const [range, setRange] = useState(10);
```
OzzieOrca, line 36: I don't think `range` is doing anything useful anymore...

src/components/Contacts/ContactDetails/ContactDontationsTab/ContactDonationsList/ContactDonationsList.tsx
```diff
@@ -0,0 +1,135 @@
...
+    data,
+    loading,
+    fetchMore,
+    networkStatus,
+  } = useContactDonationsListQuery({
+    variables: {
+      accountListId: accountListId,
+      contactId: contactId,
+    },
+    notifyOnNetworkStatusChange: true,
```
OzzieOrca, line 48: Can we set this globally?

src/components/Contacts/ContactDetails/ContactDontationsTab/ContactDonationsList/ContactDonationsList.tsx
```diff
@@ -0,0 +1,135 @@
...
+  });
+
+  const { t } = useTranslation();
+
+  return (
+    <Box>
+      {loading || networkStatus === 3 ? (
```
OzzieOrca, line 55: If `notifyOnNetworkStatusChange` is set globally:
```suggestion
      {loading ? (
```

src/components/Contacts/ContactDetails/ContactDontationsTab/ContactDonationsTab.graphql
```diff
@@ -26,6 +24,14 @@ query GetContactDonations($accountListId: ID!, $contactId: ID!) {
...
 
 fragment DonationsContact on DonationConnection {
+  pageInfo {
+    endCursor
+    hasNextPage
+    hasPreviousPage
+    startCursor
+  }
+  totalCount
+  totalPageCount
```
OzzieOrca, line 34: Any of these unused? Seems like some of them might be after your changes but maybe not.

OzzieOrca (changes requested):
Looks great! Couple small things. Thanks for working through and figuring out that pagination stuff :)

OzzieOrca (changes requested):
Looks good :)

Can you enable `notifyOnNetworkStatusChange` globally like I've mentioned? I think that would be good to have enabled for the whole project and you were using it here. Here's that comment again:

Can you add this: (Copied from https://github.com/CruGlobal/missionhub-react-native/blob/develop/src/apolloClient.ts#L110)
```
    defaultOptions: {
      watchQuery: {
        notifyOnNetworkStatusChange: true,
      },
    },
```
to https://github.com/CruGlobal/mpdx-react/blob/3a42be0bdd753957d562c7a76122b51780e6b8d5/src/lib/client.ts#L52-L58
[Docs](https://www.apollographql.com/docs/react/data/queries/#inspecting-loading-states)

apollo.config.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
     service: {
       name: 'MPDX',
       url: 'https://api.stage.mpdx.org/graphql',
+      notifyOnNetworkStatusChange: true,
```
OzzieOrca, line 6: This isn't where it goes...

It's a field on query options. But instead of specifying it on every query you can default the query options with https://www.apollographql.com/docs/react/api/core/ApolloClient/#defaultoptions.

Add
```
    defaultOptions: {
      watchQuery: {
        notifyOnNetworkStatusChange: true,
      },
    },
```
after https://github.com/CruGlobal/mpdx-react/blob/main/src/lib/client.ts#L57 instead of here.

I don't see anything about including it in `apollo.config.js` https://www.apollographql.com/docs/devtools/apollo-config/#clientservice.

OzzieOrca (changes requested):

OzzieOrca (approved):
🚀

PR #88, opened 2021-05-18
MPDX-6979: Update route when contact details panel changes state

OzzieOrca:


src/components/Layouts/Primary/TopBar/Items/ProfileMenu/ProfileMenu.tsx
```diff
@@ -72,7 +72,6 @@ const ProfileMenu = (): ReactElement => {
       state.user?.id !== data.user.id &&
       // TODO: Use fragments to ensure all required fields are loaded
       dispatch({ type: 'updateUser', user: data.user as User });
-    console.log(data);
```
TheNoodleMoose, line 75: I've been meaning to ask about this haha. Thanks for removing it 👍

pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  if (Array.isArray(accountListId)) {
+    throw new Error('accountListId should not be an array');
+  }
+  if (contactId !== undefined && !Array.isArray(contactId)) {
+    throw new Error('contactId should be an array or undefined');
+  }
+
+  useEffect(() => {
+    if (isReady) {
```
Comment (TheNoodleMoose) [inline pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx:27]: I looked up the docs for ```isReady``` and it mentioned ```Whether the router fields are updated client-side and ready for use.``` What does it mean specifically by ```router fields```? Does that mean the dynamic route params?

pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx
```diff
@@ -0,0 +1,90 @@
...
+    const { contactId: _, ...queryWithoutContactId } = query;
+    push(
+      id
+        ? {
+            pathname: '/accountLists/[accountListId]/contacts/[contactId]',
+            query: { ...queryWithoutContactId, contactId: id },
+          }
+        : {
+            pathname: '/accountLists/[accountListId]/contacts/',
+            query: queryWithoutContactId,
```
Comment (TheNoodleMoose) [inline pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx:50]: ```query``` is the data object that is passed between routes to provide data for that particular page/component right?

TheNoodleMoose (approved):

pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx
Comment (OzzieOrca) [reply pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx:50]: Ya `query` is all the routing params in the url (like `[accountListId]`) and any query params. The router parses the url and provides all the variables in `query`.

pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx
Comment (OzzieOrca) [reply pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx:27]: I wish there was a cleaner way to do this. `query` is an empty object during the first render due to rehydration. Which means you can't pass query to `useState`'s default value as it only takes a default value during render.

My understanding of the renders is:
1. At build time we do static site generation. At this point there's no url and no query params to extract from it so there is nothing in `query` for this render.
2. React rehydrates on the client but has the exact same data during the first client render so it's easy and quick for React to reconcile the existing DOM from build time with the DOM from the first render. Sounds like the goal is to make them identical.
3. React rerenders and starts handling data that is actually different on the client than it was during build time. This is where `useRouter()` finally returns some real data in `query`.

https://www.joshwcomeau.com/react/the-perils-of-rehydration/ is interesting and I think helped me understand what was going on. Not quite sure why there can't be a better solution though. Kinda breaks hooks that need a value on first render. You can't just return `null` or something before the hooks cuz React throws an error saying a different number of hooks were encountered on the subsequent render.

PR #90, opened 2021-05-18
Remove-Duplicate-Starred-Definition-in-TaskSchema

TheNoodleMoose:
I assumed that we did not need this to be required in the task schema, as the api will automatically set it to false upon task creation(?). It also looks like we aren't currently grabbing the ```starred``` field in the query ```GetTaskForTaskDrawer```, so it was preventing that update of the task. Not sure if that field needs to be added. It also wasn't set in the ```initialTask``` values we use when creating a new task, but I assume the api will set that field upon the task creation so it feels like extra work for the client to set it.


OzzieOrca (approved):
Looks fine to me. Do we even need starred in here? Does the form let you change the starred value or is that another mutation?

TheNoodleMoose:
It is needed because of the types from the ```TaskCreateInput``` and ```TaskUpdateInput```. But it's not needed in the form, either when creating or updating a task. There's another mutation called ```SetTaskStarred``` that uses ```updateTask``` to update that ```starred``` value.

PR #94, opened 2021-05-27
MPDX-6998-Edit-Contact-Tags

TheNoodleMoose:
I started the contact editing work by making tags editable. Was a little bit more than I expected so I decided to make an individual pr for this work. Also having issues with testing the mutation logic so could use help with that as well.


src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+    //         id: contactId,
+    //         tagList: [...contactTags, 'tag4'],
+    //         people: {
+    //           nodes: [],
+    //         },
+    //         name: 'Lname, Fname',
+    //       },
+    //     },
+    //   }),
+    // );
```
TheNoodleMoose, line 121: @OzzieOrca Not sure what's up with this test. I've set it up similar to other mutations test, but the value of ```readQuery``` is always null. Therefore, the ```writeQuery``` logic never fires. But while debugging the test, I can see data in the cache
Screen Shot 2021-05-27 at 3 06 53 PM

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
```diff
@@ -15,43 +24,181 @@ const ContactTagChip = styled(Chip)(({ theme }) => ({
...
+          }}
+        >
+          {({
+            values: { tag },
+            handleChange,
+            handleSubmit,
+            isSubmitting,
+          }): ReactElement => (
+            <form onSubmit={handleSubmit} noValidate>
+              <ContactTagInput
```
TheNoodleMoose, line 186: I'd appreciate it if you guys took a look at the storybook and let me know what you think of the design. In figma there's a design for the ```add tag``` text, but no design for actually adding a tag, so I improvised 🙂 I'm pretty happy with the results, but would like to hear others' feedback.

TheNoodleMoose: Here's the link to the specific [storybook](https://5f112cf86767b10022b5b991-pfwkalpnmq.chromatic.com/?path=/story/contacts-tab-contactdetailstab-contacttags--default). But for some reason the styling is different in the storybook... not sure why 🤷‍♂️ Here's a picture of it actually in the app:
Screen Shot 2021-05-27 at 4 04 27 PM
Screen Shot 2021-05-27 at 4 04 34 PM

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.test.tsx
```diff
@@ -41,11 +43,10 @@ describe('ContactTags', () => {
   });
 
   it('should add a tag', async () => {
-    const cache = new InMemoryCache();
+    const cache = new InMemoryCache({ addTypename: false });
```
TheNoodleMoose, line 46: I have this test now set up very similar to the commentList form test that I updated in this commit. Still not sure why it won't run as expected.

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.graphql
```diff
@@ -0,0 +1,17 @@
+mutation UpdateContactTags(
+  $accountListId: ID!
+  $contactId: ID!
+  $tagList: [String!]
```
OzzieOrca, line 4:
```suggestion
  $tagList: [String!]!
```

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.stories.tsx
```diff
@@ -2,37 +2,47 @@ import React, { ReactElement } from 'react';
...
-      <MuiThemeProvider theme={theme}>
-        <CssBaseline />
-        <ContactTags
-          contactId="contact_id"
-          contactTags={array('contactTags', ['help', 'something'])}
-        />
-      </MuiThemeProvider>
+      <GqlMockedProvider<UpdateContactTagsMutation>>
+        <MuiThemeProvider theme={theme}>
+          <CssBaseline />
```
OzzieOrca, line 21: This could be why some styling is different?

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
```diff
@@ -15,43 +24,189 @@ const ContactTagChip = styled(Chip)(({ theme }) => ({
...
+      const tagList = [...contactTags];
+      tagList.splice(index, 1);
+      try {
+        updateContactTags({
+          variables: {
+            accountListId,
+            contactId,
+            tagList,
+          },
+          update: (cache, { data: updateData }) => {
```
OzzieOrca, line 84: I think this whole update function is unnecessary. Apollo seems to handle updating the query fine on it's own. Probably because this isn't a relationship to another type in the cache and is just an array of strings.

You can `await` the update and have the snackbar call after.

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
```diff
@@ -15,43 +24,189 @@ const ContactTagChip = styled(Chip)(({ theme }) => ({
...
+        },
+        optimisticResponse: {
+          updateContact: {
+            contact: {
+              id: contactId,
+              tagList: [...contactTags, tag],
+            },
+          },
+        },
+        update: (cache, { data: updateData }) => {
```
OzzieOrca, line 138: Same here, I don't think you need the update function.

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
```diff
@@ -15,43 +24,189 @@ const ContactTagChip = styled(Chip)(({ theme }) => ({
...
+      resetForm();
+      updateContactTags({
+        variables: {
+          accountListId,
+          contactId,
+          tagList: [...contactTags, tag],
+        },
+        optimisticResponse: {
+          updateContact: {
+            contact: {
```
OzzieOrca, line 132: This was throwing runtime errors about `Invariant Violation: Missing field '__typename' `
```suggestion
            __typename: 'ContactUpdateMutationPayload',
            contact: {
              __typename: 'Contact',
```
solves it. Idk if there's a better way to do that.

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
OzzieOrca: Looks good to me. I haven't looked at storybook but can if you're still struggling with it there.

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
```diff
@@ -15,43 +24,189 @@ const ContactTagChip = styled(Chip)(({ theme }) => ({
...
+          onSubmit={onSubmit}
+        >
+          {({
+            values: { tag },
+            handleChange,
+            handleSubmit,
+            isSubmitting,
+          }): ReactElement => (
+            <form onSubmit={handleSubmit} noValidate>
+              <ContactTagInput
```
OzzieOrca, line 194: I think this is going to need some kind of autocomplete dropdown. Seems like in the existing app you can make up new arbitrary tags and then they are suggested when you start typing them on other contacts. Idk if it should be a new PR or not.

OzzieOrca (changes requested):

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
TheNoodleMoose: Ah, interesting. So unless I'm updating some other query that may contain data that was altered, I don't need to manually update the cache.

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
TheNoodleMoose: Do you think using an optimistic response is worth it? I've seen it in a few other places but wasn't sure if we wanted to continue to use it. Those other spots don't have the ```_typename``` as well so I'm curious if they are throwing a similar errror.

OzzieOrca: I think optimistic responses are a great feature. They are a little more work to write in maintain but hopefully it's not much. Are they strictly typed?

OzzieOrca (approved):

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
TheNoodleMoose: I think I'm going to merge this in and make a follow-up ticket to add the autocomplete.

PR #95, opened 2021-06-03
MPDX-6998-Delete-Contact

TheNoodleMoose:


src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.tsx
```diff
@@ -51,6 +85,25 @@ const ContactDetailLoadingPlaceHolder = styled(Skeleton)(({ theme }) => ({
...
+  textAlign: 'center',
+}));
+
+const CloseButton = styled(IconButton)(({ theme }) => ({
+  position: 'absolute',
+  right: theme.spacing(1),
+  top: theme.spacing(1),
+  color: theme.palette.text.primary,
+  '&:hover': {
+    backgroundColor: '#EBECEC',
```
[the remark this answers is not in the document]
TheNoodleMoose: Ah good point, yeah I'll use that instead

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.tsx
```diff
@@ -65,89 +118,189 @@ export const ContactDetailsTab: React.FC<ContactDetailTabProps> = ({
...
+        <DeleteContactTitle>
+          {t('Delete Contact')}
+          <CloseButton onClick={() => changeDialogOpen(false)}>
+            <CloseIcon titleAccess={t('Close')} />
+          </CloseButton>
+        </DeleteContactTitle>
+        <DialogContent dividers>
+          <DialogContentText>
+            {t(
+              'Are you sure you want to permanently delete this contact? Doing so will permanently delete this contacts information, as well as task history. This cannot be undone. If you wish to keep this information, you can try hiding this contact instead.',
```
LizzyRandall, line 193: Do you guys keep long strings like this in a constant file or something? I started doing this so they could all be in one spot just in case wording needed to get changed.

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.tsx
```diff
@@ -65,89 +118,189 @@ export const ContactDetailsTab: React.FC<ContactDetailTabProps> = ({
...
+        <ContactDetailSectionContainer>
+          <ContactDetailHeadingContainer>
+            <ContactDetailHeadingText variant="h6">
+              {t('Other')}
+            </ContactDetailHeadingText>
+            <ContactDetailHeadingIcon />
+          </ContactDetailHeadingContainer>
+          {loading || !data ? (
+            <>
+              <ContactDetailLoadingPlaceHolder variant="rect" />
```
LizzyRandall, line 286: So MPDx is using the material ui lab? Do you find any of the other libraries useful besides the skeleton? I've been thinking of adding it for the skeleton at least.

LizzyRandall (commented):
Looks good to me, just had a few questions.

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.tsx
TheNoodleMoose: We've used ```Autocomplete``` in a few spots. I haven't personally worked with it yet, but I have a ticket to add it to the contact tags input. The Skeleton component is really nice while waiting for data to load.

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.tsx
TheNoodleMoose: I believe this would live in a translations.json file and would have a specific key to match it with. But I'm not sure if we have that set up yet.

OzzieOrca: Tatai started this with some sort of extraction tooling in mind.
https://github.com/CruGlobal/mpdx-react/blob/main/package.json#L19
https://github.com/CruGlobal/mpdx-react/blob/main/i18next-parser.config.js

Idk if it's one of these or something else https://www.i18next.com/overview/plugins-and-utils#extraction-tools. I haven't tried running any of it yet.

I think it's similar to gettext where it just goes looking for every string inside `t('<string>')` and extracts those into a file to upload. That way all the strings are colocated with the code instead of having to keep some giant file updated.

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.tsx
```diff
@@ -65,90 +118,190 @@ export const ContactDetailsTab: React.FC<ContactDetailTabProps> = ({
   });
 
   const { t } = useTranslation();
+  const { enqueueSnackbar } = useSnackbar();
+  const { push, query } = useRouter();
+  const [deleteContact, { loading: deleting }] = useDeleteContactMutation();
+  const [dialogOpen, changeDialogOpen] = useState(false);
```
OzzieOrca, line 124: This seems like a very general name for a very specific modal.

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.tsx
```diff
@@ -65,90 +118,190 @@ export const ContactDetailsTab: React.FC<ContactDetailTabProps> = ({
...
+      });
+    } catch (error) {
+      enqueueSnackbar(error.message, { variant: 'error' });
+      throw error;
+    }
+  };
+
+  const renderDeleteContactDialog = () => {
+    return (
+      <Dialog
```
OzzieOrca, line 178: Are there any simpler modals? Should we make a component for a confirmation dialog so we don't have to have all this HTML and CSS in every component? Seems like it could take a title, body, action label, action callback, and maybe a cancel label if that needs to be customized too.

OzzieOrca (changes requested):

pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx
```diff
@@ -37,17 +37,17 @@ const ContactsPage: React.FC = () => {
     setFilterPanelOpen(!filterPanelOpen);
   };
 
-  const setContactFocus = (id?: string) => {
+  const setContactFocus = (id?: string, searchTerm?: string) => {
```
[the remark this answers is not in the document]
Comment (TheNoodleMoose) [reply pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx:40]: The search term is needed to update the cache for the ```ContactsQuery```. Yeah, I don't like this implementation and I'm glad to explore better options 🙂 I'll mess around with it more and see if there is a better way to do it.

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.tsx
```diff
@@ -51,6 +85,25 @@ const ContactDetailLoadingPlaceHolder = styled(Skeleton)(({ theme }) => ({
   margin: theme.spacing(2, 0),
 }));
 
+const DeleteContactTitle = styled(DialogTitle)(() => ({
+  textTransform: 'uppercase',
+  textAlign: 'center',
+}));
+
+const CloseButton = styled(IconButton)(({ theme }) => ({
+  position: 'absolute',
```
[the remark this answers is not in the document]
TheNoodleMoose: Umm.. I think I grabbed this from something else Tatai had done in the past and I've been using it all over. But I'm going to try to make the generic modal component like you suggested so I'll mess around with the styling there.

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.tsx
```diff
@@ -65,90 +118,190 @@ export const ContactDetailsTab: React.FC<ContactDetailTabProps> = ({
...
+              accountListId,
+              searchTerm,
+            },
+          };
+
+          const dataFromCache = cache.readQuery<ContactsQuery>(query);
+
+          if (dataFromCache) {
+            const data = {
+              contacts: {
```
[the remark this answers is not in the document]
TheNoodleMoose: Yeah, I may not need this... I'll double-check though.

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.tsx
TheNoodleMoose: Yeah, I'm going to make a generic modal component and switch this one out for that. Then I'll make a separate pr/ticket to update all the other modals to use the new generic one.

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.tsx
TheNoodleMoose: Haha true, I can update it.

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.tsx
TheNoodleMoose: @OzzieOrca Should we make a ticket to investigate this?

OzzieOrca: Ya it's work to do before deploying to production. I haven't captured it in Jira yet.

OzzieOrca (approved):
Any luck with the `searchTerm` stuff? If a bunch of things are going to need it for refreshing, maybe we should store the graphql variables for the contact filters in a context somewhere. I guess that can be refactoring for later. Seems like other components should be able to just read from it. I'm not quite following why you're setting it during the route change to contact details.

TheNoodleMoose:
@OzzieOrca @reldredge71 @LizzyRandall As I mentioned in the comments above, I went ahead and created a generic modal component to use throughout the app. I'd appreciate it you guys could take a look at the [component](https://github.com/CruGlobal/mpdx-react/blob/e3830d8cd101f6092c57469581768dc4d7d81403/src/components/common/Modal/Modal.tsx) and let me know what you guys think 🙂 I think what I have as far as props for the component is a good starting place for now, but open to suggestions/changes.

src/components/common/Modal/Modal.tsx
```diff
@@ -19,6 +19,9 @@ const ModalTitle = styled(DialogTitle)(() => ({
 }));
 
 const CloseButton = styled(IconButton)(({ theme }) => ({
+  position: 'absolute',
+  right: theme.spacing(1),
+  top: theme.spacing(1),
```
TheNoodleMoose, line 23: @OzzieOrca I had switched to using grids, but after pushing noticed everything was slightly off. Trying to get it perfect/exact was giving me a headache and felt like more work than just using position absolute. I can try again if you'd rather not use ```position: 'absolute'```

src/components/common/Modal/Modal.tsx
```diff
@@ -59,12 +57,16 @@ const Modal = ({
   const { t } = useTranslation();
   return (
     <Dialog open={isOpen} fullWidth={fullWidth} maxWidth={size}>
-      <ModalTitle>
-        {title}
-        <CloseButton onClick={() => handleClose()}>
-          <CloseIcon titleAccess={t('Close')} />
-        </CloseButton>
-      </ModalTitle>
+      <Grid container alignItems="center">
```
TheNoodleMoose, line 60: Was having issues with the grids, so I went back to ```position: absolute```. But I figured out the issue. I was missing this ```alignItems="center"```

TheNoodleMoose:
> Any luck with the `searchTerm` stuff? If a bunch of things are going to need it for refreshing, maybe we should store the graphql variables for the contact filters in a context somewhere. I guess that can be refactoring for later. Seems like other components should be able to just read from it. I'm not quite following why you're setting it during the route change to contact details.

I have not yet, but I'm going to take a look at it now and see if I can find a better way to handle it.

pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx
```diff
@@ -37,22 +37,41 @@ const ContactsPage: React.FC = () => {
...
         : {
             pathname: '/accountLists/[accountListId]/contacts/',
-            query: { ...queryWithoutContactId, searchTerm },
+            query: { ...queryWithoutContactId },
           },
     );
     setContactDetailsId(id);
   };
 
+  const setSearchTerm = (searchTerm?: string) => {
```
Comment (TheNoodleMoose) [inline pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx:56]: @OzzieOrca I think something like this is better than what I had before. Probably can still be improved but it works well for now.

pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx
```diff
@@ -47,12 +47,31 @@ const ContactsPage: React.FC = () => {
...
+              ...queryWithoutContactId,
+              contactId: contactDetailsId,
+              searchTerm,
+            },
+          }
+        : {
+            pathname: '/accountLists/[accountListId]/contacts/',
+            query: { ...queryWithoutContactId, searchTerm },
+          },
+    );
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx:72]: Let's get the contact id logic out of here.

```suggestion
    push({
      pathname,
      query: {
        ...query,
        searchTerm,
      },
    });
```

Needs
```ts
const { query, push, isReady, pathname } = useRouter();
```
above.

Setting the url could happen in the child...

OzzieOrca (changes requested):

pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx
```diff
@@ -47,12 +47,22 @@ const ContactsPage: React.FC = () => {
           }
         : {
             pathname: '/accountLists/[accountListId]/contacts/',
-            query: queryWithoutContactId,
+            query: { ...queryWithoutContactId },
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx:50]:
```suggestion
            query: queryWithoutContactId,
```

pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx
```diff
@@ -47,12 +47,22 @@ const ContactsPage: React.FC = () => {
...
           },
     );
     setContactDetailsId(id);
   };
 
+  const setSearchTerm = (searchTerm?: string) => {
+    push({
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx:57]: Do we want `push` or `replace`? Is this updating on every character press?

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.tsx
```diff
@@ -65,90 +98,187 @@ export const ContactDetailsTab: React.FC<ContactDetailTabProps> = ({
...
+                nodes: dataFromCache.contacts.nodes.filter(
+                  (contact) => contact.id !== deletedContactId,
+                ),
+              },
+            };
+            cache.writeQuery({ ...query, data });
+
+            push({
+              pathname: '/accountLists/[accountListId]/contacts',
+              query: { accountListId, searchTerm },
```
OzzieOrca, line 141: Do we want to pass all except `contactDetailsId`? Might be more future proof.

OzzieOrca (changes requested):

pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx
Comment (TheNoodleMoose) [reply pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx:57]: Yeah, it updates on every character press, but it's using the debounced input.

src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.tsx
TheNoodleMoose: So something like ``` const { contactId: _, searchTerm, ...queryWithoutContactId } = query;``` and then 
```
push({
     pathname: '/accountLists/[accountListId]/contacts',
     query: { searchTerm, ...queryWithoutContactId },
});
```

OzzieOrca: Ya. `searchTerm` should already be in `query` though.

TheNoodleMoose: I extract it out for the query to do the cache update
```
const query = {
    query: ContactsDocument,
    variables: {
        accountListId,
        searchTerm,
    },
};
```

pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx
Comment (TheNoodleMoose) [reply pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx:57]: Maybe ```replace``` would be better so the history isn't filled up?

TheNoodleMoose:
@OzzieOrca I think this is in a good spot. Like I mentioned, I'll do the additional cleanup for the modal in a separate pr(as long as you don't see anything urgent that needs to be addressed in this pr). Sorry for going out of scope here 🙂

pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx
Comment (OzzieOrca) [reply pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx]: 👍 I was talking with Spencer about contact filters and I think we want to eventually just store them in the url but don't want to clutter the history. So ya I think replace is a good choice.

OzzieOrca (approved):

PR #97, opened 2021-06-10
Cleanup-Modal-Component

TheNoodleMoose:
After discussion with others, I made some changes to the generic modal component. Also added comments for the props, which is pretty neat when viewing in VSCode.
Screen Shot 2021-06-10 at 10 27 14 AM

### DIFF
--- a/src/components/common/Modal/Modal.tsx
+++ b/src/components/common/Modal/Modal.tsx
@@ -1,5 +1,6 @@
 import React, { ReactElement } from 'react';
 import {
+  Button,
   Dialog,
   DialogActions,
   DialogContent,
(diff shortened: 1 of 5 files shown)


src/components/common/Modal/Modal.tsx
```diff
@@ -69,26 +68,7 @@ const Modal = ({
...
-              disabled={disableActionButtons}
-              variant="contained"
-              color="primary"
-              onClick={() => handleConfirm()}
-            >
-              {confirmText ? confirmText : t('Save')}
-            </Button>
-          </>
-        )}
+        {customActionSection ? customActionSection : null}
```
OzzieOrca, line 71: So by default there are no buttons? Is the only way out to click the X?

You could just make an "OK" button that is there by default that calls the same handler as the close button.

OzzieOrca (approved):

OzzieOrca (approved):

PR #100, opened 2021-06-11
Add 14 Month Report to GraphQL schema

OzzieOrca:
I tried to make the schema better. Let me know what you think :) Specifically:
1. Merged donor infos to be able to load both donation info and contact info at the same time.
2. Tried to remove all the converted stuff and make it conditional on what type of `currencyType` is requested.
3. Remove maximum which isn't used by the UI
4. Compute average and minimum totals for each currency group
5. Compute salaryCurrencyTotal as it's always used in tooltip and conditionally return it as the total

This thing is complex so I probably missed something maybe related to currency conversion. I tried to compare it to the existing one and I think it's close. I'll test it more when we have some UI to directly compare.

To test:
1. Sign in at https://mpdx-react-4zvrhx9do-cru.vercel.app/login.
2. Go to https://mpdx-react-4zvrhx9do-cru.vercel.app/api/graphql.
3. Run this query:
    ```gql
    query FourteenMonthReport {
      fourteenMonthReport(
        accountListId: "3c051f91-7806-4ce3-9132-314eaba7c5f4"
        currencyType: salary
      ) {
        currencyType
        salaryCurrency
        currencyGroups {
          currency
          totals {
            year
            months {
              month
              total
            }
            average
            minimum
          }
          contacts {
            id
            name
            total
            average
            minimum
            months {
              month
              total
              salaryCurrencyTotal
              donations {
                amount
                currency
                date
                paymentMethod
              }
            }
            accountNumbers
            lateBy30Days
            lateBy60Days
            pledgeAmount
            pledgeCurrency
            pledgeFrequency
            status
          }
        }
      }
    }
    ```
4. To switch from salary currency to partner currency, you can try changing `currencyType: salary` to `currencyType: donor`.

### DIFF
--- a/pages/api/Schema/TaskAnalytics/taskAnalytics.graphql
+++ b/pages/api/Schema/TaskAnalytics/taskAnalytics.graphql
@@ -13,8 +13,6 @@ type TaskAnalytics {
   updatedInDbAt: ISO8601DateTime!
 }
 
-scalar ISO8601DateTime
-
 type OverdueOrDueTodayTaskAnalytic {
   label: String!
   count: Int!
--- a/pages/api/Schema/reports/fourteenMonth/fourteenMonth.graphql
+++ b/pages/api/Schema/reports/fourteenMonth/fourteenMonth.graphql
@@ -11,10 +11,8 @@ enum FourteenMonthReportCurrencyType {
 }
 
 type FourteenMonthReport {
-  # id: ID!
   currencyType: FourteenMonthReportCurrencyType
   currencyGroups: [FourteenMonthReportCurrencyGroup!]!
-  ## months: [ISO8601Date!]!
   salaryCurrency: String!
 }
 
--- a/pages/api/Schema/reports/fourteenMonth/fourteenMonth.graphql
+++ b/pages/api/Schema/reports/fourteenMonth/fourteenMonth.graphql
@@ -27,6 +27,8 @@ type FourteenMonthReportCurrencyGroup {
 type FourteenMonthReportCurrencyGroupTotals {
   year: Float!
   months: [FourteenMonthReportCurrencyGroupTotalsMonth!]!
+  average: Float!
+  minimum: Float!
 }
 
 type FourteenMonthReportCurrencyGroupTotalsMonth {
@@ -36,10 +38,10 @@ type FourteenMonthReportCurrencyGroupTotalsMonth {
 
 type FourteenMonthReportContact {
   id: ID!
+  name: String!
   total: Float!
   average: Float!
   minimum: Float!
-  maximum: Float!
   months: [FourteenMonthReportContactMonths!]
   accountNumbers: [String!]!
   lateBy30Days: Boolean!
@@ -59,5 +61,6 @@ type FourteenMonthReportContactMonths {
 type FourteenMonthReportContactDonation {
   amount: Float!
   date: ISO8601Date!
+  currency: String!
   paymentMethod: String
 }
--- a/pages/api/Schema/reports/fourteenMonth/fourteenMonth.graphql
+++ b/pages/api/Schema/reports/fourteenMonth/fourteenMonth.graphql
@@ -55,6 +55,7 @@ type FourteenMonthReportContact {
 type FourteenMonthReportContactMonths {
   month: ISO8601Date!
   total: Float!
+  salaryCurrencyTotal: Float!
   donations: [FourteenMonthReportContactDonation!]!
 }
 
--- a/pages/api/Schema/reports/fourteenMonth/fourteenMonth.graphql
+++ b/pages/api/Schema/reports/fourteenMonth/fourteenMonth.graphql
@@ -1,0 +1,63 @@
+extend type Query {
+  fourteenMonthReport(
+    accountListId: ID!
+    currencyType: FourteenMonthReportCurrencyType!
+  ): FourteenMonthReport!
+}
+
+enum FourteenMonthReportCurrencyType {
+  salary
+  donor
+}
+
+type FourteenMonthReport {
+  # id: ID!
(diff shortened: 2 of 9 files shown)


pages/api/graphql-rest.page.ts
```diff
@@ -110,6 +118,25 @@ class MpdxRestApi extends RESTDataSource {
 
     return getTaskAnalytics(data);
   }
+
+  async getFourteenMonthReport(
```
OzzieOrca, line 122: @TheNoodleMoose Did you have any thoughts about how to get these out of the main file? I didn't explore it at all but this file will probably keep growing. I keep meaning to look into https://www.graphql-modules.com/ but haven't. Idk if it's compatible with `apollo-datasource-rest` which is what we haven't split up yet.

pages/api/Schema/scalarResolvers.ts
```diff
@@ -0,0 +1,19 @@
...
+  name: 'ISO8601Date',
+  description: 'An ISO 8601-encoded date',
+});
+
+const ISO8601DateTimeScalar = new GraphQLScalarType({
+  name: 'ISO8601DateTime',
+  description: 'An ISO 8601-encoded date time',
+});
+
+export const ScalarResolvers: Resolvers = {
```
OzzieOrca, line 14: I was getting some errors with the custom scalars and added some code that seems to have made them work. Not quite sure if there's a simpler solution.

pages/api/graphql-rest.page.ts
TheNoodleMoose: Yeah, I tried out some solution where I extracted the function to a separate file, and passed the ```this.get``` as an argument. If I remember it worked but didn't feel like the best solution. The issue with that solution is that the file is still growing, just a little slower haha. I think looking into graphql modules would be the next step to help improve this. A few articles I looked at recommended it, but of course, they had a completely different setup and were not using ```apollo-data-source-rest``` so I'm not sure of the compatibility.

pages/api/Schema/scalarResolvers.ts
TheNoodleMoose: So I think what we need to do is make a ```Schema/commonTypes``` folder and extract these custom scalars into that. Which is essentially what you did here lol. The error you got could have been because it was getting declared in more than one spot? https://github.com/CruGlobal/mpdx-react/blob/50024c25cfe7ac741daadfcf22b90a3c4e4796de/pages/api/Schema/TaskAnalytics/taskAnalytics.graphql#L16 Robert ran into a similar issue in his coaching questions pr, if I'm remembering correctly.

pages/api/Schema/TaskAnalytics/taskAnalytics.graphql
```diff
@@ -13,8 +13,6 @@ type TaskAnalytics {
   updatedInDbAt: ISO8601DateTime!
 }
 
-scalar ISO8601DateTime
```
TheNoodleMoose, line 16: Oh yeah, looks like it was because it was declared in more than one spot. I think extracting to a separate file like you did was a good idea. Question is, do we want a specific file/folder for custom types or just one common folder that holds everything? Not sure how many other custom scalars we actually have...

OzzieOrca: These are the only custom scalars that I know of... I think the default types are good enough for most of our use cases.

pages/api/graphql-rest.page.ts
OzzieOrca: Ya we should look into that a little more. Not urgent though.

pages/api/Schema/reports/fourteenMonth/fourteenMonth.graphql
```diff
@@ -0,0 +1,65 @@
...
+  average: Float!
+  minimum: Float!
+  months: [FourteenMonthReportContactMonths!]
+  accountNumbers: [String!]!
+  lateBy30Days: Boolean!
+  lateBy60Days: Boolean!
+  pledgeAmount: Float
+  pledgeCurrency: String
+  pledgeFrequency: String
+  status: String
```
TheNoodleMoose, line 50: Should this always return a status?

TheNoodleMoose (approved):

hotcakedev628 (approved):

pages/api/Schema/reports/fourteenMonth/fourteenMonth.graphql
```diff
@@ -0,0 +1,65 @@
...
+}
+
+type FourteenMonthReportContactMonths {
+  month: ISO8601Date!
+  total: Float!
+  salaryCurrencyTotal: Float!
+  donations: [FourteenMonthReportContactDonation!]!
+}
+
+type FourteenMonthReportContactDonation {
```
soberstadt, line 60: Why this new donation type instead of the already existing `Donation`?

OzzieOrca: Cuz I was making all new stuff. I'll look into reusing that... Good point.

pages/api/Schema/reports/fourteenMonth/fourteenMonth.graphql
OzzieOrca: Looks like currently in the schema, `Contact` has a status that is nullable. We're having to find the contact by id so if we don't find an id that matches, the contact could be null.

pages/api/Schema/reports/fourteenMonth/fourteenMonth.graphql
OzzieOrca: I don't think I have all the data to fill all the required fields of it ☹️ 
Screen Shot 2021-06-14 at 12 22 54 PM

PR #105, opened 2021-06-17
Mpdx 7037 14 month report salary

hotcakedev628:


hotcakedev628:
I think conditional rendering elements for `loading`, `error` statues of Report table should have the correct table structure like table->tr->td ...

If not, `React-DOM` errors will be occurred.

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,285 @@
...
+              <TableCell>
+                <Typography variant="h6">{salaryCurrency}</Typography>
+              </TableCell>
+              {currencyGroups?.map((currencyGroup) => {
+                const allYears = currencyGroup.totals.months.map(
+                  (month) => month.month.split('-')[0],
+                );
+                const monthCount = allYears.reduce(
+                  (year, count) => (
+                    (year[count] = (year[count] || 0) + 1), year
```
OzzieOrca, line 209: This is a comma expression right? I think it's a little hard to read. We could use spread instead:
```suggestion
                    {
                      ...year,
                      [count]: (year[count] || 0) + 1,
                    } 
```

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,285 @@
...
+            <TableRow>
+              <TableCell>
+                <Typography variant="h6">{salaryCurrency}</Typography>
+              </TableCell>
+              {currencyGroups?.map((currencyGroup) => {
+                const allYears = currencyGroup.totals.months.map(
+                  (month) => month.month.split('-')[0],
+                );
+                const monthCount = allYears.reduce(
+                  (year, count) => (
```
OzzieOrca, line 208: You're using these variables correctly but I think you've got your names flipped. The accumulator should be the first argument. Make sure you flip the names where you're using them too.
```suggestion
                  (count, year) => (
```

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,285 @@
...
+                    (year[count] = (year[count] || 0) + 1), year
+                  ),
+                  Object.create(null),
```
OzzieOrca, line 211:
```suggestion
                  {},
```
seems simpler.

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,285 @@
...
+                );
+
+                return Object.keys(monthCount).map((year) => {
+                  const count = monthCount[year];
```
OzzieOrca, line 215: Use `Object.entries` when you want both the key and the value
```suggestion
                return Object.entries(monthCount).map(([year, count]) => {
```

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,285 @@
...
+                  console.log('count, year-------------', count, year);
```
OzzieOrca, line 216: Still using this?

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,285 @@
...
+                )),
+              )}
+              <TableCell align="right">{t('Total')}</TableCell>
+            </TableRow>
+          </TableHead>
+          <TableBody>
+            {error && renderError()}
+            {loading
+              ? renderLoading()
+              : !(currencyGroups && currencyGroups.length > 0)
```
OzzieOrca, line 243:
```suggestion
              : !(currencyGroups?.length > 0)
```

OzzieOrca (changes requested):
Looks good :)

Do we want the loading/error stuff to show up in the table? Or should we just replace the table with an indicator/icon. I think the current plan is to handle GraphQL errors globally in an Apollo Link that shows a snackbar so maybe that one is less important to do something with.

A requirement I forgot to mention is we'd like the header to be sticky. Looks like there might be Material support for that.

If you want you could handle some of the additional functionality (like sticky header, printing, CSV, expanding, etc) in a new PR.

pages/accountLists/[accountListId]/reports/salaryCurrency.page.tsx
```diff
@@ -1,14 +1,27 @@
...
   const { t } = useTranslation();
   const accountListId = useAccountListId();
+  const router = useRouter();
+
+  useEffect(() => {
+    dispatch({
+      type: 'updateAccountListId',
+      accountListId: router.query.accountListId?.toString() ?? '',
+    });
+  }, []);
```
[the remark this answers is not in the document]
Comment (hotcakedev628) [reply pages/accountLists/[accountListId]/reports/salaryCurrency.page.tsx:24]: @OzzieOrca Yep, I think the hook `useAccountListId` is enough as well.

pages/accountLists/[accountListId]/reports/salaryCurrency.page.tsx
```diff
@@ -31,4 +44,18 @@ const SalaryReportPage = (): ReactElement => {
...
 };
 
+export const getServerSideProps: GetServerSideProps = async ({ req, res }) => {
+  const session = await getSession({ req });
+
+  if (!session?.user['token']) {
+    res.writeHead(302, { Location: '/' });
+    res.end();
+    return { props: {} };
+  }
```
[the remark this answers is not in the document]
Comment (hotcakedev628) [reply pages/accountLists/[accountListId]/reports/salaryCurrency.page.tsx:54]: @OzzieOrca That's a good point. We could build a hook named `useAuth` which connects and manages contexts specifically using JWT token for user auth. And also need to wrap the entire app with the Auth Provider.

I will look into the Onekey auth.

Is there any specific reason to use Onekey for the user authentication?

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,285 @@
...
+import PrintIcon from '@material-ui/icons/Print';
+import { CSVLink } from 'react-csv';
+import { DateTime } from 'luxon';
+import { useTranslation } from 'react-i18next';
+import {
+  FourteenMonthReportContact,
+  FourteenMonthReportCurrencyType,
+  // eslint-disable-next-line import/extensions
+} from 'graphql/types.generated';
+// eslint-disable-next-line import/extensions
```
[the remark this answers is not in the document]
hotcakedev628: @OzzieOrca Yeah, maybe we need to include `ts` files in the `tsconfig` since using `baseUrl` doesn't know the 'multiple dotted' extension file. I will have a look.

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,285 @@
...
 
 interface Props {
+  className?: string;
   accountListId: string;
+  title: string;
 }
 
-export const SalaryReportTable: React.FC<Props> = () => {
-  return <Box />;
+const useStyles = makeStyles(() => ({
```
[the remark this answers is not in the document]
hotcakedev628: @OzzieOrca Is there any specific reason to use the `styled` api better?

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,285 @@
...
+  root: {},
+  downloadCsv: {
+    color: 'inherit',
+    textDecoration: 'none',
+  },
+}));
+
+const applyPagination = (
```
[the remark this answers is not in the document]
hotcakedev628: Ok, that's pretty simple.

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,285 @@
...
+      currencyType: FourteenMonthReportCurrencyType.Salary,
+    },
+  });
+
+  const salaryCurrency = data?.fourteenMonthReport.salaryCurrency;
+  const currencyGroups = data?.fourteenMonthReport.currencyGroups;
+
+  useEffect(() => {
+    if (currencyGroups) {
+      setContacts(_.flatten(_.map(currencyGroups, 'contacts')));
```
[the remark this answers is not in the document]
hotcakedev628: @OzzieOrca Yeah, I agree with you. I used `lodash` after looking at the module has already installed.

pages/accountLists/[accountListId]/reports/salaryCurrency.page.tsx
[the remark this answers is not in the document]
Comment (OzzieOrca) [reply pages/accountLists/[accountListId]/reports/salaryCurrency.page.tsx:54]: I'm not sure what you mean by "Onekey".

We're using The Key (based on CAS) currently which is an internal SSO system. As an organization we are migrating to Okta so that change might be coming soon.

Ya getting this somewhere globally so we don't have to have auth redirect code in every page would be great.

src/components/Reports/SalaryReport/SalaryReportTable.tsx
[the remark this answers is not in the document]
OzzieOrca: We like it and want some consistency. Material v5 seem to be moving towards Emotion to power the CSS stuff so the styled API seems more in line with that. I think it's easier to maintain as when a component using the styled API is no longer used, it's obvious that a variable is unused. With the object styles, unused styles stick around easier.

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,238 @@
...
+        >
+          Loading
+        </Box>
+      </TableCell>
+    </TableRow>
+  );
+
+  const renderEmpty = () => (
+    <TableRow>
+      <TableCell colSpan={16}>No Data</TableCell>
```
OzzieOrca, line 94: Can you make this somewhat match the existing codebase?
Screen Shot 2021-06-18 at 4 30 14 PM
Some sort of icon (doesn't have to be that exact one), the message text, and the two action buttons. For "Connect Services" you can use the handoff link component since I don't think we're going to get to rewriting the settings section for a while. I created https://jira.cru.org/browse/MPDX-7053 for the add donations modal so you can just add a placeholder button for now.

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,238 @@
...
+                const allYears = currencyGroup.totals.months.map(
+                  (month) => month.month.split('-')[0],
+                );
+                const monthCount = allYears.reduce(
+                  (count, year) => ({
+                    ...count,
+                    [year]:
+                      ((count as { [key: string]: number })[year] || 0) + 1,
+                  }),
+                  {},
```
OzzieOrca, line 174:
```suggestion
                const monthCount = allYears.reduce<{ [key: string]: number }>(
                  (count, year) => ({
                    ...count,
                    [year]:
                      (count[year] || 0) + 1,
                  }),
                  {},
```

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,238 @@
...
+                );
+
+                return Object.entries(monthCount).map(([year, count]) => (
+                  <TableCell
+                    key={year}
+                    colSpan={count as number}
```
OzzieOrca, line 180: Should be typed now that the reduce is typed.
```suggestion
                    colSpan={count}
```

OzzieOrca (changes requested):
I'm not sure what happened to the donations data on staging. I've messaged Spencer about it.

hotcakedev628:
> I'm not sure what you mean by "Onekey".

Sorry, I was meaning "Thekey".

hotcakedev628:
@OzzieOrca Let me create another PR for building the Auth hook.

pages/accountLists/[accountListId]/reports/salaryCurrency.page.tsx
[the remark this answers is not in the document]
Comment (hotcakedev628) [reply pages/accountLists/[accountListId]/reports/salaryCurrency.page.tsx:54]: @OzzieOrca Sorry, that's my bad. I was meaning "The Key".
Let me work on the auth stuff.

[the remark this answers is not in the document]
Comment (wrandall22) [reply pages/accountLists/[accountListId]/reports/salaryCurrency.page.tsx:54]: I wonder if we shouldn't just skip The Key in the React version and go straight to Okta. I think we will migrate to Okta before the rewrite is done.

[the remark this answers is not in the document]
Comment (hotcakedev628) [reply pages/accountLists/[accountListId]/reports/salaryCurrency.page.tsx:54]: That would be great.

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,285 @@
...
+    variables: {
+      accountListId,
+      currencyType: FourteenMonthReportCurrencyType.Salary,
+    },
+  });
+
+  const salaryCurrency = data?.fourteenMonthReport.salaryCurrency;
+  const currencyGroups = data?.fourteenMonthReport.currencyGroups;
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
hotcakedev628: @OzzieOrca Btw, I guess this would be not needed if we don't filter the `data`.

src/components/Reports/SalaryReport/SalaryReportTable.tsx
hotcakedev628: @OzzieOrca There was already the component so I just updated a little bit.

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,285 @@
...
+  }, [currencyGroups]);
+
+  const renderLoading = () => (
+    <TableRow>
+      <TableCell colSpan={16}>
+        <Box
+          height="100%"
+          alignItems="center"
+          justifyContent="center"
+          bgcolor={colors.green[600]}
```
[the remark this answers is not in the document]
hotcakedev628: I used existing circular loading element for it.

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,285 @@
...
+import CodeIcon from '@material-ui/icons/Code';
+import PrintIcon from '@material-ui/icons/Print';
+import { CSVLink } from 'react-csv';
+import { DateTime } from 'luxon';
+import { useTranslation } from 'react-i18next';
+import {
+  FourteenMonthReportContact,
+  FourteenMonthReportCurrencyType,
+  // eslint-disable-next-line import/extensions
+} from 'graphql/types.generated';
```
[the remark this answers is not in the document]
hotcakedev628: @OzzieOrca I was not able to pull those types out of the query specific generated file.
I think they should be from the same file which includes the query.
https://github.com/CruGlobal/mpdx-react/blob/9680247e4f8448667c38b1a081b56f081fa6e4c5/src/components/Reports/SalaryReport/SalaryReportTable.tsx#L32
But I didn't have any luck to get those types from the file.

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,214 @@
...
+
+export const SalaryReportTable: React.FC<Props> = ({
+  className,
+  accountListId,
+  title,
+  ...rest
+}) => {
+  const classes = useStyles();
+  const { t } = useTranslation();
+  const [contacts, setContacts] = useState<FourteenMonthReportContact[]>([]);
```
OzzieOrca, line 64: Using this type works. It's kinda verbose but 🤷‍♂️ 
```suggestion
  const [contacts, setContacts] = useState<
    FourteenMonthReportQuery['fourteenMonthReport']['currencyGroups'][0]['contacts'][]
  >([]);
```

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,214 @@
...
+    variables: {
+      accountListId,
+      currencyType: FourteenMonthReportCurrencyType.Salary,
+    },
+  });
+
+  const salaryCurrency = data?.fourteenMonthReport.salaryCurrency;
+  const currencyGroups = data?.fourteenMonthReport.currencyGroups;
+
+  useMemo(() => {
```
OzzieOrca, line 76: `useMemo` should return a value that is cached. You shouldn't need to stick it in state anymore.

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,214 @@
...
+        </Grid>
+      </Box>
+      {loading ? (
+        <Loading loading />
+      ) : error ? (
+        <Notification type="error" message={error.toString()} />
+      ) : currencyGroups?.length === 0 ? (
+        <EmptyReport
+          title="You have received no donations in the last thirteen months"
+          subTitle="You can setup an organization account to import them or add a new donation."
```
OzzieOrca, line 141: Translate

OzzieOrca (changes requested):

src/components/Reports/EmptyReport/EmptyReport.tsx
```diff
@@ -0,0 +1,49 @@
...
+  flexDirection: 'column',
+  justifyContent: 'center',
+  alignItems: 'center',
+}));
+
+export const EmptyReport: React.FC<Props> = ({ title, subTitle }) => {
+  const { t } = useTranslation();
+
+  const addNewDonation = () => {
+    return;
```
OzzieOrca, line 29:
```suggestion
    // To be implemented by https://jira.cru.org/browse/MPDX-7053
    return;
```

src/components/Reports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -1,10 +1,219 @@
...
+export const SalaryReportTable: React.FC<Props> = ({
+  className,
+  accountListId,
+  title,
+  ...rest
+}) => {
+  const classes = useStyles();
+  const { t } = useTranslation();
+
+  const { data, loading, error } = useFourteenMonthReportQuery({
```
OzzieOrca, line 65: Since this is where the query is happening, can you move the `.graphql` file to live in the same folder as this file? Unless it's being shared or something.

OzzieOrca (changes requested):
Looks pretty good :)

Noticed a few things:
1. Table is spinning? Somehow it got the loading indicator spin animation applied to the table.
    

    https://user-images.githubusercontent.com/756501/123707517-af80bb80-d81e-11eb-94cd-86d93cdeaf22.mov


2. Center and add padding to loading indicator
3. Add padding to null state buttons
    Screen Shot 2021-06-28 at 2 12 39 PM


There are a few missing features but they can be tackled in separate PRs so we can get this one in:
1. Columns need to be sortable
2. Print button needs to work
3. Expand report button needs to work
4. Header needs to be sticky (can't tell if that's the case or not with the spinning)
5. CSV export is currently just a dump of the JSON. It needs to be formatted like the table/like the current production version.

src/components/Reports/FourteenMonthReports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -0,0 +1,330 @@
...
+                    <CodeIcon />
+                  </SvgIcon>
+                }
+                onClick={handleExpandToggle}
+              >
+                {t(
+                  `${isExpanded ? 'Hide' : 'Expand'}${
+                    isMobile ? '' : ' Partner Info'
+                  }`,
+                )}
```
OzzieOrca, line 211: Translation keys can't be dynamic. You can just flip where you're calling `t`.
```suggestion
                {isExpanded ? t('Hide') : t('Expand')}
                {isMobile ? '' : t('Partner Info')}
```

OzzieOrca (changes requested):
Looks like you've got some status checks failing.

src/components/Layouts/SidePanelsLayout.tsx
```diff
@@ -2,6 +2,10 @@ import { FC, ReactElement } from 'react';
 import { Box, styled, Theme } from '@material-ui/core';
 import { Breakpoint } from '@material-ui/core/styles/createBreakpoints';
 
+type ScrollBoxProps = {
+  isscroll?: 1 | 0;
```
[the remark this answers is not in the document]
hotcakedev628: As I mentioned in the last scrum, `isScrollable` will face `React-DOM` issue until we use the latest version of styled-components. That's why I used all lowercased props and the other numbers rather than `boolean` type.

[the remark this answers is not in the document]
hotcakedev628: If we use the latest version of styled-components, we can pass the props as `$isScrollable`.
Please refer to the [transient props](https://styled-components.com/docs/api#transient-props).

src/components/Reports/FourteenMonthReports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -0,0 +1,330 @@
...
+): (
+  a: { [key in Key]: number | string },
+  b: { [key in Key]: number | string },
+) => number {
+  return order === 'desc'
+    ? (a, b) => descendingComparator(a, b, orderBy)
+    : (a, b) => -descendingComparator(a, b, orderBy);
+}
+
+function stableSort<T>(array: T[], comparator: (a: T, b: T) => number) {
```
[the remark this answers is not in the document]
hotcakedev628: My code is came from here. https://material-ui.com/components/tables/#sorting-amp-selecting
Anyways, I updated my codes using `localeCompare`. :)

src/components/Reports/FourteenMonthReports/TableHead/TableHead.tsx
```diff
@@ -12,7 +12,13 @@ import {
 import { FourteenMonthReportQuery } from '../GetFourteenMonthReport.generated';
 import { TableHeadCell } from './TableHeadCell/TableHeadCell';
 
+export type Unarray<T> = T extends Array<infer U> ? U : T;
+export type Contacts = FourteenMonthReportQuery['fourteenMonthReport']['currencyGroups'][0]['contacts'];
+export type Contact = Contacts[number];
+export type Months = Contact['months'];
+export type Month = Unarray<Months>;
```
OzzieOrca, line 19: Is this the same as `export type Month = Months[0];`?

Can you use `ElementOf`? https://github.com/krzkaczor/ts-essentials#elementof-type I think it's the same and I like the name better. I've already brought that `ts-essentials` library in for something else.

We seem to need to do this a lot. I kinda forgot about these helpers. Might be useful elsewhere too. Thanks for the reminder :)

src/components/Reports/FourteenMonthReports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -0,0 +1,362 @@
...
+                  a['months'][orderBy]['total'].toString(),
+                  undefined,
+                  {
+                    numeric: true,
+                  },
+                );
+            }
+          }
+        }
+      });
```
OzzieOrca, line 172: The TS error is you don't have default case. There are logic paths which don't specifically return something and therefore return `undefined`. You could add some default case at the bottom like `return 0` or something.

This is a lot of similar code. I might clean it up like this:
```suggestion
      const getSortValue = (contact: Contact) =>
        (typeof orderBy === 'number'
          ? contact['months']?.[orderBy]['total'].toString()
          : contact[orderBy]) ?? contact.name;

      return contacts.sort((a, b) => {
        const compare = getSortValue(a).localeCompare(
          getSortValue(b),
          undefined,
          {
            numeric: true,
          },
        );

        return order === 'asc' ? compare : -compare;
      });
```

OzzieOrca (changes requested):

src/components/Reports/FourteenMonthReports/PartnerReport/PartnerReportTable.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+            accountListId={accountListId}
+            isNavListOpen={true}
+            title={title}
+            onNavListToggle={onNavListToggle}
+          />
+        </GqlMockedProvider>
+      </ThemeProvider>,
+    );
+
+    expect(getByText('test title')).toBeInTheDocument();
```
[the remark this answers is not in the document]
hotcakedev628: It is just a try to avoid the deployment errors. Let me update the test with the partner report page.

src/components/Reports/FourteenMonthReports/TableHead/TableHead.tsx
hotcakedev628: Yeah, I will use the "essential type" rather use `Month[0]`.
But I guess it needs a little tricky since it has `Maybe` type for the `Months`.
Thank you. :)

OzzieOrca (approved):
Looks good! You can merge it. Printing looks a little cut off and the expanded columns need to gain status, commitment, avg, and min but feel free to address those in a followup PR.

src/components/Reports/FourteenMonthReports/SalaryReport/SalaryReportTable.tsx
```diff
@@ -232,6 +233,7 @@ export const SalaryReportTable: React.FC<Props> = ({
                 </DownloadCsvLink>
               </Button>
               <Button
+                data-testid="PrintButton"
```
OzzieOrca, line 236: You can set `titleAccess` on the icon instead of having to resort to a test id.

PR #112, opened 2021-06-29
Mpdx 7035 donations report

rathclayton1:
Build donations page UI


wrandall22:
Commits/History is much more solid in this PR, great improvement!

wrandall22:
@reldredge71 is there a reason we're storing the yarn dependencies in the repo? It seems odd to me since they are brought in during build.

src/components/Reports/DonationsReport/DonationsReportTable.tsx
```diff
@@ -1,10 +1,242 @@
...
+    const row = params.row as Donation;
+
+    return <Typography>{row.convertedAmount + ' ' + row.currency}</Typography>;
+  };
+  const foreignAmount = (params: GridCellParams) => {
+    const row = params.row as Donation;
+    return row.foreignCurrency !== row.currency ? (
+      <Typography>{row.foreignAmount + ' ' + row.foreignCurrency}</Typography>
+    ) : (
+      <></>
```
reldredge71, line 72: Could this be `null`?

reldredge71:
@wrandall22 Maybe we need to add them to the .gitignore?  @OzzieOrca What do you think?

src/components/Reports/DonationsReport/DonationsReportTable.tsx
```diff
@@ -1,10 +1,242 @@
...
+}
 
 interface Props {
+  data: Donation[];
   accountListId: string;
 }
 
-export const DonationsReportTable: React.FC<Props> = () => {
-  return <Box />;
+const Data = styled(Box)(({ theme }) => ({
```
reldredge71, line 41: Maybe rename this `DataTable`?

reldredge71 (changes requested):

reldredge71:
@rathclayton1 Some other things:
- It looks like there are some merge conflicts here.  This happens when two PRs make changes to the same files, and Git is confused about which changes to keep.  Could you resolve these conflicts?  You can do this by going to VSCode, do the command `git merge origin/main` in the terminal, and then go to the files with conflicts and clean them up.  VSCode should guide you through choosing which changes to keep.  After you fix those and save the changes, make a commit and then push to origin.  If you get confused, let me know and I can guide you through it.
- Could you add some simple tests to the donations report files?  Some of the checks here won't pass unless you have some unit tests written out.  For now, you can look at the test files I created for the Expected Monthly Totals files for inspiration, I can go over what unit tests are when we meet on Friday.

OzzieOrca:
@wrandall22 We have https://yarnpkg.com/features/zero-installs setup for this project. I'm open to discuss the merits of using it but it is an approved Yarn workflow.

package.json
```diff
@@ -32,9 +32,11 @@
     "@babel/runtime": "^7.12.13",
     "@date-io/luxon": "1.x",
     "@material-ui/core": "^4.11.3",
+    "@material-ui/data-grid": "^4.0.0-alpha.32",
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.57",
     "@material-ui/pickers": "^3.2.10",
+    "@material-ui/styles": "^4.11.4",
```
OzzieOrca, line 39: Doesn't look like you are using this specifically. It should be imported by other Material UI packages.
```suggestion
```

pages/accountLists/[accountListId]/reports/donations.page.tsx
```diff
@@ -3,12 +3,16 @@ import Head from 'next/head';
...
+
 import Loading from '../../../../src/components/Loading';
 import { useAccountListId } from '../../../../src/hooks/useAccountListId';
+import { GetDashboardQuery } from '../../GetDashboard.generated';
 
-const DonationsReportPage = (): ReactElement => {
+interface Props {
+  data: GetDashboardQuery;
+}
+const DonationsReportPage = ({ data }: Props): ReactElement => {
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/reports/donations.page.tsx:15]: I think this is a better way to type components. You get the return value and the `children` prop for free.
```suggestion
const DonationsReportPage: React.FC<Props> = ({ data }) => {
```

src/components/Reports/DonationsReport/DonationsReportTable.tsx
```diff
@@ -1,10 +1,242 @@
...
+}));
+export const DonationsReportTable: React.FC<Props> = ({
+  data,
+  /*accountListId,*/
+}) => {
+  const { t } = useTranslation();
+  const accountCurrency = data[0].currency;
+  const link = (params: GridCellParams) => (
+    <Typography>
+      <Link href="contact.html">{params.value}</Link>
```
OzzieOrca, line 59: Is this a valid link?

src/components/Reports/DonationsReport/DonationsReportTable.tsx
```diff
@@ -1,10 +1,242 @@
...
+
+  const totalDonations = data.reduce((total, current) => {
+    return total + current.convertedAmount;
+  }, 0);
+
+  const calculateForeignTotals = () => {
+    const foreignTotals: {
+      [name: string]: { convertedTotal: number; foreignTotal: number };
+    } = {};
+    data.forEach((donation) => {
```
OzzieOrca, line 146: I think a `.reduce` could make this cleaner.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce

src/components/Reports/DonationsReport/DonationsReportTable.tsx
```diff
@@ -1,10 +1,242 @@
...
+      <Box style={{ display: 'flex', margin: 8 }}>
+        <Typography variant="h6">{title}</Typography>
+        <Button
+          style={{ marginLeft: 'auto' }}
+          variant="contained"
+          startIcon={<ChevronLeftIcon />}
+          size="small"
+          onClick={() => setPrevMonth()}
+        >
+          Previous Month
```
OzzieOrca, line 176: Translate
```suggestion
          t('Previous Month')
```

src/components/Reports/DonationsReport/DonationsReportTable.tsx
```diff
@@ -1,10 +1,242 @@
...
+        </Button>
+        <Button
+          variant="contained"
+          endIcon={<ChevronRightIcon />}
+          size="small"
+          onClick={() => setNextMonth()}
+          disabled={hasNext}
+        >
+          Next Month
```
OzzieOrca, line 185: Translate

src/components/Reports/DonationsReport/DonationsReportTable.tsx
```diff
@@ -1,10 +1,242 @@
...
+  const accountCurrency = data[0].currency;
+  const link = (params: GridCellParams) => (
+    <Typography>
+      <Link href="contact.html">{params.value}</Link>
+    </Typography>
+  );
+  const amount = (params: GridCellParams) => {
+    const row = params.row as Donation;
+
+    return <Typography>{row.convertedAmount + ' ' + row.currency}</Typography>;
```
OzzieOrca, line 65: Use JSX for string interpolation here
```suggestion
    return <Typography>{row.convertedAmount} {row.currency}</Typography>;
```

Between JSX and ES6 template literals, we shouldn't need to reach for ` + ` to concatenate strings anymore.

src/components/Reports/DonationsReport/DonationsReportTable.tsx
```diff
@@ -1,10 +1,242 @@
...
+  };
+  const foreignAmount = (params: GridCellParams) => {
+    const row = params.row as Donation;
+    return row.foreignCurrency !== row.currency ? (
+      <Typography>{row.foreignAmount + ' ' + row.foreignCurrency}</Typography>
```
OzzieOrca, line 70:
```suggestion
      <Typography>{row.foreignAmount} {row.foreignCurrency}</Typography>
```

src/components/Reports/DonationsReport/DonationsReportTable.tsx
```diff
@@ -1,10 +1,242 @@
...
+      headerName: t('Appeal'),
+      width: 130,
+      renderCell: button,
+    },
+  ];
+  const isEmpty = data.length === 0;
+
+  const [time, setTime] = useState(DateTime.now().startOf('month'));
+
+  const title = time.monthLong + ' ' + time.year;
```
OzzieOrca, line 126:
```suggestion
  const title = `${time.monthLong} ${time.year}`;
```

src/components/Reports/DonationsReport/DonationsReportTable.tsx
```diff
@@ -1,10 +1,242 @@
...
+            autoHeight
+            disableSelectionOnClick
+            hideFooter
+          />
+          <Table>
+            {Object.entries(foreignTotals).map(([currency, total]) => (
+              <TableRow key={currency}>
+                <TableCell style={{ width: 375 }}>
+                  <Typography style={{ float: 'right', fontWeight: 'bold' }}>
+                    {t('Total ' + currency + ' Donations:')}
```
OzzieOrca, line 204: Translation keys can't be dynamic. Some tool is going to go through and pull out all the strings that are inside `t('')` and the value of `currency` isn't known at compile time.

Use interpolation instead https://www.i18next.com/translation-function/interpolation
```suggestion
                    {t('Total {{currency}} Donations:', { currency })}
```

src/components/Reports/DonationsReport/DonationsReportTable.tsx
```diff
@@ -1,10 +1,242 @@
...
+                  </Typography>
+                </TableCell>
+                <TableCell style={{ width: 160 }}>
+                  <Typography style={{ float: 'left', fontWeight: 'bold' }}>
+                    {' ' + total.convertedTotal + ' ' + accountCurrency}
+                  </Typography>
+                </TableCell>
+                <TableCell style={{}}>
+                  <Typography style={{ float: 'left', fontWeight: 'bold' }}>
+                    {' ' + total.foreignTotal + ' ' + currency}
```
OzzieOrca, line 214: Can you just use JSX for these instead of `+`? The left space should probably be done with CSS padding instead of extra characters.

src/components/Reports/DonationsReport/DonationsReportTable.tsx
```diff
@@ -1,10 +1,242 @@
...
+                </Typography>
+              </TableCell>
+
+              <TableCell />
+            </TableRow>
+          </Table>
+        </Data>
+      ) : (
+        <EmptyDonationsTable
+          title={'No donations received in ' + time.monthLong + ' ' + time.year}
```
OzzieOrca, line 237: Needs translation and interpolation
```suggestion
          title={t('No donations received in {{month}} {{year}}', { month: time.monthLong, year: time.year })}
```

OzzieOrca (changes requested):

reldredge71:
@rathclayton1 It looks like the eslint, typescript, prettier and other checks are still failing.  Could you run the commands `yarn lint` and `yarn lint:ts` to see what the issue is?

src/components/Reports/DonationsReport/Table/DonationsReportTable.tsx
```diff
@@ -0,0 +1,252 @@
...
+        ? theme.palette.common.white
+        : theme.palette.cruGrayLight.main,
+  },
+}));
+export const DonationsReportTable: React.FC<Props> = ({
+  data,
+  accountListId,
+}) => {
+  const { t } = useTranslation();
+  const accountCurrency = data[0].currency;
```
reldredge71, line 55: This is not the cleanest way to get the currency... I hope that when we work with real data from GraphQL, there will be an easier way to access this, separate from the donations array.  But for now we can keep this.

src/components/Reports/DonationsReport/Table/DonationsReportTable.tsx
```diff
@@ -0,0 +1,252 @@
...
+    const row = params.row as Donation;
+    return (
+      <Typography>
+        <Link href={`../../${accountListId}/contacts/${row.partnerId}`}>
+          {params.value}
+        </Link>
+      </Typography>
+    );
+  };
+  const amount = (params: GridCellParams) => {
```
reldredge71, line 66: Could you add spaces between these methods? It makes it a little bit easier to read.

src/components/Reports/DonationsReport/MonthlyActivity/MonthlyActivitySection.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import DonationHistories from '../../../Dashboard/DonationHistories';
+import { GetDashboardQuery } from '../../../../../pages/accountLists/GetDashboard.generated';
+
+interface Props {
+  data: GetDashboardQuery;
+}
+
```
reldredge71, line 8: Could you add a comment here so that other developers know why some of this is commented out?

```
\\TODO: Query for actual data and load into DonationHistories
```

reldredge71 (approved):

PR #115, opened 2021-07-12
Bring back @material-ui/styles

reldredge71:
@material-ui/data-grid depends on @material-ui/styles.  Having @material-ui/styles caused an issue before, but maybe that was only related to to version of it that was being imported?


OzzieOrca (approved):

OzzieOrca (approved):
Looks like the main branch's tests are broken without this?

hotcakedev628:
I think the rest of codecov issues caused by tests since they are not fully covered.

reldredge71:
@OzzieOrca I don't think I saw breaking tests... although the donations page throws an error when you run the app.  This solves that issue.

PR #116, opened 2021-07-12
MPDX-6976: Contact Details - Notes tab

j2trumpet:
I got a component created! However, I have no idea how to hook it up to data saving and retrieval. I imagine it has something to do with the `accountListId={accountListId} contactId={contactId}` props that are being passed to the other contact info components, but even just trying to add them to my component threw an error. I also don't know if there are any styles that you guys have been applying to form fields other than the default MUI styling. Can you guys help me with finishing this up?

### DIFF
--- a/src/components/Contacts/ContactDetails/ContactNotesTab/ContactNotesTab.graphql
+++ b/src/components/Contacts/ContactDetails/ContactNotesTab/ContactNotesTab.graphql
@@ -20,6 +20,7 @@ mutation UpdateContactNotes(
     contact {
       id
       notes
+      notesSavedAt
     }
   }
 }
--- a/src/components/Contacts/ContactDetails/ContactNotesTab/ContactNotesTab.graphql
+++ b/src/components/Contacts/ContactDetails/ContactNotesTab/ContactNotesTab.graphql
@@ -1,0 +1,24 @@
+query GetContactNotes($contactId: ID!, $accountListId: ID!) {
+  contact(id: $contactId, accountListId: $accountListId) {
+    id
+    notes
+  }
+}
+
+mutation UpdateContactNotes(
+  $contactId: ID!
+  $accountListId: ID!
+  $notes: String!
+) {
+  updateContact(
+    input: {
+      accountListId: $accountListId
+      attributes: { id: $contactId, notes: $notes }
+    }
+  ) {
+    contact {
+      id
+      notes
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


src/components/Contacts/ContactDetails/ContactNotesTab/ContactNotesTab.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { TextField } from '@material-ui/core';
+
+export const ContactNotesTab: React.FC = () => {
+  return (
+    <TextField
+      fullWidth
+      multiline
+      placeholder="Add contact notes"
```
OzzieOrca, line 9: Since this is a constant string, it should be setup for translation into other languages.

Add this to the top
```ts
```
call the `useTranslation` hook in the component to get access to the `t` function
```ts
```
and then call the `t` function with the string to be translated
```suggestion
      placeholder={t('Add contact notes')}
```

OzzieOrca (changes requested):
Looks good :)

src/components/Contacts/ContactDetails/ContactNotesTab/ContactNotesTab.tsx
j2trumpet: Nice catch. I did see that translation being used elsewhere. I'll get this in place.

OzzieOrca (approved):

j2trumpet:
@OzzieOrca Should we hook up the data saving and retrieval before merging? If so, I have no idea how to do that.

src/components/Contacts/ContactDetails/ContactNotesTab/ContactNotesTab.graphql
```diff
@@ -0,0 +1,25 @@
...
+) {
+  updateContact(
+    input: {
+      accountListId: $accountListId
+      attributes: { id: $contactId, notes: $notes }
+    }
+  ) {
+    contact {
+      id
+      notes
```
OzzieOrca, line 22: Can you add `notesSavedAt` to the mutation response too? That way when we save the notes, we fetch the new date. Since Apollo updates the cache automatically for use, the `useQuery` will update and rerender the component when the cache updates and show the new date.

j2trumpet: I thought this was the magic you were talking about yesterday. When the notes get updated, the `notesSavedAt` field is updated automatically and the output text re-renders. No?

OzzieOrca (approved):

PR #117, opened 2021-07-13
MPDX-7072- Insert data into ExpectedMonthlyTotal

rathclayton1:


src/components/Reports/ExpectedMonthlyTotalReport/Table/ExpectedMonthlyTotalReportTable.tsx
```diff
@@ -17,70 +17,61 @@ import {
...
-  const total = () => {
-    return 3920;
-  };
-
   const showTotalPartners = () => {
     if (visible) {
-      return t(' Show ' + totalPartners() + ' Partners');
+      return t(
+        ' Show ' + data.length + ` Partner${data.length > 1 ? 's' : ''}`,
+      );
```
reldredge71, line 47:
```suggestion
      return  data.length > 1 ? t('Show {{partnerCount}} Partners', { partnerCount: data.length }) : t('Show 1 Partner')
```
@OzzieOrca How does this look to you?

src/components/Reports/ExpectedMonthlyTotalReport/Table/ExpectedMonthlyTotalReportTable.tsx
```diff
@@ -17,70 +17,61 @@ import {
...
     }
   };
-  return (
+
+  const isEmpty = data.length === 0;
```
reldredge71, line 51: Actually you can use this variable to help with the logic for `showTotalPartners`.

src/components/Reports/ExpectedMonthlyTotalReport/Table/ExpectedMonthlyTotalReportTable.test.tsx
```diff
@@ -2,18 +2,53 @@ import React from 'react';
...
+      <ExpectedMonthlyTotalReportTable
+        accountListId={'abc'}
+        title={'Donations So Far This Month'}
+        data={data.expectedMonthlyTotalReport.received.donations}
+        donations={true}
+        total={0}
+        currency={'USD'}
+      />
+    </ThemeProvider>,
+  );
```
reldredge71, line 49: Could you add a test for `donations={false}`?

reldredge71 (changes requested):

reldredge71 (approved):
looks good!

src/components/Reports/ExpectedMonthlyTotalReport/Table/ExpectedMonthlyTotalReportTable.tsx
OzzieOrca: I was going to suggest i18next's plural feature but if we're extracting text directly from the source code, I'm not sure how you'd specify other keys... This is fine.

PR #118, opened 2021-07-14
MPDX-7036: Fourteen Month Report - Partner Currency

hotcakedev628:
This PR is for fourteen month report page which has both salary and partner currency.


src/components/Reports/FourteenMonthReports/FourteenMonthReport.tsx
```diff
@@ -0,0 +1,170 @@
...
+          ? FourteenMonthReportCurrencyType.Salary
+          : FourteenMonthReportCurrencyType.Donor,
+    },
+  });
+
+  const contacts = useMemo(() => {
+    return data?.fourteenMonthReport.currencyGroups?.flatMap(
+      (currencyGroup) => [...currencyGroup?.contacts],
+    );
+  }, [data?.fourteenMonthReport.currencyGroups]);
```
OzzieOrca, line 57: So there are 2 main differences between the partner currency and salary currency reports.
1. The API returns numbers converted into the salary currency depending on `currencyType` - implemented
2. For partner currency, the frontend shows multiple tables grouped by partner currency. So the salary currency one is implemented correctly (combining all `currencyGroups`). But for partner currency, we want to show each `data?.fourteenMonthReport.currencyGroups` in a separate table.

Feel free to work on that in a new PR. Sorry if we didn't explain that as well as we could have.

OzzieOrca (approved):

src/components/Reports/FourteenMonthReports/FourteenMonthReport.tsx
hotcakedev628: Makes sense to me. :)

.eslintrc.js
```diff
@@ -31,6 +31,7 @@ module.exports = {
     ],
     'react/prop-types': 'off',
     'react/react-in-jsx-scope': 'off',
+    'react/display-name': 'off', // Disable to write display name for forwardRef
```
OzzieOrca, line 34: Where did you use this? If it's just one place you need this, it's better to just disable that single line.
```ts
// eslint-disable-next-line react/display-name
```

src/components/Reports/FourteenMonthReports/Layout/Header/Actions/Actions.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { useTranslation } from 'react-i18next';
+import { Button, ButtonGroup, styled, SvgIcon } from '@material-ui/core';
+import GetAppIcon from '@material-ui/icons/GetApp';
+import FullscreenIcon from '@material-ui/icons/Fullscreen';
+import FullscreenExitIcon from '@material-ui/icons/FullscreenExit';
+import PrintIcon from '@material-ui/icons/Print';
+import { CSVLink } from 'react-csv';
+import { DateTime } from 'luxon';
+
+export type CurrencyType = 'salary' | 'donor';
```
OzzieOrca, line 11: Could this just be `FourteenMonthReportCurrencyType`, the enum from GraphQL?

src/components/Reports/FourteenMonthReports/Layout/Table/Table.tsx
```diff
@@ -0,0 +1,130 @@
...
+  <Typography {...props} />
+))(({ theme, expanded }: { theme: Theme; expanded: boolean }) => ({
+  marginLeft: expanded ? 0 : theme.spacing(1),
+}));
+
+const PrintableContainer = styled(TableContainer)(() => ({
+  // First style set size as landscape
+  // ['@media print']: {
+  //   ['@page']: { size: 'landscape' },
+  // },
```
OzzieOrca, line 37: This style still relevant?

src/components/Reports/FourteenMonthReports/Layout/Table/TableHead/TableHead.tsx
```diff
@@ -9,18 +9,19 @@ import {
   Typography,
 } from '@material-ui/core';
 // eslint-disable-next-line import/extensions
-import { FourteenMonthReportQuery } from '../GetFourteenMonthReport.generated';
+import { FourteenMonthReportQuery } from '../../../GetFourteenMonthReport.generated';
 import { TableHeadCell } from './TableHeadCell/TableHeadCell';
 
 export type Unarray<T> = T extends Array<infer U> ? U : T;
```
OzzieOrca, line 15: Does `ElementOf` work here? https://github.com/krzkaczor/ts-essentials#elementof-type

OzzieOrca (changes requested):
I think this is really close :) Noticed a couple small things.

src/components/Reports/FourteenMonthReports/Layout/Table/Table.tsx
hotcakedev628: Yeah, making `landscape` as the default print option would be good, but it is not working.

src/components/Reports/FourteenMonthReports/Layout/Table/TableHead/TableHead.tsx
hotcakedev628: I am a bit afraid `ElementOf` would not work since the type needs to be `Maybe`.

OzzieOrca: Wrap it in `Maybe`? 🤷‍♂️

.eslintrc.js
hotcakedev628: Ok. :)

OzzieOrca (approved):
🚀 Thanks for all your work on this!

hotcakedev628:
> 🚀 Thanks for all your work on this!

My plasure. But it seems stuck.

OzzieOrca (approved):

PR #119, opened 2021-07-15
MPDX-6919 - Removing vertical scrollbar

j2trumpet:
I'm genuinely not sure what the purpose of the additional `AppBar` was, but it was causing the page to needlessly extend beyond the height of the browser. Feel free to yell at me to put it back.

### DIFF
--- a/src/components/Layouts/Basic/TopBar/TopBar.tsx
+++ b/src/components/Layouts/Basic/TopBar/TopBar.tsx
@@ -44,11 +44,6 @@ const TopBar = ({ children }: Props): ReactElement => {
           </Grid>
         </Toolbar>
       </AppBar>
-      <AppBar position="static" className={classes.appBar} elevation={0}>
-        <Toolbar className={classes.toolbar}>
-          <Grid className={classes.container} />
-        </Toolbar>
-      </AppBar>
     </>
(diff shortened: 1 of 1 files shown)


j2trumpet:
@OzzieOrca @reldredge71 @TheNoodleMoose Can you guys help me figure out why codecov is complaining?

reldredge71:
@j2trumpet I think this is failing because the percent of code coverage has dropped.  It looks like you deleted something that was previously covered, so technically speaking that lowers the overall coverage of the app.  I don't think there's anything that you can do to increase the coverage, though, so you should be fine avoiding this failed test?  @OzzieOrca @TheNoodleMoose what do you think?

TheNoodleMoose:
@j2trumpet Yep looks like it went down because you deleted the component: https://app.codecov.io/gh/CruGlobal/mpdx-react/compare/119/diff. Should be fine to merge.

OzzieOrca (approved):
Looks good. Not necessarily always true but the Codecov patch check matters more than the project one for PRs so it's the only required one. It enforces that the stuff in the PR was actually tested regardless of the rest of the state of the project. Deleting stuff may decrease the project check if the code was previously tested.

PR #120, opened 2021-07-16
Fix-Contact-Right-Panel-Scroll

TheNoodleMoose:
So I just realized for the first time today as I was trying to scroll down to the mailing section of the contact details tab... that I couldn't actually scroll lol. This was true for the whole right panel. Adding this scroll box around the right panel seemed to fix that, but let me know if there's a better way to wrap this.


OzzieOrca (approved):
The lowercase `isscroll` and number instead of a boolean is a hack to avoid React warnings about passing transient style props through to the React DOM. It would be better to just throw away those props and use normal types which we've done in some places. But 🤷‍♂️. It got added to support something in the reports section.

Idk how  I neglected scrolling on the right side. Thanks for fixing it :)

PR #122, opened 2021-07-19
MPDX-6909: Coaching Answer Sets Query

reldredge71:
Create a Rest to GraphQL implementation of the Coaching Answer Sets query.  Will be used for the Coaching Questions modal.

### DIFF
--- a/pages/accountLists/GetDashboard.graphql
+++ b/pages/accountLists/GetDashboard.graphql
@@ -29,3 +29,21 @@ query GetDashboard($accountListId: ID!) {
     }
   }
 }
+
+query GetQuestions($accountListId: ID!) {
+  coachingAnswerSets(accountListId: $accountListId) {
+    id
+    answers {
+      id
+      response
+    }
+    completedAt
+    questions {
+      id
+      position
+      prompt
+      required
+      responseOptions
+    }
+  }
+}
--- a/pages/accountLists/GetDashboard.graphql
+++ b/pages/accountLists/GetDashboard.graphql
@@ -30,8 +30,8 @@ query GetDashboard($accountListId: ID!) {
   }
 }
 
-query GetQuestions($accountListId: ID!) {
-  coachingAnswerSets(accountListId: $accountListId) {
+query GetQuestions($accountListId: ID!, $completed: Boolean = false) {
+  coachingAnswerSets(accountListId: $accountListId, completed: $completed) {
     id
     answers {
       id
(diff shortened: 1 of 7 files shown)


pages/accountLists/GetDashboard.graphql
```diff
@@ -29,3 +29,21 @@ query GetDashboard($accountListId: ID!) {
     }
   }
 }
+
+query GetQuestions($accountListId: ID!, $completed: Boolean = false) {
```
reldredge71, line 33: This is temporary-- just to test the query.  I will remove this before merging the PR.

pages/api/Schema/CoachingAnswerSets/dataHandler.ts
```diff
@@ -0,0 +1,141 @@
...
+): CoachingAnswer[] => {
+  const answers: CoachingAnswer[] = [];
+  debugger;
+  data.forEach(
+    ({
+      id,
+      attributes: {
+        //created_at,
+        //response,
+        //updated_at
```
reldredge71, line 90: These will be uncommented, once I can confirm that the REST query fetches all the needed data.

pages/api/Schema/CoachingAnswerSets/dataHandler.ts
```diff
@@ -0,0 +1,141 @@
...
+    ({
+      id,
+      /*attributes: {
+        created_at,
+        position,
+        prompt,
+        required,
+        response_options,
+        updated_at,
+      }*/
```
reldredge71, line 122: Same here.

pages/api/graphql-rest.page.ts
```diff
@@ -123,6 +127,15 @@ class MpdxRestApi extends RESTDataSource {
     return getTaskAnalytics(data);
   }
 
+  async getCoachingAnswerSets(accountListId: string, completed?: boolean) {
+    debugger;
+    const { data }: { data: CoachingAnswerSetResponse[] } = await this.get(
+      `coaching/answer_sets?filter[account_list_id]=${accountListId}&filter[completed]=${completed}&include=answers,answers.response&sort=-completed_at`,
```
reldredge71, line 133: @OzzieOrca @TheNoodleMoose This is the problem spot.  I need more help debugging this.  Christian and I figured out that the `answers` and `questions` relationships only contain an id, no other associated data.  It should just be as simple as explicitly including those fields in this query, but nothing I've tried has worked.  I tried adding `fields[answers]=response`, `fields[coaching_answers]=response`, and `include=answers.answers.response`, still no data.  Any thoughts?

PR #129, opened 2021-07-27
MPDX-6909: Coaching Answer Sets GraphQL

reldredge71:
Providing Coaching Answer Sets data using REST to GraphQL.

### DIFF
--- a/pages/api/Schema/CoachingAnswerSets/dataHandler.ts
+++ b/pages/api/Schema/CoachingAnswerSets/dataHandler.ts
@@ -61,10 +61,8 @@ interface CoachingQuestionResponse {
 const getCoachingAnswerSets = (
   data: CoachingAnswerSetData,
   included: CoachingAnswerSetIncluded,
-): CoachingAnswerSet[] => {
-  const response: CoachingAnswerSet[] = [];
-
-  data.forEach((answerSetData) => {
+): CoachingAnswerSet[] =>
+  data.map((answerSetData) => {
     const {
       id,
       attributes: { completed_at, created_at, updated_at },
@@ -75,7 +73,7 @@ const getCoachingAnswerSets = (
     //create questions
     const questions = createCoachingQuestionsList(answerSetData, included);
 
-    const answerSet: CoachingAnswerSet = {
+    return {
       id,
       answers,
       completedAt: completed_at,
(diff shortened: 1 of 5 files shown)


pages/api/Schema/CoachingAnswerSets/dataHandler.ts
```diff
@@ -0,0 +1,171 @@
...
+  };
+}
+
+const getCoachingAnswerSets = (
+  data: CoachingAnswerSetData,
+  included: CoachingAnswerSetIncluded,
+): CoachingAnswerSet[] => {
+  const response: CoachingAnswerSet[] = [];
+
+  data.forEach((answerSetData) => {
```
OzzieOrca, line 67: I think a `.map` would be better here.

pages/api/Schema/CoachingAnswerSets/dataHandler.ts
```diff
@@ -0,0 +1,171 @@
...
+): CoachingAnswer[] => {
+  const answers: CoachingAnswer[] = [];
+
+  const ids = answerSetData.relationships.answers.data.map(
+    (answer) => answer.id,
+  );
+
+  ids.forEach((id) => {
+    const answerItem = includedItems.find(
+      (item) => item.id === id,
```
OzzieOrca, line 105: I think you want the type condition in here also. There's a chance (very improbably) that you find an id that is from another type and stop looking for the one you really want too soon.

Same in your question function below.

OzzieOrca (changes requested):
Looks pretty good :)

pages/api/Schema/CoachingAnswerSets/dataHandler.ts
```diff
@@ -74,22 +72,17 @@ const getCoachingAnswerSets = (
     const answers = createCoachingAnswersList(answerSetData, included);
     //create questions
     const questions = createCoachingQuestionsList(answerSetData, included);
-
-    const answerSet: CoachingAnswerSet = {
+    debugger;
```
TheNoodleMoose, line 75: ```debugger``` here. Could probably squash the commit after removing it.

OzzieOrca: 👍 PR looks good to me. I'm happy to approve it after this is removed :) Or anyone else can approve it.

TheNoodleMoose (approved):

PR #130, opened 2021-07-27
MPDX-7038: UI for designation accounts report

hotcakedev628:
This is a PR for designation accounts report in progess.


OzzieOrca (commented):
Looks pretty good so far. :)

Here's your Vercel error:
```
10:40:56.137 | Failed to compile.
-- | --
10:40:56.138 | ./src/components/Reports/DesignationAccountsReport/Layout/List/ListItem/ListItem.tsx:60:13
10:40:56.138 | Type error: Type '(event: MouseEvent<unknown, MouseEvent>, checked: boolean) => void' is not assignable to type '(event: ChangeEvent<HTMLInputElement>, checked: boolean) => void'.
10:40:56.138 | Types of parameters 'event' and 'event' are incompatible.
10:40:56.138 | Type 'ChangeEvent<HTMLInputElement>' is missing the following properties from type 'MouseEvent<unknown, MouseEvent>': altKey, button, buttons, clientX, and 14 more.
10:40:56.138 | 58 \|           <Checkbox
10:40:56.138 | 59 \|             edge="end"
10:40:56.138 | > 60 \|             onChange={onCheckToggle}
10:40:56.138 | \|             ^
10:40:56.138 | 61 \|             checked={designationAccount.active}
10:40:56.138 | 62 \|             inputProps={{ 'aria-labelledby': 'list-1' }}
10:40:56.138 | 63 \|           />
10:40:56.222 | Error: Command "yarn run build" exited with 1
```

hotcakedev628:
> Looks pretty good so far. :)
> 
> Here's your Vercel error:
> 
> ```
> 10:40:56.137 | Failed to compile.
> -- | --
> 10:40:56.138 | ./src/components/Reports/DesignationAccountsReport/Layout/List/ListItem/ListItem.tsx:60:13
> 10:40:56.138 | Type error: Type '(event: MouseEvent<unknown, MouseEvent>, checked: boolean) => void' is not assignable to type '(event: ChangeEvent<HTMLInputElement>, checked: boolean) => void'.
> 10:40:56.138 | Types of parameters 'event' and 'event' are incompatible.
> 10:40:56.138 | Type 'ChangeEvent<HTMLInputElement>' is missing the following properties from type 'MouseEvent<unknown, MouseEvent>': altKey, button, buttons, clientX, and 14 more.
> 10:40:56.138 | 58 \|           <Checkbox
> 10:40:56.138 | 59 \|             edge="end"
> 10:40:56.138 | > 60 \|             onChange={onCheckToggle}
> 10:40:56.138 | \|             ^
> 10:40:56.138 | 61 \|             checked={designationAccount.active}
> 10:40:56.138 | 62 \|             inputProps={{ 'aria-labelledby': 'list-1' }}
> 10:40:56.138 | 63 \|           />
> 10:40:56.222 | Error: Command "yarn run build" exited with 1
> ```

Thank you. :) Still in progress.

src/components/Reports/DesignationAccountsReport/DesignationAccountsReport.tsx
```diff
@@ -0,0 +1,105 @@
...
+      variables: {
+        input: {
+          accountListId,
+          active: event.target.checked,
+          designationAccountId,
+        },
+      },
+    });
+
+    return refetch();
```
OzzieOrca, line 56: There should be better ways to do this. See my other comment. Hopefully you can just remove this all together.

If you can get the other suggestion working, might be nice to provide an [optimistic response](https://www.apollographql.com/docs/react/performance/optimistic-ui/#the-optimisticresponse-option) so you don't have to wait for a round trip to the server before the box gets checked.

pages/api/Schema/reports/designationAccounts/designationAccounts.graphql
```diff
@@ -4,9 +4,7 @@ extend type Query {
 
 extend type Mutation {
   setActiveDesignationAccount(
-    accountListId: ID!
-    active: Boolean!
-    designationAccountId: ID!
+    input: SetActiveDesignationAccountInput!
   ): SetActiveDesignationAccountResponse!
```
OzzieOrca, line 8: If you could make this return a `DesignationAccount`, the Apollo Client should be able to update its cache after the mutation and rerender the query.

That being said, it might be not be doable if you can't get all the fields you need from the REST PUT request. But hopefully you can.

OzzieOrca (changes requested):
Looks pretty good :)

src/components/Reports/DesignationAccountsReport/DesignationAccountsReport.tsx
```diff
@@ -39,21 +43,60 @@ export const DesignationAccountsReport: React.FC<Props> = ({
...
-    });
+      optimisticResponse: {
+        setActiveDesignationAccount: {
+          id: designationAccountId,
+          __typename: 'SetActiveDesignationAccountResponse',
+          active: event.target.checked,
+        },
+      },
+      update: (cache) => {
+        if (!optimisticResponse) return;
```
OzzieOrca, line 67: What is this condition doing? Seems like it's always true so it would always continue.

src/components/Reports/DesignationAccountsReport/DesignationAccountsReport.tsx
```diff
@@ -39,21 +43,60 @@ export const DesignationAccountsReport: React.FC<Props> = ({
...
+          designationAccounts: dataFromCache?.designationAccounts.map(
+            (designationAccountsGroup) => ({
+              ...designationAccountsGroup,
+              designationAccounts: designationAccountsGroup.designationAccounts.map(
+                (designationAccount) => {
+                  return designationAccount.id === designationAccountId
+                    ? {
+                        ...designationAccount,
+                        id: designationAccountId,
+                        active: event.target.checked,
```
OzzieOrca, line 88: I think this would be better to get from the mutation response than the client event in case the server doesn't save it or gets 2 updates at once or something. Unlikely but the server should be the source of truth.

pages/api/Schema/reports/designationAccounts/designationAccounts.graphql
OzzieOrca: Any thoughts on making this return the same type as the query? You might be able to avoid having to do the manual cache updates you wrote.

src/components/Reports/DesignationAccountsReport/SetActiveDesignationAccount.graphql
```diff
@@ -0,0 +1,13 @@
...
+  $input: SetActiveDesignationAccountInput!
+) {
+  setActiveDesignationAccount(input: $input) {
+    active
+    balanceUpdatedAt
+    convertedBalance
+    currency
+    designationNumber
+    id
+    name
```
OzzieOrca, line 11: Does this need all these fields? I assume `active` is the only thing that would change. And we probably need `id` to make the cache update work.

src/components/Reports/DesignationAccountsReport/DesignationAccountsReport.tsx
```diff
@@ -0,0 +1,113 @@
...
+      variables: {
+        input: {
+          accountListId,
+          active: event.target.checked,
+          designationAccountId: designationAccount.id,
+        },
+      },
+      optimisticResponse: {
+        setActiveDesignationAccount: {
+          ...designationAccount,
```
OzzieOrca, line 59: Does this need all the fields? It might only need `id` and `active`.

src/components/Reports/DesignationAccountsReport/DesignationAccountsReport.tsx
```diff
@@ -0,0 +1,113 @@
...
+  return (
+    <Box>
+      <Header
+        isNavListOpen={isNavListOpen}
+        onNavListToggle={onNavListToggle}
+        title={title}
+        totalBalance={
+          data &&
+          currencyFormat(
+            data.designationAccounts[0].balance ?? 0,
```
OzzieOrca, line 75: This balance needs to update when accounts get checked and unchecked. Idk if you want to do it with some sort of client side code or GraphQL refreshing...

OzzieOrca (changes requested):
Looks pretty good :)

src/components/Reports/DesignationAccountsReport/SetActiveDesignationAccount.graphql
hotcakedev628: Basically I used all fields while trying to get the `DesignationAccount` type so the cache can be automatically updated.

hotcakedev628: If the response type is not `DesignationAccount`, the cache didn't update.

OzzieOrca (approved):
Looks great! Thanks :) 🚀

src/components/Reports/DesignationAccountsReport/SetActiveDesignationAccount.graphql
OzzieOrca: Right. I was thinking the server absolutely needs to provide all these fields and use the `DesignationAccount` type. But I'm not sure the client actually needs to request all those fields, besides `id` and `active`. But if it's needed, it's not hurting anything 🤷‍♂️

PR #131, opened 2021-08-05
MPDX-7027-Contacts-List-Pagination

TheNoodleMoose:
Contacts page now has infinite scroll 🥳 Tried to use https://www.npmjs.com/package/react-infinite-scroll-component first but ran into issues with loading large lists(which in theory, there can be a lot of contacts). Eventually, it would actually crash the page lol. So I instead went with https://www.npmjs.com/package/react-virtualized which actually does some smart stuff under the hood to only render the elements that are visible in the viewport. Now we can load 1000+ contacts with no effect on performance! I do still have a test that is broken, not sure why it's broken. I'm going to keep trying to fix it, but let me know what you guys think in the meantime 🙂


src/components/Contacts/ContactsTable/ContactsTable.test.tsx
```diff
@@ -127,11 +161,19 @@ describe('ContactFilters', () => {
   });
 
   it('simulate row click with searchTerm', async () => {
```
TheNoodleMoose, line 163: @OzzieOrca Not sure what's up with this test. It says the ```onContactSelected``` function is not being called. But the one above works fine. If I remove everything related to ```searchTerm``` in this test, then it works. Not sure what the connection is and why the search term would affect it. It appears to works fine actually using it in the browser.

gyasistory: @TheNoodleMoose outside of the test failing everything looks good to me.

OzzieOrca (commented):
Looks good. You've got a double scrollbar thing going on. You can probably kill the outer scroll bar if you're able to or you may need to compute the available height. Not sure how the autosizer works.

I'll try and look at your test here.

src/components/Contacts/ContactsTable/ContactsTable.test.tsx
OzzieOrca: Idk what's going on with the test. It never fires the click event on the contact row. It almost seems like after the click the whole page rerenders as the row isn't there right after the click but is again after the next waitFor...

TheNoodleMoose:
I went ahead and removed the outer scrollbar, but I occasionally still see both if I scroll down and then back up really fast. I also just went ahead and commented out that one part of the test that's giving us trouble and added a comment. Hopefully, we can revisit it and figure out what's up with it.

OzzieOrca:
I'm still seeing the main page scrollbar show up. I think the virtualized list height isn't subtracting the contact actions header somehow.

I was looking through the docs for some sort of a header component/prop but didn't see any. Maybe it needs some CSS to get it to not grow past the parent container. Or maybe the parent needs a size or something...

A little late, but did you see this https://github.com/bvaughn/react-window#how-is-react-window-different-from-react-virtualized? Looks like it still supports the autosizer https://github.com/bvaughn/react-window#can-a-list-or-a-grid-fill-100-the-width-or-height-of-a-page. Haven't done any other reading or experimenting.

TheNoodleMoose:
@OzzieOrca Ah I must have missed ```react-window```. I actually just went ahead switched to that locally and got it working. I'll push it up once I get the tests fixed, etc.

LizzyRandall:
@OzzieOrca @TheNoodleMoose Have you found react-window to be good? Just curious. I found the base one to be gigantic in file size with too much stuff and then the slimmed down version didn't have enough of what I needed. I use Virtuoso https://virtuoso.dev/ and found it pretty good, it doesn't have to force the same height for each element in a list like I was seeing for react-window.

TheNoodleMoose:
> @OzzieOrca @TheNoodleMoose Have you found react-window to be good? Just curious. I found the base one to be gigantic in file size with too much stuff and then the slimmed down version didn't have enough of what I needed. I use Virtuoso https://virtuoso.dev/ and found it pretty good, it doesn't have to force the same height for each element in a list like I was seeing for react-window.

To be honest, I just got it switched over about five minutes before I left that comment. Seems to work pretty well so far from the limited testing I've done. Very similar api so it was easy to transition to react-window. But I plan to mess with it more on Monday. I havent looked into Virtuoso, but does it do similar smart rendering like react-window? Yeah I'm not the biggest fan of having to use the auto sizer, and forced row height.

hotcakedev628:
@TheNoodleMoose What should it look like? Both `react-window` and `virtuoso` are simply alike a list.
Not sure, but Material-UI list is not enough for you?
https://material-ui.com/components/lists/#pinned-subheader-list

TheNoodleMoose:
@hotcakedev628 It looks exactly the same as it did before, but just has the infinite scroll ability now. ```react-window``` just provides the components needed to make that logic work and run efficiently.

TheNoodleMoose:
@OzzieOrca  I took a look at @LizzyRandall's suggestion of https://virtuoso.dev/ and found https://virtuoso.dev/endless-scrolling/. Gave it a try and prefer it over ```react-window``` 🙂  I also believe I fixed the double scroll bar. I wasn't accounting for both the contact page header and the navbar as well. The one test still seems to not work though for whatever reason 🤷‍♂️

OzzieOrca (approved):
Sorry for the library suggestion churn haha, I haven't used any of those before. Thanks for suggesting that @LizzyRandall :) If you guys like it better, sounds like a good way to go. Looks a little more like what React Native had to offer with end reached events, item renderer function, and footer components.

I was playing with it a little, and it seems like it can get stuck in the loading state sometimes. If you scroll back up a little and down, it sends the network request again and works but I can reproduce it pretty consistently.
Screen Shot 2021-08-09 at 3 06 46 PM

src/components/Contacts/ContactsTable/ContactsTable.tsx
```diff
@@ -84,14 +84,42 @@ export const ContactsTable: React.FC<Props> = ({
...
+                      if (!fetchMoreResult) {
+                        return prev;
+                      }
+                      return {
+                        ...prev,
+                        ...fetchMoreResult,
+                        contacts: {
+                          ...prev.contacts,
+                          ...fetchMoreResult.contacts,
+                          pageInfo: fetchMoreResult.contacts.pageInfo,
```
OzzieOrca, line 103: Is this line needed? Seems like the previous line should have just set it.

OzzieOrca (commented):
Seems like the data is repeating. Somehow there are multiple Frodo Baggins and probably all the other contacts too. Maybe related to the last issue but sometimes the response for`endCursor` is `null` which  may make it reset back to the last known cursor somehow. I haven't followed that specifically. But is that an API bug? Even if `hasNextPage` is `false`, we should still get an `endCursor` right?

src/components/Contacts/ContactsTable/ContactsTable.tsx
TheNoodleMoose: Good point, I'll remove this line.

TheNoodleMoose:
@OzzieOrca Nice catch! I went ahead and updated the key value to be the index instead of the contact id. Could you give it another look and see if you're encountering any issues or repeating data? I don't seem to run into this locally. Yeah, I would think there should always be an ```endCursor```.

TheNoodleMoose:
> Seems like the data is repeating. Somehow there are multiple Frodo Baggins and probably all the other contacts too. Maybe related to the last issue but sometimes the response for`endCursor` is `null` which may make it reset back to the last known cursor somehow. I haven't followed that specifically. But is that an API bug? Even if `hasNextPage` is `false`, we should still get an `endCursor` right?

@soberstadt

OzzieOrca:
@TheNoodleMoose I can't reproduce it in playground so maybe it isn't an API issue...

Are you using `hasNextPage` anywhere? I see it in the tests and the query but not in the component. Maybe I missed it. But we shouldn't send more requests for pages when there aren't any...

Actually if you send the `endCursor` from a response with `hasNextPage === false`, you get `null`.
Screen Shot 2021-08-10 at 2 18 18 PM

TheNoodleMoose:
@OzzieOrca Ah good point. I was before when I was using ```react-virtualized``` but I don't think I brought that logic over to here. I'll add ```hasNextPage``` to the ```endReached``` logic. Hopefully, that should fix the issue we are running into.

OzzieOrca (approved):
That seems to work way better and loads 2 pages, then stops. Thanks!

Could you think about a better loading indicator at some point? Maybe MUI has something that would look good there. Can be a new PR ticket but should be somewhat trivial.

TheNoodleMoose:
@OzzieOrca Thanks for helping me test this and finding those issues! I think my problem was the account I was using had over a thousand contacts and I wasn't scrolling all the way to see the issue happen. Glad we caught this sooner than later.

OzzieOrca (approved):
Ya :) Makes sense

PR #132, opened 2021-08-05
Mpdx 7101 tools landing and sidebar

CSimbulan:
Created a landing page with the side menu for the tools section. The box sizes should adjust when the sidebar is open, but only for large screens.

image
image

Should also be responsive:

image

Since there are no pages for the individual tools, the links in the side bar don't lead anywhere yet but they should once the pages are ready.


pages/accountLists/[accountListId]/tools.page.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const { t } = useTranslation();
+  const router = useRouter();
+
+  useEffect(() => {
+    dispatch({ type: 'updateBreadcrumb', breadcrumb: t('Tools') });
+    dispatch({
+      type: 'updateAccountListId',
+      accountListId: router.query.accountListId?.toString() ?? '',
+    });
+  }, []);
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/tools.page.tsx:19]: Just toss this. We're moving away from a global app state and I don't  think the breadcrumb is used anywhere. If you need the `accountListId`, use the `useAccountListId` hook. If we need some global state in the future, we can make dedicated hooks for the specific 
 feature needed.

src/components/Tool/Home/Home.tsx
```diff
@@ -0,0 +1,95 @@
+import React, { ReactElement, useState } from 'react';
+import { makeStyles, Theme, Grid, Container, Box } from '@material-ui/core';
+import { motion } from 'framer-motion';
+
+import NavToolDrawer from '../NavToolList/NavToolDrawer';
+import Tool from './Tool';
+import { ToolsList } from './ToolList';
+
+const useStyles = makeStyles((theme: Theme) => ({
```
OzzieOrca, line 9: We kinda like the [styled api](https://material-ui.com/styles/basics/#styled-components-api) but this is ok too.

src/components/Tool/Home/Home.tsx
```diff
@@ -0,0 +1,95 @@
...
+    },
+    [theme.breakpoints.down('sm')]: {
+      paddingLeft: 45,
+    },
+  },
+  toolIcon: {
+    height: '40px',
+    width: '40px',
+    color: theme.palette.cruGrayDark.main,
+  },
```
OzzieOrca, line 30: Can you use the [spacing helper](https://material-ui.com/customization/spacing/#spacing) for any of these magic numbers? It's ok if you need a specific value, just having some consistency and fewer magic numbers is nice.

src/components/Tool/Home/Home.tsx
```diff
@@ -0,0 +1,95 @@
...
+  outter: {
```
OzzieOrca, line 31:
```suggestion
  outer: {
```

src/components/Tool/Home/Home.tsx
```diff
@@ -0,0 +1,95 @@
...
+      animate="animate"
+      exit="exit"
+      variants={variants}
+    >
+      <Box className={classes.outter}>
+        <NavToolDrawer open={isNavListOpen} toggle={handleNavListToggle} />
+        <Container
+          className={classes.container}
+          style={{
+            minWidth: isNavListOpen ? 'calc(97.5vw - 290px)' : '97.5vw',
```
OzzieOrca, line 71: Where did `97.5vw` come from?

src/components/Tool/Home/ToolList.ts
```diff
@@ -0,0 +1,82 @@
...
+  mdiMap,
+  mdiPhone,
+  mdiNewspaperVariantOutline,
+  mdiHome,
+  mdiAccountGroup,
+} from '@mdi/js';
+
+export const ToolsList = [
+  {
+    tool: 'Appeal',
```
OzzieOrca, line 17: Can you translate these?
```ts
import i18n from 'i18next';
```
```suggestion
    tool: i18n.t('Appeal'),
```

If you're in a component (which you're not here, which is fine) you can use the hook:
```ts

export ExampleComponent = () => {
  return <p>{t('Some English String')}</p>
}
```

src/components/Tool/Home/Tool.tsx
```diff
@@ -0,0 +1,85 @@
...
+            <Box
+              className={classes.iconBG}
+              display="flex"
+              justifyContent="center"
+              alignItems="center"
+            >
+              <Icon path={icon} size={1.5} />
+            </Box>
+            <Typography variant="h6">{t(tool)}</Typography>
+            <Typography variant="body2">{t(desc)}</Typography>
```
OzzieOrca, line 77: Because we're going to extract these strings by just scanning the codebase for a regex (which is how I assume https://www.i18next.com/how-to/extracting-translations#2-using-an-extraction-tool works), dynamic keys won't work here. It's better to translate them where the string gets declared, not at runtime. You can just remove the translation here and see my other comment.

src/components/Tool/NavToolList/NavToolList.tsx
```diff
@@ -0,0 +1,82 @@
...
+  return (
+    <Box component="div" className={classes.list}>
+      <List>
+        <ListItem className={classes.li}>
+          <ListItemIcon>
+            <BuildIcon />
+          </ListItemIcon>
+          <ListItemText primary={t('Tools')} />
+        </ListItem>
+        {ToolNavItems.slice(0, 1).map((tool) => (
```
OzzieOrca, line 47: Since you're parsing your own JSON, could you just group them in the JSON and iterate through the groups? Slicing here with magic numbers seems a little weird and error prone here. It could also make this list component more reusable.
```ts
const ToolsListGroups = [
  {
    groupName: 'Tools',
    items: [
      {
        tool: 'Appeal',
        desc: 'Set goals, create asks, and track progress for one-time needs',
        icon: mdiTrophy,
        id: 'appeal',
      },
      ...
    ],
  },
  ...
];
```

(I wrote this before I discovered you had 2 sets of JSON tools. They could probably still be combined like this. When you don't want them grouped, you could use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap to get rid of the groups.)

src/components/Tool/NavToolList/ToolNavItems.ts
```diff
@@ -0,0 +1,46 @@
+export const ToolNavItems = [
```
OzzieOrca, line 1: Could you combine/share this and [`ToolsList`](https://github.com/CruGlobal/mpdx-react/pull/132/files#diff-e25bacfc25a4c1c9b4558c1d31a1b232268dbc405f7a522e56f0f130095b5017R15) somehow?

OzzieOrca (changes requested):
Looks pretty good :) Just a few smaller things

pages/accountLists/[accountListId]/tools.page.tsx
Comment (CSimbulan) [reply pages/accountLists/[accountListId]/tools.page.tsx]: Done

src/components/Tool/Home/Home.tsx
CSimbulan: Changed

src/components/Tool/Home/Home.tsx
CSimbulan: It was by trial and error to find an appropriate size. I can find another way to calculate this.

src/components/Tool/Home/Tool.tsx
CSimbulan: Made these changes

src/components/Tool/NavToolList/NavToolList.tsx
CSimbulan: Grouped the tools like above and made it sharable between the landing page and the side bar. Also removed the other tool list file. The order of the tools on the landing page is now the same as the side bar, but if we want it to be like in the original, I can add some extra code to do it.

OzzieOrca: I don't think the order matters. The sidebar order is probably better anyways.

src/components/Tool/Home/ToolList.ts
```diff
@@ -10,73 +10,109 @@ import {
...
 } from '@mdi/js';
+import i18n from 'i18next';
 
 export const ToolsList = [
   {
-    tool: 'Appeal',
-    desc: 'Set goals, create asks, and track progress for one-time needs',
-    icon: mdiTrophy,
-    id: 'appeal',
+    groupName: 'Appeals',
```
OzzieOrca, line 19: Translate

OzzieOrca (changes requested):
Thanks for working on those things :) Just some trivial translates needed for the group names.

src/components/Tool/Home/ToolList.ts
CSimbulan: Done

OzzieOrca (approved):

OzzieOrca:
Feel free to merge :)

PR #133, opened 2021-08-06
Fix-Navbar-Visibility-Bug

TheNoodleMoose:
This bug was occurring because on certain pages, the state from ```useApp``` did not contain the ```accountListId```. We are moving away from putting stuff in this global state. I'm going to try to clean up other places that still use this in a separate pr but unfortunately, more components are using this than I thought...


hotcakedev628:
Right, it seems like we are getting `accountListId` from the router hooks.
I think it would be good if we move it to new global Auth hooks while migrating Okta.

OzzieOrca (approved):

PR #135, opened 2021-08-10
added ability to query for appeals given an accountlistid

CSimbulan:
Added ability to query for Appeals given an accountListId


pages/api/graphql-rest.page.ts
```diff
@@ -133,25 +129,6 @@ class MpdxRestApi extends RESTDataSource {
     return getTaskAnalytics(data);
   }
 
-  async getCoachingAnswerSets(
```
OzzieOrca, line 136: I assume this came from a bad merge? Or your branch isn't up to date? I think this was merged to main today.

It's also why you've got TS errors.

pages/api/Schema/Appeals/datahandler.ts
```diff
@@ -0,0 +1,66 @@
...
+  };
+  relationships: {
+    account_list: {
+      data: {
+        id: string;
+        type: string;
+      };
+    };
+    contacts: {
+      data: Contact[];
```
OzzieOrca, line 30: I assume this isn't the right type since `Contact` is from GraphQL and this is typing the REST response.

You don't seem to be using any of the relationships so idk if it's worth typing.

pages/api/Schema/Appeals/datahandler.ts
```diff
@@ -0,0 +1,66 @@
...
+        item.attributes.pledges_amount_received_not_processed,
+      pledgesAmountTotal: item.attributes.pledges_amount_total,
+      amountCurrency: item.attributes.total_currency,
+      updatedAt: item.attributes.updated_at,
+      createdAt: item.attributes.created_at,
+    };
+
+    response.push(appeal);
+  });
+  return response;
```
OzzieOrca, line 63: I think a [`.map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) would be better here
```suggestion
  return data.map((item) => ({
    id: item.id,
    name: item.attributes.name,
    amount: parseFloat(item.attributes.amount) || 0,
    pledgesAmountNotReceivedNotProcessed:
      item.attributes.pledges_amount_not_received_not_processed,
    pledgesAmountProcessed: item.attributes.pledges_amount_processed,
    pledgesAmountReceivedNotProcessed:
      item.attributes.pledges_amount_received_not_processed,
    pledgesAmountTotal: item.attributes.pledges_amount_total,
    amountCurrency: item.attributes.total_currency,
    updatedAt: item.attributes.updated_at,
    createdAt: item.attributes.created_at,
  });
```

OzzieOrca (changes requested):
Looks great :)

pages/api/Schema/Appeals/datahandler.ts
CSimbulan: Fixed

pages/api/Schema/Appeals/datahandler.ts
CSimbulan: Yeah I wasn't sure what type to use but for the individual appeal pages there is a list of contacts. I didn't check if that was that same array or if its coming from somewhere else but either way I just removed all the extra stuff.

OzzieOrca (approved):
Looks great :)

Maybe do a Squash and merge with Github so the `getCoachingAnswerSets` aren't removed and readded in the git history.

PR #137, opened 2021-08-12
Remove-openTaskDrawer

TheNoodleMoose:
Slowly getting rid of ```useApp```, piece by piece 🙂 We determined that the task drawer state/logic still has to exist in a context that is accessible, so now it's in its own context instead of a global one.


OzzieOrca (approved):
✨

PR #138, opened 2021-08-13
Mpdx 7102 list appeals

CSimbulan:
image
image

Because there is currently no way to directly determine if an appeal is the primary one or not, the setup might be kind of messy. If we plan on adding that ability I can change it later. The form on the right does not do anything, I just made it for the visual for now (nothing in there translated either).

Also not sure if we want to add pagination for the list of appeals, but I did not do that for now.


pages/accountLists/[accountListId]/tools/GetAppeals.graphql
```diff
@@ -0,0 +1,15 @@
...
+      name
+      amount
+      amountCurrency
+      pledgesAmountNotReceivedNotProcessed
+      pledgesAmountReceivedNotProcessed
+      pledgesAmountProcessed
+      pledgesAmountTotal
+    }
+  }
+}
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/tools/GetAppeals.graphql:15]: I think you can fetch everything you need in a single query instead of firing off 2 at the same time:
```suggestion
query GetAppeals($accountListId: ID!) {
  nonPrimaryAppeals: appeals(accountListId: $accountListId, primary: false) {
    nodes {
      ...AppealFields
    }
  }
  primaryAppeals: appeals(
    accountListId: $accountListId
    primary: true
    first: 1
  ) {
    nodes {
      ...AppealFields
    }
  }
}

fragment AppealFields on Appeal {
  id
  name
  amount
  amountCurrency
  pledgesAmountNotReceivedNotProcessed
  pledgesAmountReceivedNotProcessed
  pledgesAmountProcessed
  pledgesAmountTotal
}
```
Idk what to call `nonPrimaryAppeals`. `otherAppeals`? Just `appeals`? You can pick a name.

pages/accountLists/[accountListId]/tools/GetAppeals.graphql
```diff
@@ -0,0 +1,15 @@
+# Write your query or mutation here
+query GetAppeals($accountListId: ID!) {
+  appeals(accountListId: $accountListId, primary: false) {
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/tools/GetAppeals.graphql:3]: This is probably going to need to support pagination eventually. By default the API will only return 25 items.

OzzieOrca (changes requested):
Looks great. Just got a suggestion to improve the query. The pagination could happen in another PR.

src/components/Tool/Appeal/Appeals.tsx
```diff
@@ -0,0 +1,105 @@
...
+    <Box mb={1}>
+      <Box m={1}>
+        <Typography variant="h6">{t('Primary Appeal')}</Typography>
+      </Box>
+      <Divider />
+      {loading ? (
+        <Box display="flex" justifyContent="center" mt={10}>
+          <LoadingIndicator color="primary" size={40} />
+        </Box>
+      ) : data && data.primaryAppeal ? (
```
OzzieOrca, line 36:
```suggestion
      ) : data?.primaryAppeal ? (
```

OzzieOrca (approved):

src/components/Tool/Appeal/Appeals.tsx
```diff
@@ -0,0 +1,108 @@
...
+        <Typography variant="h6">{t('Primary Appeal')}</Typography>
+      </Box>
+      <Divider />
+      {loading ? (
+        <Box display="flex" justifyContent="center" mt={10}>
+          <LoadingIndicator color="primary" size={40} />
+        </Box>
+      ) : data?.primaryAppeal && data.primaryAppeal.nodes.length > 0 ? (
+        <>
+          {console.log(data)}
```
OzzieOrca, line 38: Still using this?

CSimbulan: I tried changing it before and was still giving me an error, but I just tried again while removing the uneeded wrapper and now its fine 🤷‍♂️

PR #140, opened 2021-08-13
remove-useApp-accountListId

TheNoodleMoose:
After a slow and somewhat painful process, ```accountListId``` has been removed from ```useApp```! 🥳 One step closer to being able to get rid of it completely.


src/components/HandoffLink/HandoffLink.test.tsx
```diff
@@ -1,18 +1,27 @@
...
   beforeEach(() => {
     open = jest.fn();
     originalOpen = window.open;
     window.open = open;
+    (useRouter as jest.SpyInstance<
+      Pick<nextRouter.NextRouter, 'query' | 'isReady'>
+    >).mockImplementation(() => ({
+      query: { accountListId: 'accountListId' },
+      isReady: true,
+    }));
```
TheNoodleMoose, line 24: For some reason, this test didn't like the other way I was mocking ```useRouter``` 🤷‍♂️

OzzieOrca (approved):
✨

PR #141, opened 2021-08-16
UI for responsibility centers report with updates for designation accounts

hotcakedev628:


src/components/Reports/ResponsibilityCentersReport/GetFinancialAccounts.graphql
```diff
@@ -0,0 +1,22 @@
+query FinancialAccounts($accountListId: ID!) {
+  financialAccounts(accountListId: $accountListId) {
+    edges {
```
OzzieOrca, line 3: You should be able to use `nodes` directly instead of `edges`. `edges` can be useful if the API provides some sort of join table data for the relationship or you need a cursor for every item but you probably don't need either here.

OzzieOrca (approved):
👍

PR #142, opened 2021-08-17
Remove-useApp

TheNoodleMoose:
Finally... ```useApp``` has been removed 🥳 Let me know how you guys feel about the naming convention for ```useUser```. I can change it if need be. I also have one test failing and I'm not sure why it's failing. Let me know if you have any ideas/suggestions 🙂


src/components/HandoffLink/HandoffLink.test.tsx
```diff
@@ -30,15 +29,11 @@ describe('HandoffLink', () => {
 
   it('default', async () => {
     const { getByRole } = render(
-      <TestWrapper
-        initialState={{
-          user: { id: 'userId', firstName: 'Bob', lastName: 'Jones' } as User,
-        }}
-      >
+      <GqlMockedProvider<GetUserQuery>>
```
TheNoodleMoose, line 32: @OzzieOrca On this test, it seems that the click is happening when the ```user``` is undefined. When I debug I see that ```user``` goes from ```undefined``` -> ```has a value``` -> ```undefined``` -> ```has a value```. Not sure the best way to address this in the test but it works fine in the app.

src/components/Constants/UseApiConstants.tsx
```diff
@@ -1,7 +1,12 @@
...
+import {
+  LoadConstantsQuery,
+  useLoadConstantsQuery,
+} from './LoadConstants.generated';
 
 // Use this call to force loading from cache storage first
-export const useApiConstants = () => {
+export const useApiConstants = ():
+  | LoadConstantsQuery['constant']
+  | undefined => {
```
OzzieOrca, line 9: TypeScript couldn't infer these types?

src/components/HandoffLink/HandoffLink.test.tsx
OzzieOrca: Weird. It's rerendering for some reason? Idk... You forced only one test to run?

OzzieOrca (approved):
✨

src/components/Constants/UseApiConstants.tsx
TheNoodleMoose: No, it was throwing a warning not an error about not having a return type. And for some reason, it couldn't infer that it could return ```undefined``` 🤷‍♂️

src/components/HandoffLink/HandoffLink.test.tsx
TheNoodleMoose: Yeah even running just the one test it fails. You're right about it rerendering. Not sure what else to try, I even tried creating a ```cache``` with the query but that did not work as well.

OzzieOrca (approved):

PR #143, opened 2021-08-17
Tools tests

CSimbulan:
Tests for tools home page that werent added before


src/components/Tool/Home/Tool.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+            id={'test'}
+          />
+          ,
+        </TestRouter>
+      </ThemeProvider>,
+    );
+    expect(getByTestId('ToolNameTypography').textContent).toEqual('Test');
+    expect(getByTestId('ToolDescriptionTypography').textContent).toEqual(
+      'Test description!!!',
+    );
```
OzzieOrca, line 34: I think it's usually best to use test ids as last resort. I mean don't sink tons of time into avoiding them if something isn't very easy to get at but there should better ways to get elements in ways that rely on what the user experiences.

Take a look at https://testing-library.com/docs/queries/about#priority and it has a bunch of suggestions.

In this case I think you could `getByText` and assert that the returned element is in the document. Or maybe there's a way to use a role and a name filter with `getByRole`.

src/components/Tool/Home/Home.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+  it('default', () => {
+    const { getByTestId } = render(
+      <ThemeProvider theme={theme}>
+        <TestRouter router={router}>
+          <Home />
+        </TestRouter>
+      </ThemeProvider>,
+    );
+
+    expect(getByTestId('Home')).toBeInTheDocument();
```
OzzieOrca, line 25: This isn't much of a test but at least it's checking that the component didn't throw an error. Idk that the home screen is all that prone to bugs so idk that it's worth spending more time on.

OzzieOrca (approved):

PR #147, opened 2021-08-19
MPDX-7010-Add-Apollo-Link-Display-Snackbar-on-Errors

TheNoodleMoose:


src/components/Contacts/ContactDetails/ContactDetailsTab/ContactDetailsTab.tsx
```diff
@@ -156,7 +156,6 @@ export const ContactDetailsTab: React.FC<ContactDetailTabProps> = ({
       setDeleteModalOpen(false);
       onClose();
     } catch (error) {
-      enqueueSnackbar(error.message, { variant: 'error' });
       throw error;
     }
```
OzzieOrca, line 160: I think you can just remove the catch altogether. Catching and rethrowing doesn't doesn't add anything that the original error being thrown didn't do.

OzzieOrca (changes requested):
Looks good :) Just think the catches can be removed.

OzzieOrca (approved):

PR #154, opened 2021-08-23
MPDX-7091-Add-Logic-To-Create-New-Person

TheNoodleMoose:


src/components/Contacts/ContactDetails/ContactDetailsTab/People/Items/PersonModal/PersonModal.tsx
```diff
@@ -168,70 +180,150 @@ export const EditPersonModal: React.FC<EditPersonModalProps> = ({
...
-    facebookAccounts: personFacebookAccounts,
-    twitterAccounts: personTwitterAccounts,
-    linkedinAccounts: personLinkedinAccounts,
-    websites: personWebsites,
-    legalFirstName: person.legalFirstName,
-    deceased: person.deceased,
-  };
+  const initialPerson: PersonCreateInput | PersonUpdateInput = person
+    ? {
+        id: person.id,
```
LizzyRandall, line 215: Couldn't something like this have {...person} to get all these properties? A lot of them look similar to PersonUpdateInput

src/components/Contacts/ContactDetails/ContactDetailsTab/People/Items/PersonModal/PersonPhoneNumber/PersonPhoneNumber.tsx
```diff
@@ -150,6 +153,7 @@ export const PersonPhoneNumber: React.FC<PersonPhoneNumberProps> = ({
                             }}
                             fullWidth
                           >
+                            <MenuItem selected value=""></MenuItem>
```
LizzyRandall, line 156: So this is making the empty menu item be the default right?

LizzyRandall (commented):
Looks good to me, I had a few questions.

src/components/Contacts/ContactDetails/ContactDetailsTab/People/Items/PersonModal/PersonPhoneNumber/PersonPhoneNumber.tsx
TheNoodleMoose: Yep, since the default value I set is ```null``` it was throwing warnings in the console about not matching an option in the ```Select```. And the value of a ```Select``` cannot be ```null``` so I have to change it to an empty string. And to get rid of that warning I just added an option with the value of an empty string.

src/components/Contacts/ContactDetails/ContactDetailsTab/People/Items/PersonModal/PersonModal.tsx
TheNoodleMoose: Ahh yep good idea I could definitely do ```...person,``` and then just keep the fields that I need to specifically set like ```phoneNumbers```, ```emailAddresses```, etc.

TheNoodleMoose: Strange but for some reason, whenever I used the spread it seems to work fine but all the tests fail 🤷‍♂️ I may just leave as-is for now.

TheNoodleMoose:
So I'm having an issue fully testing the logic for updating the cache after creating a person. I believe it has to do with the fact that we are using fragments in the ```ContactDetailsTab``` query. If I got rid of the ```...ContactPeople``` fragment in favor of: 
```graphql
 primaryPerson {
    ...ContactPerson
  }
  people {
    nodes {
      ...ContactPerson
    }
  }
```

I would then get data for ```primaryPerson``` and ```people``` when debugging. However, since ```...ContactPerson``` is a fragment, ```nodes``` and ```primaryPerson```  would just be an empty object. Not sure how to resolve this, but would be glad to hear any suggestions 🙂 Here's what the data looks like when using the fragments(note this is just how it is in the test, works fine in the app):
Screen Shot 2021-08-23 at 2 59 13 PM
VS when I get rid of the ```...ContactPerson``` fragment in favor of what I mentioned above:
Screen Shot 2021-08-23 at 3 02 34 PM

src/components/Contacts/ContactDetails/ContactDetailsTab/People/Items/PersonModal/PersonModal.tsx
```diff
@@ -54,21 +63,24 @@ const LoadingIndicator = styled(CircularProgress)(({ theme }) => ({
...
 }
 
-export const EditPersonModal: React.FC<EditPersonModalProps> = ({
+export const PersonModal: React.FC<PersonModalProps> = ({
   person,
+  contactId,
   accountListId,
   handleClose,
-}): ReactElement<EditPersonModalProps> => {
+}): ReactElement<PersonModalProps> => {
```
OzzieOrca, line 78: Doesn't `React.FC` provide this?
```suggestion
}) => {
```

OzzieOrca (approved):
Which test/line is making that happen? Looks like you've got several tests and queries that look similar.

src/components/Contacts/ContactDetails/ContactDetailsTab/People/Items/PersonModal/PersonModal.tsx
TheNoodleMoose: Yeah, I'll update this.

src/components/Contacts/ContactDetails/ContactDetailsTab/People/Items/PersonModal/PersonModal.test.tsx
```diff
@@ -0,0 +1,955 @@
...
+      );
+      expect(
+        operation.variables.attributes.linkedinAccounts[1].destroy,
+      ).toEqual(true);
+      expect(operation.variables.attributes.websites[1].destroy).toEqual(true);
+    });
+  });
+
+  describe('Creating', () => {
+    it('should handle creating a new person', async () => {
```
TheNoodleMoose, line 790: @OzzieOrca This is the test that is having issues. As I said, the test works fine and I'm able to test all functionality except the cache update logic.

PR #156, opened 2021-08-23
Appeal details page

CSimbulan:
Part of the UI for the appeal details page. It's very big so I want to do it in parts. Nothing connected to GQL yet. 

image
image
image
image
image
image


src/components/Tool/Appeal/Appeals.tsx
```diff
@@ -33,7 +33,7 @@ const Appeals = (): ReactElement => {
         <Box display="flex" justifyContent="center" mt={10}>
           <LoadingIndicator color="primary" size={40} />
         </Box>
-      ) : data?.primaryAppeal ? (
+      ) : data?.primaryAppeal && data.primaryAppeal.nodes.length > 0 ? (
```
CSimbulan, line 36: I had to add this back because just having 

`data?.primaryAppeal`

Will always return true even if the list is empty so it would crash since there's no .nodes[0]

src/components/Tool/Appeal/Appeals.tsx
```diff
@@ -61,7 +61,7 @@ const Appeals = (): ReactElement => {
         <Box display="flex" justifyContent="center" mt={10}>
           <LoadingIndicator color="primary" size={40} />
         </Box>
-      ) : data?.regularAppeals ? (
+      ) : data?.regularAppeals && data.regularAppeals.nodes.length > 0 ? (
```
CSimbulan, line 64: See comment on line 36

src/components/Tool/Appeal/AppealDetails/AppealDetailsAsked.tsx
```diff
@@ -0,0 +1,139 @@
...
+      contact.frequency ? contact.frequency : ''
+    }`,
+  }));
+
+  const updateSelected = (e: GridSelectionModel): void => {
+    const temp: string[] = [];
+    for (const x of e) {
+      temp.push(rows[parseInt(x.toString())].contact);
+    }
+    setAppealState({ ...appealState, selected: [...temp] });
```
OzzieOrca, line 116: I think `map` could be better here.

OzzieOrca (approved):

PR #158, opened 2021-08-24
Rest to gql entry histories of financial accounts

hotcakedev628:
This PR is a rest-to-gql for entry histories of financial accounts.


OzzieOrca (approved):
Looks good :)

PR #167, opened 2021-08-30
Fix mailing address UI

CSimbulan:
image
image
image
image

Mobile View:

image


pages/accountLists/[accountListId]/tools/fixMailingAddresses.page.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { ReactElement } from 'react';
+import { useTranslation } from 'react-i18next';
+import Head from 'next/head';
+import FixMailingAddresses from '../../../../src/components/Tool/FixMailingAddresses/FixMailingAddresses';
+
+const fixMailingAddresses = (): ReactElement => {
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/tools/fixMailingAddresses.page.tsx:6]: I think `React.FC<MyProps>` is usually better and you get the children prop for free.
```suggestion
const fixMailingAddresses: React.FC = () => {
```

src/components/Tool/FixMailingAddresses/Contact.tsx
```diff
@@ -0,0 +1,303 @@
...
+  locationType: string;
+  city: string;
+  state?: string;
+  zip: string;
+  country: string;
+  region?: string;
+  metro?: string;
+  primary: boolean;
+  valid: boolean;
+  new?: boolean;
```
OzzieOrca, line 110: `new` is a reserved keyword in JS. At the very least it's throwing off the Github code formatting...

OzzieOrca (changes requested):
Looks great. A couple places still need translation but other than that, it's good.

OzzieOrca (approved):

PR #171, opened 2021-09-02
Mpdx 7041 UI partner giving analysis report

hotcakedev628:


src/components/Reports/PartnerGivingAnalysisReport/PartnerGivingAnalysisReport.tsx
```diff
@@ -0,0 +1,412 @@
...
+  id: string;
+  name: string;
+  lifeTimeTotal: number;
+};
+
+type OrderBy = keyof Contact | null;
+
+const applyFilters = (contacts: Contact[], query: string): Contact[] => {
+  return contacts.filter((contact) => {
+    let matches = true;
```
OzzieOrca, line 35: I think this function might be more readable if you just return true or false instead of storing that in a variable and returning it later. The `.forEach` could be converted to a `.some` or a `.every` and the return value could be used.

Although filter/sort/pagination eventually needs to be handled on the API so if this is temporary code, no need to spend more time on it.

OzzieOrca (approved):

OzzieOrca (approved):

PR #173, opened 2021-09-03
MPDX-7029: Contact Filters Connected to Contacts Query

reldredge71:
Use ContactFiltersGroup to populate the Contacts Filters panel, and pass the selected filters into the Contacts query.


pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx
```diff
@@ -37,12 +38,16 @@ const ContactsPage: React.FC = () => {
 
   const [filterPanelOpen, setFilterPanelOpen] = useState<boolean>(false);
   //TODO: Connect these to ContactFilters, and use actual filter data for activeFilters
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx:40]: Is this resolved?

pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx
```diff
@@ -37,12 +38,16 @@ const ContactsPage: React.FC = () => {
...
+  const [activeFilters, setActiveFilters] = useState<ContactFilterSetInput>({});
 
   const toggleFilterPanel = () => {
     setFilterPanelOpen(!filterPanelOpen);
   };
 
+  const updateSelectedFilters = (filters: ContactFilterSetInput) => {
+    setActiveFilters(filters);
+  };
+
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx:50]: This doesn't seem like a very useful function and is just indirection.

src/components/Contacts/ContactFilters/ContactFilters.tsx
```diff
@@ -14,13 +14,38 @@ import {
...
+      outputKey += part;
+      return;
+    }
+
+    const capitalizedPart = part.charAt(0).toUpperCase() + part.slice(1);
+
+    outputKey += capitalizedPart;
+  });
+
+  return outputKey;
```
OzzieOrca, line 46:
```suggestion
  return stringParts.reduce((outputKey, part, index) => {
    if (index === 0) {
      return part;
    }
    return `${outputKey)${part.charAt(0).toUpperCase()}${part.slice(1)}1;
  }, '');
```

src/components/Contacts/ContactFilters/ContactFilters.tsx
```diff
@@ -59,14 +86,18 @@ export const ContactFilters: React.FC<Props & BoxProps> = ({
...
-    [name: string]: boolean | string | Array<string>;
-  }>({});
+  const [selectedFilters, setSelectedFilters] = useState<ContactFilterSetInput>(
+    {},
+  );
   const [showAll, setShowAll] = useState(false);
 
+  useEffect(() => {
+    onSelectedFiltersChanged(selectedFilters);
+  }, [selectedFilters]);
```
OzzieOrca, line 96: This could be done in `updateSelectedFilter` too. Which do you think is better? I'm ok with this, just usually try and justify extra `useEffect` hooks.

src/components/Contacts/ContactFilters/ContactFilters.tsx
```diff
@@ -81,9 +112,11 @@ export const ContactFilters: React.FC<Props & BoxProps> = ({
   };
 
   const getSelectedFilters = (group: FilterGroup) =>
-    group.filters.filter((value) => selectedFilters[value.name]);
+    group.filters.filter(
+      (value) => selectedFilters[value.filterKey as ContactFilterKey],
```
OzzieOrca, line 116: Does this need to be camel cased?

src/components/Layouts/SidePanelsLayout.tsx
```diff
@@ -129,7 +129,7 @@ export const SidePanelsLayout: FC<SidePanelsLayoutProps> = ({
           flexBasis={leftWidth}
           breakpoint="sm"
         >
-          {leftPanel}
+          <ScrollBox isscroll={isScrollBox ? 1 : 0}>{leftPanel}</ScrollBox>
```
OzzieOrca, line 132: There's been some churn on this line. Is this a change you made or a bad merge?

src/components/Shared/Filters/FilterListItem.tsx
```diff
@@ -1,57 +1,75 @@
...
+import { FilterListItemNumericRange } from './FilterListItemNumericRange';
 import { FilterListItemSelect } from './FilterListItemSelect';
 import { FilterListItemTextField } from './FilterListItemTextField';
 
 interface Props {
   filter: Filter;
-  value?: boolean | string | Array<string>;
-  onUpdate: (value?: boolean | string | Array<string>) => void;
+  value?: ContactFilterValue;
+  onUpdate: (value?: boolean | string | Array<string> | DateRangeInput) => void;
```
OzzieOrca, line 25: Is this the right type? Should it be this?
```suggestion
  onUpdate: (value?: ContactFilterValue) => void;
```

src/components/Shared/Filters/FilterListItem.tsx
```diff
@@ -1,57 +1,75 @@
...
 }
 
 export const FilterListItem: React.FC<Props> = ({
   filter,
   value,
   onUpdate,
 }: Props) => {
-  return filter.type === 'text' ? (
+  return (filter as TextFilter).__typename === 'TextFilter' ? (
```
OzzieOrca, line 33: There should be a way to make TS happy here without casting. Worst case we could use this https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates. But there should be a better way. What does it do without the casts? My hunch is it might just work but there's probably something wrong with it if you had to cast things.

OzzieOrca (changes requested):
Looking good. Thanks for working through all the filter stuff :)

src/components/Layouts/SidePanelsLayout.tsx
reldredge71: Yes I made this change.  I don't know if the left panel used to scroll and then the scrolling was disabled, but I think we need it to scroll again -- the list of filters sometimes extends far past the bottom of the screen.  Especially since we render MultiSelect filters as a long list of checkboxes -- the city filter in particular becomes a verbose list of every individual city that our contacts are from.  I think we need this panel to scroll, independent of the main view.

pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx
Comment (reldredge71) [reply pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx:40]: Yes, I can remove that.

src/components/Contacts/ContactFilters/ContactFilters.tsx
reldredge71: Yeah that sounds good.  I know what you mean, adding more hooks in general could bog down the screen with extra rerenders.  `useEffect` made sense to me because we're mostly responding to changes in the `selectedFilters` state, but calling this callback directly after setting the state is good because it reduces the need for hooks.

src/components/Contacts/ContactFilters/ContactFilters.tsx
reldredge71: Good catch, it probably does.

src/components/Shared/Filters/FilterListItem.tsx
reldredge71: I think it doesn't work because of the way `ContactFilters.graphql` is set up... `__typename` is not part of the main filter type, but it gets added among the type-specific properties.  I can try reworking this and see if there is a better solution.

src/components/Contacts/ContactFilters/ContactFilters.tsx
OzzieOrca: I think this is still outstanding. Everything else looks good :)

OzzieOrca (changes requested):

src/components/Contacts/ContactFilters/ContactFilters.tsx
reldredge71: I think I made this change... could you check again?

OzzieOrca: I missed that somehow... maybe I clicked the link to the outdated file... Looks good.

OzzieOrca (approved):

PR #175, opened 2021-09-03
Navbar tools dropdown

CSimbulan:
Dropdown menu for tools. There was an issue with the translations, I left a specific comment for that section

image
image

### DIFF
--- a/src/components/Layouts/Primary/TopBar/Items/NavMenu/NavMenu.tsx
+++ b/src/components/Layouts/Primary/TopBar/Items/NavMenu/NavMenu.tsx
@@ -4,7 +4,6 @@ import {
   Grid,
   MenuItem,
   ListItemText,
-  Theme,
   Popper,
   Grow,
   ClickAwayListener,
@@ -22,8 +21,9 @@ import { ReportNavItems } from '../../../../../Reports/NavReportsList/ReportNavI
 import { ToolsList } from '../../../../../Tool/Home/ToolList';
 import { useAccountListId } from '../../../../../../hooks/useAccountListId';
 import { useCurrentToolId } from '../../../../../../hooks/useCurrentToolId';
+import theme from '../../../../../../theme';
 
-const useStyles = makeStyles((theme: Theme) => ({
+const useStyles = makeStyles(() => ({
   navListItem: {
     order: 2,
     [theme.breakpoints.down('sm')]: {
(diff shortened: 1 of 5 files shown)


src/components/Tool/Home/ToolList.ts
```diff
@@ -12,104 +12,99 @@ import {
   mdiAccountGroup,
   mdiWrench,
 } from '@mdi/js';
-import i18n from 'i18next';
```
CSimbulan, line 15: Translating directly in this file was giving an error when trying to get them onto the navbar, they would always come up as undefined. It was even affecting the tools landing page. I changed it to the interpolation method but I'm not sure if that is the best way. The "reports" dropdown translates them as a dynamic key I think also

src/components/Layouts/Primary/TopBar/Items/NavMenu/NavMenu.tsx
```diff
@@ -36,28 +40,52 @@ const useStyles = makeStyles((theme: Theme) => ({
...
   const anchorRef = React.useRef<HTMLLIElement>(null);
+  const anchorRefTools = React.useRef<HTMLLIElement>(null);
 
   const handleReportsMenuToggle = () => {
     setReportsMenuOpen((prevOpen) => !prevOpen);
+    handleToolsMenuClose();
   };
 
-  const handleReportsMenuClose = (event: React.MouseEvent<EventTarget>) => {
-    if (anchorRef.current?.contains(event.target as HTMLElement)) {
```
CSimbulan, line 54: I'm not sure how important this is but in order to get the codecov to pass I removed this since idk how to test this line. Everything still seems to work fine

src/components/Tool/Home/ToolList.ts
OzzieOrca: Hmm... I wonder if there's a circular dependency or something... 

What you have won't translate anything... We could come back to it... Or we could just hardcode them in the menu instead of looping through them. I don't think we need the tools landing page anymore so we could just have them in one place.

src/components/Layouts/Primary/TopBar/Items/NavMenu/NavMenu.tsx
OzzieOrca: Seems like maybe this was to prevent the menu from closing when something inside it was clicked? Maybe there was a submenu at one point. Idk. Seems to behave fine. Idk that I see a reason for this either.

OzzieOrca (approved):

src/components/Tool/Home/ToolList.ts
CSimbulan: Hmm ok I'll just leave it for now incase we do want to keep the landing page for some reason, since its not really needed in the original app either but its still there

PR #177, opened 2021-09-08
MPDX-7095-Add-Logic-For-Editing-Partnership-Info

TheNoodleMoose:
Most of the work for this modal is here in this PR, except the option to edit ```Referred By```. That part is a bit more complex so I decided to make a [separate ticket](https://jira.cru.org/browse/MPDX-7121) to address it since this PR is already pretty big. Feel free to ask questions or give suggestions 🙂


src/components/Contacts/ContactDetails/ContactDontationsTab/PartnershipInfo/PartnershipInfo.tsx
```diff
@@ -50,27 +60,47 @@ const PartnershipTitle = styled(Typography)(({ theme }) => ({
...
+        >
+          <PartnershipEditIcon />
+        </IconButton>
+      </Box>
       <IconAndTextContainer>
         <IconContainer>
           <HandshakeIcon />
         </IconContainer>
         <Box style={{ margin: 0, padding: 0 }} role="cell">
-          <LabelsAndText variant="subtitle1">{status}</LabelsAndText>
```
TheNoodleMoose, line 68: I'm not sure what ```status``` this is exactly. It definitely wasn't the contact status 🤷‍♂️

src/components/Contacts/ContactDetails/ContactDontationsTab/PartnershipInfo/EditPartnershipInfoModal/EditPartnershipInfoModal.tsx
```diff
@@ -0,0 +1,431 @@
...
+    setFieldValue: (name: string, value: StatusEnum | number | null) => void,
+  ) => {
+    if (status !== StatusEnum.PartnerFinancial) {
+      setFieldValue('status', status);
+      setFieldValue('pledgeAmount', 0);
+      setFieldValue('pledgeFrequency', null);
+    } else {
+      setFieldValue('status', status);
+    }
+  };
```
TheNoodleMoose, line 150: The expected behavior here is that there should only be an amount and frequency when the ```status``` is ```PartnerFinancial```. So this clears the input fields and I disable them from being edited and show a hover icon explaining the behavior, but only when the new status is not ```PartnerFinancial```. I wanted to set ``` pledgeAmount``` to ```null``` but since the input type is ```type="number"``` it causes some really strange behavior if you set the value to ```null```, so I just settled for ```0```.

src/components/Contacts/ContactDetails/ContactDontationsTab/PartnershipInfo/EditPartnershipInfoModal/EditPartnershipInfoModal.tsx
```diff
@@ -0,0 +1,431 @@
...
+                        horizontal: 'left',
+                      },
+                      getContentAnchorEl: null,
+                      PaperProps: {
+                        style: {
+                          maxHeight: '300px',
+                          overflow: 'auto',
+                        },
+                      },
+                    }}
```
TheNoodleMoose, line 266: I found this fix for the select menus being positioned strange, especially for larger lists. So now it's only ```300px``` and is directly/correctly positioned under the select input.

src/components/Contacts/ContactDetails/ContactDontationsTab/PartnershipInfo/EditPartnershipInfoModal/EditPartnershipInfoModal.tsx
```diff
@@ -0,0 +1,431 @@
...
+                    />
+                  }
+                  label={t('Commitment Recieved')}
+                />
+              </ContactInputWrapper>
+              <ContactInputWrapper>
+                <CheckboxLabel
+                  control={
+                    <Checkbox
+                      checked={!!!noAppeals}
```
LizzyRandall, line 372: This is interesting logic? Wouldn't !noAppeals be the same? or is that some javascript magic I don't know which is probable 😄

src/components/Contacts/ContactDetails/ContactDontationsTab/PartnershipInfo/EditPartnershipInfoModal/EditPartnershipInfoModal.tsx
```diff
@@ -0,0 +1,431 @@
...
+
+  return (
+    <Modal
+      isOpen={isOpen}
+      title={t('Edit Partnership')}
+      handleClose={handleClose}
+    >
+      <Formik
+        initialValues={{
+          id: contact.id,
```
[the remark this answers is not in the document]
TheNoodleMoose: In this case, we only want these specific values from the contact since these are the only values that can be edited in this modal. If we added ```...contact```, 1. I think it would throw a type error because the other values are excluded in our validation schema, 2. That would imply to the Formik component that those fields can be edited and will be needed for this form.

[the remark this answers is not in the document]
LizzyRandall: Ah ok 👍

src/components/Contacts/ContactDetails/ContactDontationsTab/PartnershipInfo/EditPartnershipInfoModal/EditPartnershipInfoModal.tsx
```diff
@@ -0,0 +1,431 @@
...
+                    disabled={status !== StatusEnum.PartnerFinancial}
+                    aria-readonly={status !== StatusEnum.PartnerFinancial}
+                    onChange={handleChange('pledgeAmount')}
+                    inputProps={{ 'aria-label': t('Amount') }}
+                    endAdornment={
+                      <Tooltip
+                        title={
+                          <Typography>
+                            {t(
+                              'Commitments can only be set if status is Partner - Financial',
```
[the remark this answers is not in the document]
TheNoodleMoose: From what I could pick up from the existing UI, yeah only financial partners can have commitments. If you switch their status to something that is not ```Partner - Financial```, it clears the amount and frequency fields.

src/components/Contacts/ContactDetails/ContactDontationsTab/PartnershipInfo/EditPartnershipInfoModal/EditPartnershipInfoModal.tsx
TheNoodleMoose: When I read your comment here I started getting really confused trying to think through this, but I remembered why I did this! ```noAppeals``` can be ```null```, so the double bang or ```!!``` can convert non-booleans to be booleans. So ```!!null``` would be ```false```.  And then it gets a little weird. So it's a strange situation, the value is ```noAppeals```, but the UI is ```Send Appeals```, so you have to do the opposite of ```noAppeals```. So let's say ```noAppeals``` was false, meaning the UI would need to display ```Send Appeals``` as checked, or ```true```. Adding ```!!``` would make it stay ```false``` or would convert ```null``` to ```false```. But the UI needs it to be true,  so we add one more bang to get ```!!!``` and that makes it ```true```. Confusing...yes, but it works haha!

src/components/Contacts/ContactDetails/ContactDontationsTab/PartnershipInfo/EditPartnershipInfoModal/EditPartnershipInfoModal.tsx
```diff
@@ -0,0 +1,431 @@
...
+          pledgeStartDate: contact.pledgeStartDate,
+          nextAsk: contact.nextAsk,
+          noAppeals: contact.noAppeals,
+        }}
+        validationSchema={contactPartnershipSchema}
+        onSubmit={onSubmit}
+      >
+        {({
+          values: {
+            status,
```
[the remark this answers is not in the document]
TheNoodleMoose: If we added the rest of the contact, all those other values would be accessible here. But we don't need them since only certain values are going to be edited by this modal. The best practice I've discovered is to only pass in what is going to be edited in the form.

src/components/Contacts/ContactDetails/ContactDontationsTab/PartnershipInfo/EditPartnershipInfoModal/EditPartnershipInfoModal.tsx
OzzieOrca: I think this is the same as
```suggestion
                      checked={!noAppeals}
```
The first bang should return a boolean and invert the value. Everything after just inverts it since it's already a boolean.

src/components/Contacts/ContactDetails/ContactDontationsTab/PartnershipInfo/EditPartnershipInfoModal/EditPartnershipInfoModal.tsx
```diff
@@ -0,0 +1,431 @@
...
+      | 'pledgeAmount'
+      | 'pledgeFrequency'
+      | 'pledgeCurrency'
+      | 'pledgeReceived'
+      | 'pledgeStartDate'
+      | 'nextAsk'
+      | 'noAppeals'
+    >,
+  ) => {
+    console.log(attributes.nextAsk);
```
OzzieOrca, line 125: Still needed?

OzzieOrca (approved):

src/components/Contacts/ContactDetails/ContactDontationsTab/PartnershipInfo/EditPartnershipInfoModal/EditPartnershipInfoModal.tsx
TheNoodleMoose: Ah thanks, I'll remove that.

src/components/Contacts/ContactDetails/ContactDontationsTab/PartnershipInfo/EditPartnershipInfoModal/EditPartnershipInfoModal.tsx
TheNoodleMoose: Ah, so I was overcomplicating it haha. Thanks, I'll change that!

OzzieOrca (approved):

PR #179, opened 2021-09-13
MPDX-7121-Add-Ability-To-Edit-ReferredBy-In-Edit-Partnership-Modal

TheNoodleMoose:
This was a bit complicated to work on, more so than I initially thought lol. What's complicated is that you have to persist the data as the user searches for new contacts, as the data is continually changing because of the auto-complete component. And then if they decide to delete a contact referral that was already created, I had to figure out a way to remove it from the UI, but still persist it as it's not truly removed until ```destroyed``` is set to ```true``` and is sent to the api. There's just a lot of moving parts and pieces. Feel free to try it out and make sure it works. I tried to cover all the scenarios but maybe I missed something 🙂 I'll be adding tests in a separate commit as I wanted others to take an initial look at the logic for this.

### DIFF
--- a/src/components/Contacts/ContactDetails/ContactDontationsTab/PartnershipInfo/EditPartnershipInfoModal/EditPartnershipInfoModal.tsx
+++ b/src/components/Contacts/ContactDetails/ContactDontationsTab/PartnershipInfo/EditPartnershipInfoModal/EditPartnershipInfoModal.tsx
@@ -333,6 +333,23 @@ export const EditPartnershipInfoModal: React.FC<EditPartnershipInfoModalProps> =
                     onChange={(e) =>
                       updateStatus(e.target.value as StatusEnum, setFieldValue)
                     }
+                    MenuProps={{
+                      anchorOrigin: {
+                        vertical: 'bottom',
+                        horizontal: 'left',
+                      },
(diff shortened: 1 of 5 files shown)


src/components/Contacts/ContactDetails/ContactDontationsTab/PartnershipInfo/EditPartnershipInfoModal/EditPartnershipInfoModal.tsx
```diff
@@ -242,38 +374,41 @@ export const EditPartnershipInfoModal: React.FC<EditPartnershipInfoModalProps> =
...
+                        getContentAnchorEl: null,
+                        PaperProps: {
+                          style: {
+                            maxHeight: '300px',
+                            overflow: 'auto',
+                          },
+                        },
+                      }}
+                    >
+                      <MenuItem value={''} disabled></MenuItem>
```
TheNoodleMoose, line 400: I was getting errors about the value not existing in the select. Turns out MUI Select components do not support ```null``` and they don't plan on it. So you have to turn the value into an empty string and then create this dummy menu item so the error won't happen.

reldredge71 (approved):

PR #180, opened 2021-09-14
update ui adding responsive drawer

hotcakedev628:
This PR is to make the current ui responsive by adding mobile side menu


src/components/Layouts/Primary/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,196 @@
...
+  pathname,
+  depth = 0,
+}: {
+  items: Item[];
+  pathname: string;
+  depth?: number;
+}) {
+  return (
+    <List disablePadding>
+      {items?.reduce(
```
OzzieOrca, line 87:
```suggestion
      {items?.reduce<ReactElement[]>(
```

src/components/Layouts/Primary/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,196 @@
...
+}
+
+function reduceChildRoutes({
+  acc,
+  pathname,
+  item,
+  depth,
+}: {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  acc: any[];
```
OzzieOrca, line 108:
```suggestion
  acc: ReactElement[];
```

src/components/Layouts/Primary/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,196 @@
...
+  pathname: string;
+  item: Item;
+  depth: number;
+}) {
+  const key = item.title + depth;
+  const accountListId = useAccountListId();
+  const { t } = useTranslation();
```
OzzieOrca, line 115: I think these violate the [rules of hooks](https://reactjs.org/docs/hooks-rules.html). Hooks should be called at the root level of the component before any conditions so they are always executed in the same order.

OzzieOrca (changes requested):

src/components/Layouts/Primary/NavBar/NavBar.tsx
OzzieOrca: You can keep them at the top level of a component but they just can't be in helper functions.

src/components/Layouts/Primary/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,208 @@
...
+const useStyles = makeStyles((theme: Theme) => ({
+  mobileDrawer: {
+    width: 290,
+    backgroundColor: theme.palette.cruGrayDark.main,
+    zIndex: theme.zIndex.drawer + 200,
+  },
+}));
+
+export const NavBar: FC<NavBarProps> = ({
+  accountListId,
```
OzzieOrca, line 162: This can be a hook in the `NavBar` component, it just has to be called in the component function itself and not after any conditions that would return early or in helper functions.

src/components/Layouts/Primary/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,208 @@
...
+  depth,
+}: {
+  acc: ReactElement[];
+  accountListId: string;
+  pathname: string;
+  item: Item;
+  depth: number;
+}) {
+  const key = item.title + depth;
+  const { t } = useTranslation();
```
OzzieOrca, line 120: I think this needs to be called in `NavBar` since it's a helper function. Although maybe you're treating it as a component...

OzzieOrca (changes requested):
I think this is good. The hook calls could use a little cleanup still.

OzzieOrca (changes requested):
One more translation related thing

PR #181, opened 2021-09-14
7132 - Coaching List

gyasistory:
This is the first start of the Coaching Section.

In another PR I will add Infinite scrolling and Coaching Detail Page.


reldredge71 (approved):

src/components/Coaching/CoachingList.stories.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { ReactElement } from 'react';
+import { MockedProvider } from '@apollo/client/testing';
+import { CoachingList } from './CoachingList';
+import {
+  LoadCoachingListDocument,
+  LoadCoachingListQuery,
+} from './LoadCoachingList.generated';
+import { GqlMockedProvider } from '__tests__/util/graphqlMocking';
```
TheNoodleMoose, line 8: Import doesn't look right either.

src/components/Coaching/CoachingList.tsx
```diff
@@ -0,0 +1,77 @@
...
+} from './LoadCoachingList.generated';
+
+const CoachingListWrapper = styled(Box)(({ theme }) => ({
+  width: '100%',
+  padding: theme.spacing(1),
+}));
+const CoachingTitleWrapper = styled(Box)(({ theme }) => ({
+  width: '100%',
+  display: 'flex',
+  flexDirection: 'row',
```
TheNoodleMoose, line 19: Shouldn't need this as the direction's initial value should be ```row``` by default.

src/components/Coaching/CoachingList.tsx
```diff
@@ -0,0 +1,77 @@
...
+      <Box>
+        {loading ? (
+          <>
+            <CoachingLoading />
+            <CoachingLoading />
+            <CoachingLoading />
+          </>
+        ) : (
+          <>
+            {coaches?.nodes.map((coach, _index) => {
```
TheNoodleMoose, line 64: Does this page need some sort of empty or "No accounts" view in case ```coaches.nodes.length``` is ```0```?

src/components/Coaching/CoachingRow/CoachingRow.stories.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { ReactElement } from 'react';
+import { CoachFragment, CoachFragmentDoc } from '../LoadCoachingList.generated';
+import { CoachingRow } from './CoachingRow';
+import { gqlMock } from '__tests__/util/graphqlMocking';
```
TheNoodleMoose, line 4: Import doesn't look right.

src/components/Coaching/CoachingRow/CoachingRow.tsx
```diff
@@ -0,0 +1,141 @@
+import { Box, styled, Typography } from '@material-ui/core';
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { CoachFragment } from '../LoadCoachingList.generated';
+import { currencyFormat, percentageFormat } from 'src/lib/intlFormat';
+import StyledProgress from 'src/components/StyledProgress';
```
TheNoodleMoose, line 6: Imports don't look right.

src/components/Coaching/CoachingRow/CoachingRow.tsx
```diff
@@ -0,0 +1,141 @@
...
+    totalPledges,
+    receivedPledges,
+    primaryAppeal,
+  } = coach;
+
+  const calculatedMonthlyGoal = monthlyGoal ? monthlyGoal : 0;
+  const percentageRecievedPledges = receivedPledges / calculatedMonthlyGoal;
+  const percentageTotalPledges = totalPledges / calculatedMonthlyGoal;
+  const recievedCurrency = currencyFormat(receivedPledges, currency)
+    ? currencyFormat(receivedPledges, currency)
```
TheNoodleMoose, line 45: Is the check here needed? I think ```currencyFormat ```with always at least return ```0``` and the currency symbol.

src/components/Coaching/LoadCoachingList.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { useLoadCoachingListQuery } from './LoadCoachingList.generated';
+
+describe('LoadCoaching', () => {
+  it('query correct', async () => {
+    const { result, waitForNextUpdate } = renderHook(
+      () => useLoadCoachingListQuery(),
+      {
+        wrapper: GqlMockedProvider,
+      },
+    );
```
TheNoodleMoose, line 12: I think that the test should be testing rendering the component and not testing the query alone. If the component renders, and we can query the DOM for the data we expect then that determines that the query ran successfully.

TheNoodleMoose (dismissed):

src/components/Coaching/CoachingList.tsx
```diff
@@ -0,0 +1,77 @@
+import { Box, Divider, styled, Typography } from '@material-ui/core';
+import { EcoOutlined } from '@material-ui/icons';
+import { Skeleton } from '@material-ui/lab';
+import React, { ReactElement } from 'react';
```
TheNoodleMoose, line 4: Suuuuper unimportant comment here, just my preference/opinion but I always like to have ```react``` be the top import for the file. Just keeps things consistent. It's fine as is 🙂

src/components/Coaching/CoachingRow/CoachingRow.stories.tsx
gyasistory: @TheNoodleMoose this is the only option I am seeing do you know the correct import?

src/components/Coaching/CoachingRow/CoachingRow.tsx
hotcakedev628: The import should be good.
We have an absolute path setup in [tsconfig](https://github.com/CruGlobal/mpdx-react/blob/main/tsconfig.json) file.
I am pretty sure it provides more clean code.
https://nextjs.org/docs/advanced-features/module-path-aliases

TheNoodleMoose: Ah interesting, I had issues locally when I had this come up. If it works for you and others then that's good with me! I may just have to figure out my local setup.

src/components/Coaching/CoachingList.tsx
hotcakedev628: @gyasistory You don't need to style Box component.
You can style it using only the apis.
```
<Box display="flex" width="100%">
...
</Box>
```

src/components/Coaching/CoachingRow/CoachingRow.stories.tsx
gyasistory: @TheNoodleMoose I don't think it is working because my mock data is no longer working.

src/components/Coaching/CoachingList.tsx
gyasistory: @TheNoodleMoose there is no design for it right now.  Also, it should never be empty because you coach yourself.

TheNoodleMoose:
Looks like the deployment failed because of the task filters issue Scotty mentioned yesterday.

src/components/Coaching/CoachingRow/CoachingRow.tsx
```diff
@@ -0,0 +1,146 @@
...
+            </Typography>
+            <Typography>
+              {appealProcessedCurrency +
+                ' (' +
+                percentageFormat(percentageAppealProcessed) +
+                ') / ' +
+                appealTotalCurrency +
+                '(' +
+                percentageFormat(percentageAppealTotal) +
+                ')'}
```
OzzieOrca, line 127: Can you use template literals for this? It's hard to read.

src/components/Coaching/LoadCoachingList.graphql
```diff
@@ -0,0 +1,35 @@
...
+    pageInfo {
+      startCursor
+      endCursor
+      hasNextPage
+      hasPreviousPage
+    }
+  }
+}
+
+fragment Coach on CoachingAccountList {
```
OzzieOrca, line 17: Should this be named Coachee or CoachedPerson or CoachedAccountList or something? Idk what the correct word is. But it's a list of people being coached, not coaches right?

OzzieOrca (changes requested):

src/components/Layouts/Primary/TopBar/Items/NavMenu/NavMenu.tsx
```diff
@@ -265,11 +264,15 @@ const NavMenu = (): ReactElement => {
             </Popper>
           </Grid>
           <Grid item className={classes.navListItem}>
-            <HandoffLink path="/coaches">
+            <NextLink
+              href="/accountLists/[accountListId]/coaches"
+              as={`/accountLists/${accountListId}/coaches`}
+              scroll={false}
```
OzzieOrca, line 270: Is there a reason we don't want to scroll to the top?

OzzieOrca (changes requested):
Looks good. I think there's a couple other places (variable names and route names) where you are calling "coached people" "coaches". Might be misleading in the future.

pages/accountLists/[accountListId]/coaches.page.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { useTranslation } from 'react-i18next';
+import { CoachingList } from 'src/components/Coaching/CoachingList';
+
+const CoachesPage = (): ReactElement => {
+  const { t } = useTranslation();
+
+  return (
+    <>
+      <Head>
+        <title>MPDX | {t('Coaches')}</title>
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/coaches.page.tsx:12]:
```suggestion
        <title>MPDX | {t('Coaching Accounts')}</title>
```
(It's what the current app has)

I think `Coaching` could work too.

src/components/Coaching/CoachingList.tsx
```diff
@@ -0,0 +1,73 @@
...
+const CoachingLoading = styled(Skeleton)(() => ({
+  width: '75%',
+  height: '50px',
+}));
+
+export const CoachingList = (): ReactElement => {
+  const { data, loading } = useLoadCoachingListQuery();
+  const { t } = useTranslation();
+
+  const coaches = data?.coachingAccountLists;
```
OzzieOrca, line 39:
```suggestion
  const coachingAccounts = data?.coachingAccountLists;
```

src/components/Coaching/CoachingList.tsx
```diff
@@ -0,0 +1,73 @@
...
+            <CoachingLoading />
+            <CoachingLoading />
+            <CoachingLoading />
+          </>
+        ) : (
+          <>
+            {coaches?.nodes.map((coach, _index) => {
+              return (
+                <>
+                  <CoachingRow coach={coach} />
```
OzzieOrca, line 63:
```suggestion
            {coachingAccounts?.nodes.map((coachingAccount, _index) => {
              return (
                <>
                  <CoachingRow coachingAccount={coachingAccount} />
```

src/components/Coaching/CoachingRow/CoachingRow.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { CoachedPersonFragment } from '../LoadCoachingList.generated';
+import {
+  currencyFormat,
+  numberFormat,
+  percentageFormat,
+} from 'src/lib/intlFormat';
+import StyledProgress from 'src/components/StyledProgress';
+
+interface Props {
+  coach: CoachedPersonFragment;
```
OzzieOrca, line 13:
```suggestion
  coachingAccount: CoachedPersonFragment;
```

pages/accountLists/[accountListId]/coaches.page.tsx
```diff
@@ -0,0 +1,19 @@
...
+      </Head>
+      <CoachingList />
+    </>
+  );
+};
+
+export default CoachesPage;
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/coaches.page.tsx:19]: Can we call this page `/coaching`? I'm open to suggestions but I think this is an opportunity to name it better.
```suggestion
export default CoachingPage;
```

OzzieOrca (changes requested):
I still feel like calling things `coach`/`coaches` is incorrect and misleading. We're working with lists of coachees here. The current user is the coach.

I threw out some suggestions but I'm open to others. Coachee seems applicable but doesn't seem like a very common word, see [here](https://english.stackexchange.com/questions/46283/what-would-you-call-a-person-being-coached) and [here](https://english.stackexchange.com/questions/15515/is-coachee-even-a-word). I think `coaching` or `coaching account` is the best I've got so far.

pages/accountLists/[accountListId]/coaches.page.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { ReactElement } from 'react';
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/coaches.page.tsx:1]: Can you rename this file to `coaching.page.tsx` too? Just to match everything else. The main nav might need a corresponding update.

OzzieOrca (changes requested):

OzzieOrca (approved):

PR #183, opened 2021-09-15
Fix commitment info graphql

CSimbulan:
This page should load the actual data now, but no pagination yet

image

### DIFF
--- a/src/components/Tool/FixCommitmentInfo/Contact.tsx
+++ b/src/components/Tool/FixCommitmentInfo/Contact.tsx
@@ -82,7 +82,7 @@ const useStyles = makeStyles(() => ({
 }));
 
 interface Props {
-  id: string;
+  id?: string;
   name: string;
   tagTitle: string;
   tagValue: string;
--- a/src/components/Tool/FixCommitmentInfo/Contact.tsx
+++ b/src/components/Tool/FixCommitmentInfo/Contact.tsx
@@ -23,7 +23,7 @@ const useStyles = makeStyles(() => ({
     display: 'flex',
     justifyContent: 'flex-start',
     alignItems: 'center',
-    [theme.breakpoints.up('md')]: {
+    [theme.breakpoints.up('lg')]: {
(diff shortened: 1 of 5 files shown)


src/components/Tool/FixCommitmentInfo/Contact.tsx
```diff
@@ -154,19 +156,19 @@ const Contact: React.FC<Props> = ({
                   value={values.tagValue}
                   onChange={(event) => handleChange(event, 'tagValue')}
                 >
-                  {contactTags.map((t) => (
+                  {Object.keys(contactTags).map((key) => (
```
OzzieOrca, line 159: Since you seem to be using both the key and the value, I would do
```suggestion
                  {Object.entries(contactTags).map(([key, value]) => (
```
(maybe there are better names for those variables in this specific context)

src/components/Tool/FixCommitmentInfo/Contact.tsx
```diff
@@ -201,21 +204,21 @@ const Contact: React.FC<Props> = ({
                   <option value="" disabled>
                     Frequency
                   </option>
-                  {frequencies.map((freq) => (
+                  {Object.keys(frequencies).map((freq) => (
```
OzzieOrca, line 207: `Object.entries` would probably be better here too

src/components/Tool/FixCommitmentInfo/GetInvalidStatuses.graphql
```diff
@@ -0,0 +1,15 @@
+query GetInvalidStatuses($accountListId: ID!) {
+  contacts(
+    accountListId: $accountListId
+    contactsFilter: { statusValid: false }
```
OzzieOrca, line 4: The current version applies the deceased filter also
```suggestion
    contactsFilter: { statusValid: false, deceased: false }
```

Although it's not in GraphQL yet. I messaged @soberstadt.

OzzieOrca (changes requested):

OzzieOrca (approved):

PR #186, opened 2021-09-15
Fix send newsletter gql

CSimbulan:
Fix newsletter page should now use real contact data. Some if it missing though because it needs pagination.

### DIFF
--- a/src/components/Tool/FixSendNewsletter/Contact.tsx
+++ b/src/components/Tool/FixSendNewsletter/Contact.tsx
@@ -146,21 +146,10 @@ const Contact = ({
                     value={newsletter}
                     onChange={(event) => handleChange(event)}
                   >
-                    <option
-                      value="PHYSICAL"
-                      selected={newsletter === 'PHYSICAL'}
-                    >
-                      {t('Physical')}
-                    </option>
-                    <option value="EMAIL" selected={newsletter === 'EMAIL'}>
-                      {t('Email')}
-                    </option>
-                    <option value="BOTH" selected={newsletter === 'BOTH'}>
-                      {t('Both')}
-                    </option>
-                    <option value="NONE" selected={newsletter === 'NONE'}>
-                      {t('None')}
-                    </option>
+                    <option value="PHYSICAL">{t('Physical')}</option>
+                    <option value="EMAIL">{t('Email')}</option>
+                    <option value="BOTH">{t('Both')}</option>
+                    <option value="NONE">{t('None')}</option>
                   </NativeSelect>
                 </Box>
               </Grid>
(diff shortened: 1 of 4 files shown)


src/components/Tool/FixSendNewsletter/FixSendNewsletter.tsx
```diff
@@ -39,101 +45,126 @@ const useStyles = makeStyles(() => ({
...
 
-  const toggleData = (): void => {
-    test.length > 0 ? setTest([]) : setTest(testData);
-  };
+  const filtered = data?.contacts.nodes.filter(
+    (el) =>
+      el.status === 'PARTNER_FINANCIAL' ||
+      el.status === 'PARTNER_SPECIAL' ||
+      el.status === 'PARTNER_PRAY',
+  );
```
CSimbulan, line 71: not sure if there is a better way to do this. from my understanding there isn't currently a way to do this directly onto the gql query

src/components/Tool/FixSendNewsletter/GetInvalidNewsletter.graphql
```diff
@@ -0,0 +1,36 @@
+query GetInvalidNewsletter($accountListId: ID!) {
+  contacts(
+    accountListId: $accountListId
+    contactsFilter: { newsletter: NO_VALUE }
```
OzzieOrca, line 4: Same deceased filter needed eventually

src/components/Tool/FixSendNewsletter/FixSendNewsletter.tsx
OzzieOrca: I think you send a bunch of statuses:
```diff
-  contactsFilter: { newsletter: NO_VALUE }
+  contactsFilter: { newsletter: NO_VALUE, status: [PARTNER_FINANCIAL, PARTNER_SPECIAL, PARTNER_PRAY] }

File: src/components/Tool/FixSendNewsletter/Contact.tsx [hunk 3]
```diff
@@ -177,16 +162,23 @@ const Contact = ({
...
+                            height: theme.spacing(7),
+                          }}
+                        />
+                      )}
                       <Box display="flex" flexDirection="column" ml={2}>
-                        <Typography variant="h6">{name || ''}</Typography>
-                        <Typography>{email || ''}</Typography>
+                        <Typography variant="h6">
+                          {`${primaryPerson.firstName} ${primaryPerson.lastName}` ||
+                            ''}
```
OzzieOrca, line 177: I don't think the `|| ''` here does anything.... When both the first name and last names are empty, `' '` is a truthy value so it will be returned instead of `''`.

OzzieOrca (changes requested):

src/components/Tool/FixSendNewsletter/GetInvalidNewsletter.graphql
CSimbulan: I'll make a note to add it later to prevent any errors for now

OzzieOrca (approved):

PR #190, opened 2021-09-18
Nav bar tool notifications

CSimbulan:
image
image


src/components/Layouts/Primary/TopBar/Items/NavMenu/NavMenu.tsx
```diff
@@ -41,27 +43,72 @@ const useStyles = makeStyles(() => ({
     transform: 'rotate(180deg)',
   },
   subMenu: {
-    backgroundColor: theme.palette.cruGrayDark.main,
+    backgroundImage: `linear-gradient(0deg, ${theme.palette.cruGrayDark.main}, ${theme.palette.cruGrayDark.main})`,
```
OzzieOrca, line 46: Can you share some context here? Seems like this is a gradient from `cruGrayDark` to itself... 🤷‍♂️

src/components/Layouts/Primary/TopBar/Items/NavMenu/NavMenu.tsx
```diff
@@ -41,27 +43,72 @@ const useStyles = makeStyles(() => ({
...
+
 const NavMenu = (): ReactElement => {
   const { t } = useTranslation();
   const classes = useStyles();
   const accountListId = useAccountListId();
   const currentToolId = useCurrentToolId();
+  const { data, loading } = useGetToolNotificationsQuery({
+    variables: { accountListId: accountListId ?? '' },
+  });
+  const [dataState, setDataState] = useState<responseData>({});
```
OzzieOrca, line 99: Why is this needed? Can you just use `data` directly?

src/components/Layouts/Primary/TopBar/Items/NavMenu/NavMenu.tsx
```diff
@@ -41,27 +43,72 @@ const useStyles = makeStyles(() => ({
...
+
+  useEffect(() => {
+    const deepCopy = data ? JSON.parse(JSON.stringify(data)) : [];
+    setDataState(deepCopy);
+    setTotal(
+      Object.entries(deepCopy)
+        .map(([tool]) => deepCopy[tool])
+        .flatMap((entry) => entry.totalCount)
+        .reduce((a, b) => a + b, 0),
+    );
```
OzzieOrca, line 110: I think `useMemo` would be better here than `useEffect` and `useState`. Honestly just doing the calculation every render probably wouldn't be the worst thing either. It probably isn't rerendered that much.

src/components/Layouts/Primary/TopBar/Items/NavMenu/NavMenu.tsx
```diff
@@ -41,27 +43,72 @@ const useStyles = makeStyles(() => ({
...
+  });
+  const [dataState, setDataState] = useState<responseData>({});
+  const [total, setTotal] = useState(0);
+
+  useEffect(() => {
+    const deepCopy = data ? JSON.parse(JSON.stringify(data)) : [];
+    setDataState(deepCopy);
+    setTotal(
+      Object.entries(deepCopy)
+        .map(([tool]) => deepCopy[tool])
```
OzzieOrca, line 107:
```suggestion
      Object.values(deepCopy)
```

OzzieOrca (changes requested):

src/components/Layouts/Primary/TopBar/Items/NavMenu/NavMenu.tsx
CSimbulan: I cant do something like data['fixCommitmentInfo'], well it does work but it gives typescript warnings and idk if we can somehow change the return type to allow it to work

src/components/Layouts/Primary/TopBar/Items/NavMenu/NavMenu.tsx
OzzieOrca: Ok... I didn't notice the `backgroundBlendMode` before. Does that not work with ` backgroundColor: theme.palette.cruGrayDark.main`?

CSimbulan: I don't think theres a way to blend two different backgroundcolors because one will just overwrite the other, but it does work if you have a background image(solid gradient) and a background color

image

OzzieOrca: Alrighty 👍

OzzieOrca (approved):

PR #191, opened 2021-09-20
Fixed error for missing accountListId

CSimbulan:
Moved accountListId retrieval up on level to prevent errors, there are PRs out for some of the other pages so I'll just do those individually

### DIFF
--- a/pages/accountLists/[accountListId]/tools/appeals.page.tsx
+++ b/pages/accountLists/[accountListId]/tools/appeals.page.tsx
@@ -10,9 +10,11 @@ import {
   Divider,
 } from '@material-ui/core';
 import { motion } from 'framer-motion';
+import { useAccountListId } from '../../../../src/hooks/useAccountListId';
 import Appeals from '../../../../src/components/Tool/Appeal/Appeals';
 
 import AddAppealForm from '../../../../src/components/Tool/Appeal/AddAppealForm';
+import Loading from 'src/components/Loading';
 
 const useStyles = makeStyles((theme: Theme) => ({
   container: {
@@ -43,6 +45,7 @@ const useStyles = makeStyles((theme: Theme) => ({
 const AppealsPage: React.FC = () => {
   const { t } = useTranslation();
   const classes = useStyles();
+  const accountListId = useAccountListId();
 
   const variants = {
     animate: {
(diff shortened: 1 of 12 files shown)


src/components/Tool/FixCommitmentInfo/FixCommitmentInfo.tsx
```diff
@@ -9,12 +9,12 @@ import {
   CircularProgress,
 } from '@material-ui/core';
 import { Trans, useTranslation } from 'react-i18next';
-import { useAccountListId } from '../../../../src/hooks/useAccountListId';
 import { useGetInvalidStatusesQuery } from './GetInvalidStatuses.generated';
 import Contact from './Contact';
 import NoContacts from './NoContacts';
 import { contactTags } from './InputOptions/ContactTags';
 import { frequencies } from './InputOptions/Frequencies';
+import theme from 'src/theme';
```
TheNoodleMoose, line 17: Hmm, this import doesn't look right..

src/components/Tool/Appeal/Appeals.tsx
```diff
@@ -8,19 +8,21 @@ import {
...
-  const accountListId = useAccountListId();
+interface Props {
+  accountListId: string;
+}
+
+const Appeals: React.FC<Props> = ({ accountListId }: Props) => {
   const { t } = useTranslation();
   const { data, loading } = useGetAppealsQuery({
-    variables: { accountListId: accountListId || '' },
+    variables: { accountListId: accountListId },
```
TheNoodleMoose, line 25: I think for all these cases you can just do: 
```suggestion
    variables: { accountListId },
```

pages/accountLists/[accountListId]/tools/appeals.page.tsx
```diff
@@ -10,9 +10,11 @@ import {
   Divider,
 } from '@material-ui/core';
 import { motion } from 'framer-motion';
+import { useAccountListId } from '../../../../src/hooks/useAccountListId';
 import Appeals from '../../../../src/components/Tool/Appeal/Appeals';
 
 import AddAppealForm from '../../../../src/components/Tool/Appeal/AddAppealForm';
+import Loading from 'src/components/Loading';
```
Comment (TheNoodleMoose) [inline pages/accountLists/[accountListId]/tools/appeals.page.tsx:17]: Import also looks not right.

TheNoodleMoose (changes requested):
Looks pretty good, just a few things. Yeah, I think something like this is needed. Ideally ```accountListId``` would always be defined but there are instances where that could not be the case as soon as the page loads.

src/components/Tool/FixCommitmentInfo/FixCommitmentInfo.tsx
TheNoodleMoose: I think I've had this happen before. Something that happened after I updated vscode, changed my default import settings.

PR #196, opened 2021-09-28
MPDX-700-Handle-Stale-Sessions

TheNoodleMoose:
After research and trial and error, I've landed on a solution for handling stale sessions/no sessions. The first iteration was to utilize ```getInitialProps``` in ```_app.page.tsx``` and handle everything there. But utilizing ```getInitialProps``` has some drawbacks/caveats: [relevant doc](https://nextjs.org/docs/advanced-features/custom-app#caveats). So after discussing with @OzzieOrca, we landed on having a ```RouterGuard``` component that continually checks the session's state and handles the logic accordingly.


pages/routerGuard.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { useEffect, useState } from 'react';
+import { getSession, signin } from 'next-auth/client';
+import { useRouter } from 'next/router';
+import { DateTime } from 'luxon';
+
+export const RouterGuard: React.FC = ({ children }) => {
+  const [isAuthed, setIsAuthed] = useState(false);
+  const { push, pathname, query, isReady } = useRouter();
+  const getSess = async () => {
```
wrandall22, line 9: Is there a reason not to fully spell out `getSession`?

wrandall22: Nevermind, I see we're pulling in a `getSession` from next-auth

pages/accountLists.page.tsx
```diff
@@ -40,9 +40,12 @@ export const getServerSideProps: GetServerSideProps = async ({
...
   if (!token) {
-    res.writeHead(302, { Location: '/' });
-    res.end();
-    return { props: {} };
+    return {
+      redirect: {
+        destination: '/login',
+        permanent: false,
+      },
+    };
```
TheNoodleMoose, line 48: In the docs, I found a better/updated way to handle redirects. If the session was expired or not present, the ```res.writeHead``` would consistently throw errors.

pages/routerGuard.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+    setIsAuthed(!!session);
+    // If they do not have a valid session and they're not on the login page, navigate them to it
+    if (!session && pathname !== '/login') {
+      push('/login');
+    }
+    // Check if the session has a user
+    if (session?.user) {
+      // If the session is expired, sign them in again to fresh token
+      // TODO eventually will have to accomidate for okta, google signin, etc
```
TheNoodleMoose, line 20: We'll have to have some logic to determine what provider they authed with, and then resign them in accordingly.  Maybe that's provided by ```next-auth``` or maybe we have to store it somewhere. Maybe in the user object? Also heard mention that Google sign-in could be coming to mpdx in the future, but not sure when.

pages/routerGuard.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { useEffect, useState } from 'react';
+import { getSession, signin } from 'next-auth/client';
+import { useRouter } from 'next/router';
+import { DateTime } from 'luxon';
+
+export const RouterGuard: React.FC = ({ children }) => {
+  const [isAuthed, setIsAuthed] = useState(false);
```
hotcakedev628, line 7: Is there any way to get the status `isAuthed` from a hook or an api directly without using `useState`?

TheNoodleMoose: There is a ```useSession``` hook and I attempted a very similar setup. However, it didn't work as expected. ```useSession``` grabs the context from the session ```Provider``` in ```_app.page.tsx```,  while ```getSession``` is making an api call to ```/api/auth/session``` to get the most recent and updated state of the session. I noticed if I deleted my session from cookies for example and then navigated to another page, ```useSession``` would still return that I have a valid session, while ```getSession``` would have the correct and expected value.

OzzieOrca: So essentially you're wanting to update the session on every page navigation? That fires off a network request to the Next.js backend every time. I guess that's ok... Seems like a little overkill maybe. I think the initial app load will send this request twice, once in `SessionProvider` and again immediately here. We could wait until a GraphQL request fails, which probably happens around the same time since most pages fire off a query during render.

[The docs](https://next-auth.js.org/getting-started/client#options) say
>  If you need to, you can trigger an update of the session object across all tabs/windows by calling getSession() from a client side function.

That sounds like you may not need to store anything locally. My assumption is `getSession` should make updates that `useSession` picks up which would rerender the component. Maybe it doesn't work that way but that sounds cleaner.

[`refetchInterval`](https://next-auth.js.org/getting-started/client#options) is also an option. Idk that we have any specific need for it but might be a cleaner way to accomplish session updating.

pages/routerGuard.tsx
```diff
@@ -0,0 +1,34 @@
...
+      }
+    }
+  };
+
+  useEffect(() => {
+    getSess();
+  }, [pathname, query]);
+
+  return (
+    <>{(isReady && isAuthed) || pathname === '/login' ? children : null}</>
```
OzzieOrca, line 32: Do we want a loading indicator here? I don't want the user's screen to flicker though if we aren't here very long.

pages/routerGuard.tsx
```diff
@@ -0,0 +1,34 @@
...
+    setIsAuthed(!!session);
+    // If they do not have a valid session and they're not on the login page, navigate them to it
+    if (!session && pathname !== '/login') {
+      push('/login');
+    }
+    // Check if the session has a user
+    if (session?.user) {
+      // If the session is expired, sign them in again to fresh token
+      // TODO eventually will have to accomidate for okta, google signin, etc
+      if (DateTime.now().toISO() > session?.expires) {
```
OzzieOrca, line 21: `session.user.token` is the API's JWT and it also has a expiration time. Might be worth checking. You probably need something to decode the JWT (use https://jwt.io/ as a debugger if needed). The `next-auth/jwt` package might have a decode function.

OzzieOrca (commented):
Looks pretty good :) Thanks for working on this. Just had a few questions/suggestions.

pages/routerGuard.tsx
TheNoodleMoose: Yeah, I could add a large loading indicator in the center of the page, just so the user knows that something is happening.

OzzieOrca (approved):

PR #197, opened 2021-09-29
Add appeal contact tags

CSimbulan:
Added autocomplete function with multi-tags for the contacts section of adding appeal form. 

image

If the accountlist has no contact tags:
image

Need to implement how it affects the appeal creation


src/components/Tool/Appeal/AddAppealForm.tsx
```diff
@@ -105,7 +154,32 @@ const AddAppealForm = (): ReactElement => {
...
+
+  const contactTagsList = data?.accountList.contactTagList ?? [];
+
+  const selectAllStatuses = (): void => {
+    setFilterTags((prevState) => ({
+      ...prevState,
+      statuses: contactStatusesFormatted.slice(
+        0,
+        contactStatusesFormatted.length - 2,
+      ),
```
OzzieOrca, line 169: What is this doing? Removing the last status? Why? I'm just don't have the context here.

src/components/Tool/Appeal/AddAppealForm.tsx
```diff
@@ -281,55 +355,92 @@ const AddAppealForm = (): ReactElement => {
...
+                    id="tags-standard"
+                    options={contactStatusesFormatted.filter(
+                      ({ value: id1 }) =>
+                        !filterTags.statuses.some(
+                          ({ value: id2 }) => id2 === id1,
+                        ),
+                    )}
+                    getOptionLabel={(option) => option.title}
+                    value={filterTags.statuses}
+                    onChange={({}, values) => handleChange(values, 'statuses')}
```
OzzieOrca, line 373: You just throwing the event away? I think the underscore prefix is a better syntax. ESLint shouldn't complain about unused function params that start with an underscore.
```suggestion
                    onChange={(_event, values) => handleChange(values, 'statuses')}
```

src/components/Tool/FixCommitmentInfo/Contact.tsx
```diff
@@ -156,17 +156,17 @@ const Contact: React.FC<Props> = ({
                   value={values.tagValue}
                   onChange={(event) => handleChange(event, 'tagValue')}
                 >
-                  {Object.entries(contactTags).map(
-                    ([statusValue, statusTranslated]) => (
+                  {Object.entries(contactTags)
+                    .slice(0, Object.entries(contactTags).length - 3)
```
OzzieOrca, line 160: This seems like another magic number. Why are you removing a certain number of them?

OzzieOrca (changes requested):

src/components/Tool/Appeal/AddAppealForm.tsx
CSimbulan: Oh yeah the "Select All" option doesn't actually select all since I think there are contradicting options ("all active" and "all hidden"). I moved them to the end to make it easier and just cut them off.

src/components/Tool/FixCommitmentInfo/Contact.tsx
CSimbulan: In the original app the statuses "all hidden, all active, and none" aren't options for this page but are in the adding appeal form. Does it makes more sense to have a separate list for both these pages?

src/components/Tool/Appeal/AddAppealForm.tsx
```diff
@@ -105,7 +154,32 @@ const AddAppealForm = (): ReactElement => {
     name: yup.string().required('Please enter a name'),
   });
 
-  return (
+  const contactStatusesFormatted = Object.entries(
+    contactTags,
+  ).map(([tag, translatedTag]) => ({ value: tag, title: translatedTag }));
```
OzzieOrca, line 159: I think these can be loaded from the API. Since they are loaded from the API in the current app, we should probably keep doing that. (Also they are statuses, not tags 😄 your variable names are misleading).

You could reuse [ContactFilters.graphql](https://github.com/CruGlobal/mpdx-react/blob/e4ac3bbb40b0c0e27d8918dc3ce9df4be55888d2/src/components/Contacts/ContactFilters/ContactFilters.graphql) or make a new one that specified fewer fields. There's currently no way to only ask for certain filters. Maybe we could get that added but it's probably not too much overfetching to just use it as is. Here's my stab at a slightly simpler one but it's not that many fewer fields...
Screen Shot 2021-10-01 at 4 49 13 PM

src/components/Tool/Appeal/AddAppealForm.tsx
OzzieOrca: Even before mentioning loading them from the API, I was going to suggest filtering them out by id/name. I think it's confusing to read the code and error prone to rely on indexes and sort order if future changes are made to the list.

src/components/Tool/FixCommitmentInfo/Contact.tsx
OzzieOrca: Gotcha. Again, I think filtering them out by id/name and loading them from the API would be better.

OzzieOrca (changes requested):

CSimbulan:
I'll change the ones related to the fixcommitinfo contact component in the other PR

src/components/Tool/Appeal/AddAppealForm.tsx
```diff
@@ -97,13 +98,31 @@ const AddAppealForm = (): ReactElement => {
...
+  } = useContactFiltersQuery({
+    variables: {
+      accountListId,
+    },
+  });
+
+  const contactStatuses = contactFilterGroups?.accountList.contactFilterGroups
+    ? (contactFilterGroups.accountList.contactFilterGroups.filter(
+        (group) => group.name === 'Status',
+      )[0].filters[0] as MultiselectFilter).options
```
OzzieOrca, line 119: If you only want one, you can use `find`:
```suggestion
    ? (contactFilterGroups.accountList.contactFilterGroups.find(
        (group) => group.name === 'Status',
      ).filters[0] as MultiselectFilter).options
```
Might be good to find the status filter by `filterKey === 'status'` instead of index, just in case the API response changes in the future.

OzzieOrca (changes requested):
Looks better, thanks :)

OzzieOrca (approved):

PR #198, opened 2021-09-29
moved accountlistid up a level from Navbar to prevent error

CSimbulan:
Fixed an error where on page load there would be a brief moment where accountListId = '' causing an error in the navbar

### DIFF
--- a/src/components/Layouts/Primary/TopBar/TopBar.tsx
+++ b/src/components/Layouts/Primary/TopBar/TopBar.tsx
@@ -59,9 +59,11 @@ const TopBar = ({
               <img src={logo} alt="logo" style={{ cursor: 'pointer' }} />
             </NextLink>
           </Hidden>
-          <Box ml={10} flexGrow={1}>
-            <NavMenu />
-          </Box>
+          {accountListId && (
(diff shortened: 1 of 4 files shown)


src/components/Layouts/Primary/TopBar/TopBar.tsx
```diff
@@ -59,9 +59,11 @@ const TopBar = ({
               <img src={logo} alt="logo" style={{ cursor: 'pointer' }} />
             </NextLink>
           </Hidden>
-          <Box ml={10} flexGrow={1}>
-            <NavMenu />
-          </Box>
+          {accountListId && (
+            <Box ml={10} flexGrow={1}>
+              <NavMenu accountListId={accountListId} />
```
OzzieOrca, line 64: I see that you're waiting to display the menu until an accountListId is loaded. I think you could still use the `useAccountListId` hook in the child component. It'd be one less prop. 🤷‍♂️ Guess it's fine either way.
```suggestion
              <NavMenu />
```

OzzieOrca (approved):

OzzieOrca (approved):

PR #200, opened 2021-09-29
MPDX-7156 Coaching Detail Container

gyasistory:


OzzieOrca (approved):

OzzieOrca:
The coverage seems good enough to me. Feel free to just merge it (or let me know if you don't have permission to).

The one line I see that isn't covered is https://app.codecov.io/gh/CruGlobal/mpdx-react/compare/200/diff#D2L138 which is displaying something different if there are no primary appeals.

PR #202, opened 2021-09-30
Tools notificattion limit

CSimbulan:
Added limit to the notifications to 9. If there is more than 9 it will show 9+ (like in the original app)


OzzieOrca (approved):
👍

PR #203, opened 2021-10-01
Mpdx 7104 change primary appeal

CSimbulan:
Added ability to change your primary appeal

### DIFF
--- a/src/components/Tool/Appeal/Appeals.tsx
+++ b/src/components/Tool/Appeal/Appeals.tsx
@@ -85,6 +85,7 @@ const Appeals: React.FC<Props> = ({ accountListId }: Props) => {
             0
           }
           total={data.primaryAppeal.nodes[0].pledgesAmountTotal || 0}
+          changePrimary={changePrimary}
         />
       ) : (
         <NoAppeals primary />
--- a/src/components/Tool/Appeal/Appeals.tsx
+++ b/src/components/Tool/Appeal/Appeals.tsx
@@ -52,8 +52,6 @@ const Appeals: React.FC<Props> = ({ accountListId }: Props) => {
         { query: GetAppealsDocument, variables: { accountListId } },
       ],
     });
-    console.log(`New Primary ID: ${newPrimaryId}`);
-    console.log(data?.primaryAppeal.nodes[0].id);
     enqueueSnackbar(t('Primary Appeal Updated'), {
       variant: 'success',
     });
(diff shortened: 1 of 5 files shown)


src/components/Tool/Appeal/ChangePrimaryAppeal.graphql
```diff
@@ -0,0 +1,20 @@
...
+      id
+    }
+  }
+  unsetPrimary: updateAppeal(
+    input: { accountListId: $accountListId, attributes: $attributes2 }
+  ) {
+    appeal {
+      id
+    }
+  }
```
OzzieOrca, line 19: The API seems to automatically take care of this. I tested it with
```graphql
query LoadAppeals {
  appeals(accountListId: "0587c86d-3b8a-4c42-8394-8f63007fdd1e") {
    nodes {
      ...AppealAttributes
    }
  }
}

mutation TestUpdateAppeal {
  updateAppeal(
    input: {
      accountListId: "0587c86d-3b8a-4c42-8394-8f63007fdd1e"
      attributes: { id: "dbc4b0c7-eb8a-44fd-bf06-eb8c689e1c53", primary: true }
    }
  ) {
    appeal {
      ...AppealAttributes
    }
  }
}

fragment AppealAttributes on Appeal {
  id
  name
  primary
}
```

src/components/Tool/Appeal/Appeals.tsx
```diff
@@ -21,17 +26,44 @@ interface Props {
...
+  const { enqueueSnackbar } = useSnackbar();
+  const [
+    changePrimaryAppeal,
+    { loading: updating },
+  ] = useChangePrimaryAppealMutation();
   const { data, loading } = useGetAppealsQuery({
     variables: { accountListId },
   });
 
+  const changePrimary = async (newPrimaryId: string): Promise<void> => {
```
OzzieOrca, line 38:
```suggestion
  const changePrimary = async (newPrimaryId: string): void => {
```
If TS wants it to be a promise, that's fine. Just seems like an unnecessary detail since the return value  is never used.

OzzieOrca (changes requested):

src/components/Tool/Appeal/Appeals.tsx
CSimbulan: Yeah TS gives error:

The return type of an async function or method must be the global Promise<T> type. Did you mean to write 'Promise<void>'?ts(1064)

but changing it in the props to void is fine

OzzieOrca (approved):

PR #204, opened 2021-10-01
Fix commitment info mutation

CSimbulan:
Mutation for updating the contact status, statusValid, pledgeAmount, pledgeCurrency, and pledgeFrequency.


CSimbulan:
I left the frequencies the same because the ones returned from the API are different to how they're used here:

image

The value is a float but for the commitment info they are saved like "EVERY_6_MONTHS" or "EVERY_2_WEEKS" so unless there is something in the API that will auto convert it I will have to map them anyways.

PR #207, opened 2021-10-01
Add config for Codespaces

OzzieOrca:
- Update suggested extension list
- Update vscode-jest config after deprecation
- Enable vscode native bracketPairColorization

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,40 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.192.0/containers/javascript-node
+{
+  "name": "Node.js",
+  "build": {
+    "dockerfile": "Dockerfile",
+    // Update 'VARIANT' to pick a Node version: 12, 14, 16
+    "args": { "VARIANT": "16" }
+  },
+
+  // Set *default* container specific settings.json values on container create.
+  "settings": {},
+
+  // Add the IDs of extensions you want installed when the container is created.
+  "extensions": [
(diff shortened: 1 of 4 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,40 @@
...
+    "streetsidesoftware.code-spell-checker",
+    "waderyan.nodejs-extension-pack",
+    "ziyasal.vscode-open-in-github"
+  ],
+
+  // Use 'forwardPorts' to make a list of ports inside the container available locally.
+  "forwardPorts": [3000],
+
+  // Use 'postCreateCommand' to run commands after the container is created.
+  "postCreateCommand": "yarn install && yarn gql && yarn gql:server",
```
TheNoodleMoose, line 36: ```&&``` runs these sequentially, while ```&``` runs them in parallel right? The only reason I ask is that I believe ```yarn gql:server``` has to come before ```yarn gql``` because the server codegen adds it to the main ```schema.graphql```, and then gets generated via ```yarn gql```.

OzzieOrca: Huh ya we can probably make this faster by running them in parallel. I've usually run it in this order and it's been fine... But now you've got me questioning things... CI runs in this order... I think it works cuz the client codegen just looks at the individual `.graphql` files in the API directory. 🤷‍♂️ 

I have some issue that is crashing the container a minute or so after it starts. Idk if it's too many extensions or the yarn install or what. I've been meaning to get back to debugging it.

TheNoodleMoose: Ah, interesting. When I went through the process of figuring out the rest -> graphql modularization I discovered I had to run it in that order or else the types wouldn't get completely generated.

PR #209, opened 2021-10-06
Mpdx 7160 contact flows view

CSimbulan:
Flows view for contact page, just currently taking the flows settings and showing the correct contacts for them. 

image
image


OzzieOrca (approved):

OzzieOrca:
Looks like most of the [missing coverage](https://app.codecov.io/gh/CruGlobal/mpdx-react/compare/209/diff) is with pagination and drag and drop. Do you think those would be worth testing? I've never tested a drag interaction before... If it seems difficult, I'm open to just merging it soon.

CSimbulan:
> 
> 
> Looks like most of the [missing coverage](https://app.codecov.io/gh/CruGlobal/mpdx-react/compare/209/diff) is with pagination and drag and drop. Do you think those would be worth testing? I've never tested a drag interaction before... If it seems difficult, I'm open to just merging it soon.

I've been trying to do it but its not working so far. I have the drop zone as its own so component so I was trying to test just the component but due to the way its structured I couldn't find any good examples and other people were having trouble testing with the dnd hooks. So I think testing the lines from a parent component would be easier, but if you're okay with force merging it for now we can come back to it later

PR #217, opened 2021-10-13
Add active contact sort to dashboard this week queries

OzzieOrca:
Checks are failing cuz https://github.com/CruGlobal/mpdx_api/pull/2444 isn't deployed to production yet.


dbenton9:
I'm working on getting https://github.com/CruGlobal/mpdx_api/pull/2444 into production and then once we see that clear we can follow up with this 👍🏻

PR #227, opened 2021-10-18
Mpdx 7139 contacts/tasks null state

CSimbulan:
image


pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx
```diff
@@ -74,6 +75,10 @@ const ContactsPage: React.FC = () => {
     skip: !accountListId,
   });
 
+  const isFiltered =
+    Object.keys(activeFilters).length > 0 ||
+    Object.values(activeFilters).filter((filter) => filter !== []).length > 0;
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/contacts/[[...contactId]].page.tsx:80]:
```suggestion
    Object.values(activeFilters).some((filter) => filter !== []);
```

pages/accountLists/[accountListId]/tasks/[[...contactId]].page.tsx
```diff
@@ -88,6 +82,11 @@ const TasksPage: React.FC = () => {
     skip: !accountListId,
   });
 
+  const isFiltered =
+    Object.keys(activeFilters).length > 0 ||
+    Object.values(activeFilters).filter((filter) => filter !== []).length > 0;
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/tasks/[[...contactId]].page.tsx:87]:
```suggestion
    Object.values(activeFilters).some((filter) => filter !== []);
```

pages/accountLists/[accountListId]/tasks/[[...contactId]].page.tsx
```diff
@@ -88,6 +82,11 @@ const TasksPage: React.FC = () => {
...
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Comment (OzzieOrca) [inline pages/accountLists/[accountListId]/tasks/[[...contactId]].page.tsx:89]: This seems to be used. You still having errors without this?
```suggestion
```

src/components/Shared/Filters/NullState/NullState.tsx
```diff
@@ -0,0 +1,115 @@
...
+    <NullStateBox data-testid={`${page}-null-state`}>
+      <Icon
+        path={page === 'contact' ? mdiHome : mdiFormatListBulleted}
+        size={1.5}
+      />
+      {filtered ? (
+        <>
+          <Typography variant="h5">
+            <Trans
+              defaults="You have {{count}} total {{page}}s"
```
OzzieOrca, line 44: This isn't a very robust way to translate as plurals probably don't work this way in every language... Maybe a simple way would be to pass in a plural prop too.

OzzieOrca (changes requested):

src/components/Shared/Filters/NullState/NullState.tsx
CSimbulan: I was reading the doc for react-i18n and it seems for plurals they normally use the _plural extension on the key in the translation json, but since one of the plurals is a variable itself, it doesn't seem to work and they don't provide an example on this case so I'm not sure if its possible. The way I just changed it to is passing a page or page+s when needed which should be fine since its either contact/contacts or task/tasks. One way to get the key thing work would be to have separate keys for contacts and tasks so that we don't need the page variable.

OzzieOrca: I think your new solution isn't as good as your old version, although it didn't support the singular case. Before a translator at least had control over the string. Now it's part of the app's logic. I think I'm ok with merging your previous version and fixing it later.

Ideally I think it would be best to translate the word as a whole instead of tacking on characters. I guess you'd have to do that where you pass in the `page` prop. Idk how to do that with plurals other than passing in 2 props.

Guess we still have some translation stuff to figure out...

CSimbulan: Alright sure, by passing another prop you mean like passing in both "contact" and "contacts"?

OzzieOrca: Ya I guess so. That doesn't seem very clean... 🤷‍♂️

OzzieOrca (approved):

OzzieOrca (approved):

CSimbulan:
So I had to go on the codecov site to see whats not getting covered and I think its coming from something unrelated but its getting picked up because of the merge

PR #248, opened 2021-11-08
fixed crashing bug when data is undefined

CSimbulan:
Just a small fix for a crashing but when filters/mass selection is used


TheNoodleMoose (approved):
Thanks 🙂

PR #895, opened 2024-03-21
[no-Jira] Reduce initial client bundle size

canac:
## Description

**Main improvements:**
* Decrease bundle sizes by lazy-loading modals and panels that aren't visible until being clicked.
* Decrease bundle sizes by optimizing imports, notably lodash.
* Combine redundant Contacts queries in the contacts list.

## Checklist:


src/components/Reports/EmptyReport/EmptyReport.tsx
```diff
@@ -59,26 +59,26 @@ export const EmptyReport: React.FC<Props> = ({
...
+            onMouseEnter={preloadAddDonation}
             onClick={() => setAddDonationOpen(true)}
           >
             {t('Add New Donation')}
           </Button>
         )}
       </Box>
 
-      <DynamicModal
+      <Modal
```
canac, line 70: @dr-bizz The `AddDonation` component is the heaviest and the more important one to lazy load. And like we discussed earlier, lazy loading `Modal` here does nothing because it's used directly in so many places that it's already part of the bundle. I also think it's a better UX to lazy load the modal body but not the title, so that the user can at least see the title of the modal while the body loads.

Anyway, this commit reverts the `Modal` lazy loading you added since that change doesn't do anything, but please push back if you disagree.

dr-bizz: I agree, Lazy loading the body is a good way to go for now.

In another PR, we should make the Dynamic Modal component have a property of `title` so that when it renders, it shows the user the Modal title while loading the Component. There might be some layout shift.

canac: > In another PR, we should make the Dynamic Modal component have a property of `title` so that when it renders, it shows the user the Modal title while loading the Component. There might be some layout shift.

We can definitely look into that. Although it seems like we'd just be reimplementing `Modal` at that point. And `Modal` isn't a very heavy component. Removing it from the initial bundle only shaves off 1-2kb when I tried it. Also we'll create a network request waterfall if we lazy load `Modal` and its content.

canac:
@dr-bizz I tested everything again in staging, and it's all working correctly as far as I can tell.

dr-bizz (approved):
@canac I've tested this on staging and it looks awesome! Pages and components are loading super fast. This work will greatly increase productivity for users.

PR #900, opened 2024-03-26
MPDX-7814 - Allow users to open contact in new tab

dr-bizz:
## Description
https://jira.cru.org/secure/RapidBoard.jspa?rapidView=3&view=detail&selectedIssue=MPDX-7814#

Added a custom right-click menu, which for now only has the menu item `Open contact in new tab`.
But we could easily add quick actions to the right-click menu, making users make fewer clicks. One idea I had was for an open to allow the user to edit the contact details. which would open the contact details modal.
If users like the right-click, we could even use it on Tasks.

I also corrected some types in ContactContext

## Checklist:


dr-bizz:
I have pushed this code live in another PR https://github.com/CruGlobal/mpdx-react/pull/1090

PR #908, opened 2024-04-08
No-Jira - Implement LighthouseCI

wrandall22:
## Description

## Checklist:


.github/workflows/ci.yml
```diff
@@ -146,3 +146,17 @@ jobs:
...
           # This secret is automatically injected by GitHub
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: 🚨 Run Lighthouse
+        run: yarn lighthouse
+        env:
+          secrets: '{"JWT_SECRET":"JWT_SECRET","OKTA_CLIENT_SECRET":"OKTA_CLIENT_SECRET"}'
+          LHCI_GITHUB_APP_TOKEN: ${{ secrets.LHCI_GITHUB_APP_TOKEN }}
+          MPDX_USERNAME: ${{ secrets.MPDX_USERNAME }}
+          MPDX_PASSWORD: ${{ secrets.MPDX_PASSWORD }}
+          PREVIEW_URL: 'https://pr-${{ github.event.number }}.d3dytjb8adxkk5.amplifyapp.com'
```
canac, line 156: As I'm sure you know, this will need updated to the new preview urls once your PR merges. The following might work, but it will fail if the PR doesn't have the "Preview Environment" label. Maybe we skip the lighthouse step if there isn't a preview environment?

```suggestion
          PREVIEW_URL: 'https://${{ env.GITHUB_HEAD_REF }}.d3dytjb8adxkk5.amplifyapp.com'
```

.prettierignore
```diff
@@ -21,3 +21,6 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+# Lighthouse
+.lighthouseci
```
canac, line 26: I think this line isn't necessary because `prettier` claims to ignore files in `.gitignore` by default https://prettier.io/docs/en/ignore.html#:~:text=By%20default%20prettier,it%20is%20run.

lighthouse/lighthouse-auth.mjs
```diff
@@ -0,0 +1,150 @@
...
+      console.log(' Did not find signInButton', error);
+      return;
+    }
+  }
+}
+
+export async function loadInitialPage(page, origin, timeout = 60_000) {
+  let originResponse = await page.goto(origin);
+
+  return await new Promise((resolve) => {
```
canac, line 60: I think this will wait `timeout` even if the page succeeds the first time. Maybe you could add something like this:

```ts
if (originResponse.status && originResponse.status !== 404) {
  resolve(originResponse);
  return;
}
```

lighthouse/lighthouse-auth.test.ts
```diff
@@ -0,0 +1,274 @@
+import fs from 'fs';
+import { loadInitialPage, main } from './lighthouse-auth.mjs';
+
+describe('Lighthouse Pre-Login Steps', () => {
+  const OLD_ENV = process.env;
+  const previewUrl = 'https://pr-908.d3dytjb8adxkk5.amplifyapp.com';
```
canac, line 6: The tests will keep working once this preview environment is gone, right? This could be any URL since you mock `initialPage.goto`, right?

canac (commented):
Thanks for reorganizing the code and writing all of those tests! It's really clean and nice now.

.github/workflows/ci.yml
wrandall22: Daniel suggested we make this run on comment since it takes so long, but we can certainly ensure it can't run if the preview label isn't applied as well.

canac: I'm good with that or however you want to handle it.

wrandall22:
run lighthouse

wrandall22:
run lighthouse

wrandall22:
Comment-driven won't work until this is merged to main.

lighthouse/lighthouse-auth.mjs
```diff
@@ -0,0 +1,154 @@
...
+
+export async function loadInitialPage(page, origin, timeout = 60_000) {
+  let originResponse = await page.goto(origin);
+
+  if (originResponse.status && originResponse.status !== 404) {
+    return Promise.resolve(originResponse);
+  }
+
+  return await new Promise((resolve) => {
+    const initialLoadId = setInterval(async () => {
```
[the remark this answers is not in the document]
wrandall22: My understanding is that, if the first load does not succeed, it will then wait 60 seconds, check the status again (which won't have changed) and then reload the page, and keep doing that loop until it finds it. Perhaps we should introduce a limit of retries.

lighthouse/lighthouse-auth.mjs
```diff
@@ -57,22 +57,30 @@ async function login(page, origin) {
 export async function loadInitialPage(page, origin, timeout = 60_000) {
   let originResponse = await page.goto(origin);
 
-  if (originResponse.status && originResponse.status !== 404) {
+  if (pageFound(originResponse)) {
     return Promise.resolve(originResponse);
   }
 
   return await new Promise((resolve) => {
```
canac, line 64: You don't need to `await` returned promises.

```suggestion
  return new Promise((resolve) => {
```

lighthouse/lighthouse-auth.mjs
```diff
@@ -57,22 +57,30 @@ async function login(page, origin) {
...
     const initialLoadId = setInterval(async () => {
-      if (originResponse.status && originResponse.status !== 404) {
+      if (pageFound(originResponse)) {
         resolve(originResponse);
         clearInterval(initialLoadId);
       } else {
```
canac, line 69: This branch isn't needed anymore, right?

lighthouse/lighthouse-auth.mjs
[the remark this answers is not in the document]
canac: I think it could be simplified a little, but the logic looks great now!

lighthouse/lighthouse-auth.mjs
wrandall22: It is needed because otherwise it never re-queries the page.

lighthouse/lighthouse-auth.mjs
```diff
@@ -0,0 +1,162 @@
...
+        }
+      }
+    } catch (error) {
+      console.log(' Did not find signInButton', error);
+      return;
+    }
+  }
+}
+
+export async function loadInitialPage(page, origin, timeout = 60_000) {
```
canac, line 57: Uggh, this waiting approach of waiting on the status code to not be 404 probably won't work. I just made a PR with [a preview environment](https://menu-component-error.d3dytjb8adxkk5.amplifyapp.com), and while the first build is building, the response to any page is 200 with the following body:

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="utf-8">
    <title>Welcome</title>
    <style>
        body {
            background-color: #d5dbdb;
            margin: 5em 5em 5em 5em;
            font-family: "Amazon Ember", Helvetica, Arial, sans-serif;
        }

        div {
            background-color: #ffffff;
            padding: 1em 2em 2em 2em;
        }

        h1 {
            color: #16191f;
        }

        h3 {
            color: #16191f;
        }
    </style>
</head>
<body>
<div>
    <h1>Welcome</h1>
    <h3>Your app will appear here once you complete your first deployment.</h3>
    <p>
        Deployment didn't work? Here are some options:
        <ul>
            <li>Check out <b><a href="https://docs.aws.amazon.com/amplify/latest/userguide/welcome.html">our docs</a></b></li>
            <li>Click the <b>Feedback</b> button in the bottom-left corner of the service page</li>
        </ul>
    </p>
    <p>
        Quick tips:
        <ul>
            <li>Have you checked your build settings? The <b>baseDirectory</b> parameter in the 
                <b>artifacts</b> step of your YAML file should match your build output directory</li>
            <li>Building your app should produce an <b>index.html</b> file. Try building your app 
                locally and check a file with that name exists in the artifacts base directory.</li>
        </ul>
        </p>
</div>
</body>
</html>
```

We may have to make this method try again if the page content contains `Your app will appear here once you complete your first deployment.` and hope Amplify doesn't change that placeholder page's content.

wrandall22: Or we could cycle wait on something on all of our pages, if such a thing exists.

wrandall22: Looks like I could look for `<div id="__next">` which seems to be consistent.

wrandall22: I've made a commit to this effect, let me know what you think.

lighthouse/lighthouse-auth.mjs
```diff
@@ -54,31 +58,15 @@ async function login(page, origin) {
   }
 }
 
-export async function loadInitialPage(page, origin, timeout = 60_000) {
-  let originResponse = await page.goto(origin);
-
-  if (pageFound(originResponse)) {
-    return Promise.resolve(originResponse);
+export async function checkForInitialLoad(page, tries = 0, timeout = 60_000) {
```
canac, line 61: Should we switch these args since `timeout` is something that might be customized (like in tests), but we always want `tries` to start at 0?

lighthouse/lighthouse-auth.test.ts
```diff
@@ -197,33 +193,62 @@ describe('Lighthouse Pre-Login Steps', () => {
...
+            reject(new Error('Timed out waiting to find element'));
+          }, timeout);
         });
+      };
+
+      const maxTries = 14;
+      const timeout = 100;
+      for (let i = 0; i < maxTries; i++) {
+        initialPage.waitForSelector.mockImplementationOnce(
+          async () => await promiseWithTimeout(timeout),
```
canac, line 237:
```suggestion
          () => promiseWithTimeout(timeout),
```

lighthouse/lighthouse-auth.test.ts
```diff
@@ -197,33 +193,62 @@ describe('Lighthouse Pre-Login Steps', () => {
...
+        );
+      }
 
-      await loadInitialPage(initialPage, `${previewUrl}/accountLists`, 500);
-      expect(initialPage.goto).toHaveBeenCalledTimes(2);
+      initialPage.waitForSelector.mockImplementationOnce(() => {
+        return Promise.resolve({});
+      });
```
canac, line 243: This and the one on line 214 could be simplified to use `mockResolvedValueOnce` unless you want to keep it as-is to match the others.

```suggestion
      initialPage.waitForSelector.mockResolvedValueOnce({});
```

lighthouse/lighthouse-auth.test.ts
```diff
@@ -197,33 +193,62 @@ describe('Lighthouse Pre-Login Steps', () => {
...
+
+      try {
+        await checkForInitialLoad(initialPage, 0, timeout);
+        throw new Error('Should have thrown an error.');
+      } catch (error) {
+        /* eslint-disable jest/no-conditional-expect */
+        expect(initialPage.waitForSelector).toHaveBeenCalledTimes(maxTries);
+      }
```
canac, line 251: I believe this could use `await expect(() => checkForInitialLoad(initialPage, 0, timeout)).rejects.toEqual(...)`

canac (commented):
I like the idea of using the built-in timeout to wait before trying again!

PR #917, opened 2024-04-19
MPDX-7878 - Dynamically loading field array on Add Multiple Contacts Modal

dr-bizz:
## Description
The FieldArray on the Add Multiple Contacts Modal was taking at least 3 seconds to load. This fix shows the modal straight away and then loads in the FieldArray

## Checklist:

### DIFF
--- a/src/components/Layouts/Primary/TopBar/Items/AddMenu/Items/CreateMultipleContacts/DynamicCreateMultipleContactsFieldArray.tsx
+++ b/src/components/Layouts/Primary/TopBar/Items/AddMenu/Items/CreateMultipleContacts/DynamicCreateMultipleContactsFieldArray.tsx
@@ -1,0 +1,10 @@
+import dynamic from 'next/dynamic';
+import { DynamicComponentPlaceholder } from 'src/components/DynamicPlaceholders/DynamicComponentPlaceholder';
+
+export const DynamicContactsRightPanel = dynamic(
+  () =>
+    import(
+      /* webpackChunkName: "ContactsRightPanel" */ './CreateMultipleContactsFieldArray'
+    ),
+  { loading: DynamicComponentPlaceholder },
+);
--- a/src/components/Layouts/Primary/TopBar/Items/AddMenu/Items/CreateMultipleContacts/CreateMultipleContacts.tsx
+++ b/src/components/Layouts/Primary/TopBar/Items/AddMenu/Items/CreateMultipleContacts/CreateMultipleContacts.tsx
@@ -9,20 +9,16 @@ import {
   TableContainer,
   TableHead,
   TableRow,
-  TextField,
 } from '@mui/material';
 import { styled } from '@mui/material/styles';
-import { Field, FieldArray, FieldProps, Form, Formik } from 'formik';
+import { Form, Formik } from 'formik';
 import { useSnackbar } from 'notistack';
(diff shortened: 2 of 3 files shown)


src/components/Layouts/Primary/TopBar/Items/AddMenu/Items/CreateMultipleContacts/DynamicCreateMultipleContactsFieldArray.tsx
```diff
@@ -0,0 +1,10 @@
+import dynamic from 'next/dynamic';
+import { DynamicComponentPlaceholder } from 'src/components/DynamicPlaceholders/DynamicComponentPlaceholder';
+
+export const DynamicContactsRightPanel = dynamic(
+  () =>
+    import(
+      /* webpackChunkName: "ContactsRightPanel" */ './CreateMultipleContactsFieldArray'
+    ),
+  { loading: DynamicComponentPlaceholder },
+);
```
canac, line 10:
```suggestion
export const DynamicCreateMultipleContactsFieldArray = dynamic(
  () =>
    import(
      /* webpackChunkName: "CreateMultipleContactsFieldArray" */ './CreateMultipleContactsFieldArray'
    ),
  { loading: DynamicComponentPlaceholder },
);
```

canac:
One caveat is that this changes introduces a waterfall. Instead of loading all the scripts for the modal in one request, it now waits for the `CreateMultipleContacts` chunk to load, then waits for the `DynamicCreateMultipleContactsFieldArray` chunk to load, and only then renders the modal content. Unless I'm mistaken, that will mean it will take even longer for the form to load and be usable than it did before.

You mentioned it taking >3 seconds to load. For me in staging, it takes <200ms. Do you know what might be causing a discrepancy that significant? 
Screenshot 2024-04-22 at 1 11 15 PM

I do notice the modals hanging for a second on localhost but not in staging. Is that the 3 seconds you're referring to?

dr-bizz:
Yes, I tested it locally, as I didn't know you merged it into staging. That is where the 3 seconds came from. I do not see it taking longer with my fix, it takes about the same time.

I've merged it with staging, and it's loading a few milliseconds faster (Or appearing to due to the modal rendering straight away).

On your network tab, throttle your connection to `fast 3G` or `slow 3G`. This is where you will see the biggest difference with my fix since it shows the modal and a spinning loading icon while the `FieldArray` loads.

canac (approved):
I tested this a couple times with preloading disabled. Here's the recording. With slow 3G throttling on localhost, the old version (i.e. the `lazy-components` branch) takes 12.66s to render the modal content. The new version (your branch) shows the field headers in about 7s but takes 14.83s to render the rest of the modal content (~17% longer, probably due to the request waterfall I mentioned earlier). Personally, since they can see the modal title and the close button is functional, Since user can't start typing until the second load completes, I don't think it helps the UX to show the heading row early at the expense of slightly slowing down loading the rest of the content. But my measurements may not be representative, and I trust your judgement if you think this is a UX improvement, so I'm going to approve this.

https://github.com/CruGlobal/mpdx-react/assets/3740187/af9043e7-5b32-4a75-90be-6a321a9492d0

https://github.com/CruGlobal/mpdx-react/assets/3740187/64273a15-ec1b-4783-b726-dc6fa93a98ae

PR #920, opened 2024-04-22
No Jira - Allow react to compare component types on render

dr-bizz:
## Description

Preventing React from destroying the component tree on every render due to component types being recreated.
Updated all calls to Task Modal to use the new Enum I created.

I also fixed some imports.

## Checklist:


src/components/Task/Modal/TaskModal.tsx
```diff
@@ -45,69 +52,63 @@ const TaskModal = ({
 
   const task = data?.task;
 
-  const renderTitle = (): string => {
+  const renderTitle = useMemo((): string => {
```
canac, line 55: If this `useMemo` a performance optimization? Are you sure it makes much of a difference? I'd be surprised if executing a `switch` is more expensive than the overhead of `useMemo`.

src/components/Task/Modal/TaskModal.tsx
```diff
@@ -45,69 +52,63 @@ const TaskModal = ({
...
     }
-  };
+  }, [view]);
 
-  const renderView = (): ReactElement => {
-    switch (view) {
-      case 'complete':
-        if (task) {
-          return (
+  return loading ? (
```
canac, line 70: Are you sure that this isn't equivalent to what we had before? I don't think there's a performance difference between calling the function `renderView` and inlining it like you're doing here. If it were a `<View>` component defined inside of this component, that might have performance implications, but it's not JSX, it's just a regular function call.

canac (commented):
Using an enum instead of a string, is a great idea! Thanks for making that change!

As I noted in the comments, I don't think that the changes to `TaskModal` will improve rendering performance. Although I could definitely be mistaken.

src/components/Task/Modal/TaskModal.tsx
dr-bizz: Adding the useMemo on the `renderTitle` was "while I'm here" kind of work. It probably won't make much of a difference.

src/components/Task/Modal/TaskModal.tsx
dr-bizz: No, I believe it's different in terms of how React sees it. It has the same effect, but React reuses existing component instances and DOM nodes as much as possible during rendering. But to speed up reconciliation:
- It compares old and new component types at a location.
- If prevElement.type !== currElement.type, it assumes the entire subtree will be different.

I might be wrong, but my thinking process was the `renderView` function would be destroyed on every render from `<TaskModal />` and recreated from scratch as React would destroy that entire existing component tree section, including all DOM nodes since it sees the `renderView` function as a different function each time.

It is also better practice to write it as JSX inside a return rather than the above, as it is consistent with code throughout the repo.

canac: > It compares old and new component types at a location. If prevElement.type !== currElement.type, it assumes the entire subtree will be different.

You're completely correct about that. However, for React to consider something a component, it has to have JSX component syntax: i.e. `<Component />`. It would be a performance problem if `Modal` or `DynamicTaskModalCompleteForm` were defined inside of the component and were created inside of the component. `renderView` is redefined every render, but since it's not a component that won't matter. React does its diffing on the return value of the component function, and both versions of `Modal` will return identical output.

Here's an example:

```tsx
// TestComponent.tsx
export const OtherComponent: React.FC = () => <span>Other Component</span>;

export const TestComponent: React.FC = () => {
  const renderTree = () => (
    <div>
      <OtherComponent />
    </div>
  );

  return <span>{renderTree()}</span>;
};

export const TestComponentInlined: React.FC = () => {
  return (
    <span>
      <div>
        <OtherComponent />
      </div>
    </span>
  );
};
```

If you [compile the JSX with esbuild](https://esbuild.github.io/try/#YgAwLjIwLjIAAGUAQ29tcG9uZW50LnRzeABleHBvcnQgY29uc3QgT3RoZXJDb21wb25lbnQ6IFJlYWN0LkZDID0gKCkgPT4gPHNwYW4+T3RoZXIgQ29tcG9uZW50PC9zcGFuPjsKCmV4cG9ydCBjb25zdCBUZXN0Q29tcG9uZW50OiBSZWFjdC5GQyA9ICgpID0+IHsKICBjb25zdCByZW5kZXJUcmVlID0gKCkgPT4gKAogICAgPGRpdj4KICAgICAgPE90aGVyQ29tcG9uZW50IC8+CiAgICA8L2Rpdj4KICApOwoKICByZXR1cm4gPHNwYW4+e3JlbmRlclRyZWUoKX08L3NwYW4+Owp9OwoKZXhwb3J0IGNvbnN0IFRlc3RDb21wb25lbnRJbmxpbmVkOiBSZWFjdC5GQyA9ICgpID0+IHsKICByZXR1cm4gKAogICAgPHNwYW4+CiAgICAgIDxkaXY+CiAgICAgICAgPE90aGVyQ29tcG9uZW50IC8+CiAgICAgIDwvZGl2PgogICAgPC9zcGFuPgogICk7Cn07Cg), you get this output:

```js
export const OtherComponent = () => /* @__PURE__ */ React.createElement("span", null, "Other Component");
export const TestComponent = () => {
  const renderTree = () => /* @__PURE__ */ React.createElement("div", null, /* @__PURE__ */ React.createElement(OtherComponent, null));
  return /* @__PURE__ */ React.createElement("span", null, renderTree());
};
export const TestComponentInlined = () => {
  return /* @__PURE__ */ React.createElement("span", null, /* @__PURE__ */ React.createElement("div", null, /* @__PURE__ */ React.createElement(OtherComponent, null)));
};
```

The two versions will execute the exact same steps to build up the virtual DOM and will return a virtual DOM tree that is indistinguishable from one another.

With all of that said, even if there aren't performance implications, I'm still OK with the rearranging that you've done.

src/components/Task/Modal/TaskModal.tsx
```diff
@@ -45,69 +52,63 @@ const TaskModal = ({
...
+    <Loading loading />
+  ) : (
+    <Modal isOpen={open} title={renderTitle} handleClose={onModalClose}>
+      {accountListId ? (
+        <>
+          {view === TaskModalEnum.Complete && task && (
             <DynamicTaskModalCompleteForm
               accountListId={accountListId || ''}
```
canac, line 78: Now that you're checking `accountListId` on line 74, we can simplify this and the three other occurrences.

```suggestion
              accountListId={accountListId}
```

canac (approved):

src/components/Task/Modal/TaskModal.tsx
canac: It's slower on the first render and faster on subsequent renders as expected, but also we're talking about no more than 10s of microseconds.

The first pair is the first render, subsequent pairs are future renders.

Screenshot 2024-04-25 at 12 33 02 PM

canac, line 55: If you're keeping the memo, can you rename this variable since it's a string now and not a function?

```suggestion
  const title = useMemo((): string => {
```

dr-bizz: How did you get that view of the Memoized vs Not Memoized? That's a great feature

canac: I just wrapped it in `console.time/console.timeEnd`.

```ts
  console.time('Memoized');
  const renderTitle = useMemo((): string => {
    switch (view) {
      case TaskModalEnum.Complete:
        return t('Complete Task');
      case TaskModalEnum.Comments:
        return t('Task Comments');
      case TaskModalEnum.Log:
        return t('Log Task');
      case TaskModalEnum.Edit:
        return t('Edit Task');
      default:
        return t('Add Task');
    }
  }, [view]);
  console.timeEnd('Memoized');

  console.time('Not memoized');
  let title;
  switch (view) {
    case TaskModalEnum.Complete:
      title = t('Complete Task');
      break;
    case TaskModalEnum.Comments:
      title = t('Task Comments');
      break;
    case TaskModalEnum.Log:
      title = t('Log Task');
      break;
    case TaskModalEnum.Edit:
      title = t('Edit Task');
      break;
    default:
      title = t('Add Task');
      break;
  }
  console.timeEnd('Not memoized');
```

src/components/Task/Modal/TaskModal.tsx
dr-bizz: This is good to know

src/components/Task/Modal/TaskModal.tsx
dr-bizz: I'm going to stick with the memo. I did some more testing and found the function was called less often. I tested this by putting a console log in the function and counting the times is was logged out.
With the memorized function, the function was only logged twice (both when the modal opens), and with the un-memorized function, it was logged 6 times (4 when the modal opens and 2 when closes)

dr-bizz:
@canac, I had to rebase it to remove your changes from this PR. I've added some of the changes you recommended. I'm keeping the memo function and keeping the Modal content inside the return to keep it consistent with our other Components.

src/components/Task/Modal/TaskModal.tsx
canac: Definitely! That's what I would expect to see. I assume it's called twice instead of once because of React strict mode.

PR #921, opened 2024-04-22
No-Jira - Re-enable Previews

wrandall22:
## Description

This PR seeks to re-enable Amplify previews now that this is a public repo.

## Checklist:

### DIFF
--- a/.github/workflows/preview-close.yml
+++ b/.github/workflows/preview-close.yml
@@ -1,0 +1,35 @@
+name: 'Amplify PR Closed'
+on:
+  pull_request_target:
+    types: [closed]
+
+jobs:
+  delete:
+    permissions:
+      id-token: write
+    if: contains(github.event.pull_request.labels.*.name, 'Preview Environment')
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: configure AWS credentials
+        id: creds
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: us-east-1
+          role-to-assume: ${{ secrets.AWS_GITHUB_ACTIONS_ROLE }}
+          output-credentials: true
+      - name: set branchname env
+        id: setbranchname
+        run: |
+          # use GITHUB_HEAD_REF that is set to PR source branch
+          echo "branch=$GITHUB_HEAD_REF" >> "$GITHUB_OUTPUT"
+      - name: cleanup PR preview branch
+        uses: yinlinchen/amplify-preview-actions@v1.2
+        with:
+          branch_name: ${{ github.head_ref }}
+          amplify_command: delete
+        env:
+          AWS_ACCESS_KEY_ID: ${{ steps.creds.outputs.aws-access-key-id }}
+          AWS_SECRET_ACCESS_KEY: ${{ steps.creds.outputs.aws-secret-access-key }}
+          AmplifyAppId: ${{ secrets.AMPLIFY_APP_ID }}
+          AWS_REGION: 'us-east-1'
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -1,0 +1,40 @@
+name: Amplify Preview
+
+on:
+  pull_request:
+    types: [labeled]
+
+jobs:
+  deploy:
+    if: contains(github.event.pull_request.labels.*.name, 'Preview Environment')
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    permissions:
+      id-token: write
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v3
+      - name: configure AWS credentials
+        id: creds
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: us-east-1
+          role-to-assume: ${{ secrets.AWS_GITHUB_ACTIONS_ROLE }}
+          output-credentials: true
+      - name: set branch name env
+        id: setbranchname
+        run: |
+          # use GITHUB_HEAD_REF that is set to PR source branch
+          echo "branch=$GITHUB_HEAD_REF" >> "$GITHUB_OUTPUT"
+      - name: deploy PR preview
+        uses: CruGlobal/amplify-preview-actions@handle-existing-branches
+        with:
+          branch_name: ${{ github.head_ref }}
+          amplify_command: deploy
+        env:
+          AWS_ACCESS_KEY_ID: ${{ steps.creds.outputs.aws-access-key-id }}
+          AWS_SECRET_ACCESS_KEY: ${{ steps.creds.outputs.aws-secret-access-key }}
+          AmplifyAppId: ${{ secrets.AMPLIFY_APP_ID }}
+          AWS_REGION: 'us-east-1'
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          EnvironmentVariables: 'SITE_URL=https://${{ github.head_ref }}.${{ secrets.AMPLIFY_APP_ID }}.amplifyapp.com'
--- a/amplify.yml
+++ b/amplify.yml
@@ -10,6 +10,10 @@ frontend:
             PREVIEW_URL="https://${AWS_BRANCH_ARN#*/branches/}.${AWS_APP_ID}.amplifyapp.com";
             NEXTAUTH_URL=$PREVIEW_URL;
             echo "PREVIEW_URL=$PREVIEW_URL" >> .env;
+          elif [[ "${AWS_BRANCH}" != "main" && "${AWS_BRANCH}" != "staging" ]]; then
+            PREVIEW_URL="https://${AWS_BRANCH}.${AWS_APP_ID}.amplifyapp.com";
+            NEXTAUTH_URL=$PREVIEW_URL;
+            echo "PREVIEW_URL=$PREVIEW_URL" >> .env;
           fi
         - echo "NEXTAUTH_URL=$NEXTAUTH_URL" >> .env
         - echo "NODE_ENV=$NODE_ENV" >> .env


wrandall22:
generate preview

wrandall22:
generate preview

wrandall22:
generate preview

wrandall22:
generate preview

wrandall22:
generate preview

wrandall22:
generate preview

wrandall22:
generate preview

wrandall22:
generate preview

wrandall22:
generate preview

wrandall22:
generate preview

wrandall22:
According to the docs, this will only work once it is on `main` already. I have tested it locally using [act](https://nektosact.com/usage/index.html) and it seems to work there.

.github/workflows/preview.yml
```diff
@@ -0,0 +1,76 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: check if PR
+        run: echo "has comment with 'generate preview' in it"
+  deploy:
+    needs: prepare
+    if: |
+      github.event.issue.pull_request &&
+      (github.event.comment.author_association == 'MEMBER' || github.event.comment.author_association == 'CONTRIBUTOR') &&
```
dr-bizz, line 49: I know we don't work with contractors now, but if we did, would `member` and `contributor` allow a contractor to run this workflow or would we need to add `COLLABORATOR`?

.github/workflows/preview.yml
```diff
@@ -0,0 +1,76 @@
...
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: us-east-1
+          role-to-assume: ${{ secrets.AWS_GITHUB_ACTIONS_ROLE }}
+          output-credentials: true
+      - name: set branchname env
+        id: setenvname
+        run: |
+          # use GITHUB_HEAD_REF that is set to PR source branch
+          echo "##[set-output name=setbranchname;]$(echo ${GITHUB_HEAD_REF})"
```
dr-bizz, line 65: It says here that we shouldn't use `set-output`, but I haven't looked into for too long - https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

dr-bizz (commented):
Thank you for working on this! This will help us massively. I had a few questions/comments, but ti mostly looked great.

.github/workflows/preview.yml
wrandall22: Typically, we add contractors as members of the org when giving permissions.

.github/workflows/preview.yml
wrandall22: Thanks for seeing it, yeah, I should be using the `GITHUB_OUTPUT` env var instead.

canac:
Thanks for trying to get these working again! Will this require us to comment `preview generate` after every push to update the preview environment? It was nice that it was automatic before so you never had to worry about the preview environment being out-of-date.

If it wouldn't be automatic anymore, maybe you could change this to be a label? Then on every push, the workflow could update the preview environment if the label is present.

wrandall22:
It would be `generate preview` (vs `preview generate`). I think it would require a new comment (much like how `atlantis` works in terraform) each time. The reason I didn't use a label was because there is no way to prevent some outside person from generating infinite previews in our environment since anyone can create a PR and add a label and start pushing commits.

canac:
I believe you need at least `triage` permissions to add/remove labels https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization#:~:text=Apply/dismiss%20labels

wrandall22:
So, this is successfully creating a preview environment now, though [this action](https://github.com/yinlinchen/amplify-preview-actions?tab=readme-ov-file) creates it in the all apps section instead of the previews section, and so far it is not creating a comment here with the preview url. Do we want to merge this and make a separate PR for those improvements?

wrandall22:
It does not appear the CLI can handle creating pull_request previews, so it will have to be a branch.

canac:
Were you able to sign in? When I try to sign into the preview environment, it redirects me [here](https://next-stage.mpdx.org/api/auth/signin?error=OAuthCallback) after the Okta login and shows this error message.

Screenshot 2024-04-24 at 2 24 12 PM

canac:
Also is it possible to comment the link to the preview URL?

wrandall22:
I also have that issue with logging in, not sure why though, since the [pattern of the Okta apps in terraform](https://github.com/CruGlobal/cru-terraform/blob/master/applications/mpdx/web-react/okta.tf#L19) matches the branch pattern just like it did the pull_request previews pattern.

.github/workflows/preview.yml
```diff
@@ -35,3 +35,13 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           AmplifyAppId: ${{ secrets.AMPLIFY_APP_ID }}
           AWS_REGION: 'us-east-1'
+      - name: comment with preview url
```
canac, line 38: Is there a way to make this run only once instead of on every push? Maybe we could check whether the trigger type is "labeled"?

canac:
> I also have that issue with logging in, not sure why though, since the [pattern of the Okta apps in terraform](https://github.com/CruGlobal/cru-terraform/blob/master/applications/mpdx/web-react/okta.tf#L19) matches the branch pattern just like it did the pull_request previews pattern.

I wonder if it's a matter of getting the environment variables configured correctly. It's suspicious that after login it redirected back to next-stage.mpdx.org instead of the preview environment. @dr-bizz would know since he set it up originally.

wrandall22:
Yeah, maybe I need to pass in some env vars.

wrandall22:
Preview URL: https://..amplifyapp.com/

dr-bizz:
I believe the issue with authentication is here `amplify.yml` as we defined the `NEXTAUTH_URL` & `PREVIEW_URL` here. - We need the amplify.yml to stay the same since that is how other organizations use it.

One way to check to see if this is the issue is by going to AWS Amplify Cloud and adding an override for the `NEXTAUTH_URL` variable.

wrandall22:
The [Fixup: env vars](https://github.com/CruGlobal/mpdx-react/pull/921/commits/bd8105d175a78cac49c7451a86c30a89edd80169) commit fixed login

amplify.yml
```diff
@@ -6,8 +6,8 @@ frontend:
         - nvm install 18.13
         - nvm use 18.13
         - |
-          if [[ "${AWS_BRANCH_ARN#*/branches/}" == *"pr-"* ]]; then
-            PREVIEW_URL="https://${AWS_BRANCH_ARN#*/branches/}.${AWS_APP_ID}.amplifyapp.com";
+          if [[ "${AWS_BRANCH}" != "main" && "${AWS_BRANCH}" != "staging" ]]; then
+            PREVIEW_URL="https://${AWS_BRANCH}.${AWS_APP_ID}.amplifyapp.com";
```
dr-bizz, line 10: Instead of removing the existing PR logic, let's leave it in and add a else if `elif` command as other organizations will have the normal previews and they might not have the branches `main` and `staging`.

wrandall22:
Preview URL: $PREVIEW_URL

wrandall22:
Preview URL: https://.wee.amplifyapp.com/

wrandall22:
All the pieces are working (at least some of the time). I'll clean up the commits after approval to ensure you all can see what changed clearly. I don't know why the `amplify-preview-actions` ran the comment since I didn't give it a 3rd argument for comment_url. I could have my fork delete that commenting feature if we want.

wrandall22:
Looks like double comment could be removed by removing GITHUB_TOKEN, so I'll try that.

canac:
I see that your last commit generated this error:

```
The requested branch arn:aws:amplify:us-east-1:753232867021:apps/***/branches/enable-previews already have pending or running jobs
```

This makes me wonder if you only have to create the deployment once and then it automatically updates when the branch is pushed, like it does for `main` and `staging`. In the dashboard it says, "Continuous deploys set up", which hints that the update on push might happen automatically. The deploy workflow might only need to run on `labeled`.

Screenshot 2024-04-25 at 1 27 18 PM

wrandall22:
I'll give it a try and see.

wrandall22:
You are correct, a push automatically runs the deploy.

.github/workflows/preview.yml
```diff
@@ -0,0 +1,42 @@
+name: Amplify Preview
+
+on:
+  pull_request:
+    types: [labeled]
+
+jobs:
+  deploy:
+    if: contains(github.event.pull_request.labels.*.name, 'Preview Environment') && github.event.action == 'labeled'
```
canac, line 9:
```suggestion
    if: contains(github.event.pull_request.labels.*.name, 'Preview Environment')
```

.github/workflows/preview-close.yml
```diff
@@ -0,0 +1,37 @@
+name: 'Amplify PR Closed'
+on:
+  pull_request_target:
+    branches:
+      - main
```
canac, line 5: Is this filter needed? I assume that if we had a PR from `other-branch` into `feature-branch`, we would still want to delete the preview environment when `other-branch` merges.

.github/workflows/preview.yml
```diff
@@ -0,0 +1,42 @@
...
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    permissions:
+      id-token: write
+      contents: write
+      pull-requests: write
+      repository-projects: write
```
canac, line 16: Are all of these permissions needed? I would think it would only need `id-token` like the other workflow.

canac (commented):
This is looking great! Thanks for getting it working. The rename suggestions are just stylistic and not necessarily required.

.github/workflows/preview.yml
canac: Gotcha. Even `repository-projects`?

wrandall22: According to what I've read, yes. However, we can test it with a few more commits, but it will take the rest of the morning at least to do so.

wrandall22: According to [this doc](https://docs.github.com/en/rest/issues/comments?apiVersion=2022-11-28#create-an-issue-comment) we may be able to get away with just `pull-requests: write`

canac (approved):
Looks great to me! I'm excited to have this working again.

.github/workflows/preview-close.yml
```diff
@@ -0,0 +1,35 @@
...
+          output-credentials: true
+      - name: set branchname env
+        id: setbranchname
+        run: |
+          # use GITHUB_HEAD_REF that is set to PR source branch
+          echo "branch=$GITHUB_HEAD_REF" >> "$GITHUB_OUTPUT"
+      - name: cleanup PR preview branch
+        uses: yinlinchen/amplify-preview-actions@v1.2
+        with:
+          branch_name: ${{ steps.setbranchname.outputs.branch }}
```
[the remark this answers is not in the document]
wrandall22: I couldn't use `env`, but I was able to get it to work using the `context` directly.

.github/workflows/preview.yml
```diff
@@ -0,0 +1,40 @@
...
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: us-east-1
+          role-to-assume: ${{ secrets.AWS_GITHUB_ACTIONS_ROLE }}
+          output-credentials: true
+      - name: set branch name env
+        id: setbranchname
+        run: |
+          # use GITHUB_HEAD_REF that is set to PR source branch
+          echo "branch=$GITHUB_HEAD_REF" >> "$GITHUB_OUTPUT"
```
canac, line 28: If you make a follow-up PR, I think this step can be removed in both workflows.

wrandall22: Oh, thought I did that already, must have come back in rebase.

PR #926, opened 2024-04-25
[no-Jira] Run `yarn gql:w` alongside the Next.js dev server

canac:
## Description

* Keep GraphQL generated files in sync with the source files by committing generated files.
* Add a custom script to delete stale generated files that were left when a `.graphql` file was moved, renamed, or deleted. These stale files can cause TS errors and were a significant source of confusion.
* Add a GitHub Action that will check that the generated files are up-to-date.
* Speed up `yarn gql` by not running `prettier` on the generated files.
* The files in `src/graphql/` are not committed because their contents may differ depending on whether `yarn gql` was pointed at the staging or the production API server.
* Remove post-checkout `git` hook because it is no longer needed.

## Checklist:


canac:
@dr-bizz Sorry for another large PR. The vast majority of the files are yarn packages and generated files, which you should be able to ignore. But let me know if you want me to try to split it up more before you review.

.github/workflows/ci.yml
```diff
@@ -98,12 +98,33 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 18.13.0
-      - name: 📦 Install Dependencies
+      - name: 🔍 Check Dependencies
         run: yarn install --immutable --immutable-cache --check-cache
 
+  gql-codegen:
```
dr-bizz, line 104: Can you please put this in a separate commit to ensure commits have one very clear goal? - This also helps future devs when viewing code, as they know why changes happened due to the commit message.

delete-stale-files.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
dr-bizz, line 1: Please can you make this into a separate commit for the same reasons as above

pages/GetAccountLists.generated.ts
```diff
@@ -0,0 +1,45 @@
+import * as Types from '../src/graphql/types.generated';
```
dr-bizz, line 1: Please make all the generated files in a separate commit for the same reasons as above

src/components/Coaching/CoachingDetail/MonthlyCommitment/MonthlyCommitment.test.tsx
```diff
@@ -35,7 +35,7 @@ describe('MonthlyCommitment', () => {
     await waitForNextUpdate();
     expect(
       result.current.data?.reportPledgeHistories?.length,
-    ).toMatchInlineSnapshot(`3`);
+    ).toMatchInlineSnapshot(`2`);
```
dr-bizz, line 38: Is the commit `fixup! Commit files generated by gql-codegen` is because you're fixing tests after upgrading graphql-codegen?

If so, please make the commit message something like "Fixing changes after gql-codegen" or whatever. This will help future devs understand the changes you've made.

dr-bizz (changes requested):
Thank you for doing this. Your changes will make it easier for devs to switch between local branches.

I have left a few comments on how you can split up the changes more to ensure there is only one change per commit.
Can you also split this PR into two? One for `updating graphql-codegen` and the related files, and the other for .generated files and other commits. Having two separate PRs will ensure each PR is focused on one Conceptual change per PR.

canac:
@dr-bizz I broke the work up into 2 PRs (see #928) and a lot of separate commits. Let me know if that's what you had in mind.

.github/workflows/ci.yml
canac, line 104: I will also make this a required check in Terraform.

canac, line 104: Here is the Terraform PR to make this a required check https://github.com/CruGlobal/cru-terraform/pull/5838

.github/workflows/ci.yml
dr-bizz, line 1: Since we're adding the .generated files to the repo, I think the only job we need to run `yarn gql` is on gql-codegen, as the developer should have run this before pushing.

dr-bizz: We should add a pre-push husky to run `yarn gql` before pushing, so we don't have to run it multiple times on the GitHub Action.

codegen.yml
```diff
@@ -59,5 +59,4 @@ generates:
       - ./extractRootFields.js
 hooks:
   afterAllFileWrite:
-    - prettier --write
```
dr-bizz, line 62: I think it's nice to have the .generated files run by prettier, especially since now we're adding them to the repo. I know this adds a few seconds, but I think it's worth it.

.github/workflows/ci.yml
canac: > Since we're adding the .generated files to the repo, I think the only job we need to run yarn gql is on gql-codegen, as the developer should have run this before pushing.

Like I noted in the PR body:

> The files in `src/graphql/` are not committed because their contents may differ depending on whether `yarn gql` was pointed at the staging or the production API server.

I'm only committing these generated files (https://github.com/CruGlobal/mpdx-react/blob/0dc2785eee9831b7aff177b57c8ad842ea8ad207/codegen.yml#L12-L25) and `pages/api/graphql-rest.page.generated.ts`. If we committed the rest of the generated files under `src/graphql`, for example `schema.graphql`, the developer would commit the staging schema and CI would check that it matches the production schema. If there's anything in the GraphQL that's in production and not in staging, the generated files will have differences and CI will fail. That's why we need to keep everything in `src/graphql` not committed, which means that we still need to run `yarn gql` in every workflow step and during the Amplify deploy.

canac: > We should add a pre-push husky to run yarn gql before pushing, so we don't have to run it multiple times on the GitHub Action.

Personally, I disagree with this. I don't want to add another 5 seconds of waiting every time I push, just in case I forgot to run `yarn gql`. It already takes a while waiting for linting. If I forgot to run it, it will show up in CI and be caught pretty quickly.

codegen.yml
```diff
@@ -21,6 +21,7 @@ generates:
       - typescript-operations
       - typescript-react-apollo
     config:
+      addDocBlocks: false
```
dr-bizz, line 24: Good call

dr-bizz (commented):
Great work on this. I left a few comments.

codegen.yml
canac: Yeah, I was questioning it myself because it is kind of nice to have formatted files when you have to look inside generated TS files. I'll revert those commits.

.github/workflows/ci.yml
dr-bizz: > We should add a pre-push husky to run yarn gql before pushing, so we don't have to run it multiple times on the GitHub Action.

I understand your POV, and I know it may take longer to push due to waiting for Lint and gql to run, but it's still better, as we're setting our commits up for success, ensuring they will pass as many CI tests as possible. If anyone is like me, I rely on the `lint:ts` husky pre-push command to catch any errors I make.

To compromise, we can create a package script that runs `gql` before pushing.

```
"push": "yarn gql && git push origin"
```

You would write `yarn push [BRANCH_NAME]

dr-bizz (approved):
Everything looks good so I'm approving this.

Please can you add the package script I mentioned and add some text to the ReadMe explaining why you need to run gql before pushing.

canac:
I switched to a different approach of running `yarn gql:w` in parallel when you start the Next.js dev server. It should address all of the pain points of outdated generated files without needing to commit the generated files.

dr-bizz (approved):

PR #929, opened 2024-04-25
[no-Jira] Display cached contact details data immediately

canac:
## Description

When Apollo queries are executed, `data` will be the cached data, but `loading` will still be `true` until the network request completes and the freshest data is available. Any time we show a loading spinner or skeleton when `loading` is true, we are ignoring the cached data that is already available. This PR makes the components in the contact details panel display cached data immediately if it is available.

## Checklist:

### DIFF
--- a/src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.tsx
+++ b/src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.tsx
@@ -69,9 +69,10 @@ export const ContactDetailsHeader: React.FC<Props> = ({
   setContactDetailsLoaded,
   contactDetailsLoaded,
 }: Props) => {
-  const { data, loading } = useGetContactDetailsHeaderQuery({
+  const { data } = useGetContactDetailsHeaderQuery({
     variables: { accountListId, contactId },
   });
+  const loading = !data;
   const { t } = useTranslation();
 
   const { editModalOpen, setEditModalOpen } = React.useContext(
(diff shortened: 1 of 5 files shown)


src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.tsx
```diff
@@ -90,7 +91,7 @@ export const ContactDetailsHeader: React.FC<Props> = ({
       <HeaderBar>
         <ContactAvatar alt={data?.contact.name} src={data?.contact.avatar} />
         <HeaderBarContactWrap>
-          {loading ? (
+          {!data ? (
```
dr-bizz, line 94: You could use the `loading` variable here

dr-bizz (approved):
Great work with this! Tour explaination makes sense, there is probably a lot of other plavces we use loading when we should just check to see if the `data` variable is `null`.

I liked that you created a `loading` constant in ContactDetailsHeader.tsx we could use that in the other files. But having `!data` on each JSX If statment made it super clear what was happening, so i'm a fan of both.

src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsHeader.tsx
canac: The benefit of using `!data` instead of `loading` is that type narrowing happens correctly. TypeScript knows that `data` can't be falsy in the other branch, which means that we don't have to use optional chaining unnecessarily.

canac:
> There is probably a lot of other places we use loading when we should just check to see if the data variable is null.

Definitely! I started with contact details because it is frequently used, but there are plenty of other components that could be improved future PRs.

PR #930, opened 2024-04-26
HS-1142757 - Organizations links on Settings aren't rendering correctly

dr-bizz:
## Description
[HelpScout issue](https://secure.helpscout.net/conversation/2578696149/1142757?folderId=7296147)
The Organization links are only shown if the user is an admin. However, on the old MPDx, users who weren't admins but managed at least one organization could see the page.

This PR allows users who manage an organization and aren't an admin to view the links and visit the page.

## Checklist:

### DIFF
--- a/src/components/Shared/MultiPageLayout/MultiPageMenu/MultiPageMenu.graphql
+++ b/src/components/Shared/MultiPageLayout/MultiPageMenu/MultiPageMenu.graphql
@@ -1,0 +1,9 @@
+query ManageOrganizationsAccess {
+  user {
+    administrativeOrganizations(first: 1) {
+      nodes {
+        id
+      }
+    }
+  }
+}
--- a/src/components/Shared/MultiPageLayout/MultiPageMenu/MultiPageMenu.tsx
+++ b/src/components/Shared/MultiPageLayout/MultiPageMenu/MultiPageMenu.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react';
+import React from 'react';
 import Close from '@mui/icons-material/Close';
 import {
   Box,
@@ -9,6 +9,7 @@ import {
   Typography,
 } from '@mui/material';
 import { styled } from '@mui/material/styles';
+import { Session } from 'next-auth';
 import { useTranslation } from 'react-i18next';
 import { makeStyles } from 'tss-react/mui';
 import { useGetDesignationAccountsQuery } from 'src/components/EditDonationModal/EditDonationModal.generated';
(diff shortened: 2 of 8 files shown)


src/components/Shared/MultiPageLayout/MultiPageMenu/MultiPageMenu.tsx
```diff
@@ -54,6 +60,31 @@ const FilterList = styled(List)(({ theme }) => ({
   },
 }));
 
+type ShowMenuItemProps = {
+  item: NavItems;
+  user: Session['user'];
+  hasOrganizationsAccess: boolean;
+};
+
+const showMenuItem = ({
```
canac, line 69: I like that you pulled this out into a separate function!

src/components/Shared/MultiPageLayout/MultiPageMenu/MultiPageMenu.tsx
```diff
@@ -54,6 +60,31 @@ const FilterList = styled(List)(({ theme }) => ({
...
+  item,
+  user,
+  hasOrganizationsAccess,
+}: ShowMenuItemProps): boolean => {
+  if (item?.grantedAccess?.length) {
+    if (hasOrganizationsAccess && ['organizations'].includes(item.id)) {
```
canac, line 75: Is there a reason we're not doing this?

```suggestion
    if (hasOrganizationsAccess && item.id === 'organizations')) {
```

src/components/Shared/MultiPageLayout/MultiPageMenu/MultiPageMenu.graphql
```diff
@@ -0,0 +1,9 @@
+query ManageOrganizationsAccess {
+  user {
+    administrativeOrganizations(first: 2) {
```
canac, line 3: We only check that there is at least one org, so we only need to one, right?

```suggestion
    administrativeOrganizations(first: 1) {
```

canac (approved):
This looks great! I don't think that staging built successfully because it isn't working in staging, but I impersonated the HelpScout user, and it works great locally. I just left a couple of minor questions/suggestions.

src/components/Shared/MultiPageLayout/MultiPageMenu/MultiPageMenu.tsx
dr-bizz: Yes, as there are 3 organization links `organizations`, `organizations/accountLists` & `organizations/contacts'`

src/components/Shared/MultiPageLayout/MultiPageMenu/MultiPageMenu.graphql
dr-bizz: That is correct, I can change it to 1

src/components/Shared/MultiPageLayout/MultiPageMenu/MultiPageMenu.tsx
canac: In that case, don't you want `item.id.includes('organizations')` or `item.id.startsWith('organizations')`? What you have right how will only match 'organizations'.

dr-bizz: You're right, it's weird that the links were showing

dr-bizz:
@canac I've pushed it to staging

canac:
@dr-bizz Works great in staging. Thanks for fixing this! 🎉

PR #931, opened 2024-04-26
HS 1144092 - Fixing error of setting wrong accountList when impersonating

dr-bizz:
## Description
https://secure.helpscout.net/conversation/2580546553/1144092?folderId=7296147
Fixing error when redirected back to old MPDx after impersonating - Ensuerung the Impersonators accountListId is used.

## Checklist:

### DIFF
--- a/pages/api/handoff.page.ts
+++ b/pages/api/handoff.page.ts
@@ -23,10 +23,11 @@ export const returnRedirectUrl = async (
 
   const path = req.query.path ?? '';
 
-  const apiToken =
-    useImpersonatorToken && jwtToken?.impersonating
-      ? jwtToken?.impersonatorApiToken
-      : jwtToken?.apiToken;
+  const isImpersonating = !!(useImpersonatorToken && jwtToken?.impersonating);
+
+  const apiToken = isImpersonating
+    ? jwtToken?.impersonatorApiToken
+    : jwtToken?.apiToken;
 
   if (apiToken && jwtToken?.userID && req.query.auth !== 'true') {
     const ssrClient = makeSsrClient(apiToken);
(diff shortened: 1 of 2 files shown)


pages/api/handoff.page.ts
```diff
@@ -36,13 +37,17 @@ export const returnRedirectUrl = async (
     >({
       query: GetDefaultAccountDocument,
     });
-    let defaultAccountID = req.query?.accountListId;
-    if (!defaultAccountID) {
```
dr-bizz, line 40: This is what was causing the error. We were passing the Impersonatee's accountListId to the old MPDx causing the error.

canac: If the user is impersonating, won't `user` be the impersonated user (not the logged in user) since we are using `impersonatorApiToken`? If that user has a `defaultAccountList`, I believe this will still use the impersonated user's account list id. But maybe I'm misunderstanding.

pages/api/handoff.page.ts
```diff
@@ -36,13 +37,17 @@ export const returnRedirectUrl = async (
...
-        data: { user, accountLists },
-      } = response;
-      defaultAccountID = user?.defaultAccountList || accountLists?.nodes[0]?.id;
-    }
+    const {
+      data: { user, accountLists },
+    } = response;
+
+    const defaultAccountID =
+      isImpersonating && user.defaultAccountList
```
canac, line 45: Do we want to fall back to `accountLists?.nodes[0]?.id` too even when impersonating?

pages/api/handoff.page.ts
canac: No that makes sense, I was just misunderstanding

pages/api/handoff.page.ts
dr-bizz: That's a great point

wrandall22 (approved):

PR #933, opened 2024-04-29
[no-Jira] Import i18n from `src/lib/i18n`

canac:
## Description

@caleballdrin and I discovered a subtle gotcha with i18n. If you `import i18n from 'i18next'` the i18n instance isn't configured to load the translation files. In tests `i18n.t()` will always return undefined. The fix was to import the configured `i18n` instance from `src/lib/i18n`.

This PR fixes all other places in the codebase that import from `i18next` to import from `src/lib/i18n` to help prevent that bug from reoccurring.

## Checklist:


dr-bizz (approved):
Great catch! thank you for working  on this

PR #936, opened 2024-04-30
MPDX-7893 - Allowing users to add offline organizations to their accounts

dr-bizz:
## Description
Fixing error introduced in https://github.com/CruGlobal/mpdx-react/pull/891

Currently, on Connect Services, users aren't able to add offline organizations due to the `disableNewUsers` returning as `NULL` from GraphQL.

The issue was caused by a FormIk client error. I've enabled `disableNewUsers` to be nullable and added a test to ensure `NULL` is treated the same way as `TRUE`.

## Checklist:

### DIFF
--- a/src/components/Settings/integrations/Organization/Modals/OrganizationAddAccountModal.tsx
+++ b/src/components/Settings/integrations/Organization/Modals/OrganizationAddAccountModal.tsx
@@ -56,6 +56,41 @@ export type OrganizationFormikSchema = {
   password: string | undefined;
 };
 
+const OrganizationSchema: yup.SchemaOf<OrganizationFormikSchema> = yup.object({
+  selectedOrganization: yup
+    .object({
+      id: yup.string().required(),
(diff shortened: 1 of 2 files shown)


caleballdrin (approved):
Looks good. thanks for adding the test

src/components/Settings/integrations/Organization/Modals/OrganizationAddAccountModal.tsx
```diff
@@ -153,7 +153,7 @@ export const OrganizationAddAccountModal: React.FC<
           name: yup.string().required(),
```
canac, line 153: Is it possible to move this schema outside of the component? I don't think it uses anything from the component closure. Having it inside the component means it will be unnecessarily recreated on every render.

canac (approved):
Thanks for fixing all these bugs! I left an optional suggestion that I must have missed in previous code review.

PR #937, opened 2024-04-30
MPDX-7891 - Allow contact search on Organization Accounts without organization selected

dr-bizz:
## Description
[MPDX-7891](https://jira.cru.org/secure/RapidBoard.jspa?rapidView=3&view=detail&selectedIssue=MPDX-7891)
Allow contact search on the Organization Accounts page without an organization being selected. I added this to speed up the search, but the admin needs to search contacts without an organization being selected for a broader search.

## Checklist:


canac (approved):
It looks good! My only thought is that we should consider making `SearchOrganizationsAccountListsInput.organizationId` nullable instead of its type being `ID!` but accepting an empty string for the `organizationId`. That would be a more semantic use of the GraphQL type system. What do you think about that?

PR #943, opened 2024-05-10
[MPDX-7894] Save all of the user's past account list ids in the DataDog session

canac:
## Description

* Since `useDataDog.ts` doesn't export a hook, rename it to `dataDog.ts` and move it from `hooks` to `lib`.
* Remember the user's previous account list ids and add all of them to the DataDog session.

This PR is not immediately related to https://jira.cru.org/browse/MPDX-7894. However it lays the foundation for adding the user's language in a follow-up PR, which is relevant to that ticket.

## Checklist:


src/lib/dataDog.ts
```diff
@@ -0,0 +1,57 @@
...
+  interface Window {
+    DD_RUM: {
+      setUser: (user: Record<string, unknown>) => void;
+      clearUser: () => void;
+    };
+  }
+}
+
+export const isDataDogConfigured = (): boolean => {
+  if (typeof window === 'undefined') return false;
```
wrandall22, line 11: I think we should adopt the `curly: ['error', 'all']` lint rule. I'm a huge fan of always having curly braces around if blocks, even if they are one liners. Feel free to discuss. cc @dr-bizz

canac: I'm used to the curly braces being everywhere as well, and it always throws me off a little bit when they're not there.

canac: #945

src/components/DataDog/DataDog.tsx
```diff
@@ -11,11 +11,11 @@ const DataDog: React.FC = () => {
     ? Array.isArray(query.accountListId)
       ? query.accountListId[0]
       : query.accountListId
-    : '';
+    : null;
 
   const user = session?.user;
   useEffect(() => {
-    if (user) {
+    if (user && accountListId) {
```
dr-bizz, line 18: What if the user is on the accountListId page when they have mulitple accounts. Currently DataDog won't be saving their session

src/lib/dataDog.ts
dr-bizz: I'm good with adding it

src/lib/dataDog.test.ts
```diff
@@ -0,0 +1,107 @@
...
+    });
+  });
+
+  describe('when DataDog is configured', () => {
+    beforeEach(() => {
+      process.env.DATADOG_CONFIGURED = 'true';
+    });
+
+    //#region Default Tests
+    it('isDataDogConfigured should return false', () => {
```
dr-bizz, line 41:
```suggestion
    it('isDataDogConfigured should return true', () => {
```

dr-bizz (commented):
I had a few questions, but overall it looks great. Thank you for working on this.

src/lib/dataDog.test.ts
canac: Good catch, thanks! I must have accidentally changed that.

src/components/DataDog/DataDog.tsx
canac: You're right, I need to remove that `&& accountListId` condition. Thanks for catching that.

dr-bizz (approved):
The code looks great! Thank you for implementing this

PR #954, opened 2024-06-13
Mpdx-7583 duplicate tags

cjpastika:
## Description
Jira Ticket: Mpdx-7583

-Before when adding duplicate tags they just wouldn't save upon refresh
-Created an error snackbar and deleted the tag when a user tries to create a duplicate tag
-In the contact list view this will apply if any selected contacts have the tag the user is trying to add

## Checklist:

- [ X] I have given my PR a title with the format "MPDX-(JIRA#) (summary sentence max 80 chars)"
- [ X] I have applied the appropriate labels. (_Add the label "On Staging" to get the branch automatically merged into staging._)
- [ X] I have requested a review from another person on the project

### DIFF
--- a/src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
+++ b/src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
@@ -84,6 +84,22 @@ export const ContactTags: React.FC<ContactTagsProps> = ({
     }
   };
 
+  const handleValidation = async ({
+    tagList,
+  }: {
+    tagList: string[] & never[];
+  }): Promise<void> => {
+    for (let i = 0; i < tagList.length; i++) {
+      if (contactTags.includes(tagList[i])) {
+        enqueueSnackbar(t('Cannot add duplicate tags'), {
+          variant: 'error',
+        });
+        tagList.pop();
+      }
+    }
+    return;
+  };
+
   const onSubmit = async (
     { tagList }: { tagList: string[] & never[] },
     { resetForm }: FormikHelpers<{ tagList: string[] & never[] }>,
@@ -133,6 +149,7 @@ export const ContactTags: React.FC<ContactTagsProps> = ({
       <Formik
         initialValues={{ tagList: [] }}
         validationSchema={tagSchema}
+        validate={handleValidation}
         onSubmit={onSubmit}
       >
         {({
--- a/src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
+++ b/src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
@@ -91,9 +91,12 @@ export const ContactTags: React.FC<ContactTagsProps> = ({
   }): Promise<void> => {
     for (let i = 0; i < tagList.length; i++) {
       if (contactTags.includes(tagList[i])) {
-        enqueueSnackbar(t('Cannot add duplicate tags'), {
-          variant: 'error',
-        });
+        enqueueSnackbar(
+          t('Cannot add duplicate tags, duplicate tag has been removed'),
+          {
+            variant: 'error',
+          },
+        );
         tagList.pop();
       }
     }
--- a/src/components/Contacts/MassActions/AddTags/MassActionsAddTagsModal.test.tsx
+++ b/src/components/Contacts/MassActions/AddTags/MassActionsAddTagsModal.test.tsx
@@ -21,7 +21,6 @@ jest.mock('notistack', () => ({
     };
   },
 }));
-
 const selectedIds = ['abc'];
 const accountListId = '123456789';
 describe('MassActionsAddTags', () => {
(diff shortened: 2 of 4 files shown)


caleballdrin:
Great job on this Collin! Everything seems to work smoothly. And good work on the test, that looks like it was complicated haha.

src/components/Contacts/MassActions/AddTags/MassActionsAddTagsModal.tsx
```diff
@@ -70,6 +70,38 @@ export const MassActionsAddTagsModal: React.FC<
...
+
+    const existingTags = [];
+    contactsExistingTags.forEach((contact) => {
+      for (let i = 0; i < contact.tagList.length; i++) {
+        existingTags.push(contact.tagList[i]);
+      }
+    });
+
+    for (let i = 0; i < fields.tagList?.length; i++) {
+      existingTags.push(fields.tagList[i]);
```
caleballdrin, line 87: It looks like you have some typescript errors. You probably need to add ? optional chaining.

src/components/Contacts/MassActions/AddTags/MassActionsAddTagsModal.tsx
```diff
@@ -70,6 +70,38 @@ export const MassActionsAddTagsModal: React.FC<
...
+      const duplicates = existingTags.filter(
+        (item, index) => existingTags.indexOf(item) !== index,
+      );
+      if (duplicates.length > 0) {
+        enqueueSnackbar(
+          t('One or more selected contacts already has this tag'),
```
caleballdrin, line 93: @dr-bizz I think this should only prevent the user from adding tags if **every** selected contact already has the tag. If one or 2 of them already have the tag, I would want it to add the tag to the rest of the contacts and just skip over the ones that already have it. What do you think?

For example, if I've added a `likesBaseball` tag to a couple of contacts already and the next week I want to mass select and add it to 10 more people, I would have to go back to each of the 10 contacts and make sure I haven't already added the tag.

src/components/Contacts/MassActions/AddTags/MassActionsAddTagsModal.tsx
dr-bizz: @cjpastika can you push the changes we've made to fix the TS errors

src/components/Contacts/ContactDetails/ContactDetailsTab/Tags/ContactTags.tsx
```diff
@@ -84,6 +84,22 @@ export const ContactTags: React.FC<ContactTagsProps> = ({
...
+  }): Promise<void> => {
+    for (let i = 0; i < tagList.length; i++) {
+      if (contactTags.includes(tagList[i])) {
+        enqueueSnackbar(t('Cannot add duplicate tags'), {
+          variant: 'error',
+        });
+        tagList.pop();
+      }
+    }
+    return;
```
dr-bizz, line 100: Typically in JavaScript, When you want to edit any array, you create a new array from the original array and make edits on the new array, so you don't edit the original.

Why do you need to remove items from the tagList? Rather than just flagging to the user that they can't have dup tags?

dr-bizz: If you want to remove the tags, I would update the message to state that you removed the dup tag, so the user knows.

src/components/Contacts/MassActions/AddTags/MassActionsAddTagsModal.tsx
dr-bizz: Agreed.

src/components/Contacts/MassActions/AddTags/MassActionsAddTagsModal.tsx
```diff
@@ -70,6 +70,38 @@ export const MassActionsAddTagsModal: React.FC<
...
+          {
+            variant: 'error',
+          },
+        );
+        fields.tagList.pop();
+      }
+    }
+
+    return;
+  };
```
dr-bizz, line 103: You could use less loops by doing it this way
```suggestion
    const handleValidation = async (fields: Partial<ContactUpdateInput>) => {
    let existingTags: string[] = [];

    contactsForTags?.contacts.nodes.forEach((contact) => {
      existingTags = [...existingTags, ...contact.tagList];
    });

    for (let i = 0; i < fields.tagList?.length; i++) {
      existingTags = [...existingTags, fields.tagList[i]]
      const duplicates = existingTags.filter(
        (item, index) => existingTags.indexOf(item) !== index,
      );
      if (duplicates.length > 0) {
        enqueueSnackbar(
          t('One or more selected contacts already has this tag'),
          {
            variant: 'error',
          },
        );
        fields.tagList.pop();
      }
    }

    return;
  };
```

src/components/Contacts/MassActions/AddTags/MassActionsAddTagsModal.test.tsx
```diff
@@ -20,9 +21,9 @@ jest.mock('notistack', () => ({
     };
   },
 }));
+
```
dr-bizz, line 24: Remove extra line.

src/components/Contacts/MassActions/AddTags/MassActionsAddTagsModal.test.tsx
```diff
@@ -60,4 +61,60 @@ describe('MassActionsAddTags', () => {
...
+    expect(operation.variables.attributes[0].tagList[0]).toEqual('tag123');
+    await waitFor(() =>
+      expect(mockEnqueue).toHaveBeenCalledWith(
+        'One or more selected contacts already has this tag',
+        {
+          variant: 'error',
+        },
+      ),
+    );
+  });
```
dr-bizz, line 119: I would also check that the tag is not longer on the text input

dr-bizz (changes requested):

dr-bizz (approved):

PR #955, opened 2024-06-13
MPDX-7935 add mailing address

dr-bizz:
## Description
In this PR, I added a model for adding a mailing address, cleaned up the add button, and made some improvements to `AddAddressModal`, which I then did on `EditContactAddressModal`. I've also added tests for editing, deleting, and adding an address.

I also cleaned up the FixMailingAddress component, switched the modal state props to separate state values, and refactored the modal open/close handling.

## Checklist:

### DIFF
--- a/src/components/Tool/FixMailingAddresses/Contact.tsx
+++ b/src/components/Tool/FixMailingAddresses/Contact.tsx
@@ -207,6 +207,7 @@ const Contact: React.FC<Props> = ({
                       classes.paddingL2,
                       classes.hoverHighlight,
                     )}
+                    data-testid={`address-${address.id}`}
                     onClick={() => openEditAddressModal(address, id)}
                   >
                     <Box className={classes.address}>
@@ -248,6 +249,7 @@ const Contact: React.FC<Props> = ({
               >
                 <AddButton
                   className={classes.AddButton}
+                  data-testid={`addAddress-${id}`}
                   onClick={() => openNewAddressModal(newAddress, id)}
                 >
                   <AddIcon />
--- a/src/components/Tool/FixMailingAddresses/Contact.tsx
+++ b/src/components/Tool/FixMailingAddresses/Contact.tsx
@@ -20,6 +20,7 @@ import clsx from 'clsx';
 import { DateTime } from 'luxon';
 import { useTranslation } from 'react-i18next';
 import { makeStyles } from 'tss-react/mui';
+import { editableSources } from 'src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/EditContactAddressModal/EditContactAddressModal';
 import {
   AddButton,
   AddIcon,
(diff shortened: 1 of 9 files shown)


caleballdrin:
@dr-bizz The code looks good! Now that you have the add address button you can remove the "MPDX" text. Could you also align the button with the address boxes?
Screenshot 2024-06-13 at 5 02 53 PM

caleballdrin:
I also noticed that Tnt Imported addresses have a lock next to them. They should show an edit pen icon because they are editable.

dr-bizz:
> I also noticed that Tnt Imported addresses have a lock next to them. They should show an edit pen icon because they are editable.

I don't think they should be editable, as they weren't before on the old MPDX. I think this is due to ensuring TNT is the only place people can update their info if it's imported from TNT. 
Only is the source is MPDX, can it be edited on MPDX.

caleballdrin:
> I don't think they should be editable, as they weren't before on the old MPDX. I think this is due to ensuring TNT is the only place people can update their info if it's imported from TNT. Only is the source is MPDX, can it be edited on MPDX.

@dr-bizz They are editable in the old MPDX, it just also incorrectly shows a lock. There is no automatic update from Tnt like there is for Google and Siebel. Tnt import is more similar to a csv import where it is a one-time upload of data and then users will want to be able to edit that data in MPDX rather than going back to the csv/tnt file to re-import.

dr-bizz:
> > I don't think they should be editable, as they weren't before on the old MPDX. I think this is due to ensuring TNT is the only place people can update their info if it's imported from TNT. Only is the source is MPDX, can it be edited on MPDX.
> 
> @dr-bizz Tnt addresses **are** editable in the old MPDX, it just also incorrectly shows a lock. There is no automatic update from Tnt like there is for Google and Siebel. Tnt import is more similar to a csv import where it is a one-time upload of data and then users will want to be able to edit that data in MPDX rather than going back to the csv/tnt file to re-import.

@caleballdrin Fixed

caleballdrin:
@dr-bizz I'm pretty sure Tnt is different than DataServer. If you look at Pedros account you can see my contact was imported from Tnt.

caleballdrin (approved):
Nice work!

src/components/Tool/FixMailingAddresses/Contact.tsx
```diff
@@ -233,7 +235,9 @@ const Contact: React.FC<Props> = ({
                       <strong>{t('Source')}: </strong>
                     </Typography>
                   </Hidden>
-                  <Typography display="inline">MPDX</Typography>
+                  <Typography display="inline">
```
caleballdrin, line 243: @dr-bizz this should be removed altogether, not changed to appName. "MPDX" is still showing up across from the "Add Address" button.

caleballdrin (changes requested):

caleballdrin (approved):

PR #963, opened 2024-06-19
Amplify Preview - Preview to be named after PR instead of branch name

dr-bizz:
In this PR, I changed the way we name Amplify previews so that they are called after their PR ID instead of their branch.

Applying this fix will allow us to use previews again, as I've made the amplified build hit the production GraphQL server when building if the preview is prefixed with `pr-`.

This also lets us know what preview belongs to what PR and if we can delete it. 

An idea we should roll out in a different PR is removing the preview once the PR is merged into `main`.

## Description

## Checklist:


wrandall22 (changes requested):
run_lighthouse.yml also needs to be updated

PR #964, opened 2024-06-20
MPDX-7999 - Fix Amplify previews build

dr-bizz:
## Description
Amplify Previews are currently not building and deploying due to an error when building.

The error is because we're hitting the staging API when building the GraphQL and since our graphQL doesn't have the Task Phases changes in it, it causes an error.

 
Fix: Hit the production API when creating the GraphQL, then in the app call the staging API

## Checklist:

### DIFF
--- a/amplify.yml
+++ b/amplify.yml
@@ -6,14 +6,12 @@ frontend:
         - nvm install 18.13
         - nvm use 18.13
         - |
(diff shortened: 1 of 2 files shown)


dr-bizz:
Yes, plus I had to change the If statement some.

PR #965, opened 2024-06-25
Fix Mailing Address - Confirm buttons

dr-bizz:
## Description
In this PR, I added the functionality to confirm a single contact or bulk contacts based on their source.
The functionality should be the same as MPDX Web, but I've added many more alerts to inform the user about what is happening.

When doing the bulk, we don't have a bulkAddresses GraphQl mutation, so I'm having to loop over each single Address mutation many times, which is why I put it into a promise.all() function.

The GraphQL code currently has an error; however, I can't do much about it. I have requested Andrew update the mutation to allow me to pass `primary mailing address`, but until then, it will cause an error.

I've added tests, but I'm unable to test the successful requests due to the GraphQL error.

## Checklist:


caleballdrin (changes requested):
Small UI change that I missed on the other PR.

caleballdrin (approved):
Looks good. Thanks for the example of the bulk mutation

PR #958, opened 2024-06-17
Mpdx 7941 fix email address email validation

cjpastika:
## Description

Jira ticket: Mpdx-7941
-created an email validation field
-tweaked the submit button to have isValid properties
-other various UI changes to make the form and submit button fit better on the page
-Before these changes the pages functionality was extremely limited but I had to take out even more
-This functionality will be added back since I will have to fix them in my next few Jira tickets

## Checklist:

- [ X] I have given my PR a title with the format "MPDX-(JIRA#) (summary sentence max 80 chars)"
- [ X] I have applied the appropriate labels. (_Add the label "On Staging" to get the branch automatically merged into staging._)
- [ X] I have requested a review from another person on the project

PR #973, opened 2024-07-05
MPDX-7946 use contact mutation when adding email

cjpastika:
## Description

MPDX 7946
Related to MPDX 7941

Will add a contact mutation when users click the + button after entering a valid email

## Checklist:


dr-bizz:
It looks like you've done the same thing here: https://github.com/CruGlobal/mpdx-react/pull/958.

You need to merge main into this branch

src/components/Tool/FixEmailAddresses/FixEmailAddressPerson.tsx
```diff
@@ -132,7 +132,7 @@ export const FixEmailAddressPerson: React.FC<FixEmailAddressPersonProps> = ({
         personId: personId,
         isPrimary: email.primary,
       })) || [],
-    [emailAddresses],
+    [emailAddresses, dataState],
```
dr-bizz, line 135: Why do you need `dataState` as a dependency?

src/components/Tool/FixEmailAddresses/FixEmailAddresses.tsx
dr-bizz, line 1: I wonder if you need to use `dataState` at all. I also don't want to slow you down, so maybe we come back and fix this up afterwards.

src/components/Tool/FixEmailAddresses/FixEmailAddresses.test.tsx
```diff
@@ -115,28 +124,147 @@ describe('FixPhoneNumbers-Home', () => {
...
+    interface AddEmailAddressProps {
+      postSaveResponse: object;
+      emailAddressNodes: object[];
+      elementToWaitFor: string;
+      textFieldIndex: number;
+      addButtonId: string;
+      cache: InMemoryCache;
+    }
+
+    const addEmailAddress = async ({
```
dr-bizz, line 137: We shouldn.t have a component in a test like this. Ifn we want to test addEmailAddress we should import it rather than recreate it.

src/components/Tool/FixEmailAddresses/FixEmailAddresses.test.tsx
```diff
@@ -115,28 +124,147 @@ describe('FixPhoneNumbers-Home', () => {
...
+      });
+
+      const textFieldNew =
+        getAllByLabelText('New Email Address')[textFieldIndex];
+      userEvent.type(textFieldNew, newEmail.email);
+      const addButton = getByTestId(addButtonId);
+      userEvent.click(addButton);
+    };
+
+    it('should add an email address to the first person', async () => {
```
dr-bizz, line 195: We should make the test like how the user would interact with the component. Such as types in new email and clicks "add", this should fire a GraphQL Mutation request we can test.

dr-bizz (changes requested):
Great work, there were some changes I thought needed to happen before we role this PR liv, which I've commented below. Feel free to disagree with my thoughts.

src/components/Tool/FixEmailAddresses/FixEmailAddresses.test.tsx
wrandall22: That is what we are doing here, though I never found good ways to test what data got sent in the mutation, so I went with what changed on the screen as most other tests I've found are doing.

src/components/Tool/FixEmailAddresses/FixEmailAddressPerson.tsx
wrandall22: In my testing, `emailAddresses` did not update when changing data on it and `useMemo` never got called, but it did get called when `dataState` changed.

src/components/Tool/FixEmailAddresses/FixEmailAddresses.tsx
dr-bizz: We can leave this for now and come fix it at a later date.

wrandall22: My testing seemed to indicate it is necessary, but I could have been missing something. Basically, state changes to `person` did not re-call `useMemo`.

src/components/Tool/FixEmailAddresses/FixEmailAddresses.test.tsx
dr-bizz: I have updated the code to add some more tests and clean up the PR.

dr-bizz (approved):

PR #980, opened 2024-07-19
MPDX-8037-Adds pledge currency options to select dropdown.

wjames111:
## Description

- Tools/Fix Commitment Info needs currencies added to select field.
- Jira ticket [#8037](https://jira.cru.org/browse/MPDX-8037) 

## Checklist:

### DIFF
--- a/src/components/Tool/FixCommitmentInfo/Contact.tsx
+++ b/src/components/Tool/FixCommitmentInfo/Contact.tsx
@@ -6,18 +6,24 @@ import {
   Avatar,
   Box,
   Button,
+  FormHelperText,
   Grid,
   IconButton,
+  InputLabel,
   Link,
+  MenuItem,
   NativeSelect,
+  Select,
   TextField,
   Typography,
 } from '@mui/material';
 import { useTranslation } from 'react-i18next';
 import { makeStyles } from 'tss-react/mui';
 import { SetContactFocus } from 'pages/accountLists/[accountListId]/tools/useToolsHelper';
+import { useLoadConstantsQuery } from 'src/components/Constants/LoadConstants.generated';
 import { FilterOption } from 'src/graphql/types.generated';
 import { useAccountListId } from 'src/hooks/useAccountListId';
+import { getPledgeCurrencyOptions } from 'src/lib/getCurrencyOptions';
 import theme from '../../../theme';
 import { StyledInput } from '../StyledInput';
 import { frequencies } from './InputOptions/Frequencies';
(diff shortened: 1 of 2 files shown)


src/components/Tool/FixCommitmentInfo/Contact.tsx
```diff
@@ -127,6 +133,7 @@ const Contact: React.FC<Props> = ({
     amount: amount,
     frequencyValue: frequencyValue,
   });
+  const { data: constants } = useLoadConstantsQuery();
```
dr-bizz, line 136: You could use the react hook we created to make this easier to code and manage.

```
import { useApiConstants } from 'src/components/Constants/UseApiConstants';

const constants = useApiConstants();

const pledgeCurrencies = constants?.pledgeCurrencies;
```

src/components/Tool/FixCommitmentInfo/Contact.tsx
```diff
@@ -193,16 +200,31 @@ const Contact: React.FC<Props> = ({
...
+                  }}
+                >
+                  <MenuItem value={'Currency'} disabled>
+                    {t('Currency')}
+                  </MenuItem>
+                  {constants?.constant?.pledgeCurrency &&
+                    getPledgeCurrencyOptions(
+                      constants?.constant?.pledgeCurrency,
+                      'short',
+                    )}
```
dr-bizz, line 222: If you did the change I mentioned in the code above, you can change this to the below.
```suggestion
{getPledgeCurrencyOptions(pledgeCurrencies, 'short')}
```

dr-bizz: This makes the `src/lib/getCurrencyOptions.tsx` do the optional chaining for us

src/lib/getCurrencyOptions.tsx
```diff
@@ -3,14 +3,17 @@ import { Currency } from 'src/graphql/types.generated';
 
 export const getPledgeCurrencyOptions = (
   pledgeCurrencies: Currency[] | undefined | null,
+  format = 'long',
```
dr-bizz, line 6: To make this function safer, we could make format into a enum, to prevent values which aren't allowed being passed in.

dr-bizz: From memory it could look like this.

export enum pleadgeCurrencyOptionFormatEnum = {
  Long = 'long,
  Short = 'short'
  ...other formats
}


```suggestion
  format = pleadgeCurrencyOptionFormatEnum.long,
```

dr-bizz: I think this will work, but you might need to create a type for the function if it doesn't work. It should though.

dr-bizz (changes requested):
This is looking great! I added some comments on the code, but overall this is looking good.

PR #981, opened 2024-07-24
MPDX-7418 - Fix Email Addresses - Confirm Button

wrandall22:
## Description

[Jira](https://jira.cru.org/browse/MPDX-7418)
Implements the single Confirm button for Fix Email Addresses. While testing, I found that the primary switching was not working when all emails were primary, so I fixed that as well.

## Checklist:


src/components/Tool/FixEmailAddresses/FixEmailAddressPerson/FixEmailAddressPerson.tsx
```diff
@@ -287,7 +296,10 @@ export const FixEmailAddressPerson: React.FC<FixEmailAddressPersonProps> = ({
                               </Typography>
                             </Box>
                             {email.isPrimary ? (
-                              <Box data-testid={`starIcon-${id}-${index}`}>
+                              <Box
+                                data-testid={`starIcon-${id}-${index}`}
+                                onClick={() => handleChangePrimary(id, index)}
+                              >
```
dr-bizz, line 302: Currently you're not making a GraphQL call for just updating the primary? Will that come in a later PR?

If not, We should note that we should come back and do that. Curreently there isn't a graphQL endpoint for this, s it will need to be made or made via proxy.

dr-bizz (commented):
I made some small edits, but it all looks great

src/components/Tool/FixEmailAddresses/FixEmailAddressPerson/FixEmailAddressPerson.tsx
wrandall22: This gets handled in the confirm itself, `primary` is part of the payload that gets mutated. If I'm reading the Angular code correctly, it does this as well. The changes in the UI to primary are in memory until clicking the Confirm button.

dr-bizz (approved):

PR #982, opened 2024-07-25
MPDX-7416 - Fix Email Addresses - Confirm All Button

wrandall22:
## Description

This PR implements the Confirm All button on fix email addresses.

[Jira](https://jira.cru.org/browse/MPDX-7416)

## Checklist:


src/components/Tool/FixEmailAddresses/FixEmailAddresses.tsx
```diff
@@ -224,6 +263,45 @@ export const FixEmailAddresses: React.FC<FixEmailAddressesProps> = ({
     });
   };
 
+  const handleBulkConfirm = async () => {
+    const dataToSend = determineBulkDataToSend(dataState, defaultSource ?? '');
```
dr-bizz, line 267:
```suggestion
    const dataToSend = determineBulkDataToSend(
      dataState,
      defaultSource === appName ? 'MPDX' : defaultSource ?? '',
    );
```

src/components/Tool/FixEmailAddresses/FixEmailAddresses.tsx
```diff
@@ -103,18 +109,51 @@ interface FixEmailAddressesProps {
   setContactFocus: SetContactFocus;
 }
```
dr-bizz, line 111:
```suggestion
// TODO: We should make all the determineBulkDataToSend functions mostly the same so we work out who th default source the same.
```

src/components/Tool/FixEmailAddresses/FixEmailAddresses.tsx
```diff
@@ -224,6 +263,45 @@ export const FixEmailAddresses: React.FC<FixEmailAddressesProps> = ({
...
+        variables: {
+          input: {
+            accountListId,
+            attributes: dataToSend,
+          },
+        },
+        update: (cache) => {
+          data?.people.nodes.forEach((person) => {
+            cache.evict({ id: `Person:${person.id}` });
+          });
```
dr-bizz, line 280:
```suggestion
          data?.people.nodes.forEach((person) => {
            cache.evict({ id: `Person:${person.id}` });
          });
          cache.gc();
```

src/components/Tool/FixEmailAddresses/FixEmailAddresses.tsx
```diff
@@ -256,10 +334,12 @@ export const FixEmailAddresses: React.FC<FixEmailAddressesProps> = ({
                         handleSourceChange(event)
                       }
                     >
-                      <option value="MPDX">MPDX</option>
+                      <option value={appName}>{appName}</option>
                       <option value="DataServer">DataServer</option>
```
dr-bizz, line 338: Do we want DataServer to be translated? Or because it's a name, it should stay DataServer.

dr-bizz (approved):
This looks great. I merged `main` into this PR, so you should check that I didn't miss anything. I also fixed a few things that I broke during the merge

src/components/Tool/FixEmailAddresses/FixEmailAddresses.tsx
wrandall22: Is the source still going to be `MPDX` when it is a non-Cru org? Seems like that is going to cause problems.

src/components/Tool/FixEmailAddresses/FixEmailAddresses.tsx
wrandall22: This actually shouldn't be here at all. In the Angular version it is looping through an array with the existing sources plus `MPDX`:

```
union(
  flatMap((person) => map('source', person.email_addresses), data),
  ['MPDX'],
).sort();
```

wrandall22: This will need to change for mailing addresses too.

wrandall22: To answer the question of translation, they are not translated in the Angular app, so I think we can stick to the same here.

src/components/Tool/FixEmailAddresses/FixEmailAddresses.tsx
```diff
@@ -158,33 +159,38 @@ export const FixEmailAddresses: React.FC<FixEmailAddressesProps> = ({
...
+              emailAddresses: person.emailAddresses.nodes.map(
+                (emailAddress) => {
+                  existingSources.add(emailAddress.source);
+                  return {
+                    id: emailAddress.id,
+                    primary: emailAddress.primary,
+                    updatedAt: emailAddress.updatedAt,
+                    source: emailAddress.source,
+                    email: emailAddress.email,
+                  };
```
dr-bizz, line 183: I guess if the email only contained these fields you could do this instead, but not required.

```suggestion
                  return emailAddress;
```

dr-bizz (approved):
This looks great

src/components/Tool/FixEmailAddresses/FixEmailAddresses.tsx
wrandall22: I see what the mailing address logic is doing now and will update this code to do the same to catch any bad data.

PR #988, opened 2024-08-01
MPDX-7420 - Fix Email Addresses - add pagination

wrandall22:
## Description

[Jira](https://jira.cru.org/browse/MPDX-7420)

This PR adds an infinite list integration with the Fix Email Addresses tool. It does some improvements to the `InfiniteList` component to remove the flicker and allow for removing the gray background when we don't want it.

## Checklist:

### DIFF
--- a/pages/accountLists/[accountListId]/tools/fixEmailAddresses/[[...contactId]].page.test.tsx
+++ b/pages/accountLists/[accountListId]/tools/fixEmailAddresses/[[...contactId]].page.test.tsx
@@ -6,6 +6,7 @@ import { ErgonoMockShape } from 'graphql-ergonomock';
 import { getSession } from 'next-auth/react';
 import { SnackbarProvider } from 'notistack';
 import { I18nextProvider } from 'react-i18next';
+import { VirtuosoMockContext } from 'react-virtuoso';
 import TestRouter from '__tests__/util/TestRouter';
 import { GqlMockedProvider } from '__tests__/util/graphqlMocking';
 import { mockInvalidEmailAddressesResponse } from 'src/components/Tool/FixEmailAddresses/FixEmailAddressesMocks';
@@ -51,23 +52,27 @@ const Components = ({
 }) => (
   <ThemeProvider theme={theme}>
     <TestRouter>
-      <GqlMockedProvider<{
-        InvalidAddresses: InvalidAddressesQuery;
-      }>
-        mocks={{
-          GetInvalidEmailAddresses: {
-            people: {
-              nodes: mockNodes,
-            },
-          },
-        }}
+      <VirtuosoMockContext.Provider
+        value={{ viewportHeight: 1000, itemHeight: 100 }}
       >
-        <I18nextProvider i18n={i18n}>
-          <SnackbarProvider>
-            <FixEmailAddressesPage />
-          </SnackbarProvider>
-        </I18nextProvider>
-      </GqlMockedProvider>
+        <GqlMockedProvider<{
+          InvalidAddresses: InvalidAddressesQuery;
+        }>
+          mocks={{
+            GetInvalidEmailAddresses: {
+              people: {
+                nodes: mockNodes,
+              },
+            },
+          }}
+        >
+          <I18nextProvider i18n={i18n}>
+            <SnackbarProvider>
+              <FixEmailAddressesPage />
+            </SnackbarProvider>
+          </I18nextProvider>
+        </GqlMockedProvider>
+      </VirtuosoMockContext.Provider>
     </TestRouter>
   </ThemeProvider>
 );
@@ -85,12 +90,12 @@ describe('FixEmailAddressesPage', () => {
   });
 
   it('should open up contact details', async () => {
-    const { getByText, queryByTestId } = render(<Components />);
+    const { findByText, queryByTestId } = render(<Components />);
     await waitFor(() =>
       expect(queryByTestId('loading')).not.toBeInTheDocument(),
     );
 
-    const contactName = getByText('Test Contact');
+    const contactName = await findByText('Test Contact');
 
     expect(contactName).toBeInTheDocument();
     userEvent.click(contactName);
--- a/src/components/InfiniteList/InfiniteList.tsx
+++ b/src/components/InfiniteList/InfiniteList.tsx
@@ -109,11 +109,6 @@ export const InfiniteList = <T, C>({
       ScrollSeekPlaceholder: SkeletonItem,
       ...props.components,
     },
-    scrollSeekConfiguration: {
-      enter: (velocity) => Math.abs(velocity) > 200,
-      exit: (velocity) => Math.abs(velocity) < 10,
-      ...props.scrollSeekConfiguration,
-    },
     overscan: 2000,
   };
 
--- a/src/components/InfiniteList/InfiniteList.tsx
+++ b/src/components/InfiniteList/InfiniteList.tsx
@@ -23,7 +23,7 @@ const ListContainer: React.ComponentType<ListProps> = React.forwardRef(
   ),
 );
 
-const ItemWithBorders = styled(ListItem, {
+export const ItemWithBorders = styled(ListItem, {
   shouldForwardProp: (prop) => prop !== 'disableHover',
 })<{ disableHover?: boolean }>(({ disableHover }) => ({
   padding: `${padding}px`,
@@ -67,6 +67,7 @@ const GroupLabel = styled(Typography)(({ theme }) => ({
 export interface InfiniteListProps<T, C> {
   loading: boolean;
   EmptyPlaceholder?: ReactElement | null;
+  ItemOverride?: React.ComponentType<ItemProps> | null;
   itemContent: ItemContent<T, C>;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   context?: any;
@@ -78,6 +79,7 @@ export const InfiniteList = <T, C>({
   loading,
   data = [],
   EmptyPlaceholder = null,
+  ItemOverride = null,
   context,
   groupBy,
   itemContent,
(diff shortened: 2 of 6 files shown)


wrandall22:
FWIW, I tried upgrading `react-virtuoso` to `4.10.0`, but I still get a flicker.

src/components/InfiniteList/InfiniteList.tsx
```diff
@@ -103,15 +105,10 @@ export const InfiniteList = <T, C>({
       Footer: loading ? Loading : undefined,
       EmptyPlaceholder: loading ? undefined : () => EmptyPlaceholder,
       List: ListContainer,
-      Item,
+      Item: ItemOverride ?? Item,
       ScrollSeekPlaceholder: SkeletonItem,
       ...props.components,
     },
-    scrollSeekConfiguration: {
```
wrandall22, line 110: This was originally added [here](https://github.com/CruGlobal/mpdx-react/commit/69c74d8c786b0f3b6dad061a9f866f2c09fd56de#diff-ddcb4d4a5c2069166252f62cd5c7243f3c6b1e1e8f3e9bac5bdab0831b895404), but no reason was given as to why.

src/components/Tool/FixEmailAddresses/FixEmailAddresses.tsx
```diff
@@ -360,20 +370,47 @@ export const FixEmailAddresses: React.FC<FixEmailAddressesProps> = ({
...
+                  </Grid>
+                )}
+                endReached={() =>
+                  data.people.pageInfo.hasNextPage &&
+                  fetchMore({
+                    variables: { after: data.people.pageInfo.endCursor },
+                  })
+                }
+                EmptyPlaceholder={<NoData tool="fixEmailAddresses" />}
+                style={{
```
canac, line 404: I'm seeing a tiny amount of double scrolling. I think this calc may need to be adjusted to make the list slightly smaller.

https://github.com/user-attachments/assets/275e954c-a9d6-4a41-a5f3-0e57861bd4a1

canac (changes requested):
Let's remove the `!loading` check so that when a new batch of people are being fetched, the entire list doesn't disappear.

We also need to configure the cache policy for `people` so that new results are merged into the existing array.

Also, I noticed that the "Showing x of x" text is incorrect.

Since all those suggestions are on unchanged lines, I can't add comments in GitHub, so here's a diff of the changes I'm suggesting.

```patch
diff --git a/src/components/Tool/FixEmailAddresses/FixEmailAddresses.tsx b/src/components/Tool/FixEmailAddresses/FixEmailAddresses.tsx
index 4061facbb..2589f9de2 100644
--- a/src/components/Tool/FixEmailAddresses/FixEmailAddresses.tsx
+++ b/src/components/Tool/FixEmailAddresses/FixEmailAddresses.tsx
@@ -320,7 +320,7 @@ export const FixEmailAddresses: React.FC<FixEmailAddressesProps> = ({
 
   return (
     <Container>
-      {!loading && data && dataState ? (
+      {data && dataState ? (
         <FixEmailAddressesWrapper container>
           <Grid item xs={12}>
             <Typography variant="h4">{t('Fix Email Addresses')}</Typography>
@@ -415,9 +415,12 @@ export const FixEmailAddresses: React.FC<FixEmailAddressesProps> = ({
                 <Box width="100%" display="flex" justifyContent="center">
                   <Typography>
                     <Trans
-                      defaults="Showing <bold>{{value}}</bold> of <bold>{{value}}</bold>"
+                      defaults="Showing <bold>{{value}}</bold> of <bold>{{total}}</bold>"
                       shouldUnescape
-                      values={{ value: data.people.nodes.length }}
+                      values={{
+                        value: data.people.nodes.length,
+                        total: data.people.totalCount,
+                      }}
                       components={{ bold: <strong /> }}
                     />
                   </Typography>
diff --git a/src/lib/apollo/cache.ts b/src/lib/apollo/cache.ts
index 307a29f11..2e962a634 100644
--- a/src/lib/apollo/cache.ts
+++ b/src/lib/apollo/cache.ts
@@ -72,6 +72,7 @@ export const createCache = () =>
           contacts: paginationFieldPolicy,
           donations: paginationFieldPolicy,
           financialAccounts: paginationFieldPolicy,
+          people: paginationFieldPolicy,
           tasks: paginationFieldPolicy,
           userNotifications: paginationFieldPolicy,
           // Ignore the input.pageNumber arg so that queries with different page numbers will
```

src/components/InfiniteList/InfiniteList.tsx
canac: I've come across this a couple times and never had the guts to just remove it. Thanks for doing that! I think it makes things seem a lot more janky than it helps.

src/components/Tool/FixEmailAddresses/FixEmailAddresses.tsx
wrandall22: Changing that value can increase the double scroll but not get rid of it entirely it seems. If I hide the scrolling on `overflow-y` on the `<div class="css-1wkgudm">` I can get rid of the extra scroll bar. However, no height changes I've tried in the DOM make it go away.

wrandall22:
> Let's remove the !loading check so that when a new batch of people are being fetched, the entire list doesn't disappear.

This doesn't seem to impact this actually. Changing the values for `increaseViewportBy` makes it smoother. The larger the numbers, the less of the problem there is, but the longer the initial load takes as well.

wrandall22:
> We also need to configure the cache policy for people so that new results are merged into the existing array.

This is causing duplicate rows in the tests and breaking them. Specifically, in `change second email for second person to primary then delete it` the `getByTestId('starIcon-testid2-0')` returns multiple rows.

canac:
> > We also need to configure the cache policy for people so that new results are merged into the existing array.
> 
> This is causing duplicate rows in the tests and breaking them. Specifically, in `change second email for second person to primary then delete it` the `getByTestId('starIcon-testid2-0')` returns multiple rows.

Try updating `defaultGraphQLMock` to:

```ts
const defaultGraphQLMock = {
  GetInvalidEmailAddresses: {
    people: {
      nodes: mockInvalidEmailAddressesResponse,
      pageInfo: {
        hasNextPage: false,
      },
    },
  },
};
```

Without this `hasNextPage` is getting mocked to `true`, then it's loading another page, which will contain the same mocked data.

wrandall22:
That seems odd to me that it would default mock a boolean to `true` instead of `false`

canac:
> That seems odd to me that it would default mock a boolean to `true` instead of `false`

It's technically random with a deterministic seed. So you're happening to get a `true` random value here. But since the test requires `false` to work, it's better to make that explicit.

dr-bizz:
> const defaultGraphQLMock = {
>   GetInvalidEmailAddresses: {
>     people: {
>       nodes: mockInvalidEmailAddressesResponse,
>       pageInfo: {
>         hasNextPage: false,
>       },
>     },
>   },
> };

You could run the first and second queries as `true` and then switch them to `false`. Bill, I have seen you use this elsewhere.


```
const defaultGraphQLMock = () => {
  let queryResponse;
  if (queryResponse < 2) {
    queryResponse = {
      people: {
         nodes: mockInvalidEmailAddressesResponse,
         pageInfo: {
           hasNextPage: true,
         },
       }
    }
   else {
      people: {
         nodes: mockInvalidEmailAddressesResponse,
         pageInfo: {
           hasNextPage: true,
         },
       }
   }
}
```
You can also change the `mockInvalidEmailAddressesResponse` to be different if you don't want it to return the same results.

src/components/Contacts/ContactsList/ContactsList.tsx
```diff
@@ -22,7 +23,11 @@ export const ContactsList: React.FC = () => {
     <InfiniteList
       loading={loading}
       data={data?.contacts?.nodes ?? []}
-      style={{ height: `calc(100vh - ${navBarHeight} - ${headerHeight})` }}
+      style={{
```
canac, line 26: Interesting. I can get to the bottom without this change. In fact, adding those 8px causes double-scrolling for me. What do you think about pulling that change and putting it in a separate PR when we can figure out why it's behaving differently on different screens?

https://github.com/user-attachments/assets/f7ac383c-0dbb-4eb7-9e66-ab14f36a604f

pages/accountLists/[accountListId]/tools/fixEmailAddresses/[[...contactId]].page.test.tsx
```diff
@@ -90,7 +95,7 @@ describe('FixEmailAddressesPage', () => {
       expect(queryByTestId('loading')).not.toBeInTheDocument(),
     );
 
-    const contactName = getByText('Test Contact');
+    const contactName = await waitFor(() => getByText('Test Contact'));
```
Comment (canac) [inline pages/accountLists/[accountListId]/tools/fixEmailAddresses/[[...contactId]].page.test.tsx:98]: This can be simplified to `findByText`.

```suggestion
    const contactName = await findByText('Test Contact');
```

src/components/Tool/FixEmailAddresses/FixEmailAddresses.tsx
canac: Double scrolling issues are the bane of my existence. I can't even reproduce it anymore, so I think this is OK as-is.

canac (approved):
I'm not sure that the last commit with the contacts list helps. I recommended moving it out of this PR, but I'm going to go ahead and approve because the rest looks great!

src/components/Contacts/ContactsList/ContactsList.tsx
wrandall22: Yeah, I'll pull it out of the PR.

wrandall22: For reference, here is what I get on the contact list

https://github.com/user-attachments/assets/27c263d1-5ab1-4120-bf44-66576373fb30

wrandall22:
Getting a CORS issue all the sudden on the preview environment, but don't think that should be related to any changes here.

PR #995, opened 2024-08-06
MPDX-0000 - fix phone numbers - error validation

wjames111:
## Description

## Checklist:

### DIFF
--- a/src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
+++ b/src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
@@ -109,7 +109,7 @@ interface Props {
 }
 
 export interface FormValuesPerson extends PersonInvalidNumberFragment {
-  newPhoneNumber: PersonPhoneNumberFragment | string;
+  newPhoneNumber: string;
   isNewPhoneNumber: boolean;
 }
 
--- a/src/components/Tool/FixPhoneNumbers/helper.test.ts
+++ b/src/components/Tool/FixPhoneNumbers/helper.test.ts
@@ -74,7 +74,7 @@ const testData = [
 ];
 
 describe('FixPhoneNumbers-helper', () => {
-  it('Should return a Contact with all phoneNumbers', async () => {
+  it('Should return a Contact with all phoneNumbers', () => {
     const result = determineBulkDataToSend(testData[0], 'MPDX');
     expect(result).toEqual([
       {
@@ -102,7 +102,7 @@ describe('FixPhoneNumbers-helper', () => {
       },
     ]);
   });
-  it('Should still return a Contact with all phoneNumbers', async () => {
+  it('Should still return a Contact with all phoneNumbers', () => {
     const result = determineBulkDataToSend(testData[1], 'MPDX');
     expect(result).toEqual([
       {
@@ -130,7 +130,7 @@ describe('FixPhoneNumbers-helper', () => {
       },
     ]);
   });
-  it('Should return an empty array', async () => {
+  it('Should return an empty array', () => {
     const result = determineBulkDataToSend(testData[0], 'DataServer');
     expect(result).toEqual([]);
   });
--- a/src/components/Tool/FixPhoneNumbers/helper.ts
+++ b/src/components/Tool/FixPhoneNumbers/helper.ts
@@ -1,7 +1,4 @@
-import {
-  PersonPhoneNumberInput,
-  PersonUpdateInput,
-} from 'src/graphql/types.generated';
+import { PersonUpdateInput } from 'src/graphql/types.generated';
 import { FormValuesPerson } from './FixPhoneNumbers';
 
 export const determineBulkDataToSend = (
@@ -17,15 +14,12 @@ export const determineBulkDataToSend = (
     if (primaryNumber) {
       dataToSend.push({
         id: value.id,
-        phoneNumbers: value.phoneNumbers.nodes.map(
-          (number) =>
-            ({
-              id: number.id,
-              primary: number.id === primaryNumber.id,
-              number: number.number,
-              validValues: true,
-            } as PersonPhoneNumberInput),
-        ),
+        phoneNumbers: value.phoneNumbers.nodes.map((number) => ({
+          id: number.id,
+          primary: number.id === primaryNumber.id,
+          number: number.number,
+          validValues: true,
+        })),
       });
     }
   });


src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -113,21 +99,34 @@ interface PersonPhoneNumbers {
...
+  numberIndex: number;
+  phoneNumber: string;
+}
+
 interface Props {
   accountListId: string;
   setContactFocus: SetContactFocus;
 }
 
+const defaultDeleteModalState = {
```
caleballdrin, line 114: Right now the delete modal is looking pretty funky. It's would also be nice to know how the validation error message looks. Could you merge main into this PR?

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -207,25 +196,9 @@ const FixPhoneNumbers: React.FC<Props> = ({
...
-
-  const handleChangePrimary = (personId: string, numberIndex: number): void => {
-    const temp = { ...dataState };
-    temp[personId].phoneNumbers = temp[personId].phoneNumbers.map(
-      (number, index) => ({
-        ...number,
-        primary: index === numberIndex ? true : false,
-      }),
-    );
+    dataState[personId].phoneNumbers = numbers;
```
caleballdrin, line 201: It is hard to tell how much you changed the delete function in this PR since it looks like you removed it and then added it back in. I did notice that if you delete all of the phone numbers it throws an error because it tries to set an unknown phone number to primary.

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -207,25 +196,9 @@ const FixPhoneNumbers: React.FC<Props> = ({
     handleDeleteModalClose();
   };
 
-  const handleAdd = (personId: string, number: string): void => {
+  const handleChange = (personId: string, numbers: PhoneNumberData[]): void => {
```
caleballdrin, line 199: It looks like an invalid phone number can't be deleted. I think it should be able to be deleted.

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -113,21 +99,34 @@ interface PersonPhoneNumbers {
   toDelete: PersonPhoneNumberInput[];
 }
 
+export interface ModalState {
```
caleballdrin, line 102: @canac could you help me review this PR? I'm getting a little lost with how this is both using Formik and useState.

caleballdrin (changes requested):

src/components/Tool/FixPhoneNumbers/Contact.tsx
```diff
@@ -105,14 +108,8 @@ interface Props {
...
   personId: string;
-  handleChange: (
-    personId: string,
-    numberIndex: number,
-    event: React.ChangeEvent<HTMLInputElement>,
-  ) => void;
-  handleDelete: (personId: string, phoneNumber: number) => void;
-  handleAdd: (personId: string, number: string) => void;
-  handleChangePrimary: (personId: string, numberIndex: number) => void;
+  handleChange: (personId: string, number: PhoneNumberData[]) => void;
```
canac, line 111:
```suggestion
  handleChange: (personId: string, numbers: PhoneNumberData[]) => void;
```

canac:
I'm not sure why the preview environment wasn't being created, so I manually added it. It's still building right now. https://mpdx-0000-error-validation.d3dytjb8adxkk5.amplifyapp.com

canac (commented):
Here's my comments so far. Sorry if they become irrelevant after the merge.

src/components/Tool/FixPhoneNumbers/Contact.tsx
```diff
@@ -155,212 +136,340 @@ const Contact: React.FC<Props> = ({
...
+                              >
+                                <Box
+                                  display="flex"
+                                  justifyContent="flex-start"
+                                  className={clsx(
+                                    classes.responsiveBorder,
+                                    classes.paddingX,
+                                  )}
+                                >
+                                  <Field
```
canac, line 310: Why did you need to change this to a `<Field as={TextField}>`?

canac:
@wjames111 Sounds good. I had already started a review, so I submitted the comments I had already made. Some of them might become irrelevant.

I also noticed two labels (MPDX and DataServer) in `FixPhoneNumbers.tsx` that need to be localized.

canac:
I also see the error message on locked phone numbers, which probably isn't what we want.

Screenshot 2024-08-16 at 11 14 29 AM

src/components/Tool/FixPhoneNumbers/Contact.tsx
```diff
@@ -155,212 +136,340 @@ const Contact: React.FC<Props> = ({
...
+                              justifyContent="space-between"
+                              className={classes.paddingX}
+                            >
+                              <Box>
+                                <Hidden smUp>
+                                  <Typography display="inline">
+                                    <strong>{t('Source')}: </strong>
+                                  </Typography>
+                                </Hidden>
+                                <Typography display="inline">MPDX</Typography>
```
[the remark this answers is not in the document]
wrandall22: We shouldn't be hard-coding MPDX, but rather using `appName`. In the other tools, the list of possible sources comes from the union of app name and the existing sources on the items to fix.

src/components/Tool/FixPhoneNumbers/Contact.tsx
```diff
@@ -155,212 +136,340 @@ const Contact: React.FC<Props> = ({
     // setContactFocus(id);
   };
 
+  const phoneRegExp =
```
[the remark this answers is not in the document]
wrandall22: If in doubt, follow what the Angular app is doing for it.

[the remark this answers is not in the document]
wrandall22: In that case, I like Caleb's proposal of doing the validation the same way we're doing it in the person modal.

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -83,98 +90,78 @@ const useStyles = makeStyles()(() => ({
...
+  phoneNumbers: PersonPhoneNumberFragment[];
 }
 
 interface Props {
   accountListId: string;
   setContactFocus: SetContactFocus;
 }
 
+interface FormValuesPerson extends PersonInvalidNumberFragment {
+  newPhoneNumber: PersonPhoneNumberFragment | Record<string, never> | string;
```
canac, line 115: Why the `Record<string, never>`>

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -83,98 +90,78 @@ const useStyles = makeStyles()(() => ({
...
+  isNewPhoneNumber: boolean;
+}
+
+export interface FormValues {
+  people: FormValuesPerson[] | [];
```
canac, line 120:
```suggestion
  people: FormValuesPerson[];
```

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -189,258 +176,303 @@ const FixPhoneNumbers: React.FC<Props> = ({
...
+        data?.people.nodes.forEach((person: PersonInvalidNumberFragment) => {
+          cache.evict({ id: `Person:${person.id}` });
+        });
       },
-      onError() {
-        enqueueSnackbar(t(`Error updating ${name}'s phone numbers`), {
+      onError: () => {
+        enqueueSnackbar(t(`Error updating phone numbers`), {
           variant: 'error',
           autoHideDuration: 7000,
```
canac, line 225: Is there a good reason to customize the default auto-hide duration?

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -189,258 +176,303 @@ const FixPhoneNumbers: React.FC<Props> = ({
...
           variant: 'success',
           autoHideDuration: 7000,
         });
       },
     });
   };
 
+  const handleDelete = (
+    values: FormValues,
+    setValues: (FormValues) => void,
```
canac, line 284:
```suggestion
    setValues: (values: FormValues) => void,
```

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -189,258 +176,303 @@ const FixPhoneNumbers: React.FC<Props> = ({
...
+    const deleting = temp?.people[
+      deleteModalState.personIndex
+    ].phoneNumbers?.nodes.splice(deleteModalState.numberIndex, 1)[0];
+
+    deleting.destroy = true;
+
+    deleting.primary &&
+      temp?.people[deleteModalState.personIndex]?.phoneNumbers?.nodes.length &&
+      (temp.people[deleteModalState.personIndex].phoneNumbers.nodes[0].primary =
+        true);
```
canac, line 297:
```suggestion
    if (deleting.primary && temp?.people[deleteModalState.personIndex]?.phoneNumbers?.nodes.length) {
      temp.people[deleteModalState.personIndex].phoneNumbers.nodes[0].primary = true;
    }
```

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -189,258 +176,303 @@ const FixPhoneNumbers: React.FC<Props> = ({
...
+                    <Grid container className={classes.outter}>
+                      <Grid item xs={12}>
+                        <Typography variant="h4">
+                          {t('Fix Phone Numbers')}
+                        </Typography>
+                        <Divider className={classes.divider} />
+                      </Grid>
+                      <Grid item xs={12}>
+                        <Box mb={2}>
+                          <Typography fontWeight="fontWeightBold">
```
canac, line 369:
```suggestion
                          <Typography fontWeight="bold">
```

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -189,258 +176,303 @@ const FixPhoneNumbers: React.FC<Props> = ({
...
-            </>
+                  </>
+                );
+              }}
+            </Formik>
           ) : (
             <NoData tool="fixPhoneNumbers" />
           )}
-        </Grid>
+          ;
```
canac, line 471: This semicolon can be removed

canac:
The "Confirm Delete" modal close button overlaps the header text:

Screenshot 2024-08-20 at 4 15 35 PM

src/components/Tool/FixPhoneNumbers/Contact.tsx
```diff
@@ -279,29 +278,65 @@ const Contact: React.FC<Props> = ({
...
+                                                      ) =>
+                                                        number === phoneNumber
+                                                          ? {
+                                                              ...phoneNumber,
+                                                              primary: true,
+                                                            }
+                                                          : {
+                                                              ...number,
+                                                              primary: false,
+                                                            },
```
canac, line 331:
```suggestion
                                                        {
                                                              ...phoneNumber,
                                                              primary: number === phoneNumber,
                                                        }
```

src/components/Tool/FixPhoneNumbers/Contact.tsx
```diff
@@ -109,60 +115,47 @@ const ContactAvatar = styled(Avatar)(() => ({
...
-  avatar: string;
   handleUpdate: (
+    values: FormValues,
     personId: string,
+    personIndex: number,
     name: string,
-    numbers: PhoneNumberData[],
   ) => void;
+  errors: FormikErrors<any>;
+  setValues: (FormValues) => void;
```
canac, line 131:
```suggestion
  setValues: (values: FormValues) => void;
```

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -83,98 +90,78 @@ const useStyles = makeStyles()(() => ({
...
+  const initialValues: FormValues = {
+    people:
+      data?.people?.nodes.map((person) => ({
+        ...person,
+        isNewPhoneNumber: false,
+        newPhoneNumber: '',
+      })) || [],
+  };
 
   const determineBulkDataToSend = (
```
canac, line 151: Does this helper depend on anything inside of the component? If not, can we move it outside of the component? Or even into a separate helpers file?

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.test.tsx
```diff
@@ -218,19 +219,21 @@ describe('FixPhoneNumbers-Home', () => {
   });
 
   it('should hide contact from view', async () => {
-    const { getByTestId, getByText, queryByText } = render(<Components />);
+    const cache = new InMemoryCache();
+    const { getByTestId, getByText } = render(<Components cache={cache} />);
```
canac, line 223: Is a customized cache necessary? It passes for me without it.

src/components/Tool/FixPhoneNumbers/Contact.test.tsx
```diff
@@ -66,29 +68,20 @@ describe('FixPhoneNumbers-Contact', () => {
...
         <TestWrapper>
           <Contact
-            name={testData.name}
-            key={testData.name}
-            personId={testData.id}
-            numbers={testData.numbers}
-            toDelete={[]}
-            handleChange={handleChangeMock}
+            key={testData.id}
+            person={testData}
```
canac, line 77: I don't think the key prop is needed because it's not part of a `.map`.

src/components/Tool/FixPhoneNumbers/Contact.test.tsx
```diff
@@ -99,46 +92,62 @@ describe('FixPhoneNumbers-Contact', () => {
...
+
+    await waitFor(() => {
+      expect(setValuesMock).toHaveBeenCalledWith({
+        people: [
+          {
+            ...testData,
+            phoneNumbers: {
+              nodes: [
+                { ...testData.phoneNumbers.nodes[0], number: '35338958951' },
+                { ...testData.phoneNumbers.nodes[1] },
```
canac, line 144: You shouldn't need to clone this node.

```suggestion
                testData.phoneNumbers.nodes[1],
```

canac (changes requested):
It's looking pretty good. The performance is terrible with my 52 contacts, so we should add infinite scroll at some point, but that could be in a different PR.

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
canac: Another benefit is that it makes it a lot easier to test logic when it's in a helper method instead of inside a component. Can you add some tests after you extract the logic into a separate file?

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
canac: I saw that too after I commented. It doesn't really matter, but it seems good to have consistent auto-hide durations, which is easier to achieve by using the default values.

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
canac: If your function is in `helpers.ts`, the tests for it should go in `helpers.test.ts`.

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
wrandall22: I think it has to do with the fact that you are using the function syntax of `(params): retType => { return x; }` instead of `(params): retType => x`

src/components/Tool/FixPhoneNumbers/helper.test.ts
```diff
@@ -0,0 +1,137 @@
...
+            source: 'MPDX',
+          },
+        ],
+      },
+    },
+  ],
+];
+
+describe('FixPhoneNumbers-helper', () => {
+  it('Should return a Contact with all phoneNumbers', async () => {
```
canac, line 77: Since these tests don't contain an `await` you can remove the `async` from each of them.

```suggestion
  it('Should return a Contact with all phoneNumbers', () => {
```

src/components/Tool/FixPhoneNumbers/helper.ts
```diff
@@ -0,0 +1,33 @@
...
+      dataToSend.push({
+        id: value.id,
+        phoneNumbers: value.phoneNumbers.nodes.map(
+          (number) =>
+            ({
+              id: number.id,
+              primary: number.id === primaryNumber.id,
+              number: number.number,
+              validValues: true,
+            } as PersonPhoneNumberInput),
```
canac, line 27: This `as` type cast tells TypeScript to ignore whatever properties this object has and just treat it like a `PersonPhoneNumberInput`, which I don't think is what we want. I believe you can just remove it.

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -83,364 +87,354 @@ const useStyles = makeStyles()(() => ({
...
+  phoneNumbers: PersonPhoneNumberFragment[];
 }
 
 interface Props {
   accountListId: string;
   setContactFocus: SetContactFocus;
 }
 
+export interface FormValuesPerson extends PersonInvalidNumberFragment {
+  newPhoneNumber: PersonPhoneNumberFragment | string;
```
canac, line 112: I think this type is OK to just be a string, right?

```suggestion
  newPhoneNumber: string;
```

canac (changes requested):
Almost there! Just a couple more suggestions.

Great work refactoring the logic into a separate helper method with its own tests!

canac (approved):
🎉  Thanks for working on this!

PR #998, opened 2024-08-13
[no-Jira] Reapply the autocomplete focus patch to @mui/base

canac:
It was broken by the MUI upgrade in c4d5fa4e03a08be8685998c9386cf94e9d6ed8e2.

## Description

When you hover over autocomplete option and then move the mouse outside of the dropdown, the option was still being highlight. Then if you click outside the dropdown it selects the option, which is confusing to users. I had fixed this behavior before in a patch to @mui/base, but it was inadvertantly broken again by the MUI upgrade in c4d5fa4e03a08be8685998c9386cf94e9d6ed8e2.

### Before

https://github.com/user-attachments/assets/50156814-3f80-4ac8-96b0-f2c210ce0c52

### After

https://github.com/user-attachments/assets/0be0abc8-97eb-48e1-aa1e-1fc6a266b8df

## Checklist:


caleballdrin (approved):
It seems to work well on the preview environment!

package.json
```diff
@@ -183,8 +183,9 @@
     "package-json/got": "^11.8.5",
     "http-cache-semantics": "~4.1.1",
     "cacheable-request": "^10.2.7",
-    "@mui/base@5.0.0-alpha.98": "patch:@mui/base@npm%3A5.0.0-alpha.98#./.yarn/patches/@mui-base-npm-5.0.0-alpha.98-f4d605d753.patch",
-    "@mui/x-data-grid@^5.17.4": "patch:@mui/x-data-grid@npm%3A5.17.4#./.yarn/patches/@mui-x-data-grid-npm-5.17.4-542730a19e.patch"
+    "@mui/x-data-grid@^5.17.4": "patch:@mui/x-data-grid@npm%3A5.17.4#./.yarn/patches/@mui-x-data-grid-npm-5.17.4-542730a19e.patch",
+    "@mui/base@^5.0.0-beta.36": "patch:@mui/base@npm%3A5.0.0-beta.40#./.yarn/patches/@mui-base-npm-5.0.0-beta.40-248417914d.patch",
+    "@mui/base@5.0.0-beta.40": "patch:@mui/base@npm%3A5.0.0-beta.40#./.yarn/patches/@mui-base-npm-5.0.0-beta.40-248417914d.patch"
```
dr-bizz, line 188: Could you explain why you need 2 `@mui/base` resolutions? I see that one is `.36` and the other is `.40`. I assume our different MUI packages use different versions of `@mui/base`.

dr-bizz (approved):
Looks great, I had a question, but nothing more.

package.json
canac: `yarn` creates these resolutions automatically when you commit the patch. I believe you're right. `@mui/x-date-pickers` depends on `^5.0.0-beta.36` and `@mui/material` depends on `5.0.0-beta.40`, so I believe that this makes sure that all versions of `@mui/base` are patched. It also created a resolution for `@mui/lab`, which depends on `5.0.0-alpha.98`, that I must have removed, so I'm going to add that one back too.

canac: > It also created a resolution for `@mui/lab`, which depends on `5.0.0-alpha.98`, that I must have removed, so I'm going to add that one back too.

I forgot that that resolution causes build/test issues, and that's why I removed it.

PR #1001, opened 2024-08-15
[MPDX-8061] Google Import

caleballdrin:
## Description
https://jira.cru.org/secure/RapidBoard.jspa?rapidView=3&view=detail&selectedIssue=MPDX-8061#
- Get additional data from GoogleAccount rest proxy query
- Add Google Import UI

## Checklist:


src/components/Tool/GoogleImport/GoogleImport.tsx
```diff
@@ -0,0 +1,553 @@
...
+    margin: '0 0 10px 15px',
+    '> li': {
+      margin: '5px',
+    },
+  },
+  divider: {
+    margin: theme.spacing(2, 0),
+  },
+  headerTitle: {
+    fontSize: '1rem!important',
```
canac, line 95:
```suggestion
    fontSize: '1rem! important',
```

src/components/Tool/GoogleImport/GoogleImport.tsx
```diff
@@ -0,0 +1,553 @@
...
+    useGetContactTagListQuery({
+      variables: {
+        accountListId,
+      },
+    });
+
+  const { data, loading } = useGoogleContactGroupsQuery();
+  const googleAccounts = data?.googleAccounts;
+
+  const account = useMemo(() => {
```
canac, line 132: Can we rename this to `selectedAccount` and `selectedAccount` to `selectedAccountId`?

src/components/Tool/GoogleImport/GoogleImport.tsx
```diff
@@ -0,0 +1,553 @@
...
+    };
+  }, [account]);
+
+  useEffect(() => {
+    googleAccounts?.length && setSelectedAccount(googleAccounts[0]?.id || '');
+  }, [googleAccounts]);
+
+  const handleCloseModal = () => {
+    setShowModal(false);
+    window.location.href = `${process.env.SITE_URL}/accountLists/${accountListId}/tools`;
```
canac, line 157: You don't need to explicitly set the domain here.

```suggestion
    window.location.href = `/accountLists/${accountListId}/tools`;
```

src/components/Tool/GoogleImport/GoogleImport.tsx
```diff
@@ -0,0 +1,553 @@
...
+    setRedirecting(true);
+  };
+
+  const onSubmit = async (
+    attributes: Attributes,
+    { resetForm },
+  ): Promise<void> => {
+    resetForm();
+    alert('import mutation TODO');
```
canac, line 166: Does this TODO need addressed?

src/components/Tool/GoogleImport/GoogleImport.tsx
```diff
@@ -0,0 +1,553 @@
...
+          {loading && !data && (
+            <LoadingSpinner firstLoad={true} data-testid="LoadingSpinner" />
+          )}
+
+          {!loading && data && (
+            <>
+              {!data?.googleAccounts.length && <NoData tool="googleImport" />}
+              {!!data?.googleAccounts.length && (
+                <>
+                  {data?.googleAccounts.length > 1 && (
```
canac, line 190: This will always be true, right? Since we're already checking `!!data?.googleAccounts.length`?

src/components/Tool/GoogleImport/GoogleImport.tsx
```diff
@@ -0,0 +1,553 @@
...
+                              <Trans
+                                defaults="Importing from <bold>{{email}}</bold>"
+                                shouldUnescape
+                                values={{ email: account?.email }}
+                                components={{ bold: <strong /> }}
+                              />
+                            }
+                            action={
+                              <Button
+                                href={`${process.env.SITE_URL}/accountLists/${accountListId}/settings/integrations`}
```
canac, line 252:
```suggestion
                                href={`/accountLists/${accountListId}/settings/integrations`}
```

src/components/Tool/GoogleImport/GoogleImport.tsx
```diff
@@ -0,0 +1,553 @@
...
+                                            {t('Uncheck All')}
+                                          </Button>
+                                        </ButtonGroup>
+                                      </Grid>
+                                    </Grid>
+                                  </HeaderBox>
+                                  <Box
+                                    sx={{
+                                      marginX: theme.spacing(2),
+                                    }}
```
canac, line 341: Did you know that at least for `Box`, I believe you can write this as `<Box mx={2}>`?

src/components/Tool/GoogleImport/GoogleImport.tsx
```diff
@@ -0,0 +1,553 @@
...
+                                  </RadioGroup>
+                                </FormControl>
+                              </Section>
+                            </Box>
+                            <CardActions>
+                              <Button
+                                variant="contained"
+                                disabled={!isValid || isSubmitting}
+                                onClick={() => {
+                                  importAll
```
canac, line 475: Let's use if/else instead of a ternary since we're not doing anything with the return values.

src/components/Tool/GoogleImport/GoogleImport.tsx
```diff
@@ -0,0 +1,553 @@
...
+                                    ? setShowConfirmAllModal(true)
+                                    : handleSubmit();
+                                }}
+                              >
+                                Import
```
canac, line 480: This label needs localized.

src/components/Tool/GoogleImport/GoogleImport.tsx
```diff
@@ -0,0 +1,553 @@
...
+                            </DialogContent>
+                            <DialogActions>
+                              <Button
+                                variant="contained"
+                                onClick={() => {
+                                  setShowConfirmAllModal(false);
+                                  handleSubmit();
+                                }}
+                              >
+                                Ok
```
canac, line 510: This label needs localized.

src/components/Tool/GoogleImport/GoogleImport.tsx
```diff
@@ -0,0 +1,553 @@
+import React, { ReactElement, useEffect, useMemo, useState } from 'react';
```
canac, line 1: It's probably a good idea to add some tests for this component.

pages/api/graphql-rest.page.ts
```diff
@@ -900,14 +897,11 @@ class MpdxRestApi extends RESTDataSource {
...
   async googleAccounts() {
-    const { data }: { data: GoogleAccountsResponse[] } = await this.get(
-      'user/google_accounts',
-      {
-        sort: 'created_at',
-        include: 'contact_groups',
-      },
-    );
-    return GoogleAccounts(data);
+    const response = await this.get('user/google_accounts', {
```
canac, line 900: `GoogleAccountsResponse` doesn't seem to be used anywhere. Instead of removing it, it would be nice to update it to match the response from the REST API for the sake of documentation.

canac (commented):
Great work! I'll test the actual code tomorrow, but wanted to give you my initial suggestions now in case you want to get a head start on them.

src/components/Tool/GoogleImport/GoogleImport.tsx
caleballdrin: This is checking to see if there is more than 1 and that determines if we show a `Select` so they can choose between accounts.

src/components/Tool/GoogleImport/GoogleImport.tsx
caleballdrin: Ah I forgot about that

src/components/Tool/GoogleImport/GoogleImport.tsx
caleballdrin: Ok. Why would that be bettery than ternary here?

dr-bizz: I don't mind, but I see Caleb C's point. When using an if/else statement to decide which function to run, it does make it easier to read if it is in a if/else statement.

src/components/Tool/GoogleImport/GoogleImport.tsx
```diff
@@ -0,0 +1,553 @@
...
+  },
+}));
+
+const googleImportSchema = yup.object({
+  tagsForAllList: yup.array().of(yup.string()).default([]),
+  override: yup.string().required(),
+  importAll: yup.string().required(),
+  groupTags: yup.object(),
+  groups: yup.array().of(yup.string()).default([]),
+});
```
dr-bizz, line 105: Do you think it's worth adding error messages since all these are required.

src/components/Tool/GoogleImport/GoogleImport.tsx
canac: Sorry, yes you're correct.

canac:
The functionality on this page looks great! Amazing work!

src/components/Tool/GoogleImport/GoogleImport.tsx
caleballdrin: Most of these are radio buttons so they will always have a value. I made the button disabled for the one field they need to submit. I'll also remove .required()

dr-bizz (changes requested):
When there are no contacts to be imported, it looks odd. Could we have an info message stating there are no contacts to be imported?

The old one looked fine as it showed the table, so you could see there were no contacts, but a message would be better.
Screenshot 2024-08-16 at 2 57 20 PM
b58f-2ae5f4246ff4)

pages/accountLists/[accountListId]/tools/googleImport.page.tsx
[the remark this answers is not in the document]
Comment (caleballdrin) [reply pages/accountLists/[accountListId]/tools/googleImport.page.tsx:1]: @dr-bizz I think I'll change all the import pages so the urls will be `tools/import/google` and `tools/import/csv` like the way angular is.

pages/accountLists/[accountListId]/tools/import/google.page.test.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { ThemeProvider } from '@mui/material/styles';
+import { render } from '@testing-library/react';
+import TestWrapper from '__tests__/util/TestWrapper';
+import { useAccountListId } from 'src/hooks/useAccountListId';
+import theme from 'src/theme';
+import GoogleImportPage from './google.page';
+
+jest.mock('src/hooks/useAccountListId');
+jest.mock('src/components/Constants/UseApiConstants');
```
Comment (canac) [inline pages/accountLists/[accountListId]/tools/import/google.page.test.tsx:10]: Our tests normally use a mockRouter instead of mocking `useAccountListId` and use GraphQL mocks instead of mocking `useApiConstants`.

src/components/Tool/GoogleImport/GoogleImport.test.tsx
```diff
@@ -0,0 +1,368 @@
...
+  ...jest.requireActual('notistack'),
+  useSnackbar: () => {
+    return {
+      enqueueSnackbar: mockEnqueue,
+    };
+  },
+}));
+
+jest.mock('src/lib/deserializeJsonApi');
+jest.mock('next-auth/react');
```
canac, line 30: `getSession` from `next-auth/react` is already mocked for all tests in `__tests__/util/setup.ts`, so you don't need to mock it again.

src/components/Tool/GoogleImport/GoogleImport.test.tsx
```diff
@@ -0,0 +1,368 @@
...
+        getAllByRole,
+        queryByText,
+        getByRole,
+      } = render(
+        <TestComponent
+          mocks={{
+            GoogleContactGroups: {
+              ...mockGoogleContactGroupsResponse.GoogleContactGroups,
+            },
+          }}
```
canac, line 88:
```suggestion
          mocks={mockGoogleContactGroupsResponse}
```

src/components/Tool/GoogleImport/GoogleImport.test.tsx
```diff
@@ -0,0 +1,368 @@
...
+        />,
+      );
+      await waitFor(() => {
+        expect(getByText('Account to Import From')).toBeVisible();
+        expect(getAllByText(account1Email)[0]).toBeVisible();
+        expect(getByText(account1GroupName)).toBeVisible();
```
canac, line 94: I always prefer `await findBy...` instead of `await waitFor` + `getBy...` because it's simpler and has less noise. You should only need `waitFor` to wait for a function to be called or wait for an element to be removed from the DOM.

src/components/Tool/GoogleImport/GoogleImport.test.tsx
```diff
@@ -0,0 +1,368 @@
...
+                mockGoogleContactGroupsResponse.GoogleContactGroups
+                  .googleAccounts[2],
+              ],
+            },
+          }}
+        />,
+      );
+      await waitFor(() => {
+        expect(getByText(account3Email)).toBeVisible();
+      });
```
canac, line 155:
```suggestion
      expect(await findByText(account3Email)).toBeVisible();
```

src/components/Tool/GoogleImport/GoogleImport.test.tsx
```diff
@@ -0,0 +1,368 @@
...
+      const { getByText } = render(
+        <TestComponent
+          mocks={{
+            GoogleContactGroups: {
+              googleAccounts: [],
+            },
+          }}
+        />,
+      );
+      await waitFor(() => {
```
canac, line 179: Here's another one that could just be `await findByText`.

src/components/Tool/GoogleImport/GoogleImport.tsx
```diff
@@ -129,41 +120,97 @@ const GoogleImport: React.FC<Props> = ({ accountListId }: Props) => {
...
   useEffect(() => {
-    googleAccounts?.length && setSelectedAccount(googleAccounts[0]?.id || '');
+    googleAccounts?.length && setSelectedAccountId(googleAccounts[0]?.id || '');
   }, [googleAccounts]);
 
   const handleCloseModal = () => {
     setShowModal(false);
-    window.location.href = `${process.env.SITE_URL}/accountLists/${accountListId}/tools`;
     setRedirecting(true);
+    window.location.href = `/accountLists/${accountListId}/tools`;
```
canac, line 153: Sorry, I didn't catch this earlier, but using  `https://nextjs.org/docs/pages/api-reference/functions/use-router#routerpush` is better, at least for navigations within MPDX. Instead of a full-page reload, Next.js is able to more seamlessly transition between pages, keeping the navbar visible without the whole page going white during the reload. I believe it has to load less data over the network too.

src/components/Tool/GoogleImport/GoogleImport.tsx
```diff
@@ -129,41 +120,97 @@ const GoogleImport: React.FC<Props> = ({ accountListId }: Props) => {
...
     attributes: Attributes,
     { resetForm },
   ): Promise<void> => {
+    setShowConfirmAllModal(false);
+    const session = await getSession();
+    const apiToken = session?.user?.apiToken;
+
+    const groupTags = Object.entries(attributes.groupTags).reduce(
+      (result, value: [string, string[]]) => {
+        result[value[0]] = value[1].join(',');
```
canac, line 166: Optional, but here's a place you could use array destructuring. I think it's a little clearer. There are probably even better variable names than `key/value`. Is `groupId` and `tags` accurate?

```suggestion
      (result, [key, value]: [string, string[]]) => {
        result[key] = value.join(',');
```

src/components/Tool/GoogleImport/GoogleImport.tsx
```diff
@@ -129,41 +120,97 @@ const GoogleImport: React.FC<Props> = ({ accountListId }: Props) => {
...
+        headers: {
+          authorization: `Bearer ${apiToken}`,
+          'content-type': 'application/vnd.api+json',
+        },
+      },
+    ).catch((err) => {
+      enqueueSnackbar(t('Import has failed'), {
+        variant: 'error',
+      });
+      throw new Error(err);
```
canac, line 207: You can just rethrow the original error.

```suggestion
      throw err;
```

src/components/Tool/GoogleImport/GoogleImport.tsx
```diff
@@ -408,7 +463,29 @@ const GoogleImport: React.FC<Props> = ({ accountListId }: Props) => {
...
+                                  {t(
+                                    'You have no Google Contact groups/labels',
+                                  )}
+                                </AlertTitle>
+                                <Typography>
+                                  {t(
+                                    "If you'd like to import contacts by group/label, please add labels here: ",
+                                  )}
+                                </Typography>
+                                {
```
canac, line 478: I think these braces can be removed.

canac (changes requested):
I still can't test it, but good work with making the mutation work!

I didn't comment on every `waitFor` + `getBy` in the tests that could be refactored to a `findBy`, so I'd recommend searching through the test file for `waitFor` and replacing all the ones that only contain `getBy` assertions. `waitFor` blocks with `queryBy` assertions or ones that assert a function was called are fine to leave as-is. Hopefully that makes sense.

pages/accountLists/[accountListId]/tools/googleImport.page.tsx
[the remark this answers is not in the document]
Comment (dr-bizz) [reply pages/accountLists/[accountListId]/tools/googleImport.page.tsx:1]: That is fine, but can you update the Google Import button screenshotted above?

dr-bizz (approved):
It's looking great! 

I did find that the link to the Google Import from the initial Tools page still doesn't work. Still goes to wrong URL.

dr-bizz (commented):
I was meant to comment in my last review.

canac (approved):
I'm approving the code, trusting that Daniel tested the actual import process. Great work with this! 🎉 One tool closer to being able to launch the rewrite!

PR #1010, opened 2024-08-27
MPDX-8057 - tools sidebar

wjames111:
## Description

- Add Tools Navigation Menu to tools
- Ticket available [here](https://jira.cru.org/browse/MPDX-8057)

## Checklist:

### DIFF
--- a/src/components/Tool/NavToolList/Item/Item.test.tsx
+++ b/src/components/Tool/NavToolList/Item/Item.test.tsx
@@ -12,24 +12,45 @@ const item = {
   id: 'testItem',
 };
 
+const TestComponent = ({ totalCount = 0 }: { totalCount?: number }) => (
+  <ThemeProvider theme={theme}>
+    <TestWrapper>
+      <Item
+        url={item.url}
+        title={item.title}
+        isSelected={item.isSelected}
+        loading={false}
+        totalCount={totalCount}
+        toolId={item.id}
+      />
+    </TestWrapper>
+  </ThemeProvider>
+);
+
 describe('ToolItem', () => {
   it('default', () => {
-    const { queryByText } = render(
-      <ThemeProvider theme={theme}>
-        <TestWrapper>
-          <Item
-            url={item.url}
-            title={item.title}
-            isSelected={item.isSelected}
-            loading={false}
-            needsAttention={false}
-            totalCount={0}
-            toolId={item.id}
-          />
-        </TestWrapper>
-        ,
-      </ThemeProvider>,
-    );
+    const { queryByText } = render(<TestComponent />);
     expect(queryByText(item.title)).toBeInTheDocument();
+  });
+
+  it('renders count less then 9', () => {
+    const { queryByText } = render(<TestComponent totalCount={8} />);
+    expect(queryByText(item.title)).toBeInTheDocument();
+    expect(queryByText('8')).toBeInTheDocument();
+  });
+
+  it('renders count greater then 9', () => {
+    const { queryByText } = render(<TestComponent totalCount={10} />);
+    expect(queryByText(item.title)).toBeInTheDocument();
+    expect(queryByText('9+')).toBeInTheDocument();
+  });
+
+  it('renders without Badge', () => {
+    const { queryByText, queryByTestId } = render(<TestComponent />);
+    expect(queryByText(item.title)).toBeInTheDocument();
+
+    expect(
+      queryByTestId('fixCommitmentInfo-notifications'),
+    ).not.toBeInTheDocument();
   });
 });
--- a/src/components/Tool/NavToolList/Item/Item.tsx
+++ b/src/components/Tool/NavToolList/Item/Item.tsx
@@ -2,7 +2,6 @@ import NextLink from 'next/link';
 import React, { ReactElement } from 'react';
 import ArrowForwardIos from '@mui/icons-material/ArrowForwardIos';
 import { Box, ListItem, ListItemText, Theme, Typography } from '@mui/material';
-import { useTranslation } from 'react-i18next';
 import { makeStyles } from 'tss-react/mui';
 import { useAccountListId } from 'src/hooks/useAccountListId';
 
@@ -26,7 +25,6 @@ interface Props {
   title: string;
   isSelected: boolean;
   loading: boolean;
-  needsAttention: boolean;
   totalCount: number;
   toolId: string;
 }
@@ -36,13 +34,11 @@ export const Item = ({
   title,
   isSelected,
   loading,
-  needsAttention,
   totalCount,
   toolId,
 }: Props): ReactElement => {
   const { classes } = useStyles();
   const accountListId = useAccountListId();
-  const { t } = useTranslation();
 
   return (
     <NextLink
@@ -56,9 +52,9 @@ export const Item = ({
             variant: 'subtitle1',
             color: 'textPrimary',
           }}
-          primary={t(title)}
+          primary={title}
         />
-        {!loading && needsAttention && (
+        {!loading && !!totalCount && (
           <Box className={classes.notificationBox}>
             <Typography data-testid={`${toolId}-notifications`}>
               {totalCount < 10 ? totalCount : '9+'}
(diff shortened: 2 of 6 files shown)


src/components/Tool/NavToolList/Item/Item.tsx
```diff
@@ -22,29 +25,46 @@ interface Props {
...
-      >
+      <ListItem button selected={isSelected}>
         <ListItemText
+          data-testid={`${toolId}-list-item`}
           primaryTypographyProps={{
             variant: 'subtitle1',
             color: 'textPrimary',
           }}
-          primary={title}
+          primary={t(title)}
```
canac, line 59: Instead of localizing the title here, let's localize it where it is defined, which I believe is `src/components/Tool/Home/ToolList.ts`. With a few rare exceptions, you always want a static string as the argument to `t()`, not an expression or a variable name. That way `yarn extract` can pick up the translation string and make sure it gets translated in OneSky.

src/components/Tool/NavToolList/Item/Item.test.tsx
```diff
@@ -1,20 +1,34 @@
...
+            url={item.url}
+            title={item.title}
+            isSelected={item.isSelected}
+            loading={false}
+            needsAttention={false}
+            totalCount={0}
+            toolId={item.id}
+          />
+        </TestWrapper>
+        ,
```
canac, line 30: You can remove this comma.

src/components/Tool/NavToolList/Item/Item.test.tsx
```diff
@@ -1,20 +1,34 @@
...
 
 const item = {
   url: 'testItem',
   title: 'test title',
   isSelected: false,
+  id: 'testItem',
 };
 
 describe('ToolItem', () => {
   it('default', () => {
```
canac, line 16: Can we add some more tests for the different branches in the component? At least these 3 cases would be great!
* `needsAttention === false`
* `needsAttention === true && totalCount < 9`
* `needsAttention === true && totalCount > 9`

src/components/Tool/NavToolList/NavToolList.tsx
```diff
@@ -1,61 +1,153 @@
...
+const NavToolList = ({ selectedId, isOpen, toggle }: Props): ReactElement => {
   const { classes } = useStyles();
+  const { t } = useTranslation();
+  const accountListId = useAccountListId();
+  const { data, loading } = useGetToolNotificationsQuery({
+    variables: { accountListId: accountListId ?? '' },
+    skip: !accountListId,
+  });
+
+  const toolData: { [key: string]: { totalCount: number } } = {
```
canac, line 75: Is there a reason the values of this map are objects instead of just the `totalCount` number?

src/components/Tool/NavToolList/NavToolList.tsx
```diff
@@ -1,61 +1,153 @@
...
+    [ToolName.FixCommitmentInfo]: data?.[ToolName.FixCommitmentInfo] ?? {
```
canac, line 76: For all of these lookups, I would just use the GraphQL field name instead of the enum.

```suggestion
    [ToolName.FixCommitmentInfo]: data?.fixCommitmentInfo ?? {
```

src/components/Tool/NavToolList/NavToolList.tsx
```diff
@@ -1,61 +1,153 @@
...
+                  </ListItem>
+                  {group.items.map((tool) => {
+                    return (
+                      <Item
+                        key={tool.id}
+                        url={tool.url}
+                        title={tool.tool}
+                        isSelected={selectedId === tool.id}
+                        loading={loading}
+                        needsAttention={
```
canac, line 137: I don't think a separate `needsAttention` prop is needed. In the `Item` component, can't we just skip render the count if `totalCount` is 0?

src/components/Tool/NavToolList/Item/Item.tsx
```diff
@@ -22,29 +25,46 @@ interface Props {
   url: string;
   title: string;
   isSelected: boolean;
+  loading: boolean;
+  needsAttention: boolean;
+  totalCount: number;
```
canac, line 30: Do you think a name like `countNeedingAttention` would be more descriptive? I think it would be great if the prop name told what the count is of.

src/components/Tool/NavToolList/NavToolList.test.tsx
```diff
@@ -12,16 +14,48 @@ const router = {
...
+  <ThemeProvider theme={theme}>
+    <TestRouter router={router}>
+      <GqlMockedProvider>
+        <NavToolList toggle={toggleMock} isOpen={true} />
+      </GqlMockedProvider>
+    </TestRouter>
+  </ThemeProvider>
+);
+
 describe('NavToolList', () => {
```
canac, line 29: I think it would be great to add a test that mocks the number of things that need fixed and make sure that those counts render.

src/components/Tool/NavToolList/NavToolList.tsx
```diff
@@ -1,61 +1,153 @@
...
+import { styled } from '@mui/material/styles';
+import { useTranslation } from 'react-i18next';
 import { makeStyles } from 'tss-react/mui';
+import { useGetToolNotificationsQuery } from 'src/components/Layouts/Primary/TopBar/Items/NavMenu/GetToolNotifcations.generated';
+import { ToolName } from 'src/components/Layouts/Primary/TopBar/Items/NavMenu/NavMenu';
+import { useAccountListId } from 'src/hooks/useAccountListId';
 import { ToolsList } from '../Home/ToolList';
 import { Item } from './Item/Item';
 
 const useStyles = makeStyles()((theme: Theme) => ({
```
canac, line 24: I don't love that we're using `makeStyles` and styled components in the same component. Could you switch to styled components?

src/components/Tool/NavToolList/NavToolList.tsx
```diff
@@ -1,61 +1,153 @@
...
+  root: {
+    overflow: 'hidden',
   },
   li: {
-    borderTop: '1px solid',
-    borderBottom: '1px solid',
-    borderColor: theme.palette.cruGrayDark.main,
+    borderTop: `1px solid ${theme.palette.cruGrayLight.main}`,
+    borderBottom: `1px solid ${theme.palette.cruGrayLight.main}`,
+    borderColor: theme.palette.cruGrayLight.main,
```
canac, line 31: Is this style necessary? It looks like the border color is set by `borderTop` and `borderBottom`.

canac (changes requested):
Looks pretty good! You should just need a few more tests, and I left some refactoring suggestions too.

src/components/Tool/NavToolList/Item/Item.tsx
canac: You can `import i18n from 'src/lib/i18n';` and then use `i18n.t` instead of the `t` from `useTranslation`.

src/components/Tool/NavToolList/NavToolList.tsx
canac: I get that. I'm always looking for ways to simplify things, and this seems like a good candidate if we turn that object with just one field into a number.

src/components/Tool/NavToolList/Item/Item.test.tsx
```diff
@@ -12,24 +12,45 @@ const item = {
   id: 'testItem',
 };
 
+const TestComponent = ({ value = 0 }: { value?: number }) => (
```
canac, line 15: How about giving this prop the same name as the `Item` prop it represents?

```suggestion
const TestComponent = ({ totalCount = 0 }: { totalCount?: number }) => (
```

src/components/Tool/NavToolList/NavToolList.test.tsx
```diff
@@ -4,26 +4,40 @@ import { render, waitFor } from '@testing-library/react';
...
 
 const router = {
   query: { accountListId },
   isReady: true,
 };
 
-const toggleMock = jest.fn();
+const mocks = {
+  GetToolNotifications: {
+    contacts: { totalCount: 2 },
```
canac, line 22: Can you add a comment here something like this?

```
// graphql-ergonomock doesn't handle renamed fields, so we have to mock
// using the name of the field in the GraphQL schema, not the renamed field
```

canac (approved):
I have two small requests, but I'm going to go ahead and approve. You can merge after you address them. Thanks for working on this!

PR #1020, opened 2024-08-29
MPDX-8123 & MPDX-8125 Add Pledges, Commitments and donations flows view

dr-bizz:
## Description
This PR introduces enhancements to the flows view by adding support for pledges, commitments, and donations. The changes include cleaning up the flows view and centralizing logic into a reusable hook.

### Changes
- Fix the `ContactFlowRow` component
- Refactored the list `ContactRow` pledges functionality into a custom hook used in both the list and flows view.
- Added functionality to edit and remove commitments.
- Implemented the logic for determining pledges, commitments, and donations within the new hook.
- Updated the visual indication for overdue dates.
- Added tests for the new `contactFlowsRow` component and hook.

## Checklist:

### DIFF
--- a/src/components/Tool/Appeal/Shared/AmountAndFrequency/AmountAndFrequency.tsx
+++ b/src/components/Tool/Appeal/Shared/AmountAndFrequency/AmountAndFrequency.tsx
@@ -2,21 +2,20 @@ import { UseGetPledgeOrDonation } from 'src/components/Tool/Appeal/Shared/useGet
 import theme from 'src/theme';
 
 export const AmountAndFrequency: React.FC<
-  Omit<UseGetPledgeOrDonation, 'pledgeValues' | 'pledgeDonations'>
+  Pick<UseGetPledgeOrDonation, 'amountAndFrequency' | 'pledgeOverdue'>
 > = ({ amountAndFrequency, pledgeOverdue }) => {
-  const amount = amountAndFrequency?.length ? amountAndFrequency[0] : '';
-  const dateString =
-    amountAndFrequency?.length === 2 ? (
-      <span
-        style={{
-          color: pledgeOverdue ? theme.palette.statusDanger.main : 'inherit',
-        }}
-      >
-        {amountAndFrequency[1]}
-      </span>
-    ) : (
-      ''
-    );
+  const amount = amountAndFrequency?.amount ?? '';
+  const dateString = amountAndFrequency?.dateOrFrequency ? (
+    <span
+      style={{
+        color: pledgeOverdue ? theme.palette.statusDanger.main : 'inherit',
+      }}
+    >
+      {amountAndFrequency?.dateOrFrequency}
+    </span>
+  ) : (
+    ''
+  );
   return (
     <>
       {amount} {dateString}
--- a/src/components/Tool/Appeal/Shared/AmountAndFrequency/AmountAndFrequency.tsx
+++ b/src/components/Tool/Appeal/Shared/AmountAndFrequency/AmountAndFrequency.tsx
@@ -1,0 +1,25 @@
+import { UseGetPledgeOrDonation } from 'src/components/Tool/Appeal/Shared/useGetPledgeOrDonation/useGetPledgeOrDonation';
+import theme from 'src/theme';
+
+export const AmountAndFrequency: React.FC<
+  Omit<UseGetPledgeOrDonation, 'pledgeValues' | 'pledgeDonations'>
+> = ({ amountAndFrequency, pledgeOverdue }) => {
+  const amount = amountAndFrequency?.length ? amountAndFrequency[0] : '';
+  const dateString =
+    amountAndFrequency?.length === 2 ? (
+      <span
+        style={{
+          color: pledgeOverdue ? theme.palette.statusDanger.main : 'inherit',
+        }}
+      >
+        {amountAndFrequency[1]}
+      </span>
+    ) : (
+      ''
+    );
+  return (
+    <>
+      {amount} {dateString}
+    </>
+  );
+};
--- a/src/hooks/useGetPledgeOrDonation.ts
+++ b/src/hooks/useGetPledgeOrDonation.ts
@@ -1,0 +1,158 @@
+import { useEffect, useState } from 'react';
+import { TFunction } from 'i18next';
+import { DateTime } from 'luxon';
+import { useTranslation } from 'react-i18next';
+import { AppealStatusEnum } from 'src/components/Tool/Appeal/AppealsContext/AppealsContext';
+import { AppealContactInfoFragment } from 'src/components/Tool/Appeal/AppealsContext/contacts.generated';
+import { PledgeFrequencyEnum } from 'src/graphql/types.generated';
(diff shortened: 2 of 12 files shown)


canac:
When you delete a commitment, the edit/delete buttons are still there in the flows view. Then if you click edit, I get an error message about the pledge not being found. Is that a bug you want to fix in this PR?

https://github.com/user-attachments/assets/514132ac-e842-4536-bb83-b34dc3cba7f0

src/hooks/useGetPledgeOrDonation.ts
```diff
@@ -0,0 +1,183 @@
...
+  locale: string,
+): UseGetPledgeOrDonation => {
+  const { t } = useTranslation();
+  const [pledgeValues, setPledgeValues] =
+    useState<AppealContactInfoFragment['pledges'][0]>();
+  const [amountAndFrequency, setAmountAndFrequency] = useState<string[]>();
+  const [pledgeDonations, setPledgeDonations] = useState<string[] | null>(null);
+  const [pledgeOverdue, setPledgeOverdue] = useState(false);
+
+  useEffect(() => {
```
canac, line 84: I think that this is another `state/useEffect` combo that could be `useMemo` instead since declarative code is usually easier to reason about than imperative code. Your `useMemo` can return the values for `pledgeValues`, `amountAndFrequency`, `pledgeDonations`, `pledgeOverdue`. It also avoids potential logic bugs since there are some paths through the `useEffect` where not every state variable is being updated.

src/hooks/useGetPledgeOrDonation.ts
```diff
@@ -0,0 +1,183 @@
...
+  return {
+    amount: pledgeOrDonationAmount,
+    dateOrFrequency: pledgeOrDonationDate,
+    pledgeOverdue,
+  };
+};
+
+export interface UseGetPledgeOrDonation {
+  pledgeValues: AppealContactInfoFragment['pledges'][0] | undefined;
+  amountAndFrequency: string[] | undefined;
```
canac, line 66: Thoughts on making this type `[string] | [string, string] | undefined` to enforce the fact that it should only contain one or two values? While we're at it, is there a reason it isn't `{ amount: string, frequency?: string }`? That would be even more self-documenting for future-devs. Sorry I didn't think of that the first time I reviewed this logic.

src/hooks/useGetPledgeOrDonation.ts
```diff
@@ -0,0 +1,183 @@
...
+  pledgeDonations: string[] | null;
+  pledgeOverdue: boolean;
+}
+// The return value doesn't change until `delay` milliseconds have elapsed since the last time `value` changed
+export const useGetPledgeOrDonation = (
+  appealStatus: AppealStatusEnum,
+  contact: AppealContactInfoFragment,
+  appealId: string,
+  locale: string,
```
canac, line 75: I think that `locale` should come from `useLocale`.

src/hooks/useGetPledgeOrDonation.ts
```diff
@@ -0,0 +1,183 @@
...
+  pledgeValues: AppealContactInfoFragment['pledges'][0] | undefined;
+  amountAndFrequency: string[] | undefined;
+  pledgeDonations: string[] | null;
+  pledgeOverdue: boolean;
+}
+// The return value doesn't change until `delay` milliseconds have elapsed since the last time `value` changed
+export const useGetPledgeOrDonation = (
+  appealStatus: AppealStatusEnum,
+  contact: AppealContactInfoFragment,
+  appealId: string,
```
canac, line 74: With so many arguments, what are your thoughts on making this hook take a single object with three fields? That way when code calls this hook it's super self-documenting what each of the arguments are.

src/hooks/useGetPledgeOrDonation.ts
```diff
@@ -0,0 +1,183 @@
+import { useEffect, useState } from 'react';
```
canac, line 1: Nice work extracting this into a separate hook! I like that!

src/components/Tool/Appeal/Shared/AmountAndFrequency/AmountAndFrequency.tsx
```diff
@@ -0,0 +1,25 @@
+import { UseGetPledgeOrDonation } from 'src/hooks/useGetPledgeOrDonation';
+import theme from 'src/theme';
+
+export const AmountAndFrequency: React.FC<
+  Omit<UseGetPledgeOrDonation, 'pledgeValues' | 'pledgeDonations'>
```
canac, line 5: Instead of omitting the props you aren't using, it could be clearer to pick the props you do use.

```suggestion
  Pick<UseGetPledgeOrDonation, 'amountAndFrequency' | 'pledgeOverdue'>
```

src/hooks/useGetPledgeOrDonation.ts
```diff
@@ -0,0 +1,183 @@
...
+  };
+};
+
+export interface UseGetPledgeOrDonation {
+  pledgeValues: AppealContactInfoFragment['pledges'][0] | undefined;
+  amountAndFrequency: string[] | undefined;
+  pledgeDonations: string[] | null;
+  pledgeOverdue: boolean;
+}
+// The return value doesn't change until `delay` milliseconds have elapsed since the last time `value` changed
```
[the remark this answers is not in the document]
dr-bizz: yes

src/hooks/useGetPledgeOrDonation.ts
dr-bizz: Switched to useMemo. I'm not sure If it made better sense, maybe you could help me understand why people prefer declarative code over imperative code?

dr-bizz:
> When you delete a commitment, the edit/delete buttons are still there in the flows view. Then if you click edit, I get an error message about the pledge not being found. Is that a bug you want to fix in this PR?

This fix deserves its own PR.

dr-bizz:
Added it as a task here- [MPDX-8137](https://jira.cru.org/browse/MPDX-8137)

src/hooks/useGetPledgeOrDonation.ts
canac: In my components and hooks, I try to have as few `useState`s as possible and derive any other state that I need from those root states. That way if you want to know the value of a particular sub-state, you just have to look in one place: where it is calculated. On the other hand, if the component/hook uses `useState/useEffect` instead of derived state, I have to look at the `useState` to see its initial value, look at the `useEffect` to see under what conditions it changes and what new values it changes too, and search through the entire component to make sure there aren't any set state calls anywhere else that I need to consider. The derived state version requires me to hold less context in my head when reading the code. As an added benefit, setting state can cause an extra render (https://julvo.com/posts/react/derived-state/).

I'm not sure if I'm explaining it well. Does that make sense? Feel free to push back too.

canac (approved):
Love it!

Thanks for making a new ticket for the bug fix.

src/hooks/useGetPledgeOrDonation.ts
dr-bizz: That makes a lot of sense when reviewing code.  Thank you for explaining. 

I did wonder if the useState caused another render.

PR #1021, opened 2024-08-29
[MPDX-8096] Add default account setup page

canac:
## Description

* Add default account setup page to let the user choose their default account list
* If the user has only one account list, the server will set this account list to the default and redirect to the next step in the tour

[MPDX-8096](https://jira.cru.org/browse/MPDX-8096)

## Checklist:

### DIFF
--- a/pages/setup/account.page.test.tsx
+++ b/pages/setup/account.page.test.tsx
@@ -1,0 +1,134 @@
+import { GetServerSidePropsContext } from 'next';
+import { ApolloClient, NormalizedCacheObject } from '@apollo/client';
+import { render, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { getSession } from 'next-auth/react';
+import { session } from '__tests__/fixtures/session';
+import TestRouter from '__tests__/util/TestRouter';
+import { GqlMockedProvider } from '__tests__/util/graphqlMocking';
+import makeSsrClient from 'src/lib/apollo/ssrClient';
+import AccountPage, { getServerSideProps } from './account.page';
+
+jest.mock('src/lib/apollo/ssrClient');
+
+const push = jest.fn();
+const router = {
+  push,
+};
+
+const context = {
+  req: {},
+} as unknown as GetServerSidePropsContext;
+
+describe('Setup account page', () => {
+  it('renders account options, saves default account, and advances to the next page', async () => {
+    const accountListOptions = {
+      accountLists: {
+        nodes: [1, 2, 3].map((id) => ({
+          id: `account-list-${id}`,
+          name: `Account List ${id}`,
+        })),
+      },
+    };
+
+    const mutationSpy = jest.fn();
+    const { getByRole } = render(
+      <TestRouter router={router}>
+        <GqlMockedProvider onCall={mutationSpy}>
+          <AccountPage accountListOptions={accountListOptions} />
+        </GqlMockedProvider>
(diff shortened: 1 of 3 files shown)


pages/setup/account.page.tsx
```diff
@@ -0,0 +1,146 @@
...
+    >({
+      mutation: UpdateUserDefaultAccountDocument,
+      variables: {
+        input: {
+          attributes: {
+            defaultAccountList: defaultAccountListId,
+          },
+        },
+      },
+    });
```
dr-bizz, line 129: just a though, This might not be needed but if this request errored, it would cause the page would error out, I wonder if it were to error, to show the page and let the user choose. 

Feel free to disagree.

pages/setup/account.page.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const [updateUserDefaultAccount, { loading: isSubmitting }] =
+    useUpdateUserDefaultAccountMutation();
+
+  const [defaultAccountList, setDefaultAccountList] =
+    useState<AccountList | null>(null);
+
+  const handleSave = async () => {
+    if (!defaultAccountList) {
+      return;
+    }
```
dr-bizz, line 44: Show have some kind of error message to tell the user they need to select a default account list

pages/setup/account.page.tsx
```diff
@@ -0,0 +1,146 @@
...
+      variables: {
+        input: {
+          attributes: {
+            defaultAccountList: defaultAccountList.id,
+          },
+        },
+      },
+    });
+    push(`/accountLists/${defaultAccountList.id}/settings/preferences`);
+  };
```
dr-bizz, line 56: Do you think we need a success message as well?

pages/setup/account.page.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+const router = {
+  push,
+};
+
+const context = {
+  req: {},
+} as unknown as GetServerSidePropsContext;
+
+describe('Setup account page', () => {
+  it('renders account options and saves default account renders and button saves and advances to the next page', async () => {
```
dr-bizz, line 24: Can you add a to make sure the mutation doesn't go off when no default account list is selected?

dr-bizz (commented):
I added some comments, but it looks good.]]]]]]]]]

dr-bizz (commented):
> I added some comments, but it looks good.]]]]]]]]] 

`]]]]]]]]]` - My cat sat on my keyboard

pages/setup/account.page.tsx
canac: Personally, I think that navigating to a new page is enough of a hint to the user that clicking the button successfully saved their account list preference. If we show a toast, I think we'll have to wait a few seconds before redirecting to preferences, which feels like a surprising UX.

pages/setup/account.page.tsx
canac: Good point, I should definitely do that. They'll presumably get the same error when they try to pick one on the page, but at least they'll get a helpful message.

pages/setup/account.page.test.tsx
canac: I added a test that the save button is disabled until an account list is selected. Does that seem sufficient to you?

dr-bizz: That's fine

dr-bizz (approved):
Great work!

PR #1026, opened 2024-09-04
Update translations

dr-bizz:
## Description
In this PR, I have run the following scripts, **`yarn extract`** & **`yarn onesky:download`**, which have updated the translation files.
I've also ran **`yarn onesky upload`** to update OneSky with our latest phrases.

## Checklist:


dr-bizz:
That's a good point. I will check

dr-bizz:
@jeff-clay Since we're updating the french language code from `fr-fr` to `fr-FR`. Does that mean we will also need to migrate everyone who has their `locale` set as `fr-fr` to `fr-FR`?

`user.preferences.locale` is where it is stored on GraphQL

personal Preferences setting

src/lib/i18n.ts
```diff
@@ -10,7 +10,18 @@ i18next
   .init({
     nsSeparator: false,
     keySeparator: false,
-    fallbackLng: 'en',
+    // Since these languages have different translation folder names
```
canac, line 13: This code doesn't look wrong, but I don't think it's necessary. Even without this configuration, if the user's language is `fr-fr`, i18next still seems to normalize the casing and requests `fr-FR/translations.json`. Have you checked that it is broken without this configuration?

This configuration still seems useful to keep though since it does more than just fix the casing: `'fr-ca': ['fr-CA', 'fr-FR', 'en'],`

dr-bizz: I thought I did, but let me double check

dr-bizz: You are correct. I think I tested it without restarting my local, which affected it.

canac (approved):
Looks great and works well!

`fr-CA` is a little strange because the translation files seem to only contain English entries. I assume that's an issue inside of OneSky.

dr-bizz:
> Looks great and works well!
> 
> `fr-CA` is a little strange because the translation files seem to only contain English entries. I assume that's an issue inside of OneSky.

Yes it is, so the fallback isn't working, but once we work out OneSky it will work.

PR #1027, opened 2024-09-04
MPDX-8056 - Tools - Initial Page

wjames111:
## Description

- Fix styles
- Fix links
- Add notifications
- [Jira ticket 8056](https://jira.cru.org/browse/MPDX-8056)

## Checklist:


src/components/Tool/Home/ToolsListHome.ts
```diff
@@ -0,0 +1,106 @@
+import {
```
canac, line 1: This looks pretty similar to what's in `ToolsListNav.ts`. Would it be possible to generate this file from `ToolsListNav.ts`? If that's not feasible, that's OK.

canac (approved):
Looks good! My only concern is some duplication in the tools list. It might be unavoidable; I wasn't sure.

src/components/Tool/Home/ToolsListHome.ts
canac: Oh did you duplicate it to get a specific order for the tools? It might be OK then.

caleballdrin (approved):
Looks good to me!

src/components/Tool/Home/ToolsListHome.ts
canac: I think you can keep it.

PR #1029, opened 2024-09-04
[MPDX-8087] Take the user through the tour if their account isn't set up

canac:
## Description

* On the account's list page, do a server-side redirect to force the user to take the tour if their account isn't set up.
* On every page, do a client-side redirect to force the user to take the tour if their account isn't set up.
* Provide `settingUp` context to the entire application. It will be used in a future PR to hide navbar links when in setup mode.

[MPDX-8087](https://jira.cru.org/browse/MPDX-8087)

## Checklist:

### DIFF
--- a/pages/GetAccountLists.graphql
+++ b/pages/GetAccountLists.graphql
@@ -1,4 +1,8 @@
 query GetAccountLists {
+  user {
+    id
+    setup
+  }
   accountLists(first: 50) {
     nodes {
       id
--- a/pages/accountLists.page.tsx
+++ b/pages/accountLists.page.tsx
@@ -45,7 +45,7 @@ export const getServerSideProps = makeGetServerSideProps(async (session) => {
       query: GetAccountListsDocument,
     });
 
-    if (data.user.setup) {
+    if (data.user.setup && process.env.DISABLE_SETUP_TOUR !== 'true') {
       // The user has not finished setting up, so start them on the tour
       return {
         redirect: {
--- a/pages/accountLists.page.tsx
+++ b/pages/accountLists.page.tsx
@@ -45,6 +45,16 @@ export const getServerSideProps = makeGetServerSideProps(async (session) => {
       query: GetAccountListsDocument,
     });
 
+    if (data.user.setup) {
+      // The user has not finished setting up, so start them on the tour
+      return {
+        redirect: {
+          destination: '/setup/start',
+          permanent: false,
+        },
+      };
+    }
+
     if (data.accountLists.nodes.length === 1) {
       return {
         redirect: {
--- a/src/components/Setup/useNextSetupPage.test.tsx
+++ b/src/components/Setup/useNextSetupPage.test.tsx
@@ -1,0 +1,110 @@
+import { ReactElement } from 'react';
+import { waitFor } from '@testing-library/react';
+import { renderHook } from '@testing-library/react-hooks';
+import TestRouter from '__tests__/util/TestRouter';
+import { GqlMockedProvider } from '__tests__/util/graphqlMocking';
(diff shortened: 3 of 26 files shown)


pages/GetAccountLists.graphql
```diff
@@ -1,4 +1,8 @@
 query GetAccountLists {
+  user {
+    id
+    setup
```
caleballdrin, line 4: Did you intentionally use `setup` from `user` rather than getting the `userOptions`? I guess all we need now is a boolean.

canac: So `user.setup` is generated by the server and tells us whether they have no account lists, no organization accounts, or no default account chosen. If it is not null, they need to go through the setup screens I made. After that phase, we set `setup_position` and take the user through the tour screens you made. Does that make sense?

caleballdrin: Ah ok, that makes sense. I didn't realize what `user.setup` was.

canac: It's definitely confusing

caleballdrin (approved):
This looks good!

pages/accountLists.page.tsx
```diff
@@ -45,6 +45,16 @@ export const getServerSideProps = makeGetServerSideProps(async (session) => {
...
 
+    if (data.user.setup) {
+      // The user has not finished setting up, so start them on the tour
+      return {
+        redirect: {
+          destination: '/setup/start',
+          permanent: false,
+        },
+      };
+    }
```
dr-bizz, line 56: As the Tools part isn't built out yet, we should comment this out until we're ready to redirect people to the setup pages.

dr-bizz (commented):
I have a few questions but it looks great.

dr-bizz:
I know this isn't apart of this PR, but on the old setup notifications, we didn't have a skip button.
Also, it looks like the Tools pages tour isn't showing. this might because it is yet to be built

Screenshot 2024-09-05 at 3 22 42 PM

src/components/Setup/useNextSetupPage.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+    const { result } = renderHook(() => useNextSetupPage(), {
+      wrapper: Wrapper,
+    });
+    result.current.next();
+
+    await waitFor(() => expect(push).toHaveBeenCalledWith('/setup/account'));
+  });
+
+  it("when the user's account is set up next should set setup_position and redirect to the preferences page", async () => {
```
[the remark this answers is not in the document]
canac: This one tests when the user has account lists and an organization, i.e. `user.setup` on the server is `null`.

canac:
> I know this isn't apart of this PR, but on the old setup notifications, we didn't have a skip button.

@caleballdrin Do you have capacity to address that?

> Also, it looks like the Tools pages tour isn't showing. this might because it is yet to be built

@dr-bizz Correct. That will be done in MPDX-8095.

pages/accountLists.page.tsx
canac: I disabled the redirect functionality. Although this means that it can't be tested properly until tools go live. Another idea would be to do a feature-flag with an environment variable so the code runs in preview but not production. Let me know if you want further changes.

dr-bizz: I'm good with the feature-flag with an environment variable.

dr-bizz (approved):
I know we spoke about adding the feature-flag with an environment variable. But as I don't want to block you, I will approve this PR.

pages/accountLists.page.tsx
canac: I switched this to use a feature-flag. Set `DISABLE_SETUP_TOUR=true` to disable all setup tour logic.

PR #1033, opened 2024-09-06
MPDX-7955 Update Flows column results after change

dr-bizz:
## Description
In This PR, I make it possible to refetch the flows columns contacts, to ensure we are showing the latest data to the user.

I added the boolean`refreshFlowsView` and setState value`seRefreshFlowsView` to context to allow components to update `refreshFlowsView` after a major change so the columns are prefetched.

I added this to all the places we edit contacts on the flows view and added tests.

## Checklist:

### DIFF
--- a/src/components/Tool/Appeal/AppealsContext/AppealsContext.tsx
+++ b/src/components/Tool/Appeal/AppealsContext/AppealsContext.tsx
@@ -1,5 +1,12 @@
 import { useRouter } from 'next/router';
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import React, {
+  Dispatch,
+  SetStateAction,
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
 import { debounce, omit } from 'lodash';
 import { useContactFiltersQuery } from 'pages/accountLists/[accountListId]/contacts/Contacts.generated';
 import { PageEnum } from 'pages/accountLists/[accountListId]/tools/appeals/AppealsWrapper';
@@ -70,6 +77,8 @@ export interface AppealsType
   committedCountQueryResult: ReturnType<typeof useContactsCountQuery>;
   givenCountQueryResult: ReturnType<typeof useContactsCountQuery>;
   receivedCountQueryResult: ReturnType<typeof useContactsCountQuery>;
+  refreshFlowsView: boolean;
+  seRefreshFlowsView: Dispatch<SetStateAction<boolean>>;
 }
 
 export const AppealsContext = React.createContext<AppealsType | null>(null);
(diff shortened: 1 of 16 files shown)


src/components/Tool/Appeal/AppealsContext/AppealsContext.tsx
```diff
@@ -70,6 +77,8 @@ export interface AppealsType
   committedCountQueryResult: ReturnType<typeof useContactsCountQuery>;
   givenCountQueryResult: ReturnType<typeof useContactsCountQuery>;
   receivedCountQueryResult: ReturnType<typeof useContactsCountQuery>;
+  refreshFlowsView: boolean;
+  seRefreshFlowsView: Dispatch<SetStateAction<boolean>>;
```
canac, line 81: I see a bunch of other occurrences with this typo. I'm not going to comment on them all, but try a global find/replace to fix them all.


```suggestion
  setRefreshFlowsView: Dispatch<SetStateAction<boolean>>;
```

src/components/Tool/Appeal/Flow/ContactFlowColumn/ContactFlowColumn.tsx
```diff
@@ -77,6 +76,14 @@ export const ContactFlowColumn: React.FC<Props> = ({
     skip: !accountListId || !appealStatus,
   });
 
+  useEffect(() => {
```
canac, line 79: Instead of exposing a boolean with context, why don't we expose the `refetch` function and let the child components call `refetch` themselves? That will also avoid two extra renders to set the state to `true` and then back to `false`.

src/components/Tool/Appeal/Modals/AddExcludedContactModal/AddExcludedContactModal.test.tsx
```diff
@@ -163,6 +172,23 @@ describe('AddExcludedContactModal', () => {
     );
   });
 
+  it('should refetch Flows columns contacts after', async () => {
```
canac, line 175: Did you mean to complete this sentence?

canac (commented):
It looks like this works! I'm wondering if it would be better to expose a `refetchFlowsView` function through context instead of a flag.

src/components/Tool/Appeal/Modals/DeleteAppealContact/DeleteAppealContactModal.test.tsx
```diff
@@ -218,4 +223,38 @@ describe('DeleteAppealContactModal', () => {
     });
     expect(handleClose).toHaveBeenCalledTimes(1);
   });
+
+  it('should not refetch the flows columns after mutation', async () => {
```
canac, line 227: Can we add a little more to this title to distinguish it from the next test? Maybe `should not refetch the flows columns after mutation when not in flows view`

canac (approved):
Fixing the `se` --> `set` typo would be great before this merges, but I'm going to go ahead and approve.

src/components/Tool/Appeal/Flow/ContactFlowColumn/ContactFlowColumn.tsx
canac: Sorry didn't realize that `refetch` is in a child component, not the provider component.

PR #1040, opened 2024-09-09
MPDX-8124 Adding select and deselect all functionality onto each column in the flows view

dr-bizz:
## Description
In this pull request, I enabled users to select all and deselect all contacts within a specific column on the flows view. When selecting, contacts are added/removed from the list of previously selected contacts.

This functionality replicates how the old MPDX worked. I had to add 2 more functions to the useMassSelection hook to allow for multiple contact IDs to be passed through for checking or unchecking.

## Checklist:

### DIFF
--- a/src/components/Tool/Appeal/Flow/ContactFlowColumn/ContactFlowColumn.tsx
+++ b/src/components/Tool/Appeal/Flow/ContactFlowColumn/ContactFlowColumn.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useState } from 'react';
+import React, { useMemo, useRef, useState } from 'react';
 import MoreVertIcon from '@mui/icons-material/MoreVert';
 import {
   Box,
@@ -28,6 +28,7 @@ import {
   AppealsContext,
   AppealsType,
 } from 'src/components/Tool/Appeal/AppealsContext/AppealsContext';
+import { useGetIdsForMassSelectionQuery } from 'src/hooks/GetIdsForMassSelection.generated';
 import { appealHeaderInfoHeight } from '../../AppealDetails/AppealHeaderInfo/AppealHeaderInfo';
 import { useContactsQuery } from '../../AppealsContext/contacts.generated';
 import {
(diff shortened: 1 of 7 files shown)


src/components/Tool/Appeal/Flow/ContactFlowColumn/ContactFlowColumn.tsx
```diff
@@ -56,27 +57,52 @@ export const ContactFlowColumn: React.FC<Props> = ({
...
 
+  const contactsFilters = useMemo(
+    () => ({
+      ...sanitizedFilters,
+      ...starredFilter,
+      appeal: [appealId ?? ''],
+      appealStatus,
+      wildcardSearch: searchTerm as string,
+    }),
+    [sanitizedFilters, starredFilter, searchTerm, appealId],
```
canac, line 80: Should we add `appealStatus` to the dependency array? I guess it's not likely to change at runtime.

src/hooks/useMassSelection.ts
```diff
@@ -66,6 +68,16 @@ export const useMassSelection = (
     }
   };
 
+  const selectMultipleIds = (newIds: string[]) => {
+    setSelectionType(ListHeaderCheckBoxState.Checked);
+    setIds([...ids, ...newIds]);
+  };
+
+  const deselectMultipleIds = (IdsToRemove: string[]) => {
```
canac, line 76:
```suggestion
  const deselectMultipleIds = (idsToRemove: string[]) => {
```

src/components/Tool/Appeal/Flow/ContactFlowColumn/ContactFlowColumn.tsx
```diff
@@ -102,13 +128,13 @@ export const ContactFlowColumn: React.FC<Props> = ({
   };
 
   const handleSelectAll = () => {
+    selectMultipleIds(allContactIds);
     setAnchorEl(null);
-    // TODO implement select all
   };
 
   const handleDeselectAll = () => {
+    deselectMultipleIds(allContactIds);
```
canac, line 136: Can we not use `deselectAll` from the `useMassSelection` hook here?

src/hooks/useMassSelection.ts
```diff
@@ -66,6 +68,16 @@ export const useMassSelection = (
     }
   };
 
+  const selectMultipleIds = (newIds: string[]) => {
```
canac, line 71: It's probably good to add tests for this method. And `deselectMultipleIds` if you're not able to remove it and just use `deselectAll`.

canac (commented):
Code works great! I have a question about a possible simplification.

This might not be a good idea, but would it make sense for the `useMassSelection` hook to return a `selectAll` method instead of a `selectMultipleIds` method? Then more of the logic for updating the checkbox state could be contained within that hook instead of in components.

src/components/Tool/Appeal/Flow/ContactFlowColumn/ContactFlowColumn.tsx
dr-bizz: No as it deselects all contact that are selected not just the ones from this column

canac: That makes sense. Thanks!

canac (approved):

src/components/Tool/Appeal/Flow/ContactFlowColumn/ContactFlowColumn.tsx
dr-bizz: The appealStatus won't change

PR #1049, opened 2024-09-10
MPDX-8135-replace-referrals

caleballdrin:
## Description
https://jira.cru.org/secure/RapidBoard.jspa?rapidView=3&view=detail&selectedIssue=MPDX-8135&sprint=1259#

- Rename Referrals to Connections

## Checklist:

### DIFF
--- a/src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsMoreActions/ContactDetailsMoreActions.tsx
+++ b/src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsMoreActions/ContactDetailsMoreActions.tsx
@@ -92,8 +92,6 @@ const ActionPanel = ({
 }: {
   actionContent: AddMenuItem[];
 }): ReactElement => {
-  const { t } = useTranslation();
-
   return (
     <Box display="flex" flexDirection="column" justifyContent="center">
       {actionContent
@@ -106,7 +104,7 @@ const ActionPanel = ({
             onMouseEnter={onMouseEnter}
           >
             {icon}
-            <MenuItemText primary={t(`${text}`)} />
+            <MenuItemText primary={text} />
           </RowContainer>
         ))}
     </Box>
@@ -217,7 +215,7 @@ export const ContactDetailsMoreAcitions: React.FC<
   const actionContent = [
     {
       visibility: true,
-      text: 'Add Referrals',
+      text: t('Add Connections'),
       icon: <PersonIcon />,
       onClick: () => {
         setReferralsModalOpen(true);
@@ -227,7 +225,7 @@ export const ContactDetailsMoreAcitions: React.FC<
     },
     {
       visibility: true,
-      text: 'Add Task',
+      text: t('Add Task'),
       icon: <ListIcon />,
       onClick: () => {
         openTaskModal({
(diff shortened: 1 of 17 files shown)


src/components/Contacts/ContactDetails/ContactDetailsHeader/ContactDetailsMoreActions/ContactDetailsMoreActions.tsx
```diff
@@ -106,7 +104,7 @@ const ActionPanel = ({
             onMouseEnter={onMouseEnter}
           >
             {icon}
-            <MenuItemText primary={t(`${text}`)} />
+            <MenuItemText primary={text} />
```
canac, line 107: Thanks for fixing these unlocalized strings!

src/components/Contacts/ContactDetails/ContactDonationsTab/PartnershipInfo/PartnershipInfo.tsx
```diff
@@ -206,7 +206,7 @@ export const PartnershipInfo: React.FC<PartnershipInfoProp> = ({ contact }) => {
           <ClearIcon />
         </IconContainer>
         <LabelsAndText variant="subtitle1" color="textSecondary">
-          {t('Referred by ')}
+          {t('Connecting Partner ')}
```
canac, line 209: Do you think that this would read better with a colon?

```suggestion
          {t('Connecting Partner: ')}
```

src/components/Shared/Filters/helpers.ts
```diff
@@ -41,4 +41,15 @@ const reverseFiltersOptions = {
   user_ids: 'reverseUserIds',
 };
 
+export const renameFilterNames = (name: string | undefined) => {
```
canac, line 44: I wonder if this should be done at the API-level instead? At least eventually. @dbenton9

canac (approved):
Great work!

src/components/Shared/Filters/helpers.ts
dbenton9: What's happening here @canac?

canac: Sorry for the lack of context. We need to rename "referrals" to "connections" everywhere in MPDX. Some of the places where "referrals" are showing up is in filter options. I think that eventually we'd want to rename the filter titles in the API. But not without some coordination first because that change will affect Angular and presumably the mobile apps. So I'm not asking for any action on your part right now just mentioning that we may need to in the future.

caleballdrin: @dbenton9 It is not urgent but it looks like it would be a simple fix since each `filter` and `group` have a `title` that can be changed.
`  accountList(id: $accountListId) {
    contactFilterGroups {
      filters {
        filterKey
        title
        ... on MultiselectFilter {
          title
          options {
            
          }
        }`

src/components/Contacts/ContactDetails/ContactDonationsTab/PartnershipInfo/PartnershipInfo.tsx
caleballdrin: Right now the other items don't have colon's. Probably because they are often empty?
Screenshot 2024-09-10 at 12 28 40 PM

canac: Good point. I forgot that the label is a different color, so this is good as-is.

PR #1050, opened 2024-09-10
MPDX-8183 Simplifying the UX of the Pledge modal when status given or received

dr-bizz:
## Description
In this PR, I will make the edit Pledge model more like the old MPDX. 

**When the pledge has a status of `given`:**
- The amount should be disabled and not editable.
- The status should be `Given`, but this shouldn't show as an option to select.

**When the pledge has a status of `Received`:**
- It should show the received status, but otherwise, it should be hidden.
_This is unlike the old MPDX, the old MPDX shows the received status as a option, but when users select it, the contact/pledge stays in committed and causes a lot of confusion._
We have thought about removing the received column entirely, that is a large change. This might help with some of the confusion.

## Checklist:

### DIFF
--- a/src/components/Tool/Appeal/InitialPage/AddAppealForm/AddAppealForm.tsx
+++ b/src/components/Tool/Appeal/InitialPage/AddAppealForm/AddAppealForm.tsx
@@ -80,6 +80,27 @@ export const calculateGoal = (
   return Math.round(((initialGoal + letterCost) / adminPercent) * 100) / 100;
 };
 
+const gqlStatusesToDBStatusMap: { [key: string]: string } = {
+  NULL: 'null',
+  ACTIVE: 'active',
+  HIDDEN: 'hidden',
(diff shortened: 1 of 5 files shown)


src/components/Tool/Appeal/InitialPage/AddAppealForm/AddAppealForm.tsx
```diff
@@ -80,6 +80,24 @@ export const calculateGoal = (
   return Math.round(((initialGoal + letterCost) / adminPercent) * 100) / 100;
 };
 
+const gqlStatuesToDBStatusMap: { [key: string]: string } = {
```
canac, line 83:
```suggestion
const gqlStatusesToDBStatusMap: { [key: string]: string } = {
```

canac (approved):
I wasn't able to test editing a received pledge, but this looks good!

PR #1058, opened 2024-09-11
[MPDX-8192] Do not allow fractional admin percentages

canac:
## Description

Require the admin percentage to be a whole number.

[MPDX-8192](https://jira.cru.org/browse/MPDX-8192)

## Checklist:

### DIFF
--- a/src/components/Tool/Appeal/InitialPage/AddAppealForm/AddAppealForm.tsx
+++ b/src/components/Tool/Appeal/InitialPage/AddAppealForm/AddAppealForm.tsx
@@ -168,6 +168,9 @@ export const buildExclusionFilter = (
   });
 };
 
+const isPositiveInteger = (value: number | undefined) =>
+  typeof value === 'number' && value >= 0 && value === Math.floor(value);
+
 const appealFormSchema = yup.object({
   name: yup.string().required('Please enter a name'),
   initialGoal: yup
@@ -176,8 +179,8 @@ const appealFormSchema = yup.object({
     .required(i18n.t('Initial Goal is required'))
     .test(
       i18n.t('Is positive?'),
-      i18n.t('Must use a positive number for Initial Goal'),
-      (value) => typeof value === 'number' && value >= 0,
+      i18n.t('Must use a positive whole number for Initial Goal'),
+      isPositiveInteger,
     ),
   letterCost: yup
     .number()
@@ -185,8 +188,8 @@ const appealFormSchema = yup.object({
     .required(i18n.t('Letter Cost is required'))
     .test(
       i18n.t('Is positive?'),
-      i18n.t('Must use a positive number for Letter Cost'),
-      (value) => typeof value === 'number' && value >= 0,
+      i18n.t('Must use a positive whole number for Letter Cost'),
+      isPositiveInteger,
     ),
   adminPercentage: yup
     .number()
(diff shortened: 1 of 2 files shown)


src/components/Tool/Appeal/InitialPage/AddAppealForm/AddAppealForm.tsx
```diff
@@ -177,7 +177,7 @@ const appealFormSchema = yup.object({
     .test(
       i18n.t('Is positive?'),
       i18n.t('Must use a positive number for Initial Goal'),
-      (value) => parseFloat(value as unknown as string) >= 0,
+      (value) => typeof value === 'number' && value >= 0,
```
dr-bizz, line 180: When testing on https://8192-validate-percentage.d3dytjb8adxkk5.amplifyapp.com/
I was still able to enter a number with a decimal point.

dr-bizz (changes requested):
The Admin Percentage looks great. Can you add the rule of whole numbers to initial goal and letter cost?

I know I didn't specify that on the Task.
Screenshot 2024-09-12 at 9 28 59 AM

src/components/Tool/Appeal/InitialPage/AddAppealForm/AddAppealForm.tsx
canac: Are you saying the admin percentage validation is good but there isn't validation yet on the initial goal?

canac:
> Can you add the rule of whole numbers to initial goal and letter cost?

@dr-bizz It kind of seems weird to me that we're trying to keep users from entering fractional amounts, but the goal still probably won't probably be a whole number (e.g. with the numbers below). It feels like we should be rounding the goal amount too if we want everything to be a whole number. Let me know what you think.

Screenshot 2024-09-12 at 10 21 31 AM

Also, the goal amount is editable in Angular but not in React. That needs fixed too.

src/components/Tool/Appeal/InitialPage/AddAppealForm/AddAppealForm.tsx
dr-bizz: yes that is correct

dr-bizz:
@dr-bizz It kind of seems weird to me that we're trying to keep users from entering fractional amounts, but the goal still probably won't probably be a whole number (e.g. with the numbers below). It feels like we should be rounding the goal amount too if we want everything to be a whole number. Let me know what you think.


This is how the old MPDX was, not saying it's correct. But you can bring this up with Scott and see if he wants to make the change. I'm okay with it if he is.

dr-bizz:
I've created a task for the Goal input -  [MPDX-8208](https://jira.cru.org/browse/MPDX-8208)

canac:
If we're duplicating Angular, the Angular version allows you to enter fractional amounts for all four fields, including the admin percentage.

Screenshot 2024-09-13 at 9 04 26 AM

dr-bizz:
> If we're duplicating Angular, the Angular version allows you to enter fractional amounts for all four fields, including the admin percentage.
> 
> Screenshot 2024-09-13 at 9 04 26 AM

Have you clicked submit?

canac:
@dr-bizz My bad, you're right 🤦‍♂️. I'll add the whole number validation to the three fields.

dr-bizz:
All good. That got me the first time as well.

PR #1059, opened 2024-09-11
[MPDX-8193] Preselect new appeal options

canac:
## Description

* Initially set `Add contacts with the following status(es):` to `-- All Active --`
* Initially set `Do not add contacts who:` to `Have "Send Appeals" set to No`

## Checklist:

### DIFF
--- a/src/components/Tool/Appeal/InitialPage/AppealsInitialPage.tsx
+++ b/src/components/Tool/Appeal/InitialPage/AppealsInitialPage.tsx
@@ -10,10 +10,7 @@ import {
 } from '@mui/material';
 import { useTranslation } from 'react-i18next';
 import { makeStyles } from 'tss-react/mui';
-import AddAppealForm, {
-  ExclusionEnum,
-  contactExclusions,
-} from 'src/components/Tool/Appeal/InitialPage/AddAppealForm/AddAppealForm';
+import AddAppealForm from 'src/components/Tool/Appeal/InitialPage/AddAppealForm/AddAppealForm';
 import Appeals from 'src/components/Tool/Appeal/InitialPage/Appeals';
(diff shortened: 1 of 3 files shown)


src/components/Tool/Appeal/InitialPage/AppealsInitialPage.tsx
```diff
@@ -66,7 +69,19 @@ const AppealsInitialPage: React.FC = () => {
...
+                      {
+                        name: '-- All Active --',
+                        value: 'ACTIVE',
+                      },
+                    ]}
+                    appealExcludes={contactExclusions.filter(
+                      (exclusion) =>
+                        exclusion.value === ExclusionEnum.DoNotAskAppeals,
+                    )}
+                  />
```
dr-bizz, line 84: These changes need to be added to the form itself as this is the default.

Can you move these changes inside the `<AddAppealForm/>`?

dr-bizz (changes requested):
Looks great. Can we move the logic inside the AddAppealForm? As these settings are standard wherever the form will be used.

src/components/Tool/Appeal/InitialPage/AppealsInitialPage.tsx
dr-bizz: It accepts props as they will be used in the announcements, which can pass down appeal statements and appeal excludes.

Can we make the appealStatuses and appealExcludes have a default value of what you have above?

dr-bizz: It uses it in the old one here: `src/tools/appeals/appeals.service.ts` Line 62 "oneClickAppeal"

dr-bizz (approved):
Awesome work!

PR #1063, opened 2024-09-13
MPDX-8205 Style CSV

caleballdrin:
## Description
https://jira.cru.org/secure/RapidBoard.jspa?rapidView=3&view=detail&selectedIssue=MPDX-8205&sprint=1259#
- Add Card styling
- Use standard buttons
- Fix issue with not allowing data to import because it required full_name
- Show header data correctly formatted

## Checklist:


src/components/Tool/Import/Csv/uploadCsvFile.ts
```diff
@@ -182,7 +182,7 @@ export const useSupportedHeaders = () => {
 };
 
 export const useRequiredHeaders = () => {
-  const hardCodedRequiredHeaders = ['first_name', 'last_name', 'full_name'];
+  const hardCodedRequiredHeaders = ['first_name', 'last_name'];
```
wrandall22, line 185: See equivalent logic [here](https://github.com/CruGlobal/mpdx_web/blob/master/src/tools/import/csv/headers/headers.component.ts#L23-L31)

src/components/Tool/Import/Csv/uploadCsvFile.ts
```diff
@@ -182,7 +182,7 @@ export const useSupportedHeaders = () => {
...
 
   const [requiredHeaders, setRequiredHeaders] = useState(
```
canac, line 187: I don't fully understand the `useEffect` below, but right now it will run on every render. To only make it run once, it needs an empty dependency array `[]`.

canac (approved):
It seems to work now!

PR #1074, opened 2024-09-16
MPDX-8216 - tools - fix commitment info - suggested changes

wjames111:
## Description

- Suggested Changes Needs to be the default and not the fallback
- Ticket #8131 [MPDX-8131](https://jira.cru.org/browse/MPDX-8131)

## Checklist:

### DIFF
--- a/src/components/Tool/FixCommitmentInfo/Contact.tsx
+++ b/src/components/Tool/FixCommitmentInfo/Contact.tsx
@@ -203,13 +203,11 @@ const Contact: React.FC<Props> = ({
   const { classes } = useStyles();
   const { t } = useTranslation();
 
-  const suggestedAmount = !suggestedChanges?.pledge_amount
-    ? null
-    : typeof suggestedChanges.pledge_amount === 'string'
-    ? parseInt(suggestedChanges.pledge_amount)
-    : suggestedChanges.pledge_amount;
+  const suggestedAmount = suggestedChanges?.pledge_amount || null;
 
   const suggestedFrequency = suggestedChanges?.pledge_frequency || null;
+
+  const suggestedStatus = suggestedChanges?.status || null;
 
   const onSubmit = async ({
     status,
@@ -247,18 +245,10 @@ const Contact: React.FC<Props> = ({
     <Grid container className={classes.container}>
       <Formik
         initialValues={{
-          statusValue: statusValue,
+          statusValue: suggestedStatus || statusValue,
           pledgeCurrency: amountCurrency,
-          pledgeAmount: suggestedAmount
-            ? suggestedAmount
-            : amount
-            ? amount
-            : '',
-          pledgeFrequency: (suggestedFrequency as PledgeFrequencyEnum)
-            ? suggestedFrequency
-            : frequencyValue
-            ? frequencyValue
-            : '',
+          pledgeAmount: suggestedAmount || amount,
+          pledgeFrequency: suggestedFrequency || frequencyValue,
         }}
         validationSchema={commitmentInfoFormSchema}
         onSubmit={async (values) => {
@@ -315,13 +305,10 @@ const Contact: React.FC<Props> = ({
                               amount && amountCurrency
                                 ? currencyFormat(amount, amountCurrency, locale)
                                 : ''
-                            } ${
-                              pledgeFrequency &&
-                              getLocalizedPledgeFrequency(
-                                t,
-                                pledgeFrequency as PledgeFrequencyEnum,
-                              )
-                            }`}
+                            } ${getLocalizedPledgeFrequency(
+                              t,
+                              frequencyValue as PledgeFrequencyEnum,
+                            )}`}
                           </Typography>
                         </Box>
                       </Box>
--- a/src/components/Tool/FixCommitmentInfo/FixCommitmentInfo.tsx
+++ b/src/components/Tool/FixCommitmentInfo/FixCommitmentInfo.tsx
@@ -57,6 +57,7 @@ export interface DonationsType {
 }
 
 export interface SuggestedChangesType {
+  status: string;
   pledge_amount: string | number;
   pledge_frequency: string;
 }


dr-bizz:
Oh, I see. I would use what you have then (Not use proxy) and convert the JSON to what you need and use it.

src/components/Tool/FixCommitmentInfo/Contact.tsx
```diff
@@ -249,9 +249,10 @@ const Contact: React.FC<Props> = ({
         initialValues={{
           statusValue: statusValue,
           pledgeCurrency: amountCurrency,
-          pledgeAmount: amount || suggestedAmount,
```
caleballdrin, line 252: @canac Can you help us make sense of this? We don't know if Fix Commitment Info should set the initial values to blank, the suggested values, or the current values.  On my personal account, it appears that it sets the initial values to the suggested values and leaves it blank if there is no suggested value.

However, in the Angular code, the `mutateContacts` service appears to use `defaultTo` to set the default to the original/current value and only use the suggested value if it doesn't have one. I was thinking you might be able to understand the angular better than we can.

https://github.com/CruGlobal/mpdx_web/blob/19c03ae0689609eeee8b9b516700cc75ee4f3f79/src/tools/fix/commitment/commitment.service.ts

dr-bizz: On your account, is there a way to log out the `contact.pledge_amount`? I wonder if this is blank, which is why it's showing blank.

When you say blank, do you mean like this? where it only shows the "0" placeholder for the amount.
Screenshot 2024-09-16 at 3 48 39 PM

dr-bizz: If so, this is because the pledge amount is `null` and there is no `contact.suggested_changes.pledge_amount`
Screenshot 2024-09-16 at 3 49 45 PM

dr-bizz: Let me know if that makes sense

caleballdrin: Here's an example. Clearly, on this contact there is a current `status` and pledge `frequency`. But they are not set as the initial values in the form fields. I think this is the correct behavior, but I'm not seeing how that happens in the code.

Screenshot 2024-09-16 at 1 12 34 PM

caleballdrin: Ok, it looks like the data from angular is messed up...
``            "attributes": {
                "avatar": "https://stage.mpdx.org/images/avatar_f.png",
                "name": "Dauterman, Abigail",
                "pledge_amount": "150.0",
                "pledge_currency": "USD",
                "pledge_frequency": "6",
                "status": "partner_financial",
                "suggested_changes": {
                    "pledge_amount": null,
                    "pledge_frequency": null,
                    "status": "Partner - Special"
                }``

caleballdrin: The above was staging. So maybe because the `status` data was wrong it was showing up blank.
But here is production. It sets the initial value to `suggested_changes` `status` instead of the current `status`. What do you think is going on here?

Screenshot 2024-09-16 at 1 46 03 PM
Screenshot 2024-09-16 at 1 46 14 PM

canac: Is it possible that the discrepancy is because task phases is in staging but not production? Maybe in staging it tried to set the autocomplete value to `Partner - Special` but there's not an item with that value so it shows up as blank?

caleballdrin: Yes I think that is what is happening in staging. I think that is explainable.

But for production (most recent screenshots) that is not the case. It is initializing the suggested_changes status.

dr-bizz: Interesting. From what I'm looking at in the code, I do not think it's initializing the `suggested_changes` status as you haven't hooked up the `suggested_changes.status` to the initial value of `status`.

dr-bizz: mpdx-react/src/components/Tool/FixCommitmentInfo/Contact.tsx
```
const Contact: React.FC<Props> = ({
  statusTitle,
  statusValue,
  ...
  suggestedChanges,
}) => {
<Formik
        initialValues={{
          statusValue: statusValue,
          ...
          }}
/>
          
```


mpdx-react/src/components/Tool/FixCommitmentInfo/FixCommitmentInfo.tsx
```
<Contact
                        statusTitle={
                          contact.status &&
                          contactPartnershipStatus[contact.status]
                        }
                        statusValue={
                          getLocalizedContactStatus(t, contact.status) || ''
                        }
                        statuses={contactStatuses || []}
                        suggestedChanges={formatSuggestedChanges(
                          contact?.suggestedChanges,
                        )}
 />
```

dr-bizz: The only thing I can think of is the `contactStatuses` variable is incorrect, but as you said it shouldn't be.

caleballdrin: @dr-bizz my comments above are about angular, not React. We are trying to figure out what the desired behavior is in angular. Then we can replicate it in React. Right now it is unclear what the angular code is doing.

dr-bizz: src/tools/fix/commitment/commitment.service.ts
```
assign(contact, {
    ....
    original_status: contact.status,
    ...
    status: defaultTo(contact.status, contact.suggested_changes.status),
    ...
}),
```

src/tools/fix/commitment/item/item.html
```
<div class="contact">
                <a ng-href="/contacts/{{$ctrl.contact.id}}" target="_blank"
                    ><h4 class="name">{{ $ctrl.contact.name }}</h4></a
                >
                ...
                {{ $ctrl.contact.original_status }} {{ $ctrl.contact.original_pledge_amount | currency : '' }} {{
                $ctrl.contact.original_pledge_currency }} {{ $ctrl.contact.original_pledge_frequency }}
</div>

...

<div class="status">
                    <select
                        chosen
                        required
                        disable-search="true"
                        data-placeholder-text-single="'Status' | translate"
                        class="form-control"
                        id="status"
                        ng-options="opt.id as opt.value for opt in $ctrl.serverConstants.data.status_hashes"
                        ng-model="$ctrl.contact.status"
                    >
                        <option value=""></option>
                    </select>
</div>
```

dr-bizz: Does that make sense?

dr-bizz: status: suggestedChanges.status || statusValue

dr-bizz: Yes, that is correct

caleballdrin: Nice. So we have confirmed that the suggested values (status, frequency, amount) should be the initial form values. If they don't exist, then it will set the current values as the initial form values.

caleballdrin (approved):
Looks good!

PR #1075, opened 2024-09-16
Adds validation to fix email addresses.

wjames111:
## Description

- Tools - Fix Email Address - Add validation for edited email addresses
-  Ticket #[8162](https://jira.cru.org/browse/MPDX-8162)

## Checklist:

### DIFF
--- a/src/components/Tool/FixEmailAddresses/FixEmailAddressPerson/FixEmailAddressPerson.tsx
+++ b/src/components/Tool/FixEmailAddresses/FixEmailAddressPerson/FixEmailAddressPerson.tsx
@@ -326,8 +326,8 @@ export const FixEmailAddressPerson: React.FC<FixEmailAddressPersonProps> = ({
                           newEmail: email.email,
                         }}
                         validationSchema={validationSchema}
-                        onSubmit={async (values) => {
-                          await handleChange(id, index, values.newEmail);
+                        onSubmit={(values) => {
+                          handleChange(id, index, values.newEmail);
                         }}
                       >
                         {({
@@ -459,7 +459,7 @@ export const FixEmailAddressPerson: React.FC<FixEmailAddressPersonProps> = ({
                                     error={true}
                                     data-testid="statusSelectError"
                                   >
-                                    {errors.newEmail && errors.newEmail}
+                                    {errors.newEmail}
                                   </FormHelperText>
                                 </FormControl>
 


src/components/Tool/FixEmailAddresses/FixEmailAddressPerson/FixEmailAddressPerson.tsx
```diff
@@ -308,61 +319,36 @@ export const FixEmailAddressPerson: React.FC<FixEmailAddressPersonProps> = ({
...
-                                <>
+                      <Formik
+                        key={index}
+                        enableReinitialize={true}
+                        initialValues={{
+                          newEmail: email.email,
+                        }}
+                        validationSchema={validationSchema}
+                        onSubmit={async (values) => {
+                          await handleChange(id, index, values.newEmail);
```
canac, line 330: `handleChange` doesn't return a promise.

```suggestion
                        onSubmit={(values) => {
                          handleChange(id, index, values.newEmail);
```

src/components/Tool/FixEmailAddresses/FixEmailAddressPerson/FixEmailAddressPerson.tsx
```diff
@@ -372,88 +358,153 @@ export const FixEmailAddressPerson: React.FC<FixEmailAddressPersonProps> = ({
...
+                                      setFieldValue('newEmail', e.target.value);
+                                      handleSubmit();
+                                    }}
+                                    disabled={email.source !== appName}
+                                  />
+                                  <FormHelperText
+                                    error={true}
+                                    data-testid="statusSelectError"
+                                  >
+                                    {errors.newEmail && errors.newEmail}
```
canac, line 462:
```suggestion
                                    {errors.newEmail}
```

canac (approved):
> The only real issue I found with this way is that the confirm button isn’t blocked when an illformed address is entered.

I saw a few times where typing an invalid email address and clicking Confirm sent the invalid email in the GraphQL mutation. But it appears that the server ignored it without returning an error. It's a little weird but probably OK.

It looks like we have a form inside of a form now. I'm not sure if that's bad or not, but if it needs to be changed, we can do that when we research performance.

wrandall22:
> It looks like we have a form inside of a form now. I'm not sure if that's bad or not

Nested forms are not supported in HTML technically.

canac:
> I'm not seeing the nested form, am I missing something?

You're right, I wasn't reading carefully. I thought `FixEmailAddresses` was using Formik, but it's just using a simple `useState` (which I think is the right choice).

I have no issues with the code you have here. Sorry for slowing this down!

PR #1080, opened 2024-09-19
[MDPX-8221] Convert status string to `StatusEnum`

canac:
## Description

* Convert contact statuses from the REST API into `StatusEnum` values.
* Localize contact statuses in the 14 month reports.

[MPDX-8221](https://jira.cru.org/browse/MPDX-8221)

## Checklist:

### DIFF
--- a/pages/api/Schema/reports/fourteenMonth/fourteenMonth.graphql
+++ b/pages/api/Schema/reports/fourteenMonth/fourteenMonth.graphql
@@ -35,6 +35,23 @@ type FourteenMonthReportCurrencyGroupTotalsMonth {
   total: Float!
 }
 
+enum StatusEnum {
+  APPOINTMENT_SCHEDULED
+  ASK_IN_FUTURE
+  CALL_FOR_DECISION
+  CONTACT_FOR_APPOINTMENT
+  CULTIVATE_RELATIONSHIP
+  EXPIRED_REFERRAL
+  NEVER_ASK
+  NEVER_CONTACTED
+  NOT_INTERESTED
+  PARTNER_FINANCIAL
+  PARTNER_PRAY
+  PARTNER_SPECIAL
+  RESEARCH_ABANDONED
+  UNRESPONSIVE
+}
+
 type FourteenMonthReportContact {
   id: ID!
   name: String!
@@ -48,7 +65,7 @@ type FourteenMonthReportContact {
   pledgeAmount: Float
   pledgeCurrency: String
   pledgeFrequency: String
-  status: String
+  status: StatusEnum
 }
 
 type FourteenMonthReportContactMonths {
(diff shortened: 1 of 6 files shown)


pages/api/Schema/reports/fourteenMonth/fourteenMonth.graphql
```diff
@@ -35,6 +35,23 @@ type FourteenMonthReportCurrencyGroupTotalsMonth {
   total: Float!
 }
 
+enum StatusEnum {
```
caleballdrin, line 38: Is there a way we could import this instead of manually redefining it? This would be nice if the statues ever changed again.

src/components/Reports/FourteenMonthReports/Layout/Table/helpers.test.ts
```diff
@@ -20,10 +21,10 @@ const mockContact = (
 describe('extractSortKey', () => {
   it('extracts string values from the contact', () => {
     const contact = mockContact({
-      status: 'Partner',
+      status: StatusEnum.PartnerFinancial,
     });
 
-    expect(extractSortKey(contact, 'status')).toBe('Partner');
+    expect(extractSortKey(contact, 'status')).toBe('PARTNER_FINANCIAL');
```
caleballdrin, line 27: Could this just be:
```suggestion
.toBe(StatusEnum.PartnerFinancial);
```

pages/api/Schema/reports/fourteenMonth/fourteenMonth.graphql
canac: Unfortunately, since the proxy GraphQL server and the Rails GraphQL are completely separate servers there's not a way that I know of to share enums or types between them.

src/components/Reports/FourteenMonthReports/Layout/Table/helpers.test.ts
canac: Good point. The point of that test is just that it pulls out a string field from the contact, so I kind of wanted to `toBe` to be a string literal. I think it makes more sense to pick a different field for that test.

pages/api/Schema/reports/fourteenMonth/fourteenMonth.graphql
caleballdrin: Ok gotcha

src/components/Reports/FourteenMonthReports/useCsvData.test.ts
```diff
@@ -164,7 +165,7 @@ describe('useCsvData', () => {
       ],
       [
         'Quarterly Partner',
-        'Partner - Financial',
+        'PARTNER_FINANCIAL',
```
caleballdrin, line 168: I think we want the CSV to export the nicer-formatted version of the status. Will it be doing that?

caleballdrin: And then we can change this test to check that it gets translated/formatted.

canac:
@caleballdrin I localized the status and frequency. Good catch! I think that was all you were asking for, right?

caleballdrin (approved):
Thanks for localizing them! Looks great!

PR #1085, opened 2024-09-20
[MPDX-8221] Fix 14 month report export in-hand months calculation

canac:
## Description

* I believe the months were reversed by the API at some point. This PR flips the order to use the most recent months instead of the oldest months.
* It also now skips the first month because partners may not have given yet in the current month and we will get false positives of partners missing donations.

[MPDX-8221](https://jira.cru.org/browse/MPDX-8221)

## Checklist:


dr-bizz (approved):
This looks good. Great work! I don't see the functionality for this anywhere `This PR flips the order to use the most recent months instead of the oldest months.`.

I'm approving this as I don't see anythgint wrong with the code.

canac:
@dr-bizz Could this PR https://github.com/CruGlobal/mpdx_api/pull/2620 have been the one that reversed the order of months in the API? It looks like the most likely candidate.

dr-bizz:
That is most likely when it occured.

PR #1088, opened 2024-09-23
MPDX-8237 - Shows tasks count, updates Tasks once deleted.

wjames111:
## Ticket
- Jira Ticket [MPDX-8238](https://jira.cru.org/browse/MPDX-8238)

## Description
- Show total count on Tasks like Angular version
- Update total count when single task is deleted

## Checklist:

### DIFF
--- a/pages/accountLists/[accountListId]/tasks/Tasks.test.tsx
+++ b/pages/accountLists/[accountListId]/tasks/Tasks.test.tsx
@@ -203,14 +203,11 @@ describe('tasks page', () => {
       </MocksProviders>,
     );
     await waitFor(() => expect(getByText('Test Person')).toBeInTheDocument());
-    expect(getByTestId('task-checkbox-1')).toBeInTheDocument();
     userEvent.click(getByTestId('task-checkbox-1'));
     expect(getByText('Showing 17')).toBeInTheDocument();
-    expect(getByTestId('DeleteIconButton-1')).toBeInTheDocument();
     userEvent.click(getByTestId('DeleteIconButton-1'));
     // This is needed for some reason
     userEvent.click(getByTestId('DeleteIconButton-1'));
-    expect(getByText('Confirm')).toBeInTheDocument();
     userEvent.click(getByText('Yes'));
     await waitFor(() => {
       expect(mockEnqueue).toHaveBeenCalledWith(`Task deleted successfully`, {


src/components/Shared/Header/ListHeader.test.tsx
```diff
@@ -472,4 +472,30 @@ describe('ListHeader', () => {
...
+    userEvent.click(actionsButton);
+    expect(getByText('Delete Tasks')).toBeInTheDocument();
+    userEvent.click(getByText('Delete Tasks'));
+    await waitFor(() => {
+      expect(
+        queryByText('Are you sure you wish to delete the 3 selected tasks?'),
+      ).toBeInTheDocument();
+    });
+
+    userEvent.click(getAllByTestId('action-button')[0]);
```
caleballdrin, line 497: Instead of using a test-id here, it might be more readable if you just do `getByText("Yes")`

pages/accountLists/[accountListId]/tasks/[[...contactId]].page.tsx
```diff
@@ -344,6 +344,7 @@ const TasksPage: React.FC = () => {
                     toggleStarredFilter={setStarredFilter}
                     headerCheckboxState={selectionType}
                     massDeselectAll={deselectAll}
+                    showShowingCount
```
Comment (caleballdrin) [inline pages/accountLists/[accountListId]/tasks/[[...contactId]].page.tsx:347]: This is a nice addition, thanks for adding it.

caleballdrin (changes requested):
This is looking great so far. I think there is one more thing to fix.

src/components/Shared/Header/ListHeader.test.tsx
caleballdrin: Oh interesting. But the way you had it doesn't give the error?

caleballdrin: Oh I see. Maybe you need to pass an accountListId in the mock mutation

src/components/Shared/Header/ListHeader.test.tsx
```diff
@@ -112,9 +111,9 @@ const router = {
   push,
 };
 
-beforeEach(() => {
-  (useAccountListId as jest.Mock).mockReturnValue(router);
-});
+// beforeEach(() => {
+//   (useAccountListId as jest.Mock).mockReturnValue(router);
+// });
```
caleballdrin, line 116: You can remove this code instead of commenting it out. @canac maybe you should have a quick look at this PR as well. I think it looks good.

src/components/Shared/Header/ListHeader.test.tsx
```diff
@@ -226,6 +225,33 @@ describe('ListHeader', () => {
...
+    it('Should update count upon deletion', async () => {
+      const { queryByText, getByText } = render(
+        <Components
+          selectedIds={['a', 'b', 'c']}
+          page={PageEnum.Task}
+          totalItems={50}
+          showShowingCount={true}
+        />,
+      );
+      expect(queryByText('Showing 50')).toBeInTheDocument();
```
canac, line 237: If we're expecting the element to exist, use `getBy...` instead of `queryBy...`. `queryBy` is only for when elements *don't* exist.

src/hooks/useMassSelection.ts
```diff
@@ -26,6 +27,12 @@ export const useMassSelection = (
   );
   const [ids, setIds] = useState<string[]>([]);
 
+  useEffect(() => {
```
canac, line 30: We're removing ids from the list every time `idsToRemove` changes, right? I think it would be better to return a `removeIds` function from this hook that components can use to delete ids. What do you think about that?

caleballdrin: Looks like codecov wants more tests :/

canac: Just make sure your test calls `deselectIds` somewhere, and you should be good

pages/accountLists/[accountListId]/tasks/Tasks.test.tsx
```diff
@@ -195,4 +195,27 @@ describe('tasks page', () => {
...
+        <Tasks />
+      </MocksProviders>,
+    );
+    await waitFor(() => expect(getByText('Test Person')).toBeInTheDocument());
+    expect(getByTestId('task-checkbox-1')).toBeInTheDocument();
+    userEvent.click(getByTestId('task-checkbox-1'));
+    expect(getByText('Showing 17')).toBeInTheDocument();
+    expect(getByTestId('DeleteIconButton-1')).toBeInTheDocument();
+    userEvent.click(getByTestId('DeleteIconButton-1'));
+    // This is needed for some reason
```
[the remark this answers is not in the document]
Comment (caleballdrin) [reply pages/accountLists/[accountListId]/tasks/Tasks.test.tsx:211]: Oh I had that same issue with buttons on the task row. The only thing that worked was 2 clicks.

PR #1090, opened 2024-09-24
MPDX-8217 change the contact and task row links into actual links.

dr-bizz:
## Description
In this PR, I converted the contact and task JS click functions into actual links that can be opened in a new tab. This helps with UX

**Contact pages**
List view: the contacts on the list view are now links
Flows view:  The contacts on the flows view are now links.
Map view: the contacts are not links.
Contact details: The links to other contacts on a contact detail page are now links

**Task Page**
List view: the contacts on the list view are now links.

**Reports**
All the report pages that feature contacts names are now links.

**Coaches**
There is no need to edit anything here.

**Tools**
Appeals: I have made the appeal contacts names into links on both list and flows view

## Checklist:


src/components/Contacts/ContactsMap/ContactsMap.tsx
```diff
@@ -198,11 +198,7 @@ export const ContactsMap: React.FC = ({}) => {
                 {sourceToStr(t, selected.source ?? '')}
               </Typography>
               <Typography display="inline">{selected.date}</Typography>
-              <ContactLink
-                onClick={() =>
-                  onContactSelected(selected.id, true, false, true)
-                }
-              >
+              <ContactLink onClick={() => onContactSelected(selected.id, true)}>
                 Show Contact
```
canac, line 202: Want to localize this label real quick?

src/components/Reports/FourteenMonthReports/FourteenMonthReport.tsx
```diff
@@ -42,7 +42,7 @@ export const FourteenMonthReport: React.FC<Props> = ({
   currencyType,
   isNavListOpen,
   title,
-  onSelectContact,
+  getContactUrl,
```
canac, line 45: Instead of prop-drilling `getContactUrl` down from the parent, what do you think about using the `useGetContactUrls` hook in this component and passing in the path based on the currency type? I can see a case for leaving it as-is, so let me know what you think.

src/components/Task/TaskRow/TaskRow.tsx
```diff
@@ -50,14 +54,54 @@ const TaskType = styled(Typography)(({ theme }) => ({
...
+) => void;
+
+type TaskRowContactNameProps = {
+  contact: TaskRowFragment['contacts']['nodes'][0];
+  itemIndex: number;
+  contactsLength: number;
+  selectContact: OnClickFunction;
+} & TypographyProps;
+
+const TaskRowContactName: React.FC<TaskRowContactNameProps> = ({
```
canac, line 69: This component is getting long enough to where I'd consider putting it in a separate file.

src/components/Tool/Appeal/List/ContactRow/ContactRow.tsx
```diff
@@ -134,184 +133,194 @@ export const ContactRow: React.FC<Props> = ({
...
-                />
-              </ListItemIcon>
-            </Hidden>
-            <ListItemText
-              primary={
-                <Typography component="span" variant="h6" noWrap>
-                  <Box component="span" display="flex" alignItems="center">
-                    {name}
+        <ListButton
+          LinkComponent={'a'}
```
canac, line 147: Same question of whether this does anything without an `href` prop and since I thought the default link component is already `a`.

canac (commented):
I like that you moved some of the logic into a custom hook 🔥

canac:
Functionality looks really good too! One thing I noticed is that in task rows, the contact's name is always underlined now instead of just being underlined on hover. Also, command+clicking contacts and tasks to open them in a new tab opens the contact or task in the tab I was in and opens them in a new tab. Normally command+click doesn't change the current URL.

src/components/Reports/FourteenMonthReports/FourteenMonthReport.tsx
dr-bizz: I would prefer to keep the controls in one place and since it's only going down 1 component, I think we are okay

src/components/Task/TaskRow/TaskRow.tsx
dr-bizz: Done

src/components/Contacts/ContactsContext/ContactsContext.tsx
```diff
@@ -151,20 +150,16 @@ export const ContactsProvider: React.FC<ContactsContextProps> = ({
...
-      }
-    },
-  });
+  const [contactsView, updateOptions, { loading: userOptionsLoading }] =
+    useUserPreference({
+      key: 'contacts_view',
+      defaultValue: TableViewModeEnum.List,
+    });
+  useEffect(() => {
+    if (contactsView) {
```
canac, line 159: I think we shouldn't call `setViewMode` if `userOptionsLoading === true` because otherwise we might be trying to save the default value to the server.

dr-bizz: Gotcha, I will change it to ensure userOptions have finished loading is false

src/components/Task/TaskRow/TaskRowContactName.tsx
```diff
@@ -0,0 +1,95 @@
...
+    <>
+      {contactHrefObject && (
+        <NextLink
+          href={contactHrefObject}
+          shallow
+          legacyBehavior
+          passHref
+          key={contact.id}
+        >
+          <Link>
```
canac, line 69: These contact task links have a blue underline when not hovered and a black underline when hovered. Can we change it to be no underline when not hovered and a black underline when hovered?

src/components/Reports/PartnerGivingAnalysisReport/Table/Table.tsx
```diff
@@ -126,12 +123,18 @@ export const PartnerGivingAnalysisReportTable: FC<
...
                 </TableCell>
                 <TableCell>
-                  <ContactName
-                    onClick={() => onClick(contact.id)}
-                    onMouseEnter={preloadContactsRightPanel}
-                  >
-                    {contact.name}
-                  </ContactName>
+                  <NextLink href={contactUrl} passHref>
+                    <Link
```
canac, line 127: These links are also always underlined now instead of underlined on hover. Also, is the `component="a"` necessary? I think the default component for `Link`s is an `<a>` tag.

src/components/Reports/PartnerGivingAnalysisReport/Table/Table.tsx
```diff
@@ -126,12 +123,18 @@ export const PartnerGivingAnalysisReportTable: FC<
...
+                      component="a"
```
canac, line 128: I don't think this `component="a"` is necessary. And here's another link that's always underlined instead of on hover like most of the links in the app.

Maybe we should add `defaultProps: { underline: 'hover' }` to `MuiLink` in the theme to help with consistency.

src/components/Reports/FourteenMonthReports/Layout/Table/Table.tsx
```diff
@@ -124,69 +119,74 @@ export const FourteenMonthReportTable: React.FC<
...
+                hover
+                data-testid="FourteenMonthReportTableRow"
+              >
+                <StyledTableCell>
+                  <Box display="flex" flexDirection="column">
+                    <Box display="flex" alignItems="center">
+                      {!isExpanded && <StyledInfoIcon fontSize="small" />}
+                      <NameTypography variant="body1" expanded={isExpanded}>
+                        <NextLink
+                          href={contactUrl}
```
canac, line 136: Another link that now is always underlined instead of on hover. Maybe we should add `defaultProps: { underline: 'hover' }` to `MuiLink` in the theme to help with consistency.

canac (approved):
The only issue I'm seeing is links always being underlined instead of only being underlined on hover, so I'm going to approve.

We might want to add `defaultProps: { underline: 'hover' }` to `MuiLink` in the theme to ensure consistency in all links across the application. That could also be done later.

PR #1094, opened 2024-09-25
Setting the result based on the displayResults and the default to com…

dr-bizz:
## Description
In this PR, If the result is left blank on the log task modal form, it defaults it to `COMPLETED`.

I'm unsure if this is the correct functionality, as it's been some time since I edited this and I can't find anywhere that mentions what we should o in this case. the other option is that it always runs the getDatabaseValueFromResult, so if there is not a result selected, it defaults to `NONE`.

## Checklist:


caleballdrin (approved):
I think setting it to COMPLETED if there is no result is a good plan.

PR #1095, opened 2024-09-25
MPDX-8309 Send user's language to GraphQL server

dr-bizz:
## Description
In this PR, I send the user's language to the GraphQL API.

**Client-side GraphQL**
In my first commit, I change the client-side GraphQL to send the user's language to the API.
This includes grabbing the user's language preferences and adding it to the Next Auth session so we can use it throughout the site.

**Server-side GraphQL**
In the second commit, I ensure the server-side is also sending the user's language to the API. _(Not proxy like I put in my commit message)_

**Show message to users to re-login**
In the third commit I add a message to users asking them to logout and log back in after changing their language to ensure the language is fully updated. This is because we only get to set the NextAuth session on login.

## Checklist:

### DIFF
--- a/src/components/Settings/preferences/accordions/LanguageAccordion/LanguageAccordion.tsx
+++ b/src/components/Settings/preferences/accordions/LanguageAccordion/LanguageAccordion.tsx
@@ -55,16 +55,6 @@ export const LanguageAccordion: React.FC<LanguageAccordionProps> = ({
         enqueueSnackbar(t('Saved successfully.'), {
           variant: 'success',
         });
-
-        enqueueSnackbar(
-          t(
-            'Please log out and log back in again to ensure all language changes are fully applied.',
-          ),
-          {
-            variant: 'warning',
-            persist: true,
-          },
-        );
       },
       onError: () => {
         enqueueSnackbar(t('Saving failed.'), {
--- a/src/components/Settings/preferences/accordions/LanguageAccordion/LanguageAccordion.tsx
+++ b/src/components/Settings/preferences/accordions/LanguageAccordion/LanguageAccordion.tsx
@@ -55,6 +55,16 @@ export const LanguageAccordion: React.FC<LanguageAccordionProps> = ({
         enqueueSnackbar(t('Saved successfully.'), {
           variant: 'success',
         });
+
+        enqueueSnackbar(
+          t(
+            'Please log out and log back in again to ensure all language changes are fully applied.',
+          ),
+          {
+            variant: 'warning',
+            persist: true,
+          },
+        );
       },
       onError: () => {
         enqueueSnackbar(t('Saving failed.'), {
--- a/src/lib/apollo/link.ts
+++ b/src/lib/apollo/link.ts
@@ -51,13 +51,11 @@ export const batchLink = split(
   restProxyHttpLink,
 );
 
-export const makeAuthLink = (apiToken: string, language?: string | null) =>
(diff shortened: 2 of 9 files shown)


src/lib/apollo/link.ts
```diff
@@ -51,11 +51,13 @@ export const batchLink = split(
   restProxyHttpLink,
 );
 
-export const makeAuthLink = (apiToken: string) =>
+export const makeAuthLink = (apiToken: string, language?: string | null) =>
   new ApolloLink((operation, forward) => {
+    const languageHeader = language ? { 'Accept-Language': language } : {};
```
canac, line 56: Doesn't the server automatically honor the user's language preference? This code looks like it uses the language preference and defaults to the accept header: https://github.com/CruGlobal/mpdx_api/blob/7eff744504fe52358a0a975cb5f36ee6b5f1e5ad/app/controllers/graphql_controller.rb#L59-L68

canac: If I'm reading it right, `scope_request_to_locale` runs for every GraphQL request https://github.com/CruGlobal/mpdx_api/blob/7eff744504fe52358a0a975cb5f36ee6b5f1e5ad/app/controllers/graphql_controller.rb#L2

canac: Sorry, I misread it. The server uses the `Accept-Language` header first and falls back to the user's preference. I feel like we should ask the server people to switch that. In the meantime, can you just set `Accept-Language` to an empty string? When I do that, translated constants are coming through. I think that would be a simpler solution.

src/lib/apollo/link.ts
```diff
@@ -51,11 +51,13 @@ export const batchLink = split(
...
     operation.setContext(({ headers }) => ({
       headers: {
         ...headers,
+        ...languageHeader,
         Authorization: `Bearer ${apiToken}`,
```
canac, line 61: This worked for me

```suggestion
        Authorization: `Bearer ${apiToken}`,
        'Accept-Language': '',
```

src/components/Settings/preferences/accordions/LanguageAccordion/LanguageAccordion.tsx
```diff
@@ -55,6 +55,16 @@ export const LanguageAccordion: React.FC<LanguageAccordionProps> = ({
         enqueueSnackbar(t('Saved successfully.'), {
           variant: 'success',
         });
+
+        enqueueSnackbar(
```
canac, line 59: I think this can be removed now

src/lib/apollo/link.ts
```diff
@@ -57,6 +57,10 @@ export const makeAuthLink = (apiToken: string) =>
       headers: {
         ...headers,
         Authorization: `Bearer ${apiToken}`,
+        // We pass through a empty Accept-Language header to the API to avoid it defaulting to 'en-US'
```
canac, line 60: Great comment! Thanks! 👍

src/components/Settings/preferences/accordions/LanguageAccordion/LanguageAccordion.tsx
dr-bizz: Ah you are correct since we don't need it

src/lib/apollo/link.ts
dr-bizz: Thanks

canac (approved):
Great work! I assume we've tested that this won't break code that expects the constants to be in English?

PR #1100, opened 2024-09-26
MPDX-8306 Add phase grouping of statuses

caleballdrin:
## Description
https://jira.cru.org/secure/RapidBoard.jspa?rapidView=3&view=detail&selectedIssue=MPDX-8306&sprint=1268#
- Group statuses by phase in filters

## Checklist:

### DIFF
--- a/src/components/Shared/Filters/FilterListItemMultiselect.tsx
+++ b/src/components/Shared/Filters/FilterListItemMultiselect.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import SyncAltIcon from '@mui/icons-material/SyncAlt';
 import {
   Autocomplete,
@@ -11,12 +11,63 @@ import {
   Tooltip,
 } from '@mui/material';
 import { useTranslation } from 'react-i18next';
-import { MultiselectFilter } from 'src/graphql/types.generated';
+import { MultiselectFilter, PhaseEnum } from 'src/graphql/types.generated';
 import { useContactPartnershipStatuses } from 'src/hooks/useContactPartnershipStatuses';
 import { getLocalizedPhase } from 'src/utils/functions/getLocalizedPhase';
 import { renameFilterNames, reverseFiltersMap } from './helpers';
 
-interface Props {
+interface MultiselectAutocompleteProps {
+  filter: MultiselectFilter;
+  selected?: Array<string>;
+  toggleValue: (value?: Array<string>) => void;
+  filterTitle: string | undefined;
+  reverseSelected?: boolean;
+  groupBy?: ((option: any) => string) | undefined;
+  options: string[];
+}
+const MultiselectFilterAutocomplete: React.FC<MultiselectAutocompleteProps> = ({
+  filter,
+  selected,
+  toggleValue,
+  filterTitle,
+  reverseSelected,
+  groupBy,
+  options,
+}: MultiselectAutocompleteProps) => {
+  return (
+    <Autocomplete
+      multiple
+      autoHighlight
+      autoSelect
+      value={selected || []}
+      onChange={(_, value) => toggleValue(value)}
+      options={options}
+      getOptionLabel={(option) =>
+        filter.options?.find(({ value }) => String(value) === String(option))
+          ?.name ?? ''
+      }
+      groupBy={groupBy}
+      ChipProps={{
+        color: reverseSelected ? 'error' : 'default',
+        style: {
+          background: reverseSelected ? '#d32f2f' : '#ffffff',
+        },
+      }}
+      filterSelectedOptions
+      fullWidth
+      renderInput={(params) => (
+        <TextField
+          {...params}
+          placeholder={filterTitle}
+          label={filterTitle}
+          data-testid="multiSelectFilter"
+        />
+      )}
+    />
+  );
+};
+
+interface FilterListItemMultiselectProps {
   filter: MultiselectFilter;
   selected?: Array<string>;
   onUpdate: (value?: Array<string>) => void;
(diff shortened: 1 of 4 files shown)


src/components/Shared/Filters/FilterListItemMultiselect.tsx
```diff
@@ -122,7 +127,59 @@ export const FilterListItemMultiselect: React.FC<Props> = ({
...
+                          ?.phase || '',
+                      ) || -1,
+                )
+                ?.map(({ value }) => value) || []
+            }
+            groupBy={(option) =>
+              getLocalizedPhase(
+                t,
+                statusArray.find((status) => status.id === option)?.phase,
+              ) || ''
```
canac, line 157:
```suggestion
              )
```

src/components/Shared/Filters/FilterListItemMultiselect.tsx
```diff
@@ -122,7 +127,59 @@ export const FilterListItemMultiselect: React.FC<Props> = ({
...
+            autoHighlight
+            autoSelect
+            value={selected || []}
+            onChange={(_, value) => toggleValue(value)}
+            options={
+              filter.options
+                ?.slice()
+                .sort(
+                  (a, b) =>
+                    statusArray
```
canac, line 144: I think this could be clearer if we extracted `statusArray?.find((status) => status.id === option.value)` into a helper function that we called once for `a` and once for `b`.

src/components/Shared/Filters/FilterListItemMultiselect.tsx
```diff
@@ -122,7 +127,59 @@ export const FilterListItemMultiselect: React.FC<Props> = ({
             />
           </ListItem>
         ))
-      )}
+      ) : filter.filterKey === 'status' ||
+        filter.filterKey === 'contact_status' ? (
+        <ListItem>
+          <Autocomplete
```
canac, line 133: This is identical to the autocomplete besides the `groupBy` and `options` sorting, right? Ideally, we would extract it into a separate component that could be used in both places, and we pass in optional `groupBy` and `sortOptions` props.

src/components/Shared/Filters/FilterListItemMultiselect.tsx
```diff
@@ -122,7 +127,59 @@ export const FilterListItemMultiselect: React.FC<Props> = ({
             />
           </ListItem>
         ))
-      )}
+      ) : filter.filterKey === 'status' ||
```
canac, line 130: Do you mind adding a comment about the intention of this special case? Basically that we're grouping the status options by phase?

canac (approved):
I don't see any issues, so I'm going to approve. Sorting the options seems like something that should potentially be done on the server, so that we don't have to have as many special cases for certain filters. But we need a special case to group them, so it's not a big deal. And it's probably more work to add a `group` field on the server to the status options.

dr-bizz (commented):
Since Caleb C has approved this, I'm good merging it once he approves it

src/components/Shared/Filters/FilterListItemMultiselect.tsx
```diff
@@ -47,6 +103,38 @@ export const FilterListItemMultiselect: React.FC<Props> = ({
...
+  const isStatusFilter = useMemo(
+    () =>
+      filter.filterKey === 'status' || filter.filterKey === 'contact_status',
+    [filter.filterKey],
+  );
+
+  const groupByPhase = (option) =>
+    getLocalizedPhase(
+      t,
+      statusArray.find((status) => status.id === option)?.phase,
```
canac, line 116: This looks like the same as line 123 and could be moved into a helper function.

src/components/Shared/Filters/FilterListItemMultiselect.tsx
```diff
@@ -47,6 +103,38 @@ export const FilterListItemMultiselect: React.FC<Props> = ({
...
 
+  // If it is a status filter, group the status options by phase
+  const isStatusFilter = useMemo(
+    () =>
+      filter.filterKey === 'status' || filter.filterKey === 'contact_status',
+    [filter.filterKey],
+  );
+
+  const groupByPhase = (option) =>
+    getLocalizedPhase(
```
canac, line 114:
```suggestion
  const groupByPhase = (option: string) =>
```

src/components/Shared/Filters/FilterListItemMultiselect.tsx
```diff
@@ -11,28 +11,84 @@ import {
...
+interface MultiselectAutocompleteProps {
+  filter: MultiselectFilter;
+  selected?: Array<string>;
+  toggleValue: (value?: Array<string>) => void;
+  filterTitle: string | undefined;
+  reverseSelected?: boolean;
+  groupBy?: ((option: any) => string) | undefined;
+  options: string[];
+}
+const MultiselectFilterAutocomplete: React.FC<MultiselectAutocompleteProps> = ({
```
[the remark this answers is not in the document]
canac: Also, what do you think about naming this something like `StatusMultiselectFilter` because it is only used for statuses?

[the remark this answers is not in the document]
caleballdrin: @canac This component is used for both statuses and other Multiselect filters. It just renders differently based on the props given to it.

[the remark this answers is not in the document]
canac: My bad, you're right. You can ignore that suggestion about calling it a status filter.

src/components/Shared/Filters/FilterListItemMultiselect.tsx
```diff
@@ -47,6 +51,37 @@ export const FilterListItemMultiselect: React.FC<Props> = ({
...
+      selectedStatus: string,
+    ): PhaseEnum | null | undefined => getStatusPhaseId(selectedStatus);
+
+    return isStatusFilter
+      ? filter.options
+          ?.slice()
+          .sort(
+            (a, b) =>
+              findStatusPhase(a.value)?.localeCompare(
+                findStatusPhase(b.value) || '',
```
canac, line 78:
```suggestion
              getStatusPhaseId(a.value)?.localeCompare(
                getStatusPhaseId(b.value) || '',
```

canac (approved):

PR #1108, opened 2024-10-02
Mobile files

dr-bizz:
## Description
In this PR, I add the files which the MPDX mobile apps rely on.

## Checklist:

### DIFF
--- a/public/.well-known/apple-app-site-association.json
+++ b/public/.well-known/apple-app-site-association.json
@@ -1,0 +1,14 @@
+{
+  "applinks": {
+    "apps": [],
+    "details": [
+      {
+        "appID": "DQ48D9BF2V.org.cru.mpdx",
+        "paths": [
+          "/contacts/*",
+          "/auth/mobile"
+        ]
+      }
+    ]
+  }
+}
--- a/public/.well-known/assetlinks.json
+++ b/public/.well-known/assetlinks.json
@@ -1,0 +1,12 @@
+[
+    {
+        "relation": ["delegate_permission/common.handle_all_urls"],
+        "target": {
+            "namespace": "android_app",
+            "package_name": "org.mpdx",
+            "sha256_cert_fingerprints": [
+                "1C:10:CC:D9:97:17:F4:D9:B7:16:37:7C:6C:48:ED:0E:CB:29:C1:F9:6B:8A:FF:B1:E5:23:E3:B9:15:2A:C1:8B"
+            ]
+        }
+    }
+]


public/.well-known/apple-app-site-association.json
```diff
@@ -0,0 +1,14 @@
+{
+  "applinks": {
+    "apps": [],
+    "details": [
+      {
+        "appID": "DQ48D9BF2V.org.cru.mpdx",
+        "paths": [
+          "/contacts/*",
```
canac, line 8: I'm not sure what this does, but our contact links now look like `/accountLists/:id/contacts/*`

dr-bizz: Could it be URLs on the mobile apps?

canac: I'm guessing it makes clicking on a link (like in an email) open up the MPDX app.

dr-bizz: On MPDX Web, there isn't a `/auth/mobile` page

frett (approved):

public/.well-known/apple-app-site-association.json
dr-bizz: @frett Could you speak into this? The `paths`, is it paths on a phone or on the web?

frett: the paths property are the url paths that the app will handle deep links for.

I made a comment elsewhere we probably need to update our deep link handling in the app anyways to support the new accountList URLs, but on Android that's a bit more complicated because it could switch to a different account list which will cause the app to wipe all local data currently.

frett: We will want to support the new url structure in the mobile apps so that deep links open correctly in the app.

The inclusion of the new deep link handling is what should cause the mobile team to submit a PR adding the new URL patterns we are handling to the appropriate files here.

canac (approved):
Thanks for fixing that flaky test! Having to rerun it over and over was getting pretty annoying.

public/.well-known/apple-app-site-association.json
```diff
@@ -4,7 +4,7 @@
     "details": [
       {
         "appID": "DQ48D9BF2V.org.cru.mpdx",
-        "paths": ["/contacts/*", "/auth/mobile"]
+        "paths": ["/accountLists/*/contacts/*", "/auth/mobile"]
```
frett, line 7: @levieggertcru I'm assuming including a wildcard in the middle of the apple-app-site-association path would work correctly for iOS?

dr-bizz: When researching, this is what I found I should do when it's a dynamic route. But I could be wrong.

frett: I just checked the Universal Links [spec](https://developer.apple.com/library/archive/documentation/General/Conceptual/AppSearch/UniversalLinks.html) and the wildcard pattern is valid according to the spec.

frett (approved):

PR #1109, opened 2024-10-02
Fix Committment info fixes

caleballdrin:
## Description

- Users were getting an error when saving a contacts commitment info without a Pledge Frequency. I changed '' to null to fix this
- Statuses now have Phase categoriese
- Changing a contacts status in the select box was changing the current status which was wrong.
- Updated confirmation modal messages for better grammar and using appName

## Checklist:


dr-bizz (commented):
I might be missing something, but I don't see the phases as listHeaders in this Status dropdown.
Screenshot 2024-10-02 at 3 38 42 PM

caleballdrin:
@dr-bizz is that view from the amplify preview? Whatever screenshot you have there isn't the most updated version.

src/components/Tool/FixCommitmentInfo/Contact.tsx
```diff
@@ -167,7 +171,7 @@ interface Props {
   id: string;
   name: string;
   donations: DonationsType[];
-  status: string | undefined;
+  currentStatus: string | undefined;
```
canac, line 174: Would it make sense to switch this to the status enum instead of casting it on line 322?

```suggestion
  currentStatus: StatusEnum | undefined;
```

src/components/Tool/FixCommitmentInfo/Contact.tsx
```diff
@@ -312,8 +319,11 @@ const Contact: React.FC<Props> = ({
                           <Typography variant="subtitle2">
                             {`Current: ${getLocalizedContactStatus(
```
canac, line 320: This needs to be `t('Current: {{status}}')`

src/components/Tool/FixCommitmentInfo/Contact.tsx
```diff
@@ -312,8 +319,11 @@ const Contact: React.FC<Props> = ({
...
-                              status as StatusEnum,
-                            )} ${
+                              currentStatus,
+                            )}`}
+                          </Typography>
+                          <Typography variant="subtitle2">
+                            {`${
                               amount && amountCurrency
                                 ? currencyFormat(amount, amountCurrency, locale)
                                 : ''
```
canac, line 329: I think using a template string just adds noise. I'd write this as 

```
{amount && amountCurrency ? currencyFormat(amount, amountCurrency, locale) : null}
{getLocalizedPledgeFrequency(t, frequencyValue as PledgeFrequencyEnum)}
```

src/components/Tool/FixCommitmentInfo/Contact.tsx
```diff
@@ -357,15 +367,16 @@ const Contact: React.FC<Props> = ({
...
-                              {Object.values(StatusEnum).map((status) => (
-                                <MenuItem
-                                  value={status}
-                                  key={status}
-                                  data-testid="statusSelectOptions"
-                                >
-                                  {getLocalizedContactStatus(t, status)}
-                                </MenuItem>
-                              ))}
+                              {phases?.map((phase) => [
```
canac, line 370: It's weird that we're using an array in JSX. Did you mean to use a fragment?

src/components/Tool/FixCommitmentInfo/FixCommitmentInfo.test.tsx
```diff
@@ -84,6 +86,9 @@ const Components = ({
 describe('FixCommitmentInfo', () => {
   beforeEach(() => {
     setContactFocus.mockClear();
+    (useGetAppSettings as jest.Mock).mockReturnValue({
```
canac, line 89: Can we get this working by adding `<AppSettingsProvider>` to the `TestComponent` instead of mocking `useGetAppSettings`?

src/components/Tool/FixCommitmentInfo/Contact.tsx
```diff
@@ -240,7 +247,7 @@ const Contact: React.FC<Props> = ({
 
     showModal(
       modalContact,
-      t(`Are you sure you wish to update {{source}} commitment info?`, {
+      t(`Are you sure you wish to update the commitment info for {{source}}?`, {
         source: name,
       }),
       t('Update'),
```
canac, line 253: Can we change this title to something like "Update Commitment Info"?

canac (changes requested):
This is a lot better! The suggested statuses populate and update.

I'm getting an error though when I type an amount: `Variable $attributes of type ContactUpdateInput! was provided invalid value for pledgeAmount (Could not coerce value "100" to Float)`. I see this a lot with yup. For some reason, it tells TypeScript that the value is a number, but it is actually a string. You may have to do a manual `parseFloat`.

src/components/Tool/FixCommitmentInfo/Contact.tsx
caleballdrin: I needed that because an empty fragment was causing problems inside the `select`. It was saying `status` might be undefined and in the `select` it wouldn't show the initial status.

src/components/Tool/FixCommitmentInfo/Contact.tsx
caleballdrin: I needed the template string to add the space between the pledgeAmount and the frequency. Any other thoughts on how I could do that?

canac: Gotcha. `{expression1}{' '}{expression2}` is how to do that since prettier will trim off the trailing space at the end of a line.

src/components/Tool/FixCommitmentInfo/FixCommitmentInfo.test.tsx
canac: Can you send me your test file that isn't working? When I remove the `useGetAppSettings` mock and add `<AppSettingsProvider>` to `Components`, the tests still pass.

src/components/Tool/FixCommitmentInfo/Contact.tsx
canac: Does this work? This will skip rendering the fragment if phase is undefined.

```tsx
{phases?.map(
  (phase) =>
    phase && (
      <>
        <ListSubheader key={phase.id}>
          {getLocalizedPhase(t, phase.id)}
        </ListSubheader>
        ,
        {phase.contactStatuses.map((status) => (
          <MenuItem key={status} value={status}>
            {contactStatuses[status]?.translated}
          </MenuItem>
        ))}
      </>
    ),
)}
```

caleballdrin: No that still doesn't work

canac: 🤷 OK I guess it's fine as-is then.

canac (approved):
The float parsing error is happening in staging so you can fix it in another PR.

canac (approved):

PR #1116, opened 2024-10-03
Coaching Page - Add person's name

caleballdrin:
## Description
https://secure.helpscout.net/conversation/2724265010/1237546?folderId=7669074

- Adding the name of the person on the coaching list and detail page.

## Checklist:


src/components/Coaching/LoadCoachingList.test.tsx
```diff
@@ -17,85 +17,7 @@ describe('LoadCoaching', () => {
...
-          "endCursor": "Arm Bible Circus",
-          "hasNextPage": false,
-          "hasPreviousPage": true,
-          "startCursor": "Drum Baby Rock",
-        },
-        "totalCount": 7,
-        "totalPageCount": 59,
-      }
-    `);
+    expect(result.current.data?.coachingAccountLists.nodes.length).toBe(3);
```
canac, line 20: Honestly, I'd just remove this whole test file.

src/components/Coaching/CoachingRow/CoachingRow.tsx
```diff
@@ -61,31 +66,50 @@ export const CoachingRow: React.FC<Props> = ({
...
+            </CoachingNameText>
+            <Typography variant="subtitle1">{usersList}</Typography>
+          </Box>
+          <Box>
+            <Typography
+              variant="h6"
+              sx={{ float: 'left', marginInline: '5px' }}
+            >
+              {t('Balance:')}{' '}
+              {balance && currencyFormat(balance, currency, locale)}
```
canac, line 98: Do we want to hide the balance if it is zero? It seems like we would still want to show it. Also, this will just render 0 instead of formatting it as currency if the balance is 0.

src/components/Coaching/CoachingRow/CoachingRow.tsx
```diff
@@ -61,31 +66,50 @@ export const CoachingRow: React.FC<Props> = ({
...
-            }}
-            aria-label={t('Remove Access')}
-          >
-            <VisibilityOff />
-          </Button>
-        </CoachingNameText>
+        <Box display="flex">
+          <Box flex={1}>
+            <CoachingNameText variant="h6" color="primary">
+              <NextLink
```
canac, line 79: None of the `CoachingNameText` styles seem like they change anything. It already has `margin: 0`, it's `cursor: pointer` because it's a link, and `display: inline` doesn't change anything either. Unless I'm missing something, I'd just make this a `Typography` and remove the `CoachingNameText` styled component and the `color="primary"` attribute since the coloring comes from the nested link.

canac (commented):
Great work! A couple small fixes and this will be ready to go. Can you add a test too that the users are shown?

You can check off the "Missing second name for coached user" line from our functionality review sheet when this merges, because I'm pretty sure it's a duplicate.

src/components/Coaching/CoachingRow/CoachingRow.tsx
caleballdrin: I was thinking we should still show Balance: 0 if the balance is 0. But I guess we could show USD 0.

canac: Yeah, I think we should show the currency like for other amounts. But feel free to disagree.

canac (approved):
I wasn't thinking the `display="inline"` was necessary, but I trust you. Thanks for fixing these!

src/components/Coaching/CoachingRow/CoachingRow.tsx
caleballdrin: I like having display='inline' because it makes the hover only on the actual link text rather than the hover applying to the whole empty line space all the way across the page.

canac: That makes sense. I missed that.

src/components/Task/Modal/Form/Inputs/SuggestedContactStatus/SuggestedContactStatus.test.tsx
```diff
@@ -114,16 +114,16 @@ describe('SuggestedContactStatus', () => {
...
-          contactId: 'contact-1',
-        },
+      expect(mutationSpy).toHaveGraphqlOperation('ContactStatus', {
+        accountListId: accountListId,
+        contactId: 'contact-1',
       });
     });
 
-    expect(await findByText('Initiate for Appointment')).toBeInTheDocument();
+    await waitFor(
```
canac, line 123: You can also customize `findByText`'s timeout if you need to: `findByText('...', {}, { timeout: 3000 })`.

caleballdrin: Ah ok good to know

PR #1118, opened 2024-10-05
MPDX-8344 Tools Sources

caleballdrin:
## Description

- use sourceToStr() function to find the display name of sources
- Stop saving source values as appName, instead use 'MPDX' like Angular did.

## Checklist:


src/components/Tool/FixPhoneNumbers/Contact.tsx
```diff
@@ -29,10 +29,12 @@ import { useTranslation } from 'react-i18next';
 import { makeStyles } from 'tss-react/mui';
 import * as yup from 'yup';
 import { SetContactFocus } from 'pages/accountLists/[accountListId]/tools/useToolsHelper';
+import { editableSources } from 'src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/EditContactAddressModal/EditContactAddressModal';
```
dr-bizz, line 32: This should be moved to a helper file.

src/components/Tool/FixPhoneNumbers/Contact.tsx
```diff
@@ -483,7 +482,11 @@ const Contact: React.FC<Props> = ({
                                       setFieldValue('newPhone', e.target.value);
                                       handleSubmit();
                                     }}
-                                    disabled={phoneNumber.source !== appName}
+                                    disabled={
+                                      editableSources.indexOf(
+                                        phoneNumber.source,
+                                      ) === -1
```
dr-bizz, line 488: This should be made into a fucntion as it's used in mutiple places here and the EditContactAddressModal

src/components/Tool/FixMailingAddresses/FixMailingAddresses.tsx
```diff
@@ -396,7 +395,7 @@ const FixMailingAddresses: React.FC<Props> = ({
                           />
                           {t('Confirm {{amount}} as {{source}}', {
                             amount: totalContacts,
-                            source: defaultSource,
+                            source: sourceToStr(t, defaultSource),
```
dr-bizz, line 398: You use in 2 places, it could be made into a variable to make it easier to maintain.

src/components/Tool/FixMailingAddresses/Contact.tsx
```diff
@@ -321,9 +322,6 @@ const Contact: React.FC<Props> = ({
                       <strong>{t('Source')}: </strong>
                     </Typography>
                   </Hidden>
-                  <Typography display="inline">
-                    {t('{{appName}}', { appName })}
-                  </Typography>
                 </Box>
```
dr-bizz, line 325: You've removed the source?

src/components/Tool/FixEmailAddresses/FixEmailAddressPerson/FixEmailAddressPerson.tsx
```diff
@@ -453,7 +458,10 @@ export const FixEmailAddressPerson: React.FC<FixEmailAddressPersonProps> = ({
                                       setFieldValue('newEmail', e.target.value);
                                       handleSubmit();
                                     }}
-                                    disabled={email.source !== appName}
+                                    disabled={
+                                      editableSources.indexOf(email.source) ===
+                                      -1
+                                    }
```
dr-bizz, line 464: Use function of this here

src/components/Tool/FixEmailAddresses/FixEmailAddressPerson/FixEmailAddressPerson.tsx
```diff
@@ -463,7 +471,7 @@ export const FixEmailAddressPerson: React.FC<FixEmailAddressPersonProps> = ({
                                   </FormHelperText>
                                 </FormControl>
 
-                                {email.source === appName ? (
+                                {editableSources.indexOf(email.source) > -1 ? (
```
dr-bizz, line 474: again here

dr-bizz (changes requested):

src/components/Tool/FixMailingAddresses/Contact.tsx
caleballdrin: Yes, it is not needed there.

caleballdrin:
@dr-bizz Other orgs will have the source saved in their database as 'MPDX' as it always has been. But on their UI they will only see the name of their app. We could rename 'MPDX' to something like 'manual-input' but that would require a database migration.

wrandall22:
Ok, since it is in the data and not showing, that should be within the license agreement's bounds.

dr-bizz (changes requested):
Great work! I found a few more changes. Nothing crazy.

Also, it looks like some sources in MPDX didn't have the `sourceToStr` function or are using the old way to prevent the user from editing them.

- src/components/Tool/MergeContacts/ContactPair.tsx L 227, 242, 259
- src/components/Tool/FixSendNewsletter/Contact.tsx L 258
- src/components/Contacts/MassActions/Merge/MassActionsMergeModal.tsx L 168
- src/components/Contacts/ContactDetails/ContactDonationsTab/PartnershipInfo/PartnershipInfo.tsx L 114
- src/components/Contacts/ContactDetails/ContactDetailsTab/People/Items/PersonModal/PersonPhoneNumber/PersonPhoneNumberItem.tsx L 69-73
-src/components/Contacts/ContactDetails/ContactDetailsTab/People/Items/PersonModal/PersonEmail/PersonEmailItem.tsx L 70-72

src/components/Tool/FixMailingAddresses/Contact.tsx
wrandall22: The dropdown will be filled in with values that existing addresses have, and they can choose which source to create the address in. So, if they already have a DonorHub and an MPDX address, both will appear in the dropdown to choose. This impacts the logic for `primary` when saving as well.

dr-bizz: I think this is what Bill is referring to with the dropdown. There are 2 MPDX values. I assume this is due to one being something else like `MPDX6` but then being changed to MPDX on the UI.

Screenshot 2024-10-09 at 9 24 25 AM

dr-bizz: > In my opinion, I think it is self explanatory that if they create the address in MPDX, then the source will be MPDX. Just like if they create a new contact or person or donation or anything

I think it's best if we make sure the user knows they are creating the address on MPDX. This will avoid confusion.

I'm open to changing how we tell the user a new address will be created on MPDX.

src/utils/sourceHelper.ts
```diff
@@ -0,0 +1,50 @@
+import { TFunction } from 'react-i18next';
+
+const appName = process.env.APP_NAME ?? 'MPDX';
+
+// This value should be used in the database for a custom/manual source. Like when the user creates a new email, phone number, address, etc.
+export const manualSourceValue = 'manual';
```
dr-bizz, line 6: Is the backend team onboard with this switch? Or are they currently using it?

I see it makes sense as we don't want to keep using the word MPDX in the database.

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -129,16 +127,14 @@ const FixPhoneNumbers: React.FC<Props> = ({
     [key: string]: PhoneNumberData;
   }>({});
 
-  const [sourceOptions, setSourceOptions] = useState([appName]);
+  const [defaultSource, setDefaultSource] = useState<string>(manualSourceValue);
```
dr-bizz, line 130: you might not need to define the type as `manualSourceValue` is a string

dr-bizz (commented):
Great work on this! I have a few comments, but not much. This PR is very close

Please know it's always easier to spot mistakes with fresh eyes. Sometimes PRs do take some time to get through the review stage, which is totally fine.

dr-bizz:
I'm not sure if this is related, but the primary button doesn't work on `Fix Phone Numbers`.

1. Log into the app test account and select account, Pedra Perez.
2. Go to tools->Fix Phone Numbers
3. Select a different primary phone number and refresh.
4. The old phone number will still be set to primary.

https://github.com/user-attachments/assets/39db3885-61f3-4a39-8af6-140a016b969a

wrandall22:
My understanding of how the primary star worked in the Angular app is that it did nothing until confirming. When confirming, it would set the primary to the one chosen. On bulk save, it would set the primary with logic on the dropdown source. If we are changing how this works, it needs to be confirmed with stakeholders.

src/utils/sourceHelper.ts
caleballdrin: I just looked into this and it is the backend that sets the `source` value to `MPDX`. So I will remove `manual` and go back to `MPDX`.

src/components/Tool/FixMailingAddresses/Contact.tsx
caleballdrin: Ok I will add the source to the Add Address button so it will look like: `Add Address (MPDX)`

caleballdrin:
@wrandall22 @dr-bizz Only Fix Mailing Addresses page has the primary address saved immediately when the star is clicked. All others wait for the Confirm button to be pressed. I don't plan on changing any of that functionality in this PR.

src/utils/sourceHelper.ts
```diff
@@ -43,8 +40,5 @@ export const sourcesMatch = (
   defaultSource: string | undefined,
   itemSource: string | undefined,
 ): boolean => {
-  return (
-    itemSource === defaultSource ||
-    (defaultSource === manualSourceValue && itemSource === oldManualSourceValue)
-  );
+  return itemSource === defaultSource;
```
dr-bizz, line 43: We no longer want to match it with manualSourceValue?

src/components/Tool/FixMailingAddresses/FixMailingAddresses.tsx
```diff
@@ -94,12 +99,9 @@ const useStyles = makeStyles()(() => ({
     },
   },
 }));
-
-const appName = process.env.APP_NAME || 'MPDX';
-
 export const emptyAddress: ContactAddressFragment = {
   id: 'new',
-  source: appName,
+  source: '',
```
dr-bizz, line 104: Does this mean we're now not saving the source when a new address is made? Or does it default to the appName on the server?

dr-bizz (approved):
I left two comments. Nothing crazy. This is looking great! I'm going to approve to unblock you.

I see that the phone numbers not changing the primary when selected was happening on staging as well, so maybe the that bug was in production.

dr-bizz:
I guess I've been using the star feature wrong. In the angular, I'm positive that mailing addresses updated the primary on click, which is why I built it that way.

src/components/Tool/FixMailingAddresses/FixMailingAddresses.tsx
jbirdjavi: @caleballdrin "MPDX" is the default value for addresses, email_addresses, contacts, and phone_numbers. So if you're not sending one, it'll default to that.

caleballdrin: Ok that is good to know. Thanks!

src/utils/sourceHelper.ts
caleballdrin: We don't need that anymore because we aren't checking for both the current `manualSourceValue` and the `oldManualSourceValue`.

PR #1121, opened 2024-10-08
Fix Late Commitments

caleballdrin:
## Description
https://secure.helpscout.net/mailbox/2c1fe7d687998a44/7669074/
https://secure.helpscout.net/conversation/2725829190/1238582?folderId=7669074

Issue: 
Right now the API returns the first 3 late commitments (anything that is even 1 day late). Then the FE hides late commitments that are less than 7 days late. So sometimes those 3 late commitments get filtered out and it shows that there are no late commitments.

Fix:
- Remove front-end filtering
- Make the API only get commitments that are at least 30 days late. (This is how angular did it)

## Checklist:

### DIFF
--- a/src/components/Dashboard/ThisWeek/GetThisWeek.graphql
+++ b/src/components/Dashboard/ThisWeek/GetThisWeek.graphql
@@ -1,7 +1,7 @@
 query GetThisWeek(
   $accountListId: ID!
   $endOfDay: ISO8601DateTime!
-  $today: ISO8601Date!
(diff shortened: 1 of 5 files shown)


caleballdrin:
> What if we only filtered if there are at least 3 results?

We could. But I think this way is more simple. The angular site only showed commitments that were at least 30 days late. And then it also linked to the contacts page with a filter for contacts at least 30 days late.

canac (approved):
Thanks for the fix! I like that we're letting the server do the filtering work.

PR #1126, opened 2024-10-10
MPDX-8347 Round converted donations

caleballdrin:
## Description
https://jira.cru.org/browse/MPDX-8347
Angular rounds all donations to the nearest dollar. In React we stop rounding. However, we need to round the converted foreign currencies since they might not be accurate to .01. 

- Round donations if converted from another currency.
- Round total converted currency
- Round total donations if the contact has foreign donations
- Add a tooltip with the conversion rate

## Checklist:


canac (commented):
Great work! Can you add a test that covers the branch on 214?

Can you also add a code comment that mentions why were rounding? Something about how that the number displayed won't match the number in their account anyway due to variations in exchange rates so we're rounding to avoid confusion. Thank you!

canac (approved):

PR #1127, opened 2024-10-10
MPDX-8318 Financial Accounts Summary and Transactions pages

dr-bizz:
## Description
In this PR I add the Financial Accounts Summary and Transactions pages. There is a lot I do in this PR, but there are a lot of files which are just renamed.

1. Renamed "Responsibility Center" to "Financial Accounts" - The reason is that these pages allow you to view details on a certain Financial Account. Which is what the old Angular version had in the code and URL. "Responsibility Center" is the name by which users know it, but for us devs, it's known as "Financial Accounts", so to keep it consistent for us, I've renamed it.

2. I added a new context around the Financial Accounts pages since there's a lot of functionality that needs to be passed from parent to component.

3. Caleb and I created the framework that the two pages would sit on, this includes the Header and reports navigation.

4. I added the Summary page - This is the simplest of the 2, and shows a table without any filtering. _This table should have links to the transaction table, which will be coming today. I haven't had time to build the links, but it should be fairly easy._
- This needed a GraphQL query to be built with not much time, so I've built it with GraphQL Proxy.

5. I added the Transactions page - This page is dynamic; by default, it loads the transactions for the last month, but it also allows you to filter the transactions by date and category. This page features a search inbox, a table which is sortable and filterable and a button that allows you to export the data you see.
- This also needs a GQL proxy to fetch the entries (transactions).

### Things I still need to do:
As time is running out, since we are putting this live on Friday, I wanted to get this PR up, so people can review it, while I still work on the following items.
- Build the links from the summary page to transactions.
- Build the tests for the transactions page.

## Checklist:


canac (commented):
This is really great work so far in such a quick time! You might be aware of some of these already, but here are some bugs/discrepancies I noticed. Let me know if I can clarify any of them:

_I used the "RUSSIA ALL Ministries RC (local + Europe)" account list in my testing._

* The accounts should be sorted alphabetically on the main financial accounts list
* Crossroads Samara is unchecked in Angular and checked in React
* Last synced dates are in the past in Angular but are all today in React
* Graph and average is different for Crescendo in Angular vs React
* In Summary, date links in header and category line items aren't clickable links
* In Summary for Crescendo, "Nonlocal contr assessments-RUS" numbers are all negative in Angular but are positive in React
* In Transactions, oldest transactions are at the bottom in Angular but are at the top in React
* In Transactions, search placeholder doesn't match the tooltip
* In Transactions CSV export, the transactions are also reversed
* In Transactions CSV export, the currency amounts include `\xA0` characters instead of regular spaces, which causes those cells to render strangely in Excel
* After clicking on an account there's no way to go back to the financial accounts list

pages/api/Schema/reports/financialAccounts/financialAccounts/financialAccounts.graphql
```diff
@@ -0,0 +1,35 @@
...
+}
+
+type FinancialAccountSummaryResponse {
+  id: ID!
+  closingBalance: String
+  credits: String
+  debits: String
+  difference: String
+  endDate: String!
+  openingBalance: String
```
canac, line 19: Should these numbers be `Float`s instead? Or are we using a `String` because it is a string in the REST API?

src/components/Reports/FinancialAccountsReport/Context/FinancialAccount.graphql
```diff
@@ -0,0 +1,22 @@
...
+    categories {
+      nodes {
+        id
+        code
+        entryType
+        name
+      }
+    }
+    organization {
+      name
```
canac, line 19: Does `organization` have an `id` field?

src/components/Reports/FinancialAccountsReport/AccountSummary/AccountSummary.tsx
```diff
@@ -0,0 +1,384 @@
...
+                {/* Income header */}
+                <TableRow sx={{ background: theme.palette.mpdxGreen.main }}>
+                  <StyledTableCell>
+                    <Typography variant="body1">{t('Income')}</Typography>
+                  </StyledTableCell>
+                  {tableData.periods?.map((_, idx) => {
+                    return (
+                      <StyledTableCell
+                        key={`income-${idx}`}
+                        align="right"
```
canac, line 245: This shouldn't need styles since it's empty, right?

src/components/Reports/FinancialAccountsReport/AccountSummary/AccountSummary.tsx
```diff
@@ -0,0 +1,384 @@
...
+                {/* Expenses header */}
+                <TableRow sx={{ background: '#FF7679' }}>
+                  <StyledTableCell sx={{ color: theme.palette.common.white }}>
+                    <Typography variant="body1">{t('Expenses')}</Typography>
+                  </StyledTableCell>
+                  {tableData.periods?.map((_, idx) => {
+                    return (
+                      <StyledTableCell
+                        key={`expenses-${idx}`}
+                        align="right"
```
canac, line 297: Also empty

src/components/Reports/FinancialAccountsReport/AccountTransactions/AccountTransactions.tsx
```diff
@@ -0,0 +1,171 @@
...
+      return;
+    }
+    const columnHeaders = [
+      t('Date'),
+      t('Payee'),
+      t('Memo'),
+      t('Outflow'),
+      t('Inflow'),
+    ];
+    const convertDataToArray = data.financialAccountEntries.entries.reduce(
```
canac, line 103: Can this just be a `.map` instead of a reduce since we're creating one row per entry?

pages/accountLists/[accountListId]/reports/financialAccounts/[[...financialAccount]].page.tsx
```diff
@@ -0,0 +1,172 @@
+import Head from 'next/head';
```
Comment (canac) [inline pages/accountLists/[accountListId]/reports/financialAccounts/[[...financialAccount]].page.tsx:1]: Is there a reason we're having this one page handle all financial accounts pages instead of having an `index.page.tsx` for the list, `[financialAccountId].page.tsx` for the summary, and `[financialAccountId]/entries.page.tsx` for the entries?

src/components/Reports/FinancialAccountsReport/FinancialAccounts/FinancialAccounts.tsx
```diff
@@ -17,43 +11,40 @@ import {
...
+  FinancialAccountContext,
+  FinancialAccountType,
+} from '../Context/FinancialAccountsContext';
 import { useEntryHistoriesQuery } from './GetEntryHistories.generated';
 import {
   FinancialAccountsDocument,
   FinancialAccountsQuery,
   useFinancialAccountsQuery,
 } from './GetFinancialAccounts.generated';
 import { useSetActiveFinancialAccountMutation } from './SetActiveFinancialAccount.generated';
```
canac, line 25: Is there a reason to not put these three queries in the same file? I like having one `.graphql` file per component vs one file per operation.

canac (commented):
Again, really great work with this! Feel free to push back on any of my comments or wait until later to do them since I know we're trying to get this out this week.

dr-bizz:
> * The accounts should be sorted alphabetically on the main financial accounts list - ✅
> * Crossroads Samara is unchecked in Angular and checked in React - This is returning active true from the API. - ⚠️ Need API work
> * Last synced dates are in the past in Angular but are all today in React - ✅
> * Graph and average is different for Crescendo in Angular vs React - ✅
> * In Summary, date links in header and category line items aren't clickable links - ✅

pages/api/Schema/reports/financialAccounts/financialAccounts/financialAccounts.graphql
dr-bizz: They return from the REST API as strings

src/components/Reports/FinancialAccountsReport/AccountSummary/AccountSummary.tsx
dr-bizz: This table row needs styles and empty rows to go all away across the table

src/components/Reports/FinancialAccountsReport/AccountSummary/AccountSummary.tsx
dr-bizz: Thi table row (Opening Balance) isn't empty, I assume you mean the Income table row. This table row needs styles and empty rows to go all away across the table

pages/accountLists/[accountListId]/reports/financialAccounts/[[...financialAccount]].page.tsx
Comment (dr-bizz) [reply pages/accountLists/[accountListId]/reports/financialAccounts/[[...financialAccount]].page.tsx:1]: It's more of a preference. I'm okay to split it up into 3 pages, but that will be done in another PR, so we can get this live.

src/components/Reports/FinancialAccountsReport/AccountSummary/AccountSummary.tsx
canac: Sorry if my comment showed up in a weird place and wasn't clear. Yes, I meant the income header. I'm asking why we have `align="right"` on this cell and the expenses cell since the cell doesn't have any content?

src/components/Reports/FinancialAccountsReport/AccountSummary/AccountSummary.tsx
canac: > This table row needs styles and empty rows to go all away across the table

Definitely! I'm just wondering why we have `align="right"` on an empty cell.

src/components/Reports/FinancialAccountsReport/AccountTransactions/AccountTransactions.tsx
```diff
@@ -124,10 +124,22 @@ export const AccountTransactions: React.FC = () => {
           entry.description ?? '',
           entry.category?.name ?? entry.category?.code ?? '',
           entry.type === FinancialAccountEntryTypeEnum.Debit
-            ? currencyFormat(formatNumber(entry.amount), entry.currency, locale)
+            ? currencyFormat(
+                formatNumber(
+                  entry.amount.replace(/[\xA0\u2000-\u200B\uFEFF]/g, ''),
```
canac: For some reason this didn't work. There are still strange characters in the export.

canac (commented):
Incredible work with all of this!

The only other thing I noticed is that the Export CSV button doesn't do anything while the transactions are loading. Can we disable it?

src/components/Reports/FinancialAccountsReport/AccountTransactions/AccountTransactions.tsx
canac: I think that the `replace` needs to be on what `currencyFormat` returns because `currencyFormat` is the function creating the weird characters.

src/components/Reports/FinancialAccountsReport/AccountTransactions/AccountTransactions.tsx
```diff
@@ -0,0 +1,197 @@
...
+  useEffect(() => {
+    if (!hasActiveFilters && !query?.filters) {
+      setActiveFilters({
+        dateRange: {
+          min: defaultStartDate,
+          max: defaultEndDate,
+        },
+      });
+    }
+  }, [hasActiveFilters]);
```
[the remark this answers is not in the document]
dr-bizz: I didn't want it to be, as every time query?.filters changes, it would re-run this useEffect, which it should only run on `onMount` and `offMount`.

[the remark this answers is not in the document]
dr-bizz: This was a bug I was struggling with yesterday. Adding it doesn't seem to cause an issue, so I might leave it in.

src/components/Reports/FinancialAccountsReport/AccountTransactions/AccountTransactions.tsx
canac: Sorry, do you think we should replace the weird spaces with regular spaces instead of removing them? I think "RUB 100" would be better than "RUB100" in the export.

src/components/Reports/FinancialAccountsReport/AccountTransactions/AccountTransactions.tsx
```diff
@@ -0,0 +1,196 @@
...
+        dateRange,
+        categoryId,
+        wildcardSearch,
+      },
+    },
+  });
+
+  const handleExportCSV = () => {
+    if (!data) {
+      // Alert user there is no data
```
canac, line 112: Maybe we should pass a prop to `Header` to notify it that data is loading so that it can disable the export button?

dr-bizz: Ah sorry I missed this one

canac (approved):
There are some failing tests, but this looks really good! Amazing work! 🚀

dr-bizz:
I will create a new PR which adds all the tests for the Transactions page and components I have missed the tests due to the timeline of getting this out on production.

PR #1131, opened 2024-10-11
[MPDX-8360] Fix changing contact filters reloading the page

canac:
## Description

The root cause of buggy routing behavior throughout the app is that changes to the router don't seem to take effect immediately. For example, when clearing filters, one `useEffect` removes `filters` from the query and another `useEffect` removes `searchTerm` from the query. However, when the second effect runs, the change to the filters haven't propagated yet, and it ends up adding the filters back to the URL.

My solution, which has seemed to work well on the contacts page, is to put state that affects the URL (like filters and the search query) into a `useState`. Changes to that state will update immediately. The context wrapper then updates the URL as the state changes. Because the state variables are the source of truth instead of the URL, delays in the URL updating don't cause synchronization issues anymore.

I would like to eventually extract the logic in the wrapper component into another url filter context that can be used in contacts, tasks, partner giving analysis, and responsibility centers. But I'd like to wait on that for now and verify that this approach works better.

There is still buggy behavior on the appeals page when switching between list and flow mode. I don't believe that it is any more buggy than it was before, and it can be addressed in a future PR. It does seem like I've fixed the bugginess on the appeals list view when clicking the "Given", "Received", etc, filters ([MPDX-8239](https://jira.cru.org/browse/MPDX-8239)).

Selecting contacts and then clicking map view to only view those contacts doesn't work, but it didn't work before, so that can be done in a future PR.

I have also done some cleanup like combining variables and remove the extra parameters from `setContactFocus`. At one point, I thought that debouncing hook was buggy and wrote more tests to verify its implementation. The hook was actually correct, but I'm leaving the tests since that hook didn't have test coverage before.

[MPDX-8360](https://jira.cru.org/browse/MPDX-8360)

## Checklist:

### DIFF
--- a/pages/accountLists/[accountListId]/contacts/ContactsWrapper.tsx
+++ b/pages/accountLists/[accountListId]/contacts/ContactsWrapper.tsx
@@ -8,6 +8,7 @@ import {
   TaskFilterSetInput,
 } from 'src/graphql/types.generated';
 import { sanitizeFilters } from 'src/lib/sanitizeFilters';
+import { getQueryParam } from 'src/utils/queryParam';
 
 interface Props {
   children?: React.ReactNode;
@@ -57,16 +58,12 @@ export const ContactsWrapper: React.FC<Props> = ({ children }) => {
 
   const [activeFilters, setActiveFilters] = useState<
     ContactFilterSetInput & TaskFilterSetInput
-  >(
-    typeof query.filters === 'string'
-      ? JSON.parse(decodeURIComponent(query.filters))
-      : {},
-  );
+  >(JSON.parse(decodeURIComponent(getQueryParam(query, 'filters') ?? '{}')));
   const [starredFilter, setStarredFilter] = useState<
     ContactFilterSetInput & TaskFilterSetInput
   >({});
   const [searchTerm, setSearchTerm] = useState(
-    typeof query.searchTerm === 'string' ? query.searchTerm : '',
+    getQueryParam(query, 'searchTerm') ?? '',
   );
   const [filterPanelOpen, setFilterPanelOpen] = useState(true);
 
--- a/pages/accountLists/[accountListId]/contacts/ContactsWrapper.tsx
+++ b/pages/accountLists/[accountListId]/contacts/ContactsWrapper.tsx
@@ -1,4 +1,4 @@
-import { ParsedUrlQuery, ParsedUrlQueryInput } from 'node:querystring';
+import { ParsedUrlQuery } from 'node:querystring';
 import { useRouter } from 'next/router';
 import React, { useEffect, useMemo, useState } from 'react';
 import { ContactsProvider } from 'src/components/Contacts/ContactsContext/ContactsContext';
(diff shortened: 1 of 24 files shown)


dr-bizz (commented):
As we spoke about there seems to be a bug when switching from `list` to `flows` etc.

dr-bizz (changes requested):
This is looking great! The PR is pretty confusing as there are a lot of large changes which took me a while to review them. I will probably have to have another pass at it since I probably missed some things.

The code is looking good, I did spot an error on the Tasks filters when filtering, which will need to be fixed before we push this into production.

https://github.com/user-attachments/assets/7803c6fe-bb84-440a-9a01-8845368d5f98

canac:
> The code is looking good, I did spot an error on the Tasks filters when filtering, which will need to be fixed before we push this into production.

@dr-bizz Thanks for catching that. It should be fixed now.

Sorry that this code is still really hairy and hard to follow because the contacts context is split into context and wrapper components. Also, non-contacts pages use the contacts context. It could use some refactoring to pull out URL filters logic and contact panel logic into reusable context. Then we would only need contacts context on the contacts pages. But that will have to be a future improvement.

dr-bizz (approved):
I've tested this and it looks great

pages/accountLists/[accountListId]/tools/appeals/AppealsWrapper.tsx
```diff
@@ -81,24 +84,31 @@ export const AppealsWrapper: React.FC<Props> = ({ children }) => {
     });
   };
 
+  const setActiveFilters: AppealsContextProps['setActiveFilters'] = (value) => {
+    const filters = typeof value === 'function' ? value(activeFilters) : value;
```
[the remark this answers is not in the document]
Comment (canac) [reply pages/accountLists/[accountListId]/tools/appeals/AppealsWrapper.tsx:88]: It's just because `Dispatch<SetStateAction<_>>` means that you can pass a value or a function to the setter. But I don't think we're actually passing in a function. I could have also changed the type of `setActiveFilters` to `(filters: ContactFilterSetInput & TaskFilterSetInput) => void`. I might actually do that.

pages/accountLists/[accountListId]/contacts/ContactsWrapper.tsx
```diff
@@ -95,7 +98,7 @@ export const ContactsWrapper: React.FC<Props> = ({ children }) => {
       newQuery.searchTerm = encodeURIComponent(searchTerm);
     }
     return newQuery;
-  }, [accountListId, contactId, viewMode, activeFilters, searchTerm]);
+  }, [contactId, viewMode, activeFilters, searchTerm]);
```
Comment (dr-bizz) [inline pages/accountLists/[accountListId]/contacts/ContactsWrapper.tsx:101]: Should `addViewMode` be added as a dependency?

Comment (dr-bizz) [reply pages/accountLists/[accountListId]/contacts/ContactsWrapper.tsx:101]: I guess it won't change much

dr-bizz (approved):
The code is looking good, I'm going to review it on preview

dr-bizz:
⚠️ Might be best to push this live on Monday when we are all around for the next few days.

dr-bizz:
My bad; clicked the wrong button.

dr-bizz:
...I've caused quite a bit of drama 😅

canac:
After waiting for the branch to finish deleting, I recreated it again, so the old preview link will work again after it finishes building.

PR #1133, opened 2024-10-11
Helpscout - Allow users to accept invites

caleballdrin:
## Description
https://secure.helpscout.net/conversation/2731091026/1241412?folderId=7669074
- Accept account invites
- Redirect old links

## Checklist:


pages/account_lists/[accountListId]/accept_invite/[inviteId].page.tsx
```diff
@@ -0,0 +1,30 @@
+import { GetServerSideProps } from 'next';
+import { ReactNode } from 'react';
+import { getSession } from 'next-auth/react';
+import BaseLayout from 'src/components/Layouts/Basic';
+
+const InvitePage = (): ReactNode => null;
+
+InvitePage.layout = BaseLayout;
```
Comment (canac) [inline pages/account_lists/[accountListId]/accept_invite/[inviteId].page.tsx:8]: The page doesn't need a layout since it's never rendered.

pages/acceptInvite.page.tsx
```diff
@@ -0,0 +1,99 @@
...
+    }
+    const { accountListId, inviteCode, accountInviteId, orgInviteId, orgId } =
+      router.query;
+    const acceptInvite = async (id, code, url) => {
+      const inviteType = url.includes('organizations')
+        ? 'organization_invites'
+        : 'account_list_invites';
+
+      try {
+        const session = await getSession();
```
canac, line 29: You can do useRequiredSession to get the token. Can you move the fetch to a helper method outside of the component?

caleballdrin: For moving the fetch, are you thinking I just pass all the props to it? Like `apiToken`, `code`, `url`, `id`

pages/acceptInvite.page.tsx
```diff
@@ -0,0 +1,132 @@
...
+        } else {
+          enqueueSnackbar(t('Accepted invite successfully.'), {
+            variant: 'success',
+          });
+          router.push(`/accountLists/${currentUserAccountListId}`);
+        }
+      } catch (err) {
+        enqueueSnackbar(
+          t(
+            'Unable to accept invite. Try asking the account holder to resend the invite.',
```
dr-bizz, line 96: The organization error should say:

```
Unable to accept invite. Try asking the organization admin to resend the invite
```

pages/acceptInvite.page.tsx
```diff
@@ -0,0 +1,132 @@
...
+    const inviterAccountListId = router.query.accountListId || undefined;
+    const acceptInvite = async (id, code, url) => {
+      const inviteType = url.includes('organizations')
+        ? 'organization_invites'
+        : 'account_list_invites';
+
+      try {
+        const apiToken = session?.apiToken;
+
+        const response = await fetchAcceptInvite({
```
dr-bizz, line 69: In MPDX_web it does a check to see if `id, code or url` are empty and if so they return an error. I assume the request won't work so it would result in the same outcome.

Also, you should add the prop types of the variables to the acceptInvite function.

pages/acceptInvite.page.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+        if (!response.ok) {
+          throw new Error('Network response was not ok');
+        }
+        if (url.includes('organizations')) {
+          enqueueSnackbar(t('Accepted invite successfully.'), {
+            variant: 'success',
+          });
+          router.push(
+            `/accountLists/${currentUserAccountListId}/settings/preferences`,
```
dr-bizz, line 85: Shouldn't this redirect to the `organizations` page on settings?

pages/acceptInvite.page.tsx
```diff
@@ -0,0 +1,132 @@
...
+    if (accountInviteId && inviteCode && inviterAccountListId) {
+      const url = `account_lists/${inviterAccountListId}/invites/${accountInviteId}/accept`;
+      acceptInvite(accountInviteId, inviteCode, url);
+      hasFetchedRef.current = true;
+    } else if (orgInviteId && inviteCode && orgId) {
+      const url = `organizations/${orgId}/invites/${orgInviteId}/accept`;
+      acceptInvite(orgInviteId, inviteCode, url);
+      hasFetchedRef.current = true;
+    }
+  }, [router.query]);
```
dr-bizz, line 114: Whst if they have none of these query parameters, should we add a warning snackbar to warn the users they don't have the correct paramters?

dr-bizz (approved):
Looking great! I left a few comments, but that is it. Great work. I don't want to block you so I will approve this.

pages/acceptInvite.page.tsx
```diff
@@ -0,0 +1,132 @@
...
+  inviteType,
+  id,
+  code,
+}: FetchAcceptInviteProps): Promise<Response> => {
+  return fetch(process.env.REST_API_URL + url, {
+    method: 'PUT',
+    headers: {
+      authorization: `Bearer ${apiToken}`,
+      'content-type': 'application/vnd.api+json',
+      accept: 'application/vnd.api+json',
```
canac, line 32: I don't think that the `Accept` header is necessary. None of our other fetch calls to the API server add it.

pages/acceptInvite.page.tsx
```diff
@@ -0,0 +1,132 @@
...
+  // Ref to track if the API call has already been triggered
+  const hasFetchedRef = useRef(false);
+
+  useEffect(() => {
+    if (!router.isReady || hasFetchedRef.current) {
+      return;
+    }
+    const { inviteCode, accountInviteId, orgInviteId, orgId } = router.query;
+    const inviterAccountListId = router.query.accountListId || undefined;
+    const acceptInvite = async (id, code, url) => {
```
canac, line 61:
```suggestion
    const acceptInvite = async (id: string, code: string, url: string) => {
```

pages/acceptInvite.page.tsx
```diff
@@ -0,0 +1,132 @@
...
+      const inviteType = url.includes('organizations')
+        ? 'organization_invites'
+        : 'account_list_invites';
+
+      try {
+        const apiToken = session?.apiToken;
```
canac, line 67: One benefit of using `useRequiredSession` is that session is guaranteed to not be null/undefined.

```suggestion
        const apiToken = session.apiToken;
```

pages/acceptInvite.page.tsx
```diff
@@ -0,0 +1,132 @@
...
+  });
+};
+
+const AcceptInvitePage = (): ReactElement => {
+  const { t } = useTranslation();
+  const { appName } = useGetAppSettings();
+  const { enqueueSnackbar } = useSnackbar();
+  const router = useRouter();
+  const session = useRequiredSession();
+  const currentUserAccountListId = useAccountListId();
```
canac, line 50: This hook pulls the account list id from the URL, but this page doesn't have `accountListId` in the URL, so it will always be undefined.

pages/acceptInvite.page.tsx
canac: Angular shows an error toast and redirects to the dashboard

https://github.com/CruGlobal/mpdx_web/blob/f02ea571631feb3ff6f52e17f454d4c4af3f2896/src/acceptInvite/acceptInvite.component.ts#L107-L111

pages/acceptInvite.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+      enqueueSnackbar: mockEnqueue,
+    };
+  },
+}));
+
+describe('AcceptInvitePage', () => {
+  const mockPush = jest.fn();
+
+  beforeEach(() => {
+    (useRouter as jest.Mock).mockReturnValue({
```
canac, line 35: Can we use `<TestRouter router={router}>` instead of mocking `useRouter`?

pages/acceptInvite.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+  // @ts-ignore
+  ...jest.requireActual('notistack'),
+  useSnackbar: () => {
+    return {
+      enqueueSnackbar: mockEnqueue,
+    };
+  },
+}));
+
+describe('AcceptInvitePage', () => {
```
canac, line 31: Great test coverage!

pages/account_lists/[accountListId]/accept_invite/[inviteId].page.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const redirectURL =
+    accountListId && inviteId && code
+      ? `/acceptInvite?accountListId=${accountListId}&accountInviteId=${inviteId}&inviteCode=${code}`
+      : `/accountLists/${accountListId}`;
+
+  return {
+    redirect: {
+      destination: session ? redirectURL : '/login',
+      permanent: true,
+    },
```
Comment (canac) [inline pages/account_lists/[accountListId]/accept_invite/[inviteId].page.tsx:21]: Login redirects should technically be `permanent: false`.

```suggestion
    redirect: session ? {
      destination: redirectURL,
      permanent: true,
    } : {
      destination: '/login',
      permanent: false,
    },
```

pages/account_lists/[accountListId]/accept_invite/[inviteId].page.test.tsx
```diff
@@ -0,0 +1,59 @@
+import { GetServerSidePropsContext } from 'next';
+import { getSession } from 'next-auth/react';
+import { getServerSideProps } from './[inviteId].page';
+
+jest.mock('next-auth/react');
```
Comment (canac) [inline pages/account_lists/[accountListId]/accept_invite/[inviteId].page.test.tsx:5]: `__tests__/util/setup.ts` already mocks `getSession`. I would remove all the session mocking code, `import { session } from '__tests__/fixtures/session';`, and just assert that the session in the response matches that mocked session. Since we don't really care about what the session is in this code.

canac (commented):
Great work! I think the most important thing from my comments is removing `useAccountListId()` because unless I'm mistaken it will either always return undefined or might even be returning the inviter's account list id. It might be working because of the default account list redirect logic you implemented. But if we're relying on that, let's make it explicit and redirect the user to `/accountLists/_/settings/preferences` (with an underscore as an intentionally invalid account list id or some other clearly invalid id) and add a code comment to document what we're doing.

pages/acceptInvite.page.tsx
canac: I think so!

pages/acceptInvite.page.tsx
caleballdrin: Yeah, it should go to integrations and the organizations tab. I'll change it.

pages/acceptInvite.page.tsx
```diff
@@ -81,19 +90,24 @@ const AcceptInvitePage = (): ReactElement => {
...
         } else {
           enqueueSnackbar(t('Accepted invite successfully.'), {
             variant: 'success',
           });
-          router.push(`/accountLists/${currentUserAccountListId}`);
+          router.push(dashboardLink);
         }
       } catch (err) {
+        const inviter = url.includes('organizations')
+          ? 'organization admin'
```
canac, line 105: These two labels need to be localized

PR #1135, opened 2024-10-14
Removing nonNull items as they are causing errors when they are null …

dr-bizz:
## Description
I removed the `nonNull` items from GraphQL as some items are null for some users, which causes the page to error and not render the page.

It now seems to work on local.

Helpscout Tix: https://secure.helpscout.net/conversation/2732656504/1242288?folderId=7296147

## Checklist:


caleballdrin (approved):
Looks good. This is also blocking me from being able to push code because of the ts errors.

PR #1138, opened 2024-10-16
RC: Add tests to transactions page and components

dr-bizz:
## Description
I've added tests to the transactions page and components. I was meant to add these tests to a previous PR but due to time I had to launch without the tests.

I also fix a few errors which have been raised in a helpscout ticket:
https://secure.helpscout.net/conversation/2732656504/1242288?folderId=7296147

## Checklist:

### DIFF
--- a/pages/accountLists/[accountListId]/reports/financialAccounts/[[...financialAccount]].page.test.tsx
+++ b/pages/accountLists/[accountListId]/reports/financialAccounts/[[...financialAccount]].page.test.tsx
@@ -1,0 +1,160 @@
+import React from 'react';
+import { ThemeProvider } from '@mui/material/styles';
+import { LocalizationProvider } from '@mui/x-date-pickers';
+import { AdapterLuxon } from '@mui/x-date-pickers/AdapterLuxon';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { SnackbarProvider } from 'notistack';
+import { I18nextProvider } from 'react-i18next';
+import TestRouter from '__tests__/util/TestRouter';
+import { GqlMockedProvider } from '__tests__/util/graphqlMocking';
+import {
+  defaultFinancialAccount,
+  defaultFinancialAccountSummary,
+} from 'src/components/Reports/FinancialAccountsReport/AccountSummary/AccountSummaryMock';
+import { FinancialAccountSummaryQuery } from 'src/components/Reports/FinancialAccountsReport/AccountSummary/financialAccountSummary.generated';
+import { FinancialAccountQuery } from 'src/components/Reports/FinancialAccountsReport/Context/FinancialAccount.generated';
+import { FinancialAccountsQuery } from 'src/components/Reports/FinancialAccountsReport/FinancialAccounts/FinancialAccounts.generated';
+import { FinancialAccountsMock } from 'src/components/Reports/FinancialAccountsReport/FinancialAccounts/FinancialAccountsMocks';
+import i18n from 'src/lib/i18n';
+import theme from 'src/theme';
+import FinancialAccountsPage from './[[...financialAccount]].page';
+
+const accountListId = 'account-list-1';
+const financialAccountId = 'financialAccountId';
+const defaultRouter = {
+  query: { accountListId },
+  isReady: true,
+};
+
+const Components = ({ router = defaultRouter }: { router?: object }) => (
+  <I18nextProvider i18n={i18n}>
+    <LocalizationProvider dateAdapter={AdapterLuxon}>
+      <SnackbarProvider>
+        <ThemeProvider theme={theme}>
+          <TestRouter router={router}>
+            <GqlMockedProvider<{
+              FinancialAccountSummary: FinancialAccountSummaryQuery;
+              FinancialAccount: FinancialAccountQuery;
+              FinancialAccounts: FinancialAccountsQuery;
+            }>
+              mocks={{
+                FinancialAccountSummary: defaultFinancialAccountSummary,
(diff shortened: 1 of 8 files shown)


pages/accountLists/[accountListId]/reports/financialAccounts/[[...financialAccount]].page.test.tsx
```diff
@@ -0,0 +1,160 @@
...
+    const { findByText, findByRole, getByText, queryByText, queryByRole } =
+      render(
+        <Components
+          router={{
+            query: {
+              accountListId,
+              financialAccount: [financialAccountId, 'entries'],
+            },
+            isReady: true,
+          }}
```
Comment (canac) [inline pages/accountLists/[accountListId]/reports/financialAccounts/[[...financialAccount]].page.test.tsx:96]: This router is used a few times. You could consider extracting it into an `entriesRouter`.

```ts
const entriesRouter = {
  ...defaultRouter,
  query: {
    ...defaultRouter.query,
    financialAccount: [financialAccountId, 'entries'],
  },
};
```

src/components/Reports/FinancialAccountsReport/AccountTransactions/AccountTransactions.test.tsx
```diff
@@ -0,0 +1,346 @@
...
+      expect(setPanelOpen).toHaveBeenCalledWith(Panel.Filters);
+    });
+
+    it('should set filters to default date range if no activeFilters or url filters', () => {
+      render(<Components />);
+
+      expect(setActiveFilters).toHaveBeenCalledWith({
+        dateRange: {
+          min: defaultStartDate,
+          max: defaultEndDate,
```
canac, line 133: Instead of using the exported default start and end date, I think I'd prefer calculating what we expect the start date and end date to be based on a today of January 1, 2020 (mocked in `setup.ts`). That way we're also testing that `defaultStartDate`/`defaultStartEnd` date are calculated correctly. But feel free to disagree.

src/components/Reports/FinancialAccountsReport/AccountTransactions/AccountTransactions.tsx
```diff
@@ -27,6 +26,25 @@ const Container = styled(Box)(() => ({
...
+  isExpense?: boolean,
+): number => {
+  if (!amount) {
+    return 0;
+  }
+
+  if (amount === '0' || isExpense) {
+    return Number(amount);
+  }
+  return amount?.[0] === '-' ? Number(amount.substring(1)) : -Number(amount);
```
canac, line 45: Is there a reason this isn't just `-Number(amount)`? Because either way we're flipping the sign of `amount`, right?

src/components/Reports/FinancialAccountsReport/AccountTransactions/AccountTransactions.tsx
```diff
@@ -118,28 +136,25 @@ export const AccountTransactions: React.FC = () => {
...
+        return [
+          ...array,
+          [
+            entry.entryDate
+              ? dateFormatShort(DateTime.fromISO(entry.entryDate), locale)
+              : 'No entry date',
+            entry.description ?? '',
+            entry.category?.name ?? entry.category?.code ?? '',
+            entry.type === FinancialAccountEntryTypeEnum.Debit
+              ? `${formatAmountForExport(entry.amount, true)}`
```
canac, line 150: We're intentionally removing the currency from these numbers, right? If different entries can have different curriencies is that a good idea? Or do we know that they all end up being the same currency in practice?

canac (commented):
I have a couple questions about the code that changed, but the tests all look great! Amazing work! :100:

pages/accountLists/[accountListId]/reports/financialAccounts/[[...financialAccount]].page.test.tsx
Comment (dr-bizz) [reply pages/accountLists/[accountListId]/reports/financialAccounts/[[...financialAccount]].page.test.tsx:96]: Good catch

src/components/Reports/FinancialAccountsReport/AccountTransactions/AccountTransactions.tsx
dr-bizz: I believe each financial account has its own currency. This came up as a recommendation from Helpscout since the old MPDX also removes the currency

src/components/Reports/FinancialAccountsReport/AccountTransactions/AccountTransactions.tsx
dr-bizz: Good catch!

src/components/Reports/FinancialAccountsReport/AccountTransactions/AccountTransactions.tsx
```diff
@@ -27,6 +26,25 @@ const Container = styled(Box)(() => ({
...
+  }
+
+  if (amount === '0' || isExpense) {
+    return Number(amount);
+  }
+  return amount?.[0] === '-' ? Number(amount.substring(1)) : -Number(amount);
+};
+
 const formatDateRange = (startDate?: DateTime, endDate?: DateTime) => {
   if (!startDate) {
```
canac, line 49: I just noticed that we're reassigning the arguments here. Can you make a different variable? Or give them a default value since that's the intent of this code (this might be a bad idea since the start date expression is a little complex).

```ts
const formatDateRange = (
  startDate: DateTime = DateTime.local().minus({ months: 1 }).plus({ days: 1 }),
  endDate: DateTime = DateTime.local(),
) => `${startDate.toISODate()}..${endDate.toISODate()}`;
```

canac (approved):
Did you want to [change this](https://github.com/CruGlobal/mpdx-react/pull/1138#discussion_r1803368971) while you're working in this code?

PR #1139, opened 2024-10-17
MPDX-8373 Task Row

caleballdrin:
## Description
https://jira.cru.org/secure/RapidBoard.jspa?rapidView=3&view=detail&selectedIssue=MPDX-8373&sprint=1276#
- Move duplicate Task Phase/Action text to new component
- Add tags and tag icon hover based on screen width
- Add assignee Avatar with tooltip
- Make task subject shorten with ellipses on main task list
- Task List has condensed double lines on all screen sizes

Other unrelated fixes
- Add hover to check boxes on Contact Task rows
- Add selected count to Contact Tasks tab
- Stop autofilling assignee when editing a task that was created without one.

## Checklist:


src/components/Task/TaskRow/TaskActionPhase.tsx
```diff
@@ -0,0 +1,64 @@
...
+  fontSize: '14px',
+  letterSpacing: '0.25',
+  whiteSpace: 'nowrap',
+  fontWeight: 400,
+  marginRight: theme.spacing(0.5),
+}));
+
+interface TaskActionPhaseProps {
+  activityData: ActivityData | null | undefined;
+  activityType: ActivityTypeEnum | undefined | null;
```
canac, line 40: Just to match the prop above.

```suggestion
  activityType: ActivityTypeEnum | null | undefined;
```

src/components/Task/TaskRow/TaskActionPhase.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <SubjectWrapInner
+        data-testid="phase-action-wrap"
+        style={{
+          minWidth: '92px',
+          flexDirection: 'column',
+        }}
+      >
+        <TaskPhase>{activityData?.phase ? activityData.phase : ''}</TaskPhase>
+        <TaskType>
+          {activityType ? getLocalizedTaskType(t, activityType) : ''}
```
canac, line 59: Looks like `getLocalizedTaskType` already does the falsy check.

```suggestion
          {getLocalizedTaskType(t, activityType)}
```

pages/accountLists/[accountListId]/tasks/[[...contactId]].page.tsx
```diff
@@ -406,8 +406,8 @@ const TasksPage: React.FC = () => {
                             onTaskCheckToggle={toggleSelectionById}
                             isChecked={isRowChecked(task.id)}
                             useTopMargin={index === 0}
-                            contactDetailsOpen={contactDetailsOpen}
                             removeSelectedIds={deselectMultipleIds}
+                            filterPanelOpen={filterPanelOpen}
```
Comment (canac) [inline pages/accountLists/[accountListId]/tasks/[[...contactId]].page.tsx:410]: I think that this change is making the task rows always be in condensed mode, even with a large browser window and without having the contact details open.

canac (commented):
These are some great improvements! My main concern is that its now showing the task rows in condensed mode even without the contact details panel open.

pages/accountLists/[accountListId]/tasks/[[...contactId]].page.tsx
Comment (caleballdrin) [reply pages/accountLists/[accountListId]/tasks/[[...contactId]].page.tsx:410]: Yes, that change was intentional. It seems like the condensed view is also useful for small or medium screens. Especially if we're trying to display the same amount of data that the old UI did.

I was thinking we could just leave the condensed view for all screen sizes. But what do you think? Should we show the non-condensed view for extra large screen sizes?

Comment (caleballdrin) [reply pages/accountLists/[accountListId]/tasks/[[...contactId]].page.tsx:410]: I've asked Scott for his feedback.

Comment (caleballdrin) [reply pages/accountLists/[accountListId]/tasks/[[...contactId]].page.tsx:410]: I asked Scott and attached is his reply. 
I think the test account tends to have shorter task titles so it seems empty. But on my personal, the automated tasks that the API makes have very long titles and it is nice to have the room on larger screens.

Screenshot 2024-10-17 at 12 34 14 PM

Comment (canac) [reply pages/accountLists/[accountListId]/tasks/[[...contactId]].page.tsx:410]: Sounds good!

canac (approved):

src/components/Task/TaskRow/CommentTooltipText.tsx
```diff
@@ -0,0 +1,30 @@
...
+  comments: { id: string; body: string; updatedAt: string }[];
+}
+export const CommentTooltipText: React.FC<CommentTooltipTextProps> = ({
+  comments,
+}) => {
+  const locale = useLocale();
+
+  const latestComment = !!comments.length
+    ? comments[comments.length - 1]
+    : null;
```
canac, line 21: `.at(index)` is similar to `[index]` but it can also take negative numbers to index backwards from the end of the array, i.e. `array.at(-1)` is the last element, `array.at(-2)` is the second-to-last element, etc.

```suggestion
  const latestComment = comments.length.at(-1) ?? null;
```

src/components/Task/TaskRow/TaskRow.tsx
```diff
@@ -290,28 +301,50 @@ export const TaskRow: React.FC<TaskRowProps> = ({
...
                   numberOfComments={comments?.totalCount}
                   onClick={handleCommentButtonPressed}
                   onMouseEnter={() => preloadTaskModal(TaskModalEnum.Comments)}
-                  small
                 />
               </Box>
+            </Tooltip>
+          </Hidden>
+
+          <Hidden mdUp>
```
canac, line 324: Is this the same as the above, just with the `small` prop on `TaskCommentsButton`? If so, maybe we could do a media query to determine whether `small` should be true or false instead of having to duplicate all the components. I didn't look super carefully, so there might be more differences I missed.

caleballdrin: Great point! I didn't notice how similar they were.

PR #1141, opened 2024-10-17
Redirect-to-RC-when-switch-profiles

dr-bizz:
## Description
In this PR, I fix serval bugs and improve the codebase.

### Fixes:
Fixed bug when you switch your account on the Summary or Transaction RC pages and receive a error message. Now when you switch, it now redirects you back to the main RC page.

Fixed URL sending user to the entries page. (Simple rename of url)

### Improvements
I update the RC pages so they use 3 seperate pages instead of one catch all page. - This should help to prevent unwanted rerenders.

Since we only wanted the transactions page to use context, I had to alter components that are used on the Summary or accounts page to not use context. - This is where a lot of the changes come in.

FYI, I have merged in my recent Tests and changes which caused a lot of merge conflicts, and I had to alter some code.

## Checklist:


pages/accountLists/[accountListId]/reports/financialAccounts/[financialAccountId]/Wrapper.tsx
```diff
@@ -32,28 +33,20 @@ export const FinancialAccountsWrapper: React.FC<Props> = ({ children }) => {
...
     }
-    const length = financialAccount.length;
-    setFinancialAccountId(financialAccount[0]);
-    if (length === 1) {
-      setPage(FinancialAccountPageEnum.AccountSummaryPage);
-    } else if (length > 1) {
-      setPage(FinancialAccountPageEnum.AccountTransactionsPage);
+    if (typeof query.financialAccountId === 'string') {
+      setFinancialAccountId(query.financialAccountId);
     }
```
Comment (canac) [inline pages/accountLists/[accountListId]/reports/financialAccounts/[financialAccountId]/Wrapper.tsx:48]: You could use `getQueryParam` here to match the `useState`.

canac (approved):
I think this is really good! I briefly tested each of the pages, and they still seem to be working well.

pages/accountLists/[accountListId]/reports/financialAccounts/[financialAccountId]/Wrapper.tsx
Comment (dr-bizz) [reply pages/accountLists/[accountListId]/reports/financialAccounts/[financialAccountId]/Wrapper.tsx:48]: I can use the `getQueryParam` but this causes a typescript error as it doesn't understand `query.financialAccountId` has a defined value, so I'm thinking of keeping it as `typeof query.financialAccountId === 'string'`

Comment (canac) [reply pages/accountLists/[accountListId]/reports/financialAccounts/[financialAccountId]/Wrapper.tsx:48]: That's fine

PR #1145, opened 2024-10-18
MPDX-8351 Add UseLocalizedConstants() hook

caleballdrin:
## Description
https://jira.cru.org/secure/RapidBoard.jspa?rapidView=3&view=detail&selectedIssue=MPDX-8351&sprint=1276#
- Add new `UseLocalizedConstants` hook and replace `getLocalizedStatus`
- Add API constants mock to the test setup and remove old `LoadConstants` mocks
- Move get `getContactStatusesByPhase` to the `useContactPartnershipStatus` hook
- Consolidate functions converting old contact statuses to a new helper file.

## Checklist:


src/utils/functions/convertContactStatus.ts
```diff
@@ -0,0 +1,74 @@
...
+// Convert a status string into a StatusEnum
+// Statuses may be lowercase and underscored (i.e. "never_contacted") after task phases lands
+// Statuses may be sentence case with spaces (i.e. "Never Contacted") before task phases lands
+export const convertStatus = (
+  status: string | null | undefined,
+): StatusEnum | null => {
+  const foundStatus =
+    Object.values(StatusEnum).find((value) => {
+      return String(value.toLowerCase()) === String(status?.toLowerCase());
+    }) || null;
```
canac, line 12: Is the `String` coercion or defaulting to `null` necessary? 

```suggestion
  const foundStatus = Object.values(StatusEnum).find(
    (value) => value.toLowerCase() === status?.toLowerCase(),
  );
```

src/components/Contacts/ContactsMap/ContactsMapPanel.tsx
```diff
@@ -118,7 +118,7 @@ export const ContactsMapPanel: React.FC<ContactMapsPanelProps> = ({
     .map(([statusKey, status]) => {
       return {
         key: statusKey,
-        title: status.translated,
+        title: status.name,
```
canac, line 121: Why not `status.translated`?

src/components/Settings/integrations/Google/Modals/EditGoogleAccountModal.test.tsx
```diff
@@ -510,13 +471,12 @@ describe('EditGoogleAccountModal', () => {
...
         },
       );
-      expect(mutationSpy.mock.calls[2][0].operation.operationName).toEqual(
-        'SyncGoogleAccount',
-      );
-      expect(mutationSpy.mock.calls[2][0].operation.variables.input).toEqual({
-        googleAccountId: googleAccount.id,
-        integrationName: 'calendar',
-        googleIntegrationId: googleIntegration.id,
+      expect(mutationSpy).toHaveGraphqlOperation('SyncGoogleAccount', {
```
canac, line 474: Thanks for taking the opportunity to switch to `toHaveGraphqlOperation` so that this test never breaks again from changing the number of GraphQL calls!

src/components/Shared/Filters/FilterPanel.tsx
```diff
@@ -155,6 +154,14 @@ export const FilterPanel: React.FC<FilterPanelProps & BoxProps> = ({
     FinancialAccountContext,
   ) as FinancialAccountType;
 
+  const matchFilterContactStatuses = (status: string | null | undefined) => {
```
canac, line 157: Could this use `convertContactStatus`?

src/components/Task/Modal/Form/Inputs/SuggestedContactStatus/SuggestedContactStatus.tsx
```diff
@@ -58,14 +59,13 @@ export const SuggestedContactStatus: React.FC<SuggestedContactStatusProps> = ({
...
-        contactStatuses,
-      ).map((s) => s.id);
+      );
       // Hide suggestedStatus if the suggested status is a partner care status, otherwise, show it
       return !disabledStatuses.includes(currentContactStatus);
     }
     // Show suggestedStatus if the contact has no status
     return true;
-  }, [contactStatuses, currentContactStatus]);
+  }, [currentContactStatus]);
```
canac, line 68: Can we add `suggestedContactStatus` to the dependency array? I think we also need to do `useCallback` on `getContactStatusesByPhase` and add `getContactStatusesByPhase` to the dependency array. Otherwise, `getContactStatusesByPhase` could be using the old constants data.

src/hooks/useContactPartnershipStatuses.ts
```diff
@@ -112,10 +111,19 @@ export const useContactPartnershipStatuses = () => {
       return { id: statusKey, ...s };
     });
 
+  const getContactStatusesByPhase = (
```
canac, line 114: Can you wrap this in `useCallback` with a dependency array of `[phases]`?

src/hooks/useLocalizedConstants.ts
```diff
@@ -0,0 +1,37 @@
...
+import {
+  ContactFilterStatusEnum,
+  PhaseEnum,
+  StatusEnum,
+} from 'src/graphql/types.generated';
+
+export const useLocalizedConstants = () => {
+  const constants = useApiConstants();
+
+  const getLocalizedContactStatus = (
```
canac, line 11: These should probably also be wrapped in `useCallback` with a dependency array of `[constants]`.

canac (commented):
Amazing work! I love how many lines were deleted, how our tests can be a little simpler now, and that you were able to use basically the same function and just return it from the hook now.

src/components/Shared/Filters/FilterPanel.tsx
caleballdrin: `ContactFilterStatusEnum` is different than `StatusEnum` because it includes `NULL`, `HIDDEN`, and `ACTIVE`. So this check is for those statuses that are missing from `StatusEnum`

canac: That makes sense. Can you remove the `String` coercion here too unless there's something I'm missing and that's necessary?

src/components/Shared/Filters/FilterPanel.tsx
```diff
@@ -155,6 +154,14 @@ export const FilterPanel: React.FC<FilterPanelProps & BoxProps> = ({
...
+    return (
+      Object.values(ContactFilterStatusEnum).find((value) => {
+        return String(value.toLowerCase()) === String(status?.toLowerCase());
+      }) || null
```
canac, line 161: Actually, since `value` is always uppercase, it might make more sense to convert `status` to uppercase and then compare them.

```suggestion
      Object.values(ContactFilterStatusEnum).find((value) => value === status?.toUpperCase()) || null
```

src/components/Task/Modal/Form/LogForm/TaskModalLogForm.test.tsx
```diff
@@ -160,14 +150,7 @@ describe('TaskModalLogForm', () => {
       <LocalizationProvider dateAdapter={AdapterLuxon}>
```
caleballdrin, line 150: @canac do you have any ideas why this test isn't passing? I'm guessing it times out before the `onClose` function is called when saving. Someone added a wait time of `25000` so I'm guessing it has had problems in the past. Could waiting for useApiConstants be making it take longer?

canac: If you get add `debug` to the result that `render` returns and call `console.log(debug(undefined, Infinity))` after clicking save, you can see that there's a `disabled` attribute on the save button. My guess is that something about the form isn't valid.

src/components/Contacts/ContactsMap/ContactsMapPanel.tsx
caleballdrin: Both `status.translated` and `status.name` are the same for normal statuses. But Filter Contact Statuses have
different values for each.
`    NULL: {
      name: 'null',
      translated: t('-- None --'),
      phase: null,
    },
    ACTIVE: {
      name: 'active',
      translated: t('-- All Active --'),
      phase: null,
    },
    HIDDEN: {
      name: 'hidden',
      translated: t('-- All Hidden --'),
      phase: null,
    },`

canac: That makes sense

canac (commented):
This looks really good! I think once we refresh the constants when the language changes, this will be good to go.

src/components/Settings/preferences/accordions/LanguageAccordion/LanguageAccordion.tsx
```diff
@@ -55,6 +57,7 @@ export const LanguageAccordion: React.FC<LanguageAccordionProps> = ({
         enqueueSnackbar(t('Saved successfully.'), {
           variant: 'success',
         });
+        setLocaleStorageLanguage(attributes.locale || '');
```
canac, line 60: My hope is that this wouldn't be necessary. Changing the preference load the new preference from the server, which will update the locale in the Apollo cache, which will cause `user.preferences.locale` to update in `UseApiConstants`.

src/components/Constants/UseApiConstants.tsx
```diff
@@ -7,9 +9,22 @@ import {
...
+  const { data, refetch } = useLoadConstantsQuery({
     fetchPolicy: 'cache-first',
   });
 
+  const userSavedLanguage = data?.user.preferences?.locale;
+  const [localeStorageLanguage, setLocaleStorageLanguage] = useLocalStorage(
+    `user-language`,
+    '',
+  );
+  useEffect(() => {
```
canac, line 21: This is great! The only thing I'd suggest is making sure that this effect doesn't run when `userSavedLanguage` is undefined because it hasn't finished loading yet.

src/components/Constants/UseApiConstants.tsx
```diff
@@ -7,9 +9,22 @@ import {
...
   | LoadConstantsQuery['constant']
   | undefined => {
-  const { data } = useLoadConstantsQuery({
+  const { data, refetch } = useLoadConstantsQuery({
     fetchPolicy: 'cache-first',
   });
 
+  const userSavedLanguage = data?.user.preferences?.locale;
+  const [localeStorageLanguage, setLocaleStorageLanguage] = useLocalStorage(
+    `user-language`,
```
canac, line 18: What do you think about this to make it a little more clear that this value is supposed to be the language of the cached constants?

```suggestion
    'constants-language',
```

src/components/Constants/UseApiConstants.tsx
```diff
@@ -7,9 +9,22 @@ import {
...
+    '',
+  );
+  useEffect(() => {
+    // if the language in locale storage is different than the saved language, the user may have changed languages in a different browser. If so, refetch the constants.
```
canac, line 22: This comment is definitely true. But this effect is also responsible for refetching the language when the user changes the preference in their own browser (not just a different browser).

canac (commented):
Code looks great and is exactly what I had in mind! I'd like to spend some time testing it in preview, then I'll approve it.

src/components/Constants/UseApiConstants.tsx
```diff
@@ -14,14 +14,15 @@ export const useApiConstants = ():
...
     '',
   );
   useEffect(() => {
-    // if the language in locale storage is different than the saved language, the user may have changed languages in a different browser. If so, refetch the constants.
-    if (localeStorageLanguage !== userSavedLanguage) {
-      setLocaleStorageLanguage(userSavedLanguage || '');
+    // if the language in local storage is different than the saved language, that may mean the cached constants are in the previous language. If so, refetch the constants.
+
+    if (userSavedLanguage && localStorageLanguage !== userSavedLanguage) {
+      setLocalStorageLanguage(userSavedLanguage || '');
```
canac, line 25:
```suggestion
      setLocalStorageLanguage(userSavedLanguage);
```

src/components/Constants/UseApiConstants.tsx
```diff
@@ -14,14 +14,15 @@ export const useApiConstants = ():
   });
 
   const userSavedLanguage = data?.user.preferences?.locale;
```
canac, line 16: Sorry for the wordsmithing. Testing it in preview now.

```suggestion
  const currentLanguage = data?.user.preferences?.locale;
  const [constantsLanguage, setConstantsLanguage] = useLocalStorage(
    `constants-language`,
    '',
  );
```

canac:
In preview, it's loading the constants query every time you load the page now. I think it's because `useLocalStorage` returns `initialValue` on the first render. I would either update `useLocalStorage` `useState` to initialize to `readValue()` or just directly read from local storage without using the hook.

canac:
I tested all of the components with non-trivial changes in preview, and everything is looking great with the translated statuses.

canac (approved):
One of the PRs I just merged created a merge conflict, sorry! But this looks great now! I'm seeing the constants query whenever you change the language but otherwise using the cached constants.

PR #1146, opened 2024-10-18
[MPDX-8393] Add `useUserPreference` hook

canac:
## Description

Create a new `useUserPreference` hook that is as ergonomic as `useState` but syncs its value with a user preference.

**Usage:**

```ts
const [value, setValue] = useUserPreference({ key: 'hidden_tabs', defaultValue: [] });
```

**Features:**
* Calling `setValue` updates `value` immediately without having to wait for the server to respond.
* If `defaultValue` isn't a string, the value is transparently serialized and deserialized as JSON.
* If another component loaded all options with the `userOptions` query, this hook can extract the relevant option from the Apollo cache immediately before the `userOption` query returns.

Depends on https://github.com/CruGlobal/mpdx_api/pull/2872

I updated the code to use this new hook for the `setup_position`, and `flows` (the flows columns) preferences to prove its usefulness. It could also be used for the view mode preference once #1131 is merged. I will also use it in the upcoming PR for the dismissible beacon. 

## Checklist:


pages/accountLists/[accountListId]/contacts/flows/setup.page.tsx
```diff
@@ -138,27 +88,35 @@ const ContactFlowSetupPage: React.FC = () => {
...
+            (status) => status !== draggedStatus,
+          ),
+        };
+      }
+      if (destinationIndex > -1) {
+        temp[destinationIndex] = {
+          ...temp[destinationIndex],
+          statuses: [...temp[destinationIndex].statuses, draggedStatus],
+        };
+      }
```
Comment (canac) [inline pages/accountLists/[accountListId]/contacts/flows/setup.page.tsx:115]: Before this was updating cloning the array of flow options but still modifying the original column. This clones the column too before updating it.

src/hooks/UserPreference.graphql
```diff
@@ -0,0 +1,15 @@
+query UserOption($key: String!) {
```
canac, line 1: This query and mutation doesn't need to load the `id` field anymore because I changed the cache to use the `key` to generate `Option`s' unique id.

src/lib/apollo/cache.ts
```diff
@@ -80,6 +82,15 @@ export const createCache = () =>
           people: paginationFieldPolicy,
           tasks: paginationFieldPolicy,
           userNotifications: paginationFieldPolicy,
+          // When loading a user option, look it up from the cache by its key
+          userOption: {
```
canac, line 86: This makes it so that if some other component loaded all the user options with the `userOptions` query, the `userOption` (singular) query can pull the option out of the cache by its key instead of having to wait for it to load.

dr-bizz (approved):
This is great!

src/hooks/useUserPreference.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  const [value, setValue] = useState(defaultValue);
+
+  useEffect(() => {
+    if (!data?.userOption) {
+      return;
+    }
+
+    if (typeof defaultValue === 'string') {
+      setValue((data.userOption.value ?? defaultValue) as T);
```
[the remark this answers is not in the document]
canac: This hook assumes that if `defaultValue` is a string, the component just needs to save a simple string as the preference. For example the view mode, which could be "list", "flows", or "map". If `defaultValue` isn't an object, it assumes that the component needs to save an object or array as the preference and does the automatic JSON serialization/deserialization. This branch is the simple string code path.

PR #1150, opened 2024-10-23
[MPDX-8393] Make the Helpjuice beacon dismissable

canac:
## Description

Make the Helpjuice beacon dismissable. There is a new Hide Beacon link in the beacon widget that dismisses the beacon. Once dismissed, the beacon peeks just barely out from the right side of the screen. Hovering over it causes it to slide out, and clicking on it makes it no longer dismissed.

On most pages, the dismissed preference is saved to a persistent user preference. On login and error pages, we don't have an Apollo client, so the preference can't be persisted, but the beacon can still be temporarily dismissed until the next page load.

I had to jump through some hoops to get this to work, but the complexity is abstracted away into the `Helpjuice` component, and the comments for `ApolloBeacon` and `NoApolloBeacon` should describe what is going on. I chose not to do something like this because although it might technically work, it violates React's rules of hooks that hooks must not be conditionally called.

```tsx
const hasApolloClient = !useContext(getApolloContext());
const [dismissed, setDismissed] = hasApolloClient
  ? useUserPreference({
      key: 'beacon_dismissed',
      defaultValue: false,
    })
  : useState(false);
  return <DismissableBeacon dismissed={dismissed} setDismissed={setDismissed} />;
```

[MPDX-8393](https://jira.cru.org/browse/MPDX-8393)

## Checklist:

### DIFF
--- a/pages/_app.page.tsx
+++ b/pages/_app.page.tsx
@@ -110,6 +110,7 @@ const App = ({
     <TaskModalProvider>
       <Layout>
         <SnackbarUtilsConfigurator />
+        <Helpjuice />
         <Box
           sx={(theme) => ({
             fontFamily: theme.typography.fontFamily,
@@ -191,10 +192,7 @@ const App = ({
                             pageContent
                           ) : (
                             <RouterGuard>
-                              <GraphQLProviders>
-                                {pageContent}
-                                <Helpjuice />
-                              </GraphQLProviders>
+                              <GraphQLProviders>{pageContent}</GraphQLProviders>
                             </RouterGuard>
                           )}
                           <Loading />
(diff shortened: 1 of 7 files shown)


dr-bizz (changes requested):
The code looks great, but when I test it on the preview link, the beacon is visible after a refresh

canac:
@dr-bizz Good catch! I messed up the `hasApolloCheck`.

dr-bizz (commented):
One thought I had to improve the Help Jiuce beacon is when it is clicked, the icon changes to a close button, as it isn't obvious you have to click it again to close it, and some users might click on hide beacon thinking it will just close the beacon.

Screenshot 2024-10-23 at 3 35 40 PM

dr-bizz (approved):
This is awesome! Thank you for adding on the x icon when it's open! it adds a lot to the user experience.

PR #1151, opened 2024-10-23
[MPDX-8240] Deselect removed ids

canac:
## Description

When the list of contacts, tasks, or other entities changes, deselect any that were removed from the list.

## Testing

* Go to the tasks page and select 2 tasks. You should see "2 Selected" in the header.
* Change the tasks filters so that only one of the selected tasks is visible. You should see "1 Selected" in the header.

[MPDX-8240](https://jira.cru.org/browse/MPDX-8240)

## Checklist:


dr-bizz (commented):
The code and UI look great! On the first test, it removed all of my selected tasks, but on the second test it worked as described. - It also did this on contacts.

canac:
> On the first test, it removed all of my selected tasks.

@dr-bizz Good catch! That should be fixed now, as well as a performance improvement. Instead of waiting for the contacts or tasks to load before loading the mass selection ids, we load the mass selection ids immediately.

dr-bizz (approved):
This works well in the UI, and the code looks great. I didn't know you could fetch previousData. At first I was worried that if no tasks are present, it would default back to the previousData, but that isn't the case.

PR #1154, opened 2024-10-23
[MPDX-8405] Fix filtering the map by selected contacts

canac:
## Description

* Support filtering the contacts map by the selected contacts.
* Refactor `ContactsContext` tests to use a reusable wrapper.

## Testing

* Case 1
  * Go to the contacts list and select a few contacts
  * Switch to the map view
  * Verify that only the selected contacts show up on the map
  * Switch back to the list view
  * Verify that all contacts show up in the list view (i.e. the list is not filtered by contact ids like the map was)
* Case 2
  * Go to the contacts list (**do not** select any contacts)
  * Switch to the map view
  * Verify that all contacts show up on the map (after they finish loading)

[MPDX-8405](https://jira.cru.org/browse/MPDX-8405)

## Checklist:

### DIFF
--- a/src/components/Contacts/ContactsContext/ContactsContext.tsx
+++ b/src/components/Contacts/ContactsContext/ContactsContext.tsx
@@ -150,16 +150,16 @@ export const ContactsProvider: React.FC<ContactsContextProps> = ({
     [activeFilters],
   );
 
-  const [contactsView, updateOptions, { loading: userOptionsLoading }] =
+  const [contactsView, saveContactsView, { loading: userOptionsLoading }] =
     useUserPreference({
       key: 'contacts_view',
       defaultValue: TableViewModeEnum.List,
     });
   useEffect(() => {
-    if (contactsView && !userOptionsLoading) {
+    if (contactsView) {
       setViewMode(contactsView);
     }
-  }, [contactsView, userOptionsLoading]);
+  }, [contactsView]);
 
   const contactsFilters = useMemo(() => {
     // Remove filters in the map view
@@ -257,8 +257,15 @@ export const ContactsProvider: React.FC<ContactsContextProps> = ({
     _: React.MouseEvent<HTMLElement>,
     view: string,
   ) => {
-    setViewMode(view as TableViewModeEnum);
-    updateOptions(view as TableViewModeEnum);
+    const newViewMode = view as TableViewModeEnum;
+    saveContactsView(newViewMode);
+    if (newViewMode === TableViewModeEnum.Map) {
+      // When switching to the map, make the filter only show the selected contacts, if any
+      setActiveFilters({ ids });
+    } else if (viewMode === TableViewModeEnum.Map) {
+      // When switching away from the map, reset the filter to show all contacts
+      setActiveFilters({});
+    }
   };
   //#endregion
 
--- a/src/components/Contacts/ContactsContext/ContactsContext.test.tsx
+++ b/src/components/Contacts/ContactsContext/ContactsContext.test.tsx
@@ -7,14 +7,10 @@ import TestRouter from '__tests__/util/TestRouter';
 import { GqlMockedProvider } from '__tests__/util/graphqlMocking';
 import { ContactFiltersQuery } from 'pages/accountLists/[accountListId]/contacts/Contacts.generated';
 import { ContactsWrapper } from 'pages/accountLists/[accountListId]/contacts/ContactsWrapper';
-import { GetUserOptionsQuery } from 'src/components/Contacts/ContactFlow/GetUserOptions.generated';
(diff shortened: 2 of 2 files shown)


src/components/Contacts/ContactsContext/ContactsContext.tsx
```diff
@@ -150,16 +150,16 @@ export const ContactsProvider: React.FC<ContactsContextProps> = ({
...
 
-  const [contactsView, updateOptions, { loading: userOptionsLoading }] =
+  const [contactsView, saveContactsView, { loading: userOptionsLoading }] =
     useUserPreference({
       key: 'contacts_view',
       defaultValue: TableViewModeEnum.List,
     });
   useEffect(() => {
-    if (contactsView && !userOptionsLoading) {
+    if (contactsView) {
```
canac, line 159: @dr-bizz When I asked you to add this additional check, I was mistaken. I thought this effect saved the user preference, but it is just updating the view mode in the contacts wrapper. It was right how you had it first, so I'm switching it back.

src/components/Contacts/ContactsContext/ContactsContext.tsx
```diff
@@ -257,8 +257,15 @@ export const ContactsProvider: React.FC<ContactsContextProps> = ({
...
-    setViewMode(view as TableViewModeEnum);
-    updateOptions(view as TableViewModeEnum);
+    const newViewMode = view as TableViewModeEnum;
+    saveContactsView(newViewMode);
+    if (newViewMode === TableViewModeEnum.Map) {
+      // When switching to the map, make the filter only show the selected contacts, if any
+      setActiveFilters({ ids });
+    } else if (viewMode === TableViewModeEnum.Map) {
+      // When switching away from the map, reset the filter to show all contacts
+      setActiveFilters({});
```
dr-bizz, line 267: I wonder if there is a way to save the filters so a user doesn't lose them if they switch to the map and then back to list view.

One thought I had is to store them in local storage

dr-bizz (commented):
I was thinking of a way to inform users that the map is using the filters to filter down results, as it is not obvious. I wonder if we should highlight the map icon when we filter, but it might become annoying after some time.

Screenshot 2024-10-23 at 3 43 41 PM

src/components/Contacts/ContactsContext/ContactsContext.tsx
canac: I need to test more to make sure that I didn't break anything else, but this should be fixed now. The contacts wrapper still stores the contact filters in memory, but it only uses the `ids` filter in map view. If you switch back to the list view, it still remembers the previous filters and restores them.

pages/accountLists/[accountListId]/contacts/ContactsWrapper.tsx
```diff
@@ -77,6 +77,15 @@ export const ContactsWrapper: React.FC<Props> = ({
...
   const [filterPanelOpen, setFilterPanelOpen] = useState(true);
 
+  // Only allow the ids filter in map view, and remove the ids filter in other views
+  const activeFilters = useMemo(() => {
+    const { ids, ...otherFilters } = activeFiltersRaw;
+    if (viewMode === TableViewModeEnum.Map) {
+      return ids?.length ? { ids } : {};
+    }
+    return otherFilters;
+  }, [viewMode, activeFiltersRaw]);
```
Comment (dr-bizz) [inline pages/accountLists/[accountListId]/contacts/ContactsWrapper.tsx:87]: Does `ids` return an empty array or undefined? as you could write it as the below and save the extra If statement.
```suggestion
    const { ids, ...otherFilters } = activeFiltersRaw;
    if (viewMode === TableViewModeEnum.Map && ids?.length) {
      return { ids }
    }
    return otherFilters;
  }, [viewMode, activeFiltersRaw]);
```

dr-bizz (commented):
The code looks great, but more contacts are showing on the map than I expected. I thought only 3 (the same number of the contacts I had on the list view) would show up on maps.


https://github.com/user-attachments/assets/0b8395bc-8f05-4839-b6a1-e60ad91cbdb5

src/components/Contacts/ContactsContext/ContactsContext.tsx
```diff
@@ -257,8 +253,12 @@ export const ContactsProvider: React.FC<ContactsContextProps> = ({
     _: React.MouseEvent<HTMLElement>,
     view: string,
   ) => {
-    setViewMode(view as TableViewModeEnum);
-    updateOptions(view as TableViewModeEnum);
+    const newViewMode = view as TableViewModeEnum;
+    saveContactsView(newViewMode);
+    if (newViewMode === TableViewModeEnum.Map && ids.length > 0) {
```
dr-bizz, line 258:
```suggestion
    if (newViewMode === TableViewModeEnum.Map && ids.length) {
```

dr-bizz (approved):

PR #1157, opened 2024-10-24
[no-Jira] Fix disabled tooltip button warning

canac:
## Description

Fix MUI warning:

```
You are providing a disabled button child to the Tooltip component.
A disabled element does not fire events.
Tooltip needs to listen to the child element's events to display the title.
```

I went with the simple fix, but I could see a case for hiding the reverse button entirely instead of disabling it to reduce some clutter. This could be good since there are already so many different elements in the filter panel. Let me know your thoughts.

## Checklist:


dr-bizz (approved):
Thank you for fixing this warning

PR #1162, opened 2024-10-30
MPDX- 8231 - Announcements

dr-bizz:
## Description
In this PR, I add the announcements to MPDX. There are two versions of announcements, `Banners` and `Modals`; I've added both.

You will also see that I have moved the `<AlertBanner/>` into the announcements component; this is because we don't want both showing at the same time, and we need to have an authenticated session to use GraphQL, where AlertBanner does not, meaning we can show an alert on all pages if we wanted to. AlertBanner will primarily be used to show an alert on the login page.

You will see that I also have to use font awesome, but as we only want it for this component, I load it just after the initial render and then remove it on `unmount`.

## How to Test
The banners and modals should only show announcements that are in your language. See PR here: https://github.com/CruGlobal/mpdx_api/pull/2875.

_I found it easier to test on local with a local version of the API running, as I can alter the DB quickly, but test as you prefer._

You will need to alter an existing announcement (which you haven't acknowledged) or create a new one.
If you need additional reference of what it should look like, you can go to [old.stage-mpdx.org](https://old-stage.mpdx.org/login) to see how it used to look.

### Banner UI
1. Set up a banner announcement. 
2. Ensure it is showing
3. Alter the start and end dates to test; it doesn't show outside the allotted window.
4. Ensure the title and body are showing
5. Test the banner styles to ensure the banner changes colour
6. Add actions and tests they show as expected. Actions with the style icon should display an icon; the other styles should change how the button looks.
7. Clicking the close button should acknowledge the banner, and it will not reappear. (I still need to work out how to un-acknowledge it, so you might need to create a new banner.)
_The image doesn't show on the banner._

### Modal UI
1. Set up a modal announcement. 
2. Ensure it is showing
3. Alter the start date and end date to test it doesn't show outside the allotted window.
4. Ensure the title and body are showing
5. Test that the image is showing and is styled nicely
6. The announcement style shouldn't change the modal background, so you don't need to test.
7. Add actions and tests they show as expected. Actions with the style icon should display an icon; the other styles should change how the button looks.
8. Clicking the close buttons or outside the modal should acknowledge the banner, and it will not reappear. (I still need to work out how to un-acknowledge it, so you might need to create a new banner.)

### Announcement Actions

1. The `Go` action should redirect the user to another page. And then acknowledge the announcement so it doesn't show again.
2. The `AppealCreate` action should open the create appeal modal, prefill details about the end-of-year appeal, and acknowledge the announcement.
3. The `Track` action should `dispatch` the announcement args to analytics and then acknowledge the announcement.

## Checklist:

### DIFF
--- a/src/components/Announcements/AnnouncementAction/AnnouncementAction.test.tsx
+++ b/src/components/Announcements/AnnouncementAction/AnnouncementAction.test.tsx
@@ -1,0 +1,141 @@
+import React from 'react';
+import { ThemeProvider } from '@emotion/react';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import TestRouter from '__tests__/util/TestRouter';
+import { ActionStyleEnum } from 'src/graphql/types.generated';
+import theme from 'src/theme';
+import { ActionFragment } from '../Announcements.generated';
+import { AnnouncementAction } from './AnnouncementAction';
+
+const handlePerformAction = jest.fn();
+const accountListId = 'accountListId';
+const router = {
+  query: { accountListId },
+  isReady: true,
+};
+const defaultTextAndIconColor = '#EEEEEE';
+
+interface AnnouncementActionProps {
+  actionStyle?: ActionStyleEnum;
+  textAndIconColor?: string;
+  isBanner?: boolean;
+}
+const TestComponent: React.FC<AnnouncementActionProps> = ({
+  actionStyle = ActionStyleEnum.Link,
+  textAndIconColor = defaultTextAndIconColor,
+  isBanner = false,
+}) => {
+  const action: ActionFragment = {
+    id: 'action-1',
+    style: actionStyle,
+    label: 'Contacts',
+    args: 'coaching-account',
+  };
+
+  return (
+    <TestRouter router={router}>
+      <ThemeProvider theme={theme}>
+        <AnnouncementAction
+          action={action}
+          textAndIconColor={textAndIconColor}
+          handlePerformAction={handlePerformAction}
+          isBanner={isBanner}
+        />
+      </ThemeProvider>
+    </TestRouter>
+  );
+};
+
+describe('AnnouncementAction', () => {
+  describe('Icon', () => {
+    it('initial', () => {
+      const { getByRole, queryByText } = render(
+        <TestComponent actionStyle={ActionStyleEnum.Icon} />,
+      );
+
+      expect(queryByText('Contacts')).not.toBeInTheDocument();
+
+      const icon = getByRole('button', { name: 'Contacts' });
+      userEvent.click(icon);
+      expect(handlePerformAction).toHaveBeenCalled();
+    });
+  });
+  describe('Button', () => {
+    it('initial', () => {
+      const { getByText } = render(<TestComponent />);
+
+      const button = getByText('Contacts');
+      userEvent.click(button);
+
+      expect(handlePerformAction).toHaveBeenCalled();
+    });
+
+    describe('Styles', () => {
+      it('Primary Banner', () => {
+        const { getByText } = render(
+          <TestComponent isBanner actionStyle={ActionStyleEnum.Primary} />,
+        );
+
+        const button = getByText('Contacts');
+        expect(button).toHaveStyle({
+          'background-color': '#05699B',
+          color: defaultTextAndIconColor,
+        });
+      });
+      it('Primary Modal', () => {
+        const { getByText } = render(
+          <TestComponent actionStyle={ActionStyleEnum.Primary} />,
+        );
+
+        const button = getByText('Contacts');
+        expect(button).toHaveStyle({
+          'background-color': '#05699B',
+          color: '#FFFFFF',
+        });
+      });
+
+      it('Link Banner', () => {
+        const { getByText } = render(<TestComponent isBanner />);
+
+        const button = getByText('Contacts');
+        expect(button).toHaveStyle({
+          'background-color': 'transparent',
+          color: defaultTextAndIconColor,
+        });
+      });
+      it('Link Modal', () => {
+        const { getByText } = render(<TestComponent />);
+
+        const button = getByText('Contacts');
+        expect(button).toHaveStyle({
+          'background-color': 'transparent',
+          color: '#05699B',
+        });
+      });
+
+      it('Warning Banner', () => {
+        const { getByText } = render(
+          <TestComponent isBanner actionStyle={ActionStyleEnum.Warning} />,
+        );
+
+        const button = getByText('Contacts');
+        expect(button).toHaveStyle({
+          'background-color': '#ED6C02',
+          color: '#FFFFFF',
+        });
+      });
+      it('Warning Modal', () => {
+        const { getByText } = render(
+          <TestComponent actionStyle={ActionStyleEnum.Warning} />,
+        );
+
+        const button = getByText('Contacts');
+        expect(button).toHaveStyle({
+          'background-color': '#ED6C02',
+          color: '#FFFFFF',
+        });
+      });
+    });
+  });
+});
--- a/src/components/Announcements/AnnouncementAction/AnnouncementAction.test.tsx
+++ b/src/components/Announcements/AnnouncementAction/AnnouncementAction.test.tsx
@@ -50,13 +50,12 @@ const TestComponent: React.FC<AnnouncementActionProps> = ({
 describe('AnnouncementAction', () => {
   describe('Icon', () => {
     it('initial', () => {
-      const { getByRole, queryByText } = render(
+      const { getByRole } = render(
         <TestComponent actionStyle={ActionStyleEnum.Icon} />,
       );
 
-      expect(queryByText('Contacts')).not.toBeInTheDocument();
-
       const icon = getByRole('button', { name: 'Contacts' });
+      expect(icon.childNodes[0]).toHaveClass('material-icons');
       userEvent.click(icon);
       expect(handlePerformAction).toHaveBeenCalled();
     });
--- a/src/components/Announcements/AnnouncementAction/AnnouncementAction.tsx
+++ b/src/components/Announcements/AnnouncementAction/AnnouncementAction.tsx
@@ -83,7 +83,7 @@ export const AnnouncementAction: React.FC<AnnouncementActionProps> = ({
       onClick={handleClick}
       aria-label={action.label}
     >
-      <Icon baseClassName="far" className={`fa-${action.label}`} />
+      <Icon>{action.label}</Icon>
     </IconButton>
   ) : (
     <StyledButton {...buttonProps} onClick={handleClick}>
--- a/src/components/Announcements/AnnouncementAction/AnnouncementAction.tsx
+++ b/src/components/Announcements/AnnouncementAction/AnnouncementAction.tsx
@@ -27,7 +27,7 @@ export const AnnouncementAction: React.FC<AnnouncementActionProps> = ({
       variant: 'contained',
     };
     switch (action.style) {
-      case ActionStyleEnum.Reverse:
+      case ActionStyleEnum.ReverseAction:
         return {
           sx: { borderColor: textAndIconColor },
           variant: 'outlined',
@@ -80,8 +80,8 @@ export const AnnouncementAction: React.FC<AnnouncementActionProps> = ({
   return action.style === ActionStyleEnum.Icon ? (
     <IconButton
       sx={{ color: textAndIconColor, marginRight: 1 }}
-      key={action.id}
       onClick={handleClick}
+      aria-label={action.label}
     >
       <Icon baseClassName="far" className={`fa-${action.label}`} />
     </IconButton>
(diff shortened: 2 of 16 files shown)


wrandall22:
Is there no Material icons that can substitute for these FA icons? It seems like a step backwards to re-introduce those icons. Is this the free or paid version of FA?

dr-bizz:
> Is there no Material icons that can substitute for these FA icons? It seems like a step backwards to re-introduce those icons. Is this the free or paid version of FA?

It's a free version. I can look into seeing if we can get MUI to work, but I was trying to avoid having to update all the announcement icon actions, as they are `337` of them.

They all look like they are using the `aa-user-rating-{num}` so we could us the material icons, but that would still require us to load in the material icons:
```
<link
  rel="stylesheet"
  href="https://fonts.googleapis.com/icon?family=Material+Icons"
/>
```

wrandall22:
Ah, I assumed since the whole site was using Material icons already that we would have already loaded them.

wrandall22:
If we're using the free version of FA here, it is less of a concern. One of the benefits to moving to Material/React was dropping the dependency on a paid license of FA, in particular for non-Cru orgs.

src/theme.ts
```diff
@@ -4,7 +4,7 @@ const defaultTheme = createTheme();
 
 // https://www.cru.org/brand/color/
 const cruColors = {
-  yellow: '#FFCF07',
+  yellow: '#F5BE19',
```
canac, line 7: Is there a reason we're moving away from the official Cru yellow? https://www.cru.org/brand/color/ The old MPDX may have been using the older yellow.

src/components/Announcements/AnnouncementAction/AnnouncementAction.tsx
```diff
@@ -0,0 +1,94 @@
...
+  }, [action.style, textAndIconColor]);
+
+  const handleClick = () => {
+    handlePerformAction(action);
+  };
+
+  return action.style === ActionStyleEnum.Icon ? (
+    <IconButton
+      sx={{ color: textAndIconColor, marginRight: 1 }}
+      key={action.id}
```
canac, line 83: Normally `key` is only needed inside of a `.map` function.

src/components/Announcements/AnnouncementBanner/AnnouncementBanner.tsx
```diff
@@ -0,0 +1,151 @@
...
+}
+
+export const AnnouncementBanner: React.FC<AnnouncementBannerProps> = ({
+  announcement,
+  handlePerformAction,
+}) => {
+  const { t } = useTranslation();
+
+  useEffect(() => {
+    const bannerElement = document.getElementById('JS-announcement-banner');
```
canac, line 101: Can we use state to accomplish this so that we're doing this the React way instead of with raw DOM manipulation? Like a `shown` state that controls the `show` class that is initially `false` and we set it to `true` in the effect.

src/components/Announcements/Announcements.tsx
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    return () => {
+      const fontAwesomeStyles = document.querySelector(
+        'link[id="fontAwesomeStyles"]',
+      );
+      if (fontAwesomeStyles) {
+        document.head.removeChild(fontAwesomeStyles);
+      }
+    };
```
canac, line 89: I think you can simplify to this. I was using `.remove()` in the beacon and it worked great.

```suggestion
    if (!hasBeenLoaded) {
      const fontAwesomeStyles = document.createElement('link');
      fontAwesomeStyles.rel = 'stylesheet';
      fontAwesomeStyles.id = 'fontAwesomeStyles';
      fontAwesomeStyles.href =
        'https://use.fontawesome.com/releases/v5.14.0/css/all.css';
      document.head.appendChild(fontAwesomeStyles);
      return () => fontAwesomeStyles.remove();
    }
```

src/components/Announcements/Announcements.tsx
```diff
@@ -0,0 +1,184 @@
...
+            const userRatingMatch = action.args?.match(
+              /^aa-(user-rating)-(\d)$/,
+            );
+            if (userRatingMatch) {
+              // eslint-disable-next-line @typescript-eslint/no-unused-vars
+              const [_adobeEventName, event, rating] = userRatingMatch;
+              dispatch(event, { rating });
+            } else {
+              dispatch(action?.args ?? 'undefined event');
+            }
```
canac, line 128: I know this is the last case, but it's probably good to have a `break;` here anyway.

src/components/Announcements/AnnouncementAction/AnnouncementAction.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+  describe('Button', () => {
+    it('initial', () => {
+      const { getByText } = render(<TestComponent />);
+
+      const button = getByText('Contacts');
+      expect(button).toBeInTheDocument();
+      userEvent.click(button);
+
+      expect(handlePerformAction).toHaveBeenCalledTimes(1);
+      expect(handlePerformAction).toHaveBeenCalled();
```
canac, line 73: This `expect` is redundant.

src/components/Announcements/AnnouncementAction/AnnouncementAction.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+    });
+  });
+  describe('Button', () => {
+    it('initial', () => {
+      const { getByText } = render(<TestComponent />);
+
+      const button = getByText('Contacts');
+      expect(button).toBeInTheDocument();
+      userEvent.click(button);
```
canac, line 70: `getByText` implicitly asserts the element's existence.

```suggestion
      userEvent.click(getByText('Contacts'));
```

src/components/Announcements/AnnouncementAction/AnnouncementAction.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+    describe('Styles', () => {
+      it('Primary Banner', () => {
+        const { getByText } = render(
+          <TestComponent isBanner actionStyle={ActionStyleEnum.Primary} />,
+        );
+
+        const button = getByText('Contacts');
+        expect(button).toHaveStyle('background-color: #05699B');
+        expect(button).toHaveStyle(`color: ${defaultTextAndIconColor}`);
```
canac, line 84: I think you can check multiple styles at once:

```suggestion
        expect(button).toHaveStyle({ 'background-color': '#05699B', color: defaultTextAndIconColor });
```

src/components/Announcements/AnnouncementModal/AnnouncementModal.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+    expect(handlePerformAction).toHaveBeenLastCalledWith();
+    expect(handlePerformAction).toHaveBeenCalledTimes(2);
+  });
+
+  it('should fire handlePerformAction() with the first action', () => {
+    const { getByRole } = render(<TestComponent />);
+
+    userEvent.click(getByRole('button', { name: 'Contacts' }));
+
+    // Check to validate it is called with no arguments
```
canac, line 71: I think this comment needs updated.

```suggestion
    // Check to validate it is called with announcement action
```

src/components/Announcements/AnnouncementModal/AnnouncementModal.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+    expect(handlePerformAction).toHaveBeenCalledWith(announcement.actions[0]);
+  });
+
+  it('should fire handlePerformAction() with the second action', () => {
+    const { getByRole } = render(<TestComponent />);
+
```
canac, line 77: This comment also may need updated.

src/components/Announcements/Announcements.test.tsx
```diff
@@ -0,0 +1,268 @@
...
+      await waitFor(() => {
+        expect(mutationSpy).toHaveGraphqlOperation('AcknowledgeAnnouncement', {
+          input: { announcementId: modalAnnouncement.id, actionId: undefined },
+        });
+      });
+    });
+  });
+
+  describe('Modal and Banner actions', () => {
+    test.each([
```
canac, line 166: Great usage of `test.each` to simplify our tests!

src/components/Announcements/Announcements.test.tsx
```diff
@@ -0,0 +1,268 @@
...
+          expect(getByText('Add Appeal')).toBeInTheDocument();
+        });
+      },
+    );
+
+    test.each([
+      { announcement: defaultAnnouncement, type: 'Banner' },
+      { announcement: modalAnnouncement, type: 'Modal' },
+    ])(
+      'should track and  dispatch the click on the $type',
```
canac, line 238:
```suggestion
      'should track and dispatch the click on the $type',
```

canac (commented):
This looks great! I'll have to spend some time testing it, but the code and tests look really well-written and organized!

src/theme.ts
dr-bizz: Ah, I see; I can revert it back; it does make it a little harder to view the text but not that much.

src/components/Announcements/AnnouncementBanner/AnnouncementBanner.tsx
dr-bizz: Good point!

src/components/Announcements/Announcements.tsx
dr-bizz: Nice!

src/theme.ts
canac: You're right that we need to make sure it's legible. The style guide also says:

```
Take care to preserve legibility when using yellow. As a rule, avoid using yellow text on white and white text on yellow.
```

I haven't looked at the UI yet; I'm about to test it. But we should probably find an alternative if we're doing yellow-on-white or vice-versa.

src/components/Announcements/Announcements.tsx
dr-bizz: I'll add it in so it makes it easier to read.

src/components/Announcements/Announcements.graphql
```diff
@@ -0,0 +1,37 @@
+query Announcements {
+  announcements {
```
canac, line 2: Was this where you were going to add the filter to only load one annoucement?

```suggestion
  announcements(first: 1) {
```

src/components/Announcements/Announcements.tsx
```diff
@@ -0,0 +1,184 @@
...
+  const { push } = useRouter();
+  const accountListId = useAccountListId();
+  const [showAppealModal, setShowAppealModal] = useState(false);
+  const [hideAnnouncement, setHideAnnouncement] = useState(false);
+  const { contactStatuses } = useContactPartnershipStatuses();
+
+  const { data } = useAnnouncementsQuery();
+  const [acknowledgeAnnouncement] = useAcknowledgeAnnouncementMutation();
+
+  // TODO replace this with the ability to only limit announcements to one item.
```
canac, line 56: Just noting this remaining TODO

src/components/Announcements/Announcements.tsx
```diff
@@ -0,0 +1,184 @@
...
+import { DynamicAddAppealModal } from '../Tool/Appeal/Modals/AddAppealModal/DynamicAddAppealModal';
+import { DynamicAnnouncementBanner } from './AnnouncementBanner/DynamicAnnouncementBanner';
+import { DynamicAnnouncementModal } from './AnnouncementModal/DynamicAnnouncementModal';
+import {
+  ActionFragment,
+  useAcknowledgeAnnouncementMutation,
+  useAnnouncementsQuery,
+} from './Announcements.generated';
+
+const appealName = `${DateTime.local().year} ${i18n.t('End of Year Ask')}`;
```
canac, line 30: If we move this inside of the component, it will update if the announcement is shown on preferences and the user changes their language.

dr-bizz:
@wrandall22 @canac should we use Material Icons instead of Font-Awesome? I'm okay with switching it. I guess one of the advantages of using material icons, is that it's in the MUI family, the only down side is we would have to update 337 rows in the DB, but this shouldn't take long.

canac:
@dr-bizz It seems cleaner to just use Material Icons since that's what we're already using in this project. For the migration, do you have to create a mapping of icon names from MUI to FA? That kind of migration feels difficult. Also, will it be harder to make the MUI icons show up since it's an import instead of just adding the right classname?

canac (approved):
I couldn't find any issues during testing. Great work!

FYI, to undismiss announcements, I connected to the staging database and ran `delete from "person_announcements" where "announcement_id" = 'xxx' and "user_id" = 'xxx';` It worked really well.

dr-bizz:
> I couldn't find any issues during testing. Great work!
> 
> FYI, to undismiss announcements, I connected to the staging database and ran `delete from "person_announcements" where "announcement_id" = 'xxx' and "user_id" = 'xxx';` It worked really well.

This is great to know. It's weird that they used the word "person" instead of user. I think that is what threw me off finding the table.

I've messaged in slack about how we can use Material Icons

PR #1167, opened 2024-10-31
MPDX-8421 14month Late Circle

caleballdrin:
## Description
https://jira.cru.org/browse/MPDX-8421
- Add colored circle if the partner is late on their giving. 
- Hide it on print view

## Checklist:

### DIFF
--- a/src/components/Reports/FourteenMonthReports/Layout/Table/Table.test.tsx
+++ b/src/components/Reports/FourteenMonthReports/Layout/Table/Table.test.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { ThemeProvider } from '@mui/material/styles';
-import { render, waitFor } from '@testing-library/react';
+import { render, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import TestRouter from '__tests__/util/TestRouter';
 import { GqlMockedProvider } from '__tests__/util/graphqlMocking';
@@ -29,7 +29,7 @@ const defaultFourteenMonthReport = {
             average: 258,
             id: 'contact-1',
             lateBy30Days: false,
-            lateBy60Days: false,
+            lateBy60Days: true,
             months: [
               {
                 donations: [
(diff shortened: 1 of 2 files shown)


src/components/Reports/FourteenMonthReports/Layout/Table/Table.tsx
```diff
@@ -140,6 +141,31 @@ export const FourteenMonthReportTable: React.FC<
...
                       </NameTypography>
+                      {(contact.lateBy30Days || contact.lateBy60Days) && (
+                        <Tooltip
+                          title={t('{{daysLate}}+ days late', {
+                            daysLate: contact.lateBy60Days ? '60' : '30',
+                          })}
+                          arrow
+                        >
+                          <Box
+                            sx={{
```
canac, line 152: Since this is a rather complicated `sx` block, can we make this a styled component? For the `lateBy60Days` flag, you could either pass in a prop or add/remove a `60-days-late` classname and add custom styles when that classname is present. Something like this in the styled component.

```
bgcolor: 'orange',
'&.60-days-late': {
  bgcolor: 'red',
},
```

src/components/Reports/FourteenMonthReports/Layout/Table/Table.test.tsx
```diff
@@ -261,6 +261,29 @@ describe('FourteenMonthReportTable', () => {
...
+  it('should should show the dot if a donation is late', async () => {
+    const { getAllByTestId, queryByTestId } = render(<Components />);
+
+    await waitFor(() => {
+      expect(
+        queryByTestId('LoadingFourteenMonthReport'),
+      ).not.toBeInTheDocument();
+    });
+
+    const fourteenMonthReportRow = getAllByTestId(
```
canac, line 273:
```suggestion
    const fourteenMonthReportRows = getAllByTestId(
```

src/components/Reports/FourteenMonthReports/Layout/Table/Table.test.tsx
```diff
@@ -261,6 +261,29 @@ describe('FourteenMonthReportTable', () => {
...
+    );
+    expect(
+      within(fourteenMonthReportRow[0]).getByTestId('lateCircle60'),
+    ).toBeInTheDocument();
+
+    await waitFor(() => {
+      expect(
+        within(fourteenMonthReportRow[1]).queryByTestId('lateCircle30'),
+      ).not.toBeInTheDocument();
+    });
```
canac, line 284:
```suggestion
    expect(
        within(fourteenMonthReportRow[1]).queryByTestId('lateCircle30'),
    ).not.toBeInTheDocument();
```

canac (changes requested):
Thanks for adding the tooltip!

src/components/Reports/FourteenMonthReports/Layout/Table/Table.test.tsx
caleballdrin: I thought we usually have `waitFor` when querying something that should not be in the document. No?

canac: Yes, that's a good rule of thumb for something that needs to asynchronously load but we expect to not be in the document. However, in this case, we already waited for the table rows to load, so at that point, either the late circle will be there or it won't. There's no reason to wait longer.

caleballdrin: Ah got it

canac (approved):
Thanks for the fixup!

PR #1168, opened 2024-10-31
[MPDX-8386] Redirect after login

canac:
## Description

Redirect the user back to the page they were trying to go to after they successfully log in.

Looking at the redirect trace in the DevTools network tab (select preserve log, filter by Doc to see page loads, and look at the `Location` response header), it appears that after sign in, NextAuth, redirects back to the page that initiated the login, in our case, `/login?redirect=xxx`. Our `getServerSideProps` function on the login page notices that we have a session and a redirect URL and redirects the user one last time to their original page.

## Testing

* Go to a page that isn't a dashboard page (for example, tasks).
* Copy the URL.
* Sign out of MPDX and Okta.
* Navigate to the URL you copied.
* Ensure that when it takes you to the login page, `?redirect` is set to the tasks page.
* Ensure that you are redirected to the tasks page after logging in.

[MPDX-8386](https://jira.cru.org/browse/MPDX-8386)

## Checklist:

### DIFF
--- a/pages/accountLists.page.test.tsx
+++ b/pages/accountLists.page.test.tsx
@@ -20,10 +20,12 @@ interface GetServerSidePropsReturn {
   redirect: unknown;
 }
 
-const accountListId = 'accountID1';
+const accountListId = 'account-list-1';
 
 describe('Account Lists page', () => {
-  const context = {} as GetServerSidePropsContext;
+  const context = {
+    resolvedUrl: '/accountLists/account-list-1',
+  } as GetServerSidePropsContext;
 
   describe('NextAuth unauthorized', () => {
     it('should redirect to login', async () => {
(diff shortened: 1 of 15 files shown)


dr-bizz (commented):
This looks great! Should we automattically start to log users in when they are redirected to the login screen?

dr-bizz:
> > Should we automatically start to log users in when they are redirected to the login screen?
> 
> Sorry, but I'm not sure what you mean.

Sorry, I can add more context. When we redirect users from the page they were on to the login page, we shouldn't add another click to the user, we should start to log the users in as soon as they land on the login. They will need to add their login details.

Yes, It's using the `immediateSignIn` when users are redirected to the login screen when they are not authenticated.

dr-bizz (approved):

PR #1169, opened 2024-10-31
[MDPX-8385] Preserve previous task name

canac:
## Description

* When creating a new task after completing a different task, default the task name to the previous task name.
* If the user changes the new task's task phase or activity type, the task name will be overridden, which I believe is what we want.
* ~~If the user had a customized task name for a text message task, for example, and they make the next action a phone call, we just use the autogenerated task name. This is to prevent someone who didn't customize their task name from getting "Call to Initiate Appointment" as the initial task name for a text message task. I think this is a reasonable tradeoff, but feel free to discuss.~~ **EDIT**: Now I'm just comparing the task subject with the default task subject. If they differ, this means that the user customized the task subject. This logic will break if the server changes the default task subjects, but this should be rare, so I think this is acceptable.

## Testing

* Case 1
  * Create a new task and customize the name
  * Complete the task and make sure that the next action is set
  * Check that the task name in the Add Task modal is the customized task name
* Case 2
  * Log a new task and customize the name
  * Complete the task and make sure that the next action is set
  * Check that the task name in the Add Task modal is the customized task name
* Case 3
  * Create a new task
  * Check that change the phase and activity type changes the task subject

[MPDX-8385](https://jira.cru.org/browse/MPDX-8385)

## Checklist:

### DIFF
--- a/src/components/Task/Modal/Form/Complete/TaskModalCompleteForm.tsx
+++ b/src/components/Task/Modal/Form/Complete/TaskModalCompleteForm.tsx
@@ -55,6 +55,7 @@ import { useUpdateContactStatusMutation } from '../TaskModal.generated';
 import {
   extractSuggestedTags,
   getDatabaseValueFromResult,
+  getDefaultTaskName,
 } from '../TaskModalHelper';
 import { possibleNextActions } from '../possibleNextActions';
 import { possiblePartnerStatus } from '../possiblePartnerStatus';
@@ -219,9 +220,17 @@ const TaskModalCompleteForm = ({
     }
     onClose();
     if (attributes.nextAction) {
+      const defaultSubject = getDefaultTaskName(
+        task.activityType ?? null,
+        activityTypes,
+      );
+
       openTaskModal({
         view: TaskModalEnum.Add,
         defaultValues: {
+          // If the user changed the task subject from the default, use their customized task
+          // subject in the new task
+          subject: task.subject === defaultSubject ? undefined : task.subject,
           activityType: attributes.nextAction,
           // TODO: Use fragments to ensure all required fields are loaded
           contactIds: task.contacts.nodes.map((contact) => contact.id),
--- a/src/hooks/usePhaseData.ts
+++ b/src/hooks/usePhaseData.ts
@@ -26,8 +26,15 @@ export type ActivityData = {
   name: string;
   phaseId: PhaseEnum;
   phase: string;
+  subject?: string;
   title?: string;
 };
+
+const capitalizeWords = (sentence: string): string =>
+  sentence
+    .split(' ')
+    .map((word) => word[0].toUpperCase() + word.substring(1))
+    .join(' ');
 
 const phaseFromActivity = (
   activity: PhaseEnum | null,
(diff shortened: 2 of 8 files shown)


src/components/Task/Modal/Form/Complete/TaskModalCompleteForm.tsx
```diff
@@ -222,6 +222,10 @@ const TaskModalCompleteForm = ({
       openTaskModal({
         view: TaskModalEnum.Add,
         defaultValues: {
+          subject:
```
caleballdrin, line 225: This works as expected when I tried it. I wonder if you should also add this feature when logging a task.

I think some people will want it to keep the previous task name even if it is a different task action. I wonder if we could have it look at the previous task action, check if it was customized, and then if it was customized, keep the old one even if the action is different.

dr-bizz: ^ I agree, we should add this functionality to logs task.

src/components/Task/Modal/Form/Complete/TaskModalCompleteForm.tsx
```diff
@@ -222,6 +222,10 @@ const TaskModalCompleteForm = ({
...
+            attributes.nextAction === task.activityType
```
dr-bizz, line 226: Can you add a comment here that helps the dev understand why this is here? 

I'm thinking of devs who come to this project, might not understand this logic straight away.

dr-bizz (changes requested):
Great work on this!! The code looks good. I know you and others have been after this for some time. This works for when you're on the partner care phase, but when you're on another phase, it doesn't.

Maybe this is working as expected, and I'm testing it wrong. This is what I have found in my testing:


https://github.com/user-attachments/assets/4ce585d3-f14b-40da-96ce-ec1d011101d8

src/components/Task/Modal/Form/Complete/TaskModalCompleteForm.tsx
canac: > I wonder if we could have it look at the previous task action, check if it was customized, and then if it was customized, keep the old one even if the action is different.

You saying that made me realize that there is an easy way to check whether the task subject was customized: just check whether it matches the generated task subject or not. This logic will break if the task subjects change on the server, but I think this is OK because that should be fairly rare.

canac:
> Great work on this!! The code looks good. I know you and others have been after this for some time. This works for when you're on the partner care phase, but when you're on another phase, it doesn't.

@dr-bizz I'm wondering if it works now that I'm taking a different approach for deciding whether to use the previous task name.

src/hooks/usePhaseData.ts
```diff
@@ -112,26 +119,31 @@ export const usePhaseData = (phaseEnum?: PhaseEnum | null): GetPhaseData => {
   //   name: 'Special Gift Appeal',
   //   phase: 'Initiation',
   //   phaseId: 'INITIATION',
+  //   subject: 'Special gift appeal',
```
dr-bizz, line 122:
```suggestion
  //   subject: 'Special Gift Appeal',
```

src/hooks/usePhaseData.ts
```diff
@@ -112,26 +119,31 @@ export const usePhaseData = (phaseEnum?: PhaseEnum | null): GetPhaseData => {
...
   //   title: 'Initiation - Special Gift Appeal',
   // },
   // FOLLOW_UP_TEXT_MESSAGE: {
   //   name: 'Text Message',
   //   phase: 'Follow-Up',
   //   phaseId: 'FOLLOW_UP',
+  //   subject: 'Text Message to follow up',
```
dr-bizz, line 129:
```suggestion
  //   subject: 'Text Message To Follow Up',
```

dr-bizz (approved):
Looks great! Great work. It was a little hard to review since you did a force push, so I wasn't sure what I reviewed already, but the code looks great.

src/hooks/usePhaseData.ts
canac: Good catch 👍

PR #1170, opened 2024-11-01
Show Received on pledge modal to avoid confusion

dr-bizz:
## Description
I have added the "Received" option back onto the pledge modal. Looking back at previous messages, it looks like users can try to put the contact into "Received", ost of the time the server will override their choice but not always.

Putting it back this way will make it simplier for users until we overhaul the appeals Committed->Given process.

## Checklist:

### DIFF
--- a/src/components/Tool/Appeal/Modals/PledgeModal/PledgeModal.test.tsx
+++ b/src/components/Tool/Appeal/Modals/PledgeModal/PledgeModal.test.tsx
@@ -410,7 +410,7 @@ describe('PledgeModal', () => {
     );
   });
 
-  it('can not select the status given or received if the pledge does not have one of those statuses already', async () => {
+  it('can not select the status given if the pledge does not have the status "given"', async () => {
     const { getByRole, queryByRole } = render(
       <Components
         pledge={{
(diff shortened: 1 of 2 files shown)


src/components/Contacts/ContactsLeftPanel/ContactsLeftPanel.test.tsx
```diff
@@ -60,32 +63,50 @@ const mocks = {
...
-    };
 
     const { findByText } = render(
       <ThemeProvider theme={theme}>
         <TestRouter router={router}>
           <GqlMockedProvider<{ UserOption: UserOptionQuery }> mocks={mocks}>
             <ContactsWrapper>
-              <>
-                <Component />
+              <ContactsContext.Provider
```
canac, line 85: Do you know why this helps? `ContactsWrapper` already wraps its children in `<ContactsContext.Provider>`.

canac (approved):
I'm curious about the flaky test fix, but the pledge modal fix looks good.

src/components/Contacts/ContactsLeftPanel/ContactsLeftPanel.test.tsx
dr-bizz: My guess is that the `viewMode` was getting set randomly, but I could be wrong. I figured if I controlled all the properties, I would know for sure it would pass.

PR #1171, opened 2024-11-01
MPDX-8425 Require Task Action on Mass Actions

caleballdrin:
## Description
https://jira.cru.org/secure/RapidBoard.jspa?rapidView=3&view=detail&selectedIssue=MPDX-8425&sprint=1282#
- Require Task Action on mass edit modal

## Checklist:

### DIFF
--- a/src/components/Task/MassActions/EditTasks/MassActionsEditTasksModal.test.tsx
+++ b/src/components/Task/MassActions/EditTasks/MassActionsEditTasksModal.test.tsx
@@ -86,47 +86,80 @@ describe('MassActionsEditTasksModal', () => {
     );
     expect(getByText('Blank fields will not be affected!')).toBeInTheDocument();
   });
-});
+  it('shows correct Action field options based on the Task Type', async () => {
+    const handleClose = jest.fn();
+    const { queryByText, findByRole } = render(
+      <LocalizationProvider dateAdapter={AdapterLuxon}>
+        <ThemeProvider theme={theme}>
+          <SnackbarProvider>
+            <GqlMockedProvider>
+              <MassActionsEditTasksModal
+                accountListId={accountListId}
+                ids={['task-1', 'task-2']}
+                selectedIdCount={2}
+                handleClose={handleClose}
+              />
+            </GqlMockedProvider>
+          </SnackbarProvider>
+        </ThemeProvider>
+      </LocalizationProvider>,
+    );
+    const taskTypeField = await findByRole('combobox', { name: 'Task Type' });
+    const actionField = await findByRole('combobox', { name: 'Action' });
 
-it('shows correct Action field options based on the Task Type', async () => {
-  const handleClose = jest.fn();
-  const { queryByText, findByRole } = render(
-    <LocalizationProvider dateAdapter={AdapterLuxon}>
-      <ThemeProvider theme={theme}>
-        <SnackbarProvider>
-          <GqlMockedProvider>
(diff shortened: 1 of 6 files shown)


src/utils/functions/getLocalizedResultStrings.ts
```diff
@@ -37,7 +37,7 @@ export const getLocalizedResultString = (
       return t('No Response Yet');
 
     case DisplayResultEnum.AppointmentResultNotInterested:
-      return t('Does not want to meet');
+      return t('Not Interested');
```
dr-bizz, line 40: Good catch

dr-bizz: Was this missed when we did Task Phases?

src/components/Task/MassActions/EditTasks/MassActionsEditTasksModal.tsx
```diff
@@ -185,6 +201,11 @@ export const MassActionsEditTasksModal: React.FC<
                     }
                     // None and null are distinct values: null leaves the action unchanged and None changes the action to None
                     preserveNone
+                    required={!!taskPhase}
```
dr-bizz, line 204: It seems it is auto-selecting options when it opens the activity and then you click away.

src/components/Task/MassActions/EditTasks/MassActionsEditTasksModal.test.tsx
```diff
@@ -86,47 +86,80 @@ describe('MassActionsEditTasksModal', () => {
...
+        </ThemeProvider>
+      </LocalizationProvider>,
+    );
+    const taskTypeField = await findByRole('combobox', { name: 'Task Type' });
+    const actionField = await findByRole('combobox', { name: 'Action' });
+
+    userEvent.click(taskTypeField);
+    userEvent.click(await findByRole('option', { name: 'Appointment' }));
+
+    await waitFor(() => expect(getByText('Save')).toBeDisabled());
```
dr-bizz, line 156: For bonus points, you should make the error message show in the test

dr-bizz (changes requested):
The code is looking good. I found an error when you select a phase, it shows the activity dropdown, but if you click away from the dropdown, it autoselects the first option, which is the first issue, and the second issue is that it doesn't recognise it is selected.

If you select another option or select it properly, it works. I know these errors were here before, but since we are making the activity dropdown required if the task phase is selected, it means we will have to fix them.

Screenshot 2024-11-04 at 1 37 20 PM

src/utils/functions/getLocalizedResultStrings.ts
caleballdrin: Yeah, we missed it. In the future, we should get it from the API.

src/components/Task/MassActions/EditTasks/MassActionsEditTasksModal.test.tsx
caleballdrin: I do love bonus points, but I spent 30 mins and couldn't get it to find the text... I think as long as it is disabled we can know it is working correctly.

dr-bizz (changes requested):
It still does it somewhat. If you get to the activity and then press the downward arrow, then click on another field, it selects the item you were on, then doesn't allow the user to continue.
Screenshot 2024-11-04 at 4 35 09 PM

dr-bizz (approved):
Awesome work! Thank you for fixing this!

PR #1175, opened 2024-11-05
MPDX-8429 Fix 502 error on 14-month report

dr-bizz:
## Description
The 14-month Proxy GraphQL query always failed on `iryna.oleksiuk@cru.org`'s account. This is due to the amount of data being returned, and the server couldn't cope.

For a very quick fix, I have called the API directly from the frontend, instead of using Next.js backend. I believe this will fix the issue for now, but the main issue is that we need to make this query into a proper GraphQL query.

I hope this doesn't cause the query to be slow for others. - In my testing, it has somewhat sped it up.

## To Test:
1. Impersonate `iryna.oleksiuk@cru.org` and see if the page returns a 502.
2. Impersonate others or use your own account to ensure the report is working as it should.

Slack Message: https://cru-main.slack.com/archives/GG47BDJDU/p1730814260125489
Jira Ticket: https://jira.cru.org/browse/MPDX-8429

## Checklist:


src/components/Reports/FourteenMonthReports/Layout/Table/Table.tsx
```diff
@@ -147,12 +147,10 @@ export const FourteenMonthReportTable: React.FC<
...
                       {!isExpanded && <StyledInfoIcon fontSize="small" />}
                       <NameTypography variant="body1" expanded={isExpanded}>
-                        <NextLink
-                          href={contactUrl}
-                          passHref
-                          onMouseEnter={preloadContactsRightPanel}
-                        >
-                          <Link>{contact.name}</Link>
+                        <NextLink href={contactUrl} passHref legacyBehavior>
+                          <Link onMouseEnter={preloadContactsRightPanel}>
```
canac, line 151: Does this fix the barage of console warnings?

src/components/Reports/FourteenMonthReports/handleFourteenMonthReportData.ts
```diff
@@ -0,0 +1,148 @@
+import { FourteenMonthReport } from 'pages/api/graphql-rest.page.generated';
```
canac, line 1: Could we import this from the datahandler? Or did you have to customize some things?

src/components/Reports/FourteenMonthReports/FourteenMonthReport.tsx
```diff
@@ -48,33 +51,68 @@ export const FourteenMonthReport: React.FC<Props> = ({
...
+          .toISODate()}...${DateTime.now().toISODate()}`;
+
+        const response = await fetch(
+          `${process.env.REST_API_URL}reports/${requestUrl}`,
+          {
+            headers: {
+              authorization: `Bearer ${apiToken}`,
+              'Content-Type': 'application/vnd.api+json',
+            },
+            method: 'GET',
```
canac, line 90: `GET` is the default method and can be omitted.

canac:
I tested that the request is working now when using a direct REST API call ✅

src/components/Reports/FourteenMonthReports/Layout/Table/Table.tsx
dr-bizz: yes it does

src/components/Reports/FourteenMonthReports/handleFourteenMonthReportData.ts
canac: Oh sorry, I meant can we import everything in this file, instead of copying it?

canac (approved):
I tested again, and it still works. Thanks for fixing this!

PR #1177, opened 2024-11-05
[MPDX-8145] Make address street optional

canac:
## Description

Allow the address street to be ommitted. Users may know the city and state of a contact but not their street address. However, we do require that at least one of the address text fields be entered.

## Testing

* Create a new address
  * Check that the save button is disabled
  * Type in a field that isn't the street (i.e. the city)
  * Save the address and check that it saved successfully
* Edit the existing address you just created
  * Clear the single field with a value
  * Check that the save button is disabled because all fields are empty
  * Type in a different field that isn't the street (i.e. the state)
  * Save the address and check that it updated successfully

[MPDX-8415](https://jira.cru.org/browse/MPDX-8415)

## Checklist:


dr-bizz (approved):
This looks good

PR #1179, opened 2024-11-05
[HS-1251563] Sort and refresh contact tags

canac:
## Description

* Sort tags in the filters panel alphabetically, like Angular did.
* Refresh the contact filters when contact tags are added or deleted.
* In the future, we should do something similar for task tags.

https://secure.helpscout.net/conversation/2749833892/1251563/

## Checklist:

### DIFF
--- a/src/components/Shared/Filters/TagsSection/FilterTagDeleteModal.tsx
+++ b/src/components/Shared/Filters/TagsSection/FilterTagDeleteModal.tsx
@@ -7,7 +7,6 @@ import { ContactFiltersDocument } from 'pages/accountLists/[accountListId]/conta
 import { TaskFiltersDocument } from 'pages/accountLists/[accountListId]/tasks/Tasks.generated';
 import { DeleteButton } from 'src/components/common/Modal/ActionButtons/ActionButtons';
 import Modal from 'src/components/common/Modal/Modal';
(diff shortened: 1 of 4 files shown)


src/components/Shared/Filters/TagsSection/FilterTagDeleteModal.tsx
```diff
@@ -36,11 +34,7 @@ export const FilterTagDeleteModal: React.FC<FilterTagDeleteModalProps> = ({
         page,
       },
       refetchQueries: [
-        {
-          query:
-            page === 'contacts' ? ContactFiltersDocument : TaskFiltersDocument,
-          variables: { accountListId },
-        },
+        page === 'contacts' ? ContactFiltersDocument : TaskFiltersDocument,
```
canac, line 37: For some reason, how this was wasn't causing the data to refresh. Removing the variables makes it work though.

dr-bizz: I wonder if it saw it as a different query

dr-bizz (approved):
The code looks great. I'm a little bit confused with this message: https://secure.helpscout.net/conversation/2749833892/1251563/

Why do you need to refetch the whole contact list when you can just refetch the filters?

PR #1188, opened 2024-11-11
[no-Jira] Upgrade i18next-parser

canac:
## Description

Upgrade `i18next-parser` to fix a vulnerability in `vue-template-compiler`. Also, run `yarn extract` to verify that the output is the same with the old version and the new version.

## Testing

The `Run yarn extract` commit was made with the old version as a reference point. To verify that the the new version produces the same output, run `git restore public/locales/en/translation.json --source=main` to revert that file back to its contents on `main`. Then run `yarn extract` and verify that there are no uncommitted changes. Without the changes to the config file, v9 produces empty strings for the values of any new labels it finds.

## Checklist:


dr-bizz (changes requested):
When testing, I found 4 words which were lost in your PR. I confirmed that these words weren't in your translation.json and they were in the repo.

```
Annual
Appointment
Archive
Initiation
```

dr-bizz (approved):
You are correct, `main` also took all 4 of those phrases out.

I was reviewing it quickly, and saw annual in a test file, so there is no instances of those words in components.

PR #1196, opened 2024-11-15
Appeals - Remove appeals when switch account list

dr-bizz:
## Description
Ensure React keeps track of the appeals, to ensure they are removed correctly when the user switches account.

## Checklist:

### DIFF
--- a/src/components/Tool/Appeal/InitialPage/Appeals.tsx
+++ b/src/components/Tool/Appeal/InitialPage/Appeals.tsx
@@ -92,13 +92,11 @@ const Appeals: React.FC<AppealsProps> = ({ accountListId }) => {
             ))}
         </>
       ) : data?.regularAppeals.nodes.length ? (
-        <>
-          {data.regularAppeals.nodes.map((appeal) => (
-            <Box key={appeal.name} mb={3}>
-              <Appeal appeal={appeal} changePrimary={handleChangePrimary} />
-            </Box>
-          ))}
-        </>
+        data.regularAppeals.nodes.map((appeal) => (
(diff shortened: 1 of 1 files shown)


caleballdrin (approved):
That fixes the issue for me!

dr-bizz:
Great!

src/components/Tool/Appeal/InitialPage/Appeals.tsx
```diff
@@ -92,13 +92,11 @@ const Appeals: React.FC<AppealsProps> = ({ accountListId }) => {
...
         </>
       ) : data?.regularAppeals.nodes.length ? (
-        <>
-          {data.regularAppeals.nodes.map((appeal) => (
-            <Box key={appeal.name} mb={3}>
-              <Appeal appeal={appeal} changePrimary={handleChangePrimary} />
-            </Box>
-          ))}
-        </>
+        data.regularAppeals.nodes.map((appeal) => (
```
canac, line 95: @dr-bizz Do you know why moving it outside of the fragment changes things?

dr-bizz: The issue was the `key`. At first, it was the appeal's name, which sometimes had duplicates. I switched it to the `id`, which would be unique.

I don't think removing the fragment did anything apart from simplifying the code (in my opinion). Maybe React read the Fragment as an item and kept it in the render, but that is unlikely.

canac: Gotcha. I missed that the `key` changed. That's probably what fixed it.

PR #1197, opened 2024-11-15
[no-Jira] Deduplicate frontend yarn packages

canac:
## Description

I hope this will decrease bundle sizes because emotion packages won't be included in the bundle multiple times anymore.

## Checklist:


canac:
@dr-bizz I guess this doesn't actually change the bundle sizes. I can close this if you think it's not worth doing.

dr-bizz (commented):
How did you know it was duplicated, and then how did you remove it. Is there a yarn command?

canac:
> How did you know it was duplicated, and then how did you remove it. Is there a yarn command?

@dr-bizz `yarn dedupe --check` will show you packages that can be deduplicated, and `yarn dedupe` will do the deduplication. For this PR I ran `yarn dedupe '@emotion/*' '@mui/*' '@popper/*'` to only deduplicate packages that look like they are used on the frontend. Most of the duplicates appear to come from dev dependencies and aren't included in the client bundle.

canac:
@dr-bizz Did you have any more thoughts on this? The change in bundle size isn't enough to trigger the bundle analyzer comment, but it goes down by 235 bytes for `_app.js` and about 300 additional bytes for the `flows/setup` and `fixCommittmentInfo` pages.

dr-bizz (approved):
You are fine to merge this, even if it doesn't change that much

PR #1201, opened 2024-11-18
Fixing issue with error when deleting contact from appeal.

dr-bizz:
## Description
On appeals, when deleting a contact from an appeal, it would sometimes error. This PR fixes this bug.

The issue arose as I added functionality that tested the contact in question is on the appeal. However, since it meant we had to load all the contacts before we deleted it, if the user clicks the "Remove" button before it all loads, it would cause this error.
I could have prevented the "Remove" button from being selected until all contacts were loaded or removed the functionality and let the server check and produce the error. I opted to remove it as it adds a better UX, and as a bonus, there's less code to manage.

## Checklist:

### DIFF
--- a/src/components/Tool/Appeal/Modals/DeleteAppealContact/DeleteAppealContactModal.test.tsx
+++ b/src/components/Tool/Appeal/Modals/DeleteAppealContact/DeleteAppealContactModal.test.tsx
@@ -33,7 +33,6 @@ jest.mock('notistack', () => ({
 const accountListId = 'abc';
 const appealId = 'appealId';
 const contactId = 'contact-1';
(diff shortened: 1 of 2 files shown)


canac:
Do you mind putting a link to the HelpScout ticket in the description and putting the ticket number in the title?

canac:
> I might talk to the backend team about making the API accept a contact ID

@dr-bizz I'd recommend that if they have capacity. In most cases, I don't think that the API should make clients deal with join ids.

dr-bizz:
The backend will be adding the contact ID to GraphQL. so I will be closing this PR and will open a new one when I need to update this mutation.

PR #1206, opened 2024-11-20
[MPDX-8445] Next.js v14 upgrade

canac:
## Description

* Upgrade to Next.js 14
* Upgrade to Node.js 18.19.0 because that is the oldest version suported by Next.js 14
* Update the Amplify config and GitHub Actions workflows to use the node version defined in `.tool-versions` instead of needing to update the version number in lots of places

Code changes shouldn't be needed because we don't use any Next.js features that had breaking changes: https://nextjs.org/docs/app/building-your-application/upgrading/version-14#v14-summary

[MPDX-8445](https://jira.cru.org/browse/MPDX-8445)

## Checklist:


amplify.yml
```diff
@@ -3,8 +3,10 @@ frontend:
   phases:
     preBuild:
       commands:
-        - nvm install 18.13
-        - nvm use 18.13
+        - NODE_VERSION=$(cat .tool-versions | grep '^nodejs' | sed 's/nodejs //')
+        - echo "Using Node.js version $NODE_VERSION"
+        - nvm install "$NODE_VERSION"
+        - node -v
```
dr-bizz, line 9: This is great! It nwill allow the non-Cru orgs to not have to edit this file if they have a different version

src/lib/intlFormat.test.ts
```diff
@@ -253,7 +253,7 @@ describe('intlFormat', () => {
         locale,
       );
 
-      expect(date.trim()).toBe('Jan 16, 2024, 6:34\u202fPM UTC');
+      expect(date.trim()).toBe('Jan 16, 2024, 6:34 PM UTC');
```
dr-bizz, line 256: Did this break with the upgrade of Node.js?

dr-bizz (approved):
This works great! it Might be worth to wait until Monday to merge in case something goes wrong over the weekend.

src/lib/intlFormat.test.ts
canac: Yes, it was changed in node 18.15.0

PR #1209, opened 2024-11-21
[MPDX-8446] Automate translation download and extraction

canac:
## Description

* On every commit to main run `yarn extract` before uploading labels to OneSky. That way translation can begin immediately on any new labels instead of needing to wait for `public/locales/en/translation.json` to be updated. (The extracted labels are not committed and are discarded when the action terminates)
* During deployment to Amplify, run `yarn onesky:download` to deploy the app with the latest labels from OneSky. I tested that if that command fails for any reason, the deployment will still continue successfully.
  * I added the `ONESKY_*` environment variables to Amplify.
* Add a new workflow that runs daily to create a PR with any new labels from `yarn extract` or `yarn onesky:download`. Here is an example from my fork of the PR that will be created: https://github.com/canac/mpdx-react/pull/5. If the PR hasn't merged, the action will rebase and force-push any new labels to the branch.

[MPDX-8446](https://jira.cru.org/browse/MPDX-8446)

## Checklist:

### DIFF
--- a/.github/workflows/onesky.yml
+++ b/.github/workflows/onesky.yml
@@ -1,0 +1,46 @@
+name: OneSky Translation PR
+
+on:
+  schedule:
+    # Run every day at 10:50am UTC (5:50am EST). To avoid delays and possibly dropped jobs GitHub
+    # recommends not scheduling jobs at the top of the hour.
+    - cron: '50 10 * * *'
+  workflow_dispatch:
+
+permissions:
+  id-token: write
+  contents: write
+  pull-requests: write
+
+jobs:
+  download:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version-file: .tool-versions
+      - name: 📦 Install Dependencies
+        run: yarn install --immutable --immutable-cache
+      - name: 🌐 Extract Translations
+        run: yarn extract
+      - name: ⛅🔽 OneSky Download
+        env:
+          ONESKY_API_KEY: ${{ secrets.ONESKY_API_KEY }}
+          ONESKY_API_SECRET: ${{ secrets.ONESKY_API_SECRET }}
+          ONESKY_PROJECT_ID: ${{ secrets.ONESKY_PROJECT_ID }}
+        run: yarn onesky:download
+      - name: 🔀 Create PR
+        uses: peter-evans/create-pull-request@v7
+        env:
+          # Disable git hooks because we are only modifying translation files
+          HUSKY: 0
+        with:
+          branch: bot-update-translations
+          add-paths: |
+            public/locales/
+          # Sign commits so that the author is a bot
+          sign-commits: true
+          commit-message: Run `yarn extract` and `yarn onesky:download`
+          title: '[no-Jira] Update translations'
+          body: Update translations with the latest labels extracted from the components and downloaded from OneSky.
--- a/amplify.yml
+++ b/amplify.yml
@@ -19,6 +19,7 @@ frontend:
         - yarn
         - yarn disable-telemetry
         - yarn gql
+        - yarn onesky:download
     build:
       commands:
         - yarn build:amplify
--- a/amplify.yml
+++ b/amplify.yml
@@ -19,7 +19,7 @@ frontend:
         - yarn
         - yarn disable-telemetry
         - yarn gql
-        - yarn onesky:download
+        - timeout 1m yarn onesky:download || echo "yarn onesky:download timed out after 1 minute or failed with an error"
     build:
       commands:
         - yarn build:amplify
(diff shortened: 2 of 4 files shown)


wrandall22:
> I tested that if that command fails for any reason, the deployment will still continue successfully.

Have we tested what happens when we get a timeout from OneSky?

canac:
> Have we tested what happens when we get a timeout from OneSky?

@wrandall22 Not specifically. I tested the case where the environment variables aren't provided and it logs a 401 message and terminates with a successful error code. Are you afraid of it hanging forever if OneSky doesn't respond?

wrandall22:
Right, my concern would be it hanging forever or making the deployment timeout.

canac:
> Right, my concern would be it hanging forever or making the deployment timeout.

@wrandall22 OK, I set it up to timeout after one minute. It takes 10 seconds to run locally on my machine, so that should be plenty of time. I tested in a branch that `timeout 1s sleep 5 || echo "Timed out"` prints the message and doesn't stop the deployment.

amplify.yml
```diff
@@ -17,6 +17,7 @@ frontend:
         - yarn
         - yarn disable-telemetry
         - yarn gql
+        - timeout 1m yarn onesky:download || echo "yarn onesky:download timed out after 1 minute"
```
dr-bizz, line 20:
```suggestion
        - timeout 1m yarn onesky:download || echo "yarn onesky:download failed or timed out after 1 minute"
```

.github/workflows/onesky.yml
```diff
@@ -0,0 +1,46 @@
+name: OneSky Translation PR
+
+on:
+  schedule:
+    # Run every day at 10:50am UTC (5:10am EST). To avoid delays and possibly dropped jobs GitHub
```
dr-bizz, line 5:
```suggestion
    # Run every day at 10:10am UTC (5:10am EST). To avoid delays and possibly dropped jobs GitHub
```

.github/workflows/onesky.yml
```diff
@@ -0,0 +1,46 @@
...
+          ONESKY_API_KEY: ${{ secrets.ONESKY_API_KEY }}
+          ONESKY_API_SECRET: ${{ secrets.ONESKY_API_SECRET }}
+          ONESKY_PROJECT_ID: ${{ secrets.ONESKY_PROJECT_ID }}
+        run: yarn onesky:download
+      - name: 🔀 Create PR
+        uses: peter-evans/create-pull-request@v7
+        env:
+          # Disable git hooks because we are only modifying translation files
+          HUSKY: 0
+        with:
```
dr-bizz, line 38: Do we want to add reviewers to this?

dr-bizz: What are your thoughts about adding `delete-branch: true` to delete thre branch after each merge and so the branch is only around if changes are to be made.

.github/workflows/onesky.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: 🌐 Extract Translations
+        run: yarn extract
+      - name: ⛅🔽 OneSky Download
+        env:
+          ONESKY_API_KEY: ${{ secrets.ONESKY_API_KEY }}
+          ONESKY_API_SECRET: ${{ secrets.ONESKY_API_SECRET }}
+          ONESKY_PROJECT_ID: ${{ secrets.ONESKY_PROJECT_ID }}
+        run: yarn onesky:download
+      - name: 🔀 Create PR
+        uses: peter-evans/create-pull-request@v7
```
dr-bizz, line 34: Is there a way to run a rebase, to get the most up to date translation changes?

dr-bizz (commented):
This is looking great! I left some comments

amplify.yml
canac: `onesky/download.js` catches most errors, logs them, and exits successfully, so it would take a syntax error or something like that to get it to print this message due to a non-timeout failure. But I can update it.

.github/workflows/onesky.yml
canac: > What are your thoughts about adding `delete-branch: true` to delete thre branch after each merge and so the branch is only around if changes are to be made.

We have this repo configured to automatically delete branches from merged PRs, so that shouldn't be an issue

canac: > Do we want to add reviewers to this?

Adding team reviewers requires a Personal Access Token with `repo` scope or making a GitHub App with those permissions (https://github.com/peter-evans/create-pull-request/tree/v7/#action-inputs). If we use a PAT, we have to deal with rotating it when it expires and having to get someone else to create one if the person who created it leaves our GitHub org or the `web-engineering-js` team. It's also less secure to use a long-lived PAT instead of the short-lived automatic `GITHUB_TOKEN`. Making a GitHub App for this also doesn't feel worth it.

If the goal is to be alerted when this PR gets created, I think we can mention `@CruGlobal/web-engineering-js` in the PR body, and we'll all get a notification.

dr-bizz (approved):
Thank you for making this! Lets test it out and review after some time

canac:
> Lets test it out and review after some time

@dr-bizz Yes, it would be good to revisit this maybe in the new year to see if we would like to make any adjustments.

PR #1212, opened 2024-11-22
[MPDX-8445] Next.js v15 upgrade

canac:
## Description

Upgrade to Next.js v15. There are lots of breaking changes in the v15 release but not in features that we are using: https://nextjs.org/docs/app/building-your-application/upgrading/version-15.

I only needed to remove the `swcMinify` config option that started defaulting to true in Next.js 13 and was removed in Next.js 15.

[MPDX-8445](https://jira.cru.org/browse/MPDX-8445)

## Checklist:


dr-bizz (commented):
It mentions that fetch requests are not cached by default. I know we don't use fetch much and might not need to add caching but thought it was worth noting.

The code and the web ap look great.


Also it looks like `swcMinify` is allowed in Next.js 15 - https://nextjs.org/docs/messages/swc-minify-enabled

canac:
> It mentions that fetch requests are not cached by default. I know we don't use fetch much and might not need to add caching but thought it was worth noting.

@dr-bizz The only places I saw that we were making `fetch` requests on the server were by Apollo for GraphQL requests. Since GraphQL uses the `POST` method and Next.js only cached `GET` requests, we shouldn't be affected by this change.

> Also it looks like `swcMinify` is allowed in Next.js 15 - https://nextjs.org/docs/messages/swc-minify-enabled

That link is just additional info for an error message generated by older Next.js versions. Here's the PR that removed `swcMinify`: https://github.com/vercel/next.js/pull/65690.

dr-bizz (approved):

PR #1213, opened 2024-11-25
[no-Jira] Give bundle-analyzer permission to comment on dependabot PRs

canac:
## Description

Currently, Dependabot PRs fail in the `bundle-analzyer` step because the default GitHub token doesn't have permission to create a comment on the PR ([example run](https://github.com/CruGlobal/mpdx-react/actions/runs/11895982183/job/33146819121?pr=1200)). It appears that the fix is to give additional permissions to that workflow step ([docs](https://github.blog/changelog/2021-10-06-github-actions-workflows-triggered-by-dependabot-prs-will-respect-permissions-key-in-workflows/)).

## Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,6 +121,8 @@ jobs:
 
   bundle-analyzer:
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3


.github/workflows/ci.yml
```diff
@@ -121,6 +121,8 @@ jobs:
 
   bundle-analyzer:
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
```
dr-bizz, line 125: Can we add the permissions to just what needs it? I believe the `transferwise/actions-next-bundle-analyzer@v2` step is what creates the comment. Can we make only that step have the write permission?

dr-bizz (commented):
Thank you for fixing this

.github/workflows/ci.yml
canac: That's a good idea, but according to [the docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token#overview), it seems that you can set permissions for an entire workflow or for a specific job, but not for a specific step in a job

> You can use permissions either as a top-level key, to apply to all jobs in the workflow, or within specific jobs. When you add the permissions key within a specific job, all actions and run commands within that job that use the GITHUB_TOKEN gain the access rights you specify.

dr-bizz (approved):

PR #1216, opened 2024-12-05
MPDX-8488 - Prevent users from deleting Siebel contacts

dr-bizz:
## Description
This PR is fixing a bug reported by a user on [HelpScout](https://secure.helpscout.net/conversation/2774814376/1265656?viewId=7296147).
If a third party, such as Siebel, creates a contact, we shouldn't allow the user the ability to delete the contact. Deleting the contact will remove it, but Siebel will only add the contact back since it thinks it should exist.

To fix this issue, before the user can delete the contact, I perform a check to ensure all the contact's data is editable, and if it is removed, it will be removed for good.

Test:
1. Find a contact which was created by Siebel [You can look at the contacts on the HelpScout ticket](https://secure.helpscout.net/conversation/2774814376/1265656?viewId=7296147)
2. Go to the contact and click on more actions.
3. Click delete. This will open up the delete modal
4. You should see a message stating you can't delete this contact and need to contact donation services.

Repeat this step for a contact created on MPDX; you should be able to delete the contact. _Make sure you aren't impersonating anyone._

## Checklist:


dr-bizz:
I did the work on this PR, and then I couldn't run `yarn lint` without updating `@typescript-eslint/parser`. I'm unsure how this happened. Maybe I'm doing something wrong. Anyways, I have updated it to ensure `yarn lint` works.

src/components/Contacts/ContactDetails/ContactDetailsHeader/DeleteContactModal/ContactSource.graphql
```diff
@@ -0,0 +1,25 @@
+query ContactSource($accountListId: ID!, $contactId: ID!) {
+  contact(accountListId: $accountListId, id: $contactId) {
+    id
+    source
+    addresses {
+      nodes {
+        source
```
canac, line 7: These three sets of nodes need an `id` field.

package.json
```diff
@@ -118,7 +118,7 @@
     "@types/testing-library__jest-dom": "^5.14.5",
     "@types/uuid": "^9.0.1",
     "@typescript-eslint/eslint-plugin": "^7.5.0",
-    "@typescript-eslint/parser": "^7.5.0",
+    "@typescript-eslint/parser": "8.17.0",
```
canac, line 121: Why are we updating this package? Should that be done in a different PR?

src/components/Contacts/ContactDetails/ContactDetailsHeader/DeleteContactModal/DeleteContactModal.tsx
```diff
@@ -22,15 +25,71 @@ interface DeleteContactModalProps {
...
+  const accountListId = useAccountListId() ?? '';
+  const [canDelete, setCanDelete] = useState(true);
+
+  const { data } = useContactSourceQuery({
+    variables: { accountListId, contactId },
+    skip: !open && !contactId,
+  });
+  const contactSources = data?.contact;
+
+  useEffect(() => {
```
canac, line 48: Instead of having an effect that calculates the value of `canDelete` and sets it, this should be a `useMemo` that returns `canDelete`. Changing that will avoid an extra render.

src/components/Contacts/ContactDetails/ContactDetailsHeader/DeleteContactModal/DeleteContactModal.tsx
```diff
@@ -22,15 +25,71 @@ interface DeleteContactModalProps {
...
+          foundNonEditableEmailAddress,
+          foundNonEditablePhone,
+        };
+      },
+    );
+    const isPersonNonEditable = hasNonEditablePersonData.some((person) => {
+      return (
+        person.foundNonEditableEmailAddress || person.foundNonEditablePhone
+      );
+    });
```
canac, line 81: I believe you can combine these two loops.

```suggestion
    const hasNonEditablePersonData = contactSources.people?.nodes?.some(
      (people) => {
        const foundNonEditableEmailAddress = people.emailAddresses.nodes.some(
          (email) => !isEditableSource(email.source),
        );
        const foundNonEditablePhone = people.phoneNumbers.nodes.some(
          (phone) => !isEditableSource(phone.source),
        );
        return foundNonEditableEmailAddress || foundNonEditablePhone;
      },
    );
```

src/components/Contacts/ContactDetails/ContactDetailsHeader/DeleteContactModal/DeleteContactModal.tsx
```diff
@@ -39,9 +98,14 @@ export const DeleteContactModal: React.FC<DeleteContactModalProps> = ({
...
-          {t(
-            'Are you sure you want to permanently delete this contact? Doing so will permanently delete this contacts information, as well as task history. This cannot be undone. If you wish to keep this information, you can try hiding this contact instead.',
-          )}
+          {canDelete &&
+            t(
+              'Are you sure you want to permanently delete this contact? Doing so will permanently delete this contacts information, as well as task history. This cannot be undone. If you wish to keep this information, you can try hiding this contact instead.',
+            )}
+          {!canDelete &&
+            t(
+              "This contact cannot be deleted because part or all of the contact's data is sourced from a third Party. Please email Donation Services to request that this contact be deleted, or you can hide this contact instead.",
```
canac, line 107: What do you think of this change to match more closely the text in `EditContactAddressModal`?

```suggestion
              "This contact cannot be deleted because part or all of the contact's data syncs with Donation Services. Please email Donation Services to request that this contact be deleted, or you can hide this contact instead.",
```

src/components/Contacts/ContactDetails/ContactDetailsHeader/DeleteContactModal/DeleteContactModal.tsx
```diff
@@ -22,15 +25,71 @@ interface DeleteContactModalProps {
...
+    // Which will only recreate the data after deleting it on MPDX.
+    // To prevent this confusion, we do not allow a contact to be deleted if it has non editable data.
+
+    const isContactNonEditable = !isEditableSource(contactSources.source ?? '');
+
+    const isAddressNonEditable = contactSources.addresses?.nodes.some(
+      (address) => !isEditableSource(address.source ?? ''),
+    );
+
+    const hasNonEditablePersonData = contactSources.people?.nodes?.map(
```
canac, line 63: These checks will also prevent users from deleting contacts with information that was imported from Google (and possibly TNT). Do we want that?

package.json
dr-bizz: I wasn't able to commit anything until I upgraded this package. If you want, I can add it in a separate PR, but I need it to make future changes to this PR.

canac: Interesting, I do see this warning, but it lints successfully. Do you mind adding a caret `^` to the version?

```
WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.
```

dr-bizz: Upon investigating further, I was getting a Lint error (which wasn't obvious) with something else.

I will add the caret to the version.

dr-bizz:
Forgot to run `yarn` ...again

canac (commented):
The changes look good! Let me know if you hear back from Scott via another communications channel.

src/components/Contacts/ContactDetails/ContactDetailsHeader/DeleteContactModal/DeleteContactModal.tsx
```diff
@@ -12,37 +17,210 @@ import {
...
+    const sources: Map<string, string[]> = new Map();
+    sources.set('contact', [contactSources.source ?? '']);
+
+    const isAddressNonEditable = contactSources.addresses?.nodes.reduce(
+      (acc, address) => {
+        sources.set('address', [
+          ...(sources.get('address') ?? []),
+          address.source,
+        ]);
+        return acc || !isEditableSource(address.source ?? '');
```
canac, line 89: Address sources aren't nullable

```suggestion
        return acc || !isEditableSource(address.source);
```

src/components/Contacts/ContactDetails/ContactDetailsHeader/DeleteContactModal/DeleteContactModal.tsx
```diff
@@ -12,37 +17,210 @@ import {
...
+          phone: [],
+        },
+      };
+    }
+    // We ensure the contact was created on MPDX and that all the data is editable.
+    // If any data is not editable, this means it was created by a third party.
+    // Which will only recreate the data after deleting it on MPDX.
+    // To prevent this confusion, we do not allow a contact to be deleted if it has non editable data.
+
+    const isContactNonEditable = !isEditableSource(contactSources.source ?? '');
```
canac, line 78: Maybe I'm missing something, but this code seems unnecessarily complex. Would something like this work?

```
const addressSources = new Set<string>();
const emailSources = new Set<string>();
const phoneSources = new Set<string>();

contactSources.addresses?.nodes.forEach(address => {
  if (!isEditableSource(address.source)) {
    addressSources.set(address.source);
  }
}

// Then do the same thing for email and phone sources

return {
  canDelete: isEditableSource(contact.source) && addressSources === 0 && emailSources === 0 && phoneSources === 0,
  addressSources,
  emailSources,
  phoneSources,
};
```

By using a set of an array we can omit creating the unique sources step. And it also avoids having reduce functions that both return a value and mutate variables in the scope. (`reduce` functions are ideally pure functions).

Then in the JSX, you can do `Array.from(addressSources).join` or `[...addressSources].join`

src/components/Contacts/ContactDetails/ContactDetailsHeader/DeleteContactModal/DeleteContactModal.tsx
```diff
@@ -12,37 +17,210 @@ import {
...
+            <Typography fontWeight="bold">
+              {t(
+                `For contacts originating from Siebel or Donor Hub, email Donation Services to request deletion.`,
+              )}
+            </Typography>
+            <br />
+            <br />
+            <Typography variant="h6">{t('Data sources:')}</Typography>
+            <List dense={true}>
+              {!!dataInfo.uniqueSources.contact.length && (
```
canac, line 172: If you don't use my suggestion above, it looks like `sources.contact` is never pushed to, so `uniqueSources.contact` will always be empty.

canac (commented):
I have a suggestion for how to simplify calculating the unique sources, so let me know what you think of that. The core logic looks good though!

src/components/Contacts/ContactDetails/ContactDetailsHeader/DeleteContactModal/DeleteContactModal.tsx
dr-bizz: Looks like I missed that one.

src/components/Contacts/ContactDetails/ContactDetailsHeader/DeleteContactModal/DeleteContactModal.tsx
dr-bizz: Thank you for this recommendation. I have made these changes with some modifications.

dr-bizz:
Screenshot 2024-12-10 at 7 37 23 AM

src/components/Contacts/ContactDetails/ContactDetailsHeader/DeleteContactModal/DeleteContactModal.tsx
```diff
@@ -12,37 +17,178 @@ import {
...
+    const emailSources = new Set<string>();
+    const phoneSources = new Set<string>();
+
+    contactSources.addresses?.nodes.forEach((address) => {
+      if (!isEditableSource(address.source)) {
+        addressSources.add(address.source);
+      }
+    });
+
+    contactSources.people?.nodes?.forEach((person) => {
```
canac, line 85: `nodes` isn't nullable.

```suggestion
    contactSources.people?.nodes.forEach((person) => {
```

src/components/Contacts/ContactDetails/ContactDetailsHeader/DeleteContactModal/DeleteContactModal.tsx
```diff
@@ -12,37 +17,178 @@ import {
...
+  const accountListId = useAccountListId() ?? '';
+
+  const { data } = useContactSourceQuery({
+    variables: { accountListId, contactId },
+    skip: !open && !contactId,
+  });
+  const contactSources = data?.contact;
+
+  const dataInfo: DataInfo = useMemo(() => {
+    if (!contactSources) {
```
canac, line 60: I don't think we need a special case when `contactSources` is undefined. Just change the loops to `contactSources?.field` and no sources will be added when `contactSources` is undefined.

src/components/Contacts/ContactDetails/ContactDetailsHeader/DeleteContactModal/DeleteContactModal.tsx
```diff
@@ -12,37 +17,178 @@ import {
...
+      person.phoneNumbers.nodes.forEach((phone) => {
+        if (!isEditableSource(phone.source)) {
+          phoneSources.add(phone.source);
+        }
+      });
+    });
+
+    return {
+      canDeleteWithoutIssues:
+        isEditableSource(contactSources.source ?? '') &&
```
canac, line 100: If the source is an empty string, `isEditableSource` will return false. I think we want to convert `null` to `undefined` so that `isEditableSource` returns true.

```suggestion
        isEditableSource(contactSources.source ?? undefined) &&
```

src/components/Contacts/ContactDetails/ContactDetailsHeader/DeleteContactModal/DeleteContactModal.tsx
```diff
@@ -12,37 +17,178 @@ import {
...
+        ) : (
+          <>
+            <Typography>
+              {t(
+                `Be cautious when deleting this contact, as its data may sync with Donation Services or other third-party systems. Deleting the contact will not remove it from those systems; consider hiding the contact instead.`,
+              )}
+            </Typography>
+            <Typography fontWeight="bold">
+              {t(
+                `For contacts originating from Siebel or Donor Hub, email Donation Services to request deletion.`,
```
canac, line 135: Most users won't know what Siebel is, and "US Donation Services" is what will show up in the list. Also, we can remove the space from "Donor Hub" to match what `sourceToStr` returns.

Is it possible to include the Donation Services email address in this message? Maybe that's bad though if it shows up for non-Cru orgs.

```suggestion
                `For contacts originating from Donation Services or DonorHub, email Donation Services to request deletion.`,
```

canac (approved):
I have a couple more non-blocking suggestions, but since they're minor, I'm approving. Thanks for your work on this!

src/components/Contacts/ContactDetails/ContactDetailsHeader/DeleteContactModal/DeleteContactModal.tsx
canac: Yeah, that works. I wouldn't want to delay this PR any further.

PR #1234, opened 2024-12-19
MPDX-8506 - Prevent hydration errors being sent to RollBar

dr-bizz:
## Description
Due to the amount of errors being sent to RollBar regarding hydration and it maxing out our log rate limit, we need to prevent them from being sent to Rollbar.

In this PR, I add a check to prevent any logs regarding hydration from being sent to Rollbar.

Screenshot 2024-12-19 at 8 31 13 AM

We might just be able to get away with ignoring `Minified React error` and not have to worry about `Expected server HTML` since that only shows locally.

## Checklist:

### DIFF
--- a/pages/_app.page.tsx
+++ b/pages/_app.page.tsx
@@ -85,7 +85,7 @@ const App = ({
   const { session } = pageProps;
   const rollbarConfig: Rollbar.Configuration = {
     accessToken: process.env.ROLLBAR_ACCESS_TOKEN,
-    environment: 'react_' + process.env.NODE_ENV,
+    environment: `react_${process.env.NODE_ENV}`,
     codeVersion: React.version,
     captureUncaught: true,
     captureUnhandledRejections: true,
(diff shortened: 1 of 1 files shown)


pages/_app.page.tsx
```diff
@@ -97,6 +97,12 @@ const App = ({
       },
     },
     enabled: !!process.env.ROLLBAR_ACCESS_TOKEN,
+    checkIgnore: (_, args) => {
+      // Ignore React hydration warnings
+      return (
```
canac, line 102: Can we specifically just ignore React errors 418, 423, and 425? I wouldn't want us to miss other React errors.

pages/api/utils/rollBar.ts
```diff
@@ -10,6 +10,12 @@ const rollbar = new Rollbar({
   captureUncaught: true,
   captureUnhandledRejections: true,
   enabled: isRollBarEnabled,
+  checkIgnore: (_, args) => {
```
canac, line 13: Is this necessary? This is the server-side Rollbar config, right? I'm pretty sure that hydration errors only happen on the client and wouldn't show up here.

canac (commented):
Thanks for looking into this!

pages/api/utils/rollBar.ts
dr-bizz: I will check again, but I added it here as I was getting the error `Expected server HTML` and wanted to ensure we don't pass this to Rollbar. We might be able to remove.

canac: I don't see any occurrences of that error in Rollbar

pages/_app.page.tsx
```diff
@@ -96,8 +96,15 @@ const App = ({
         },
       },
     },
-    enabled: !!process.env.ROLLBAR_ACCESS_TOKEN,
+    enabled: Boolean(process.env.ROLLBAR_ACCESS_TOKEN),
+    checkIgnore: (_, args) =>
+      [
```
canac, line 101:
```suggestion
      // Ignore React hydration warnings
      [
```

canac (approved):
✅

PR #1239, opened 2024-12-19
[no-Jira] Disable Dependabot PRs for non-security updates

canac:
## Description

* Prevent Dependabot from creating PRs for package updates that aren't vulnerable. This config would have prevented PRs like #1235, #1236, #1237, #1238 from being created.
* Update the GitHub workflow to replace all instances of invalid URL characters with dashes in the Amplify Preview URL comment instead of just the first.

@dr-bizz Let me know if you like those PRs, but I was thinking that we ignore them so they don't add noise and we can focus on Dependabot updates that resolve security vulnerabilities.

## Checklist:

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -22,7 +22,7 @@ jobs:
           role-to-assume: ${{ secrets.AWS_GITHUB_ACTIONS_ROLE }}
           output-credentials: true
       - name: deploy PR preview
-        uses: CruGlobal/amplify-preview-actions@handle-existing-branches
+        uses: CruGlobal/amplify-preview-actions@sed-replace-all
         with:
           branch_name: ${{ github.head_ref }}
           amplify_command: deploy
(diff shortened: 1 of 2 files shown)


dr-bizz:
Yes, I'm all for making less noise if we can.

.github/workflows/preview.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           role-to-assume: ${{ secrets.AWS_GITHUB_ACTIONS_ROLE }}
           output-credentials: true
       - name: deploy PR preview
-        uses: CruGlobal/amplify-preview-actions@handle-existing-branches
+        uses: CruGlobal/amplify-preview-actions@sed-replace-all
```
dr-bizz, line 25: What does this do?

dr-bizz (approved):
As I'm out today, I don't want to block you, but I was unsure what the below does and how it is different from `CruGlobal/amplify-preview-actions@handle-existing-branches`


```
CruGlobal/amplify-preview-actions@sed-replace-all
```

PR #1242, opened 2024-12-19
[No-Jira] Upgrade Apollo Server 4

caleballdrin:
## Description

Upgrade to Apollo Server 4

## Checklist:


dr-bizz (changes requested):
Thank you for being willing to take this on! This needed to happen.
We should wait to push this live until the new year when everyone is back. I would hate for people to be called away from family to fix an error.

On a quick review, you should remove the old packages. They're also comments in the `graphql-rest.page.ts` page that are related to these packages, which need to be removed.


`apollo-server-micro`
`apollo-datasource-rest`


You could use the `depcheck` NPM package to find others that aren't used.

pages/api/graphql-rest.page.ts
```diff
@@ -677,16 +702,16 @@ class MpdxRestApi extends RESTDataSource {
         per_page: pageSize,
         sort: `${sortAscending ? '' : '-'}${camelToSnake(sortField)}`,
       },
-      {
-        headers: {
-          'Content-Type': 'application/vnd.api+json',
-          'X-HTTP-Method-Override': 'GET',
-        },
+      headers: {
+        'Content-Type': 'application/vnd.api+json',
```
canac, line 706: Can you remove this header? It is redundant.

pages/api/graphql-rest.page.ts
```diff
@@ -1363,32 +1383,21 @@ const cors = Cors({
...
-export default cors(async (req, res) => {
+export default cors((req, res) => {
   if (req.method === 'OPTIONS') {
     res.end();
     return false;
   }
-  await startServer;
-  await apolloServer.createHandler({
-    path: '/api/graphql-rest',
+  startServerAndCreateNextHandler(apolloServer, {
```
canac, line 1394: This should resolve the `API resolved without sending a response for /api/graphql-rest, this may result in stalled requests.` errors I'm seeing.

```suggestion
export default cors(async (req, res) => {
  if (req.method === 'OPTIONS') {
    res.end();
    return false;
  }
  await startServerAndCreateNextHandler(apolloServer, {
```

canac, line 1394: Also, I believe we need to create the handler once outside of this function and then call it inside this function. That should fix the `An error occurred during Apollo Server startup. All GraphQL requests will now fail. The startup error was: You should only call 'start()' or 'startInBackgroundHandlingStartupErrorsByLoggingAndFailingAllRequests()' once on your ApolloServer.` error.


```ts
const handler = startServerAndCreateNextHandler(apolloServer, {
  context: async (req: NextApiRequest): Promise<Context> => ({
    dataSources: {
      mpdxRestApi: new MpdxRestApi(req.headers.authorization ?? ''),
    },
  }),
});

export default cors(async (req, res) => {
  if (req.method === 'OPTIONS') {
    res.end();
    return false;
  }
  await handler(req, res);
});

```

canac (commented):
This looks pretty good in my basic testing! In addition to removing the packages Daniel mentioned, you also need to run `yarn` and commit the changes it makes.

canac:
> On a quick review, you should remove the old packages.
> 
> `apollo-server-micro` `apollo-datasource-rest`
> 
> You could use the `depcheck` NPM package to find others that aren't used.

You should also be able to remove `micro` because it is a peer dependency of `apollo-server-micro` and won't be needed after removing `apollo-server-micro`.

caleballdrin:
> you also need to run `yarn` and commit the changes it makes.

@canac  I ran `yarn` and there aren't any changes to commit.

dr-bizz (approved):
This is looking Great! I've reviewed some of the delete requests and they are working.

Thank you for removing the old dependencies and upgrading Apollo.

pages/api/graphql-rest.page.ts
```diff
@@ -168,9 +169,21 @@ class MpdxRestApi extends RESTDataSource {
...
+    parsedBody: unknown;
+  }): Promise<GraphQLError> {
+    const error = await super.errorFromResponse({ response, parsedBody });
+    const restError =
+      parsedBody &&
+      typeof parsedBody === 'object' &&
+      'errors' in parsedBody &&
+      Array.isArray(parsedBody.errors)
+        ? parsedBody.errors[0]
+        : null;
```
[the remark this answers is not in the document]
canac: That is because `error` had used to have the `any` type so we could access whatever we wanted to. Now `error.extensions.response` and `parsedBody` (which are equivalent) have the `unknown` type so we have to check that it is an object with the expected structure before doing anything with any of the fields.

canac (approved):
I'm not seeing the warnings anymore in the console. Great work with this!

package.json
```diff
@@ -32,7 +32,8 @@
     "@apollo/client": "^3.6.9",
     "@apollo/server": "^4.11.2",
     "@apollo/subgraph": "^2.1.3",
-    "@babel/core": "^7.26.0",
+    "@as-integrations/next": "^3.2.0",
+    "@babel/core": "^7.20.5",
```
canac, line 38: You accidentally reverted the `@babel/core` upgrade.

canac:
This is good to go as far as I'm concerned!

PR #1249, opened 2025-01-06
MPDX-8508 Add Informational Text for Hidden Columns in Donations Table

dr-bizz:
## Description
If the user is on a smaller screen, when viewing the donations table in a contact's drawer, they might not see all the columns on it and not know to scroll vertically to view all the columns. We need to add text to inform the user that there might be extra columns.

HelpScout: https://secure.helpscout.net/conversation/2782399338/1269019?viewId=7296147
Jira: https://jira.cru.org/secure/RapidBoard.jspa?rapidView=3&view=detail&selectedIssue=MPDX-8508#

## Checklist:


.gitignore
```diff
@@ -54,3 +54,6 @@ node_modules
 # Lighthouse
 .lighthouseci
 lighthouse-results.md
+
+# Yarn Plug'n'Play (PnP) files
```
canac, line 58: Was this change intentional? I think we want to keep the `.pnp.*` files committed. Otherwise, devs will have to remember to run `yarn` any time `package.json` or `yarn.lock` changes.

src/components/Contacts/ContactDetails/ContactDonationsTab/ContactDonationsTab.tsx
```diff
@@ -126,6 +126,10 @@ export const ContactDonationsTab: React.FC<ContactDonationsProp> = ({
             visibleColumnsStorageKey="contact-donations"
             hideDisplayName={true}
           />
+          <Typography variant="body2" sx={{ mt: 2, fontStyle: 'italic' }}>
+            On smaller screens, some columns may be hidden. Please scroll
```
canac, line 130: Let's localize this label

canac (changes requested):

canac:
> They were popping up on my local and would cause me hassle when checking out to a new branch.

@dr-bizz What were the changes to `.pnp.cjs`? Was it about making one of the binary packages (I can't remember which one, but I think it was `@surma/{something}`) unplugged?

.gitignore
canac: It appears that the yarn developers intend for PnP files to be committed when using zero-install mode, which we are.

https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

dr-bizz: Yes, as the pnp files would always be edited after I ran `yarn` (without me making any changes). This causes lots of issues when checking out different branches.

I'm happy to keep them on this repo, but I need to find a way to ensure they don't change after running `yarn`. I'll take a look

src/components/Contacts/ContactDetails/ContactDonationsTab/ContactDonationsTab.tsx
dr-bizz: Good catch

.gitignore
canac: In the past, I've seen running `yarn` change the `@surma/rollup-plugin-off-main-thread` entry in `.pnp.cjs`. I thought you fixed that in d8180b5c0bb173794ee4c67a1cadfa921a55a0c3 though. What changes is `yarn` making to `.pnp.cjs` when you switch branches?

PR #1252, opened 2025-01-09
Fix login loop from logout redirect

dr-bizz:
## Description
This PR fixes the error where users find themselves in an endless loop on the login screen. When you are logged out after a certain amount of time, you are redirected to the homepage with the redirect of `/logout`, which, once you log in, you will be automatically logged out.

This PR prevents the redirect for the logout page from being added to the login redirect.

## Checklist:

### DIFF
--- a/pages/api/utils/pagePropsHelpers.test.ts
+++ b/pages/api/utils/pagePropsHelpers.test.ts
@@ -27,6 +27,15 @@ describe('pagePropsHelpers', () => {
         },
       });
     });
+
+    it("doesn't redirect to logout", () => {
+      expect(loginRedirect({ ...context, resolvedUrl: '/logout' })).toEqual({
+        redirect: {
+          destination: '/login',
+          permanent: false,
+        },
+      });
+    });
   });
 
   describe('enforceAdmin', () => {
--- a/pages/api/utils/pagePropsHelpers.ts
+++ b/pages/api/utils/pagePropsHelpers.ts
@@ -21,7 +21,11 @@ export const loginRedirect = (
   context: GetServerSidePropsContext,
 ): { redirect: Redirect } => ({
   redirect: {
-    destination: `/login?redirect=${encodeURIComponent(context.resolvedUrl)}`,
+    destination: `/login${
+      context.resolvedUrl !== '/logout'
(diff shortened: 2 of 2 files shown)


canac (commented):
The code looks good to break the redirect loop, but when I click the sign out button, it takes me to an Okta login page. Is it doing that for you?

canac:
I think you also need to change all the `signOut({ callbackUrl: 'signOut' })` calls to `signOut()`. When I do that it takes me to the login page with the page I was on in the redirect param, which is helpful (i.e. `http://localhost:3000/login?redirect=%2FaccountLists%2F6bbc265f-ec81-4b23-b297-5bd298c4d619`.

dr-bizz:
> I think you also need to change all the `signOut({ callbackUrl: 'signOut' })` calls to `signOut()`. When I do that it takes me to the login page with the page I was on in the redirect param, which is helpful (i.e. `http://localhost:3000/login?redirect=%2FaccountLists%2F6bbc265f-ec81-4b23-b297-5bd298c4d619`.

We need to keep the `{callbackUrl: 'signOut'}` on the `signOut()` function as this is how we know we should redirect it to Okta to sign out of all sessions.

src/components/Reports/HealthIndicatorReport/HealthIndicatorReport.tsx Line 270 - 283
```
redirect({ url, baseUrl }) {
        if (url.startsWith(baseUrl)) {
          return url;
        }
        if (url === 'signOut' && AUTH_PROVIDER === 'OKTA') {
          return `https://signon.okta.com/login/signout?fromURI=${encodeURIComponent(
            process.env.OKTA_SIGNOUT_REDIRECT_URL ?? '',
          )}`;
        }
        if (url.startsWith('/')) {
          return new URL(url, baseUrl).toString();
        }
        return baseUrl;
      },
      ```

**Comment by dr-bizz** [comment]
I believe the reason is that Okta doesn't know how to redirect back to localhost, but I could be wrong.

We run the same code, but Okta doesn't redirect us back to MPDX on local.
```
return `https://signon.okta.com/login/signout?fromURI=${encodeURIComponent(
   process.env.OKTA_SIGNOUT_REDIRECT_URL ?? '',
)}`;
```

PR #1258, opened 2025-01-16
MPDX-8515 - Fetching all pages on the coaches page

dr-bizz:
## Description
As there was a limit on the number of coaches we could grab at one time from the GraphQL, I've implemented the `useFetchAllPages` functionality into the coaches page to ensure all coaches are rendered and not just the first 25.

This PR continues the work done on PR https://github.com/CruGlobal/mpdx-react/pull/1257 Where I adjusted the number of coaches to be fetched from 25 to 100. However, the server limits us to only fetch 50 at a time.

To Test, please impersonate Shelley and navigate to her coaches page. You can find Shelley's email address from this [HelpScout ticket](https://secure.helpscout.net/conversation/2817301546/1288380?viewId=7296147)

## Checklist:

### DIFF
--- a/src/components/Coaching/CoachingList.tsx
+++ b/src/components/Coaching/CoachingList.tsx
@@ -3,12 +3,15 @@ import { Spa } from '@mui/icons-material';
 import { Box, Divider, Skeleton, Typography } from '@mui/material';
 import { styled } from '@mui/material/styles';
 import { useTranslation } from 'react-i18next';
-import { CoachingRow } from './CoachingRow/CoachingRow';
+import { useFetchAllPages } from 'src/hooks/useFetchAllPages';
+import { CoachingRow, CoachingRowWrapper } from './CoachingRow/CoachingRow';
 import { useLoadCoachingListQuery } from './LoadCoachingList.generated';
 
-interface CoachingListProps {
-  accountListId: string;
-}
+const LoadingCoach: React.FC = () => (
+  <CoachingRowWrapper>
+    <CoachingLoading role="listitem" data-testid="loading-coaches" />
+  </CoachingRowWrapper>
+);
 
 const CoachingListWrapper = styled(Box)(({ theme }) => ({
   width: '100%',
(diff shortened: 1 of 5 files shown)


canac:
@dr-bizz Since most users don't have many people they're coaching, what do you think about just using `useFetchAllPages` to load all of the coaches if there are multiple pages? Even people like Shelly with lots of coached staff, it's only 99, not hundreds. I think that would simplify our code, and could be a better experience for our users since infinite scroll has had some issues on other pages.

dr-bizz:
> @dr-bizz Since most users don't have many people they're coaching, what do you think about just using `useFetchAllPages` to load all of the coaches if there are multiple pages? Even people like Shelly with lots of coached staff, it's only 99, not hundreds. I think that would simplify our code, and could be a better experience for our users since infinite scroll has had some issues on other pages.

@canac I've updated the query to use useFetchAllPages

src/components/Coaching/CoachingList.tsx
```diff
@@ -57,9 +63,9 @@ export const CoachingList: React.FC<CoachingListProps> = ({
       <Box>
         {loading ? (
```
canac, line 64: This won't render anything until all the pages load. Can we show the skeletons if `!data && !error`?

src/components/Coaching/CoachingList.tsx
```diff
@@ -78,3 +84,11 @@ export const CoachingList: React.FC<CoachingListProps> = ({
     </CoachingListWrapper>
   );
 };
+
+const LoadingCoach: React.FC = () => (
```
canac, line 88: Does this need the `span` wrapper? Also, can we move it above the main component like we do in most files?

src/lib/apollo/cache.ts
```diff
@@ -42,6 +42,12 @@ export const createCache = () =>
         },
         merge: true,
       },
+      CoachingAccountLists: {
+        fields: {
+          contacts: paginationFieldPolicy,
+        },
+        merge: true,
+      },
```
canac, line 50: Do we need this?

canac (commented):
This looks really good! I've got a couple questions on it.

canac (approved):
Apart from the flakiness with loading multiple pages concurrently, this looks great!

PR #1262, opened 2025-01-27
MPDX-8521 - Formatting subpremise for US and Singapore addresses

dr-bizz:
## Description
In this PR, I changed how we format the `subpremise` (Apt, Unit or Suite number) on an address.
Before, we always put it in front of the street address formatted like `Apt 100/Street Address`. This is the correct format for the UK, CA and many other countries, but the US and Singapore prefer that `subpremise` be used at the end of the address.

[USPS stated the `subpremise` should be after the street address](https://pe.usps.com/text/pub28/28c2_003.htm)
[Singapore stated the `subpresmise` should be after the address prefixed with a hashtag.](https://www.upu.int/UPU/media/upu/PostalEntitiesFiles/addressingUnit/sgpEn.pdf)

[Helpscout](https://secure.helpscout.net/conversation/2827220856/1294018?viewId=7296147)
Jira: [MPDX-8521](https://jira.cru.org/browse/MPDX-8521)

## Checklist:

### DIFF
--- a/src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/AddAddressModal/AddAddressModal.test.tsx
+++ b/src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/AddAddressModal/AddAddressModal.test.tsx
@@ -204,7 +204,7 @@ describe('AddAddressModal', () => {
     userEvent.click(
       getByRole('option', { name: '100 Lake Hart Dr, Orlando, FL 32832, USA' }),
     );
-    expect(addressAutocomplete).toHaveValue('A/100 Lake Hart Drive');
+    expect(addressAutocomplete).toHaveValue('100 Lake Hart Drive A');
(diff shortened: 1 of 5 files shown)


wrandall22:
I wonder if we shouldn't be using a library to handle this for us, like `localized-address-format`.

dr-bizz:
I looked into a formatter. I didn't find one that does what we need it to.

I installed and tested the following NPM packages. 
1. `localized-address-format`
2. `@fragaria/address-formatter`
3. `i18n-postal-address`
Unfortunately, none of these took a `subpremise` as a parameter when formatting the address. I played around with adding the `subpremise` to the start and end of the street name, but it doesn't format it correctly.

dr-bizz:
Today, I met with Scott and the other developers, and we believe we can use the Google Places API for this task. However, it will require significant effort and has been scheduled for T2.

In the meantime, I have manually added the formatting for the US, Singapore, and most countries, including the UK and Canada. However, this will be replaced in the near future so we can manage this more efficiently.

PR #1265, opened 2025-02-11
[MPDX-8538] Default to calculated goal on accounts list page

canac:
## Description

When the user hasn't set a monthly goal, use the machine calculated goal on the accounts list page.

*Caveat*: If the user's default currency is different from their organization's default, we aren't able to calculate percentages because the goal amount is in one currency and their received and committed amounts are in a different currency.

[MPDX-8538](https://jira.cru.org/browse/MPDX-8538)

## Checklist:

### DIFF
--- a/src/components/AccountLists/AccountLists.tsx
+++ b/src/components/AccountLists/AccountLists.tsx
@@ -82,24 +82,25 @@ const AccountLists = ({ data }: Props): ReactElement => {
                 currency: preferencesCurrency,
                 healthIndicatorData,
               }) => {
-                const monthlyGoal =
-                  preferencesGoal ?? healthIndicatorData?.machineCalculatedGoal;
-                const currency =
-                  typeof preferencesGoal === 'number'
-                    ? preferencesCurrency
-                    : healthIndicatorData?.machineCalculatedGoalCurrency;
+                const hasPreferencesGoal = typeof preferencesGoal === 'number';
+                const monthlyGoal = hasPreferencesGoal
+                  ? preferencesGoal
+                  : healthIndicatorData?.machineCalculatedGoal;
+                const currency = hasPreferencesGoal
+                  ? preferencesCurrency
+                  : healthIndicatorData?.machineCalculatedGoalCurrency;
 
                 // If the currency comes from the machine calculated goal and is different from the
                 // user's currency preference, we can't calculate the received or total percentages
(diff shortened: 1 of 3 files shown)


src/components/AccountLists/AccountLists.tsx
```diff
@@ -76,18 +76,37 @@ const AccountLists = ({ data }: Props): ReactElement => {
...
+                // user's currency preference, we can't calculate the received or total percentages
+                // because the numbers are in different currencies
                 const receivedPercentage =
-                  receivedPledges / (monthlyGoal ?? NaN);
-                const totalPercentage = totalPledges / (monthlyGoal ?? NaN);
+                  currency === preferencesCurrency && monthlyGoal
+                    ? receivedPledges / monthlyGoal
+                    : NaN;
+                const totalPercentage =
+                  currency === preferencesCurrency && monthlyGoal
```
dr-bizz, line 100: You could make this if statement into a variable so we're don't repeat the code.

src/components/AccountLists/AccountLists.tsx
```diff
@@ -76,18 +76,37 @@ const AccountLists = ({ data }: Props): ReactElement => {
...
-                currency,
+                currency: preferencesCurrency,
+                healthIndicatorData,
               }) => {
+                const monthlyGoal =
+                  preferencesGoal ?? healthIndicatorData?.machineCalculatedGoal;
+                const currency =
+                  typeof preferencesGoal === 'number'
+                    ? preferencesCurrency
+                    : healthIndicatorData?.machineCalculatedGoalCurrency;
```
dr-bizz, line 90: Should we make these variables use the same IF statements to kae it easier to read?

I know `monthlyGoal` will not equal `preferencesGoal`  while `currency` equals `healthIndicatorData?.machineCalculatedGoalCurrency`. but it could be easier to understand if it was like the below

```suggestion
                const usePreferencesGoal = typeof preferencesGoal === 'number';
                const monthlyGoal = usePreferencesGoal
                  ? preferencesGoal
                  : healthIndicatorData?.machineCalculatedGoal;
                const currency = usePreferencesGoal
                    ? preferencesCurrency
                    : healthIndicatorData?.machineCalculatedGoalCurrency;
```

dr-bizz (approved):
This looks great! I only made a few comments, but this is looking great

PR #1280, opened 2025-03-20
MPDX-8572 - Upgrade to Node 20

dr-bizz:
## Description
In this PR, I've upgraded Node to version `20.17.0`; it was on version 18 before. To use Node 20, I needed to upgrade Yarn to version 4, as version 3 had an issue with PNP (Plug N Play) when using Node 20.

I could have upgraded Node to `20.18.3`, but as there is no `@types/node` version available for that release since the `20.17.0` version was released this week, I opted for `20.17.0`.

Jira: [MPDX-8572](https://jira.cru.org/browse/MPDX-8572)

### Changes
- I've updated the README file to include the new node version.
- Deleted the PNP file and re-added it. I'm hoping this solves the PNP issue we've been having.
- Updated packages which needed to be updated.
- Upgraded Yarn to version 4.

I've gone through all of MPDX React's dependencies and devDependencies and confirmed that they work with Node `20.17.0`.

## Testing:
The preview link uses Node `20.17.0`, so feel free to test it out there or on your local.

## Prep before merging
Before this PR is merged, we will need to make these changes to the amplify.yml file on the AWS Amplify console.

```
...
commands:
    - nvm install 20.17.0
    - nvm use 20.17.0
    ...
    - yarn set version 4.7.0
...
```

## Checklist:


dr-bizz:
I'm unsure if the way I'm doing it is the best way.

I think the best way would be to go through keep dependency and check if the engine supports Node version `20.17.0`. 
Which will take forever 😭

dr-bizz:
> I'm unsure if the way I'm doing it is the best way.
> 
> I think the best way would be to go through keep dependency and check if the engine supports Node version `20.17.0`. Which will take forever 😭

After reviewing all the dependencies and devDependency, I found that the issue came from Yarn not being updated. However, I was able to check that each package we have supports Node 20.

package.json
```diff
@@ -140,9 +140,9 @@
     "lighthouse": "^11.7.0",
     "lint-staged": "^13.0.3",
     "next-compose-plugins": "^2.2.1",
-    "node-mocks-http": "^1.11.0",
+    "node-mocks-http": "1.16.2",
     "prettier": "^2.7.1",
-    "puppeteer": "^22.6.3",
+    "puppeteer": "24.4.0",
```
canac, line 145:
```suggestion
    "node-mocks-http": "^1.16.2",
    "prettier": "^2.7.1",
    "puppeteer": "^24.4.0",
```

.yarnrc.yml
```diff
@@ -15,4 +15,4 @@ supportedArchitectures:
     - darwin
     - win32
 
-yarnPath: .yarn/releases/yarn-3.8.6.cjs
+yarnPath: .yarn/releases/yarn-4.7.0.cjs
```
canac, line 18: I was able to get `corepack` to work locally by running `corepack enable` and removing this line. That lets us remove the `.yarn/releases/yarn-4.7.0.cjs` file. Did it not work for you?

.yarnrc.yml
```diff
@@ -1,6 +1,6 @@
-plugins:
-  - path: .yarn/plugins/@yarnpkg/plugin-interactive-tools.cjs
-    spec: "@yarnpkg/plugin-interactive-tools"
+compressionLevel: mixed
```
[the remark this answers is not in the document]
dr-bizz: Yeah, we can remove it. It's a lot faster without setting it.

.yarnrc.yml
dr-bizz: You're right it's working for me now

dr-bizz: I'm going to continue to look into this after lunch, but it looks like if you don't mention which yarn to use, it defaults to the global version, which could cause errors if we have projects on different yarn versions. Since non-orgs will be using it, it's prob best to keep the version of yarn in the repo.

.yarnrc.yml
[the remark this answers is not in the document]
dr-bizz: I removed it, and it edited 4,000 files. and GitHub prevents me from pushing these files up. so it's probably best we keep it on mixed.

dr-bizz:
When adding `cacheMigrationMode`, and clearing my yarn cache and re-installing. I didn't see a change in the amount of files it edited.

.yarnrc.yml
canac: If I run `corepack disable` then `yarn` it prints this error:

```
error This project's package.json defines "packageManager": "yarn@4.7.0". However the current global version of Yarn is 1.22.22.

Presence of the "packageManager" field indicates that the project is meant to be used with Corepack, a tool included by default with all official Node.js distributions starting from 16.9 and 14.19.
Corepack must currently be enabled by running corepack enable in your terminal. For more information, check out https://yarnpkg.com/corepack.
```

So I don't think it will accidentally let you use yarn v1 if you have `packageManager` set to v4.

canac:
> When adding `cacheMigrationMode`, and clearing my yarn cache and re-installing. I didn't see a change in the amount of files it edited.

After changing `cacheMigrationMode`, I had to set `compressionLevel: mixed` and run `git restore .yarn/cache yarn.lock --source=main && yarn`. If you already have the cache migrated on-disk, it was still using the old one until you revert the changes to the cache and the lockfile. Only after you have migrated the cache, then you can remove the `compressionLevel` config.

dr-bizz:
@canac I've updated the MPDX Node upgrade PR to use Git LFS for the cache. I had to delete all the cache zip files and then add LFS, and then re-add the cache.

It seems to ave worked, as before I wasn't able to push my changes since the files were so big.

I've updated the ReadMe with instructions on how to download the LFS files. Please let me know if you get stuck, so I can improve the ReadMe instructions.

.github/workflows/preview-close.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     if: contains(github.event.pull_request.labels.*.name, 'Preview Environment')
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
```
canac, line 13: Does this need Git LFS?

.github/workflows/preview.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       id-token: write
       pull-requests: write
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
```
canac, line 16: Does this need Git LFS?

.github/workflows/update-staging.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     if: github.ref == 'refs/heads/main' || contains(github.event.pull_request.labels.*.name, 'On Staging')
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
```
canac, line 14: Does this need Git LFS?

canac (commented):
Git LFS worked fine for me. Can you try removing the committed yarn release file and relying on corepack? Can you also add those carets to the version specifiers in `package.json`? Everything seems to be working well though.

`cacheMigrationMode: required-only` is less important to me now that we're storing files in Git LFS.

canac:
Also, the `yarn-check-cache` step worked for me locally. Why wasn't it working in CI?

.github/workflows/preview.yml
dr-bizz: I doubt it, as we're not installing any packages.

dr-bizz:
> Git LFS worked fine for me. Can you try removing the committed yarn release file and relying on corepack? Can you also add those carets to the version specifiers in package.json? Everything seems to be working well though.

If you remove the release and use corepack, the lint-staged in pre-commit no longer works.

dr-bizz:
@canac Maybe I'm doing it wrong. Let's connect after your lunch.

canac:
We're getting Amplify build errors. Can you add `corepack enable` to `amplify.yml`?

dr-bizz:
Ah now we're getting the `The remote archive doesn't match the expected checksum` error

canac (approved):

PR #1285, opened 2025-03-26
MPDX-8572 - Remove Git LFS Install as not needed

dr-bizz:
## Description
In this PR, I remove the documentation and AMplify command to install LFS, as it's not needed

Jira: MPDX-8572 

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,13 +45,10 @@ brew install git-lfs
 
 Alternatively, download from the [official site](https://git-lfs.com/).
 
-**Initialize Git LFS and pull LFS files**
+**Pull LFS files**
 
 ```bash
-# If you haven't initialized git-lfs locally before, run:
-git lfs install
-
-# Fetch the LFS files (Ensure you're inside the repo's directory):
+# Fetch the LFS files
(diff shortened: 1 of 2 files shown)


amplify.yml
```diff
@@ -24,8 +24,7 @@ frontend:
         - tar -xvzf git-lfs.tar.gz
         - mv git-lfs-3.6.1/git-lfs ./bin/
         - chmod +x ./bin/git-lfs
-        # Install git-lfs and pull down the files
-        - export PATH=$PATH:$(pwd)/bin && git-lfs install
+        # Pull down the files
         - export PATH=$PATH:$(pwd)/bin && git-lfs pull
```
canac, line 28: Should we simplify this now that we're only calling git-lfs once?

```suggestion
        - ./bin/git-lfs pull
```

canac (approved):

amplify.yml
dr-bizz: Unfortunately, that will cause an error in Amplify, since Git tries to call `git-lfs` after the pull has finished.

PR #1289, opened 2025-03-28
[MPDX-8569] Use Cru brand colors in Donation and HI graphs

canac:
## Description

The colors in the theme's `cruColors` aren't part of the Cru brand anymore. In the first commit, I updated `cruColors` to include all of the colors from the new [brand palette](https://www.cru.org/brand/color/#section-title-4). I also removed the `cru` prefix from the colors. `cruYellow`, for example, is now just `yellow`. I also moved the grays to `mpdxColors` and replaced `cruGray*` with `mpdxGray*` in the codebase. The only change in the UI is that places that used `cruYellow` changed from `#FFCF07` to `#FFD000`. (In the future, we should try to consolidate our colors and use Cru brand colors wherever possible. I added a BC item for this task.)

In the second commit, I updated the colors in the Donation graph and HI graph to use the new colors from the brand palette. There were contrast issues before when we used different shades of teal. I think the graphs pop more now too. I also updated the HI numbers in the accordions to match their colors in the graph, as requested by stakeholders.

Screenshot 2025-03-28 at 10 11 34 AM

Please let me know if you have feedback on the colors I chose.

Screenshot 2025-03-28 at 10 13 38 AM
Screenshot 2025-03-28 at 10 13 31 AM

MPDX-8569

## Checklist:


src/theme.ts
```diff
@@ -43,14 +59,33 @@ const graphColors = {
...
+    cerise: Palette['primary'];
+    ruby: Palette['primary'];
+    sky: Palette['primary'];
+    cyan: Palette['primary'];
+    turquoise: Palette['primary'];
+    navy: Palette['primary'];
+    mint: Palette['primary'];
+    green: Palette['primary'];
+    moss: Palette['primary'];
+    oliveDrab: Palette['primary'];
```
wrandall22, line 70: Should these be secondary instead of primary?

wrandall22:
@j2trumpet do you want to weigh in on color choice?

wrandall22:
In that first graph screenshot, I wonder about the committed color being the same as one of the colors in the bar (UYW). Could that get confusing or hard to read?

src/components/Dashboard/DonationHistories/DonationHistories.tsx
```diff
@@ -294,11 +297,7 @@ const DonationHistories = ({
                     <BarChart data={periods}>
                       <XAxis tickLine={false} dataKey="startDate" />
                       <Tooltip />
-                      <Bar
-                        dataKey="total"
-                        fill={palette.graphBlue1.main}
-                        barSize={10}
-                      />
+                      <Bar dataKey="total" fill={fills[0]} barSize={10} />
```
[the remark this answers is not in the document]
canac: On tiny screens, instead of the full graph showing, this condensed graph shows. Instead of having separate colors for each currency, it just shows the total in one color. Here it is just picking the first color, which is the same color users who only receive donations in one currency see.

src/theme.ts
canac: I resolved this by using an equivalent but more descriptive TypeScript type

canac:
> @j2trumpet do you want to weigh in on color choice?

I'm definitely open to feedback! FWIW, the four colors for the graphs were suggested by @rguinee.

> In that first graph screenshot, I wonder about the committed color being the same as one of the colors in the bar (UYW). Could that get confusing or hard to read?

@wrandall22 I went back and forth on that. I made sure to put the yellow last so that it will only show up if the user has received donations in 4 or more currencies. I could drop it, and only have 3 unique currency colors (they are reused if the user has more currencies than colors).

Thoughts on changing it to ruby? I think it may be too similar to graphite, the average color.

Screenshot 2025-03-31 at 11 21 14 AM

j2trumpet:
I think the colors look great, but I'm a little concerned about the yellow text. The number in your first screenshot looks fine, but the yellow text in the legend/captions under the graphs is a little harder to read because it's a thinner font weight. The [brand color guidelines](https://www.cru.org/brand/color/#section-title-6) recommend only putting yellow text on a white background if the text is bold. Can we try that and see how it looks?

canac:
@j2trumpet Thanks for reminding me about the legend text colors.

Here's how it looks in bold:

Screenshot 2025-03-31 at 1 06 49 PM

And here's how it looks in black instead of bold:

Screenshot 2025-03-31 at 1 07 54 PM

Let me know what you think, but yellow still seems a little hard to read. Maybe we drop yellow from the currencies list and just use it for the Committed line color?

wrandall22:
I think my preference is to drop yellow from the currencies list. I'm ok with replacing it with another tertiary color if we want, but not sure how many people are going to have over 3 currencies of donations.

j2trumpet:
The black text is in line with the brand guidelines. I think we'd want to use different colors for each item on the graph to avoid confusion. I'd keep yellow for committed, but if we want a fourth color option for currencies, maybe use orange instead and make it the last color to be used?

canac:
@wrandall22 When considering donations from the past year, 56 account lists have 5+ currencies. 103 have 4 currencies. Almost 300 have 3 currencies. So by removing yellow, we'd be ~tripling the number of account lists who see repeated currency colors. But in absolute terms, it's ~100 more account lists.

If the ruby goal line is OK, then I'm good with switching yellow for orange in the currencies list.

If we're OK going down to 3 currency colors, I think it would be best to reserve orange for the goal line.

The MPD health graph needs 4 colors, and I think I prefer yellow over orange there. Thoughts?

Screenshot 2025-03-31 at 2 07 17 PM
Screenshot 2025-03-31 at 2 07 09 PM

j2trumpet:
I concur with Bill, though the legend text for MPD health should match the color and font-weight from the donation chart.

canac:
Thanks everyone for the feedback so far.

* I made the legend text black instead of bold so that it is visible in the MPD Health graph (even bold yellow is tough to read).
* I made the tooltip text black and put in a colored rectangle next to the text.
* I changed the fourth currency color to orange and changed the goal color to ruby.

Screenshot 2025-03-31 at 3 33 47 PM
Screenshot 2025-03-31 at 3 33 29 PM

wrandall22 (approved):

j2trumpet:
Thanks for adding some color and formatting to the tooltip. It looks a lot better in your most recent screenshots than in the ones from the black/bold legend text examples.

caleballdrin (approved):
The colors look great to me! I like the use of the colored square instead of colored text for the tooltip.

dr-bizz:
Thank you for making these changes. 

I'm surprised the Ruby colour was picked for the goal, as red usually reflects a negative emotion. I would have gone with MPDX Blue.

Can we also change the other colours on MPDx to match the new Cru brand colour scheme?
Screenshot 2025-04-07 at 8 03 12 AM

Screenshot 2025-04-07 at 8 01 40 AM

Screenshot 2025-04-07 at 7 57 36 AM

dr-bizz (commented):
I put my comment above ^

canac:
> I'm surprised the Ruby colour was picked for the goal, as red usually reflects a negative emotion. I would have gone with MPDX Blue.

@dr-bizz I was picking colors that are the easiest to tell apart. We could try turquoise, but I think it looks a little close in color to the graphite average line. Personally, I think I still prefer ruby, but I'm curious other people's thoughts.

Screenshot 2025-04-07 at 11 41 14 AM

> Can we also change the other colours on MPDx to match the new Cru brand colour scheme?

@dr-bizz I would love to do that at some point! I've actually already added a [BC item](https://docs.google.com/spreadsheets/d/1k0OOBGivAn6RjUtgzmsTKbttmiYcXaeoRv24NIHE_cs/edit?gid=583951540#gid=583951540&range=49:49) for it. Unfortunately, I think that is out of scope for the Health Indicator work I'm allocated to.

j2trumpet:
> @dr-bizz I was picking colors that are the easiest to tell apart.

I've seen graphs like these where instead of using different colors for every line, they duplicate colors but will switch from a solid line to dashed. Is that 1) possible and 2) something we'd want to consider to effectively double our options for colors with the greatest difference?

wrandall22:
I think I like the dashed better.

rguinee:
Dashed looks cool. Good suggestion @j2trumpet !

dr-bizz:
I'm watching the recording of the Cru Brand Guidelines meeting. Rachael mentioned that the Tertiary palette should only be used on short-term things like social media posts, and also not so much by the main Cru brand. - As MPDX is under the Cru brand and is long-term, should we be using some of these colours?


[Watch Cru Brand Guidelines, talking about the tertiary palette.](https://drive.google.com/file/d/10rgmCHpju1ARc0OmpqLdo8lTfIaz221Q/view?usp=sharing&t=2155)

rguinee:
@dr-bizz In short, yes we should. The guidelines they've created are intended to protect the uniqueness of Cru's visual identity in a marketing and brand communication context. In a web and product context, we need to lean into the flexibility of the brand and it's broad color palette (arguably, even what they have now is limited) to support the hundreds of little design decisions and interaction elements of web components that we make along the way to creating something that's clearly a Cru product but still unique in its own right. These secondary and tertiary colors serve us well, but we still follow the heart of the brand guidelines and use them sparingly in accent details (the lines of a chart vs the background of a nav bar).

dr-bizz (approved):
Awesome work on this one!! Thank you for getting feedback from the teama nd implementing it. Great work.

PR #1294, opened 2025-04-03
[no-Jira] Improve panel layout navigation

canac:
## Description

* Use `aside` element for left and right panels.
* Use `main` element for main content.
* Add labels to left and right panels.

## Testing

* Go to the Contacts page
  * Use Ctrl+Option+Shift+Right/Left to navigate between the left panel, main content, and right panel (if it is moving between navbar items use Ctrl+Option+Shift+Up to navigate out of the navbar)
  * After opening a contact, Press Ctrl+Option+Shift+Up twice, then Ctrl+Option+Shift+Right to get to the contact's panel.

Other improvements are possible, but the main areas of the layout are much easier to navigate between in VoiceOver. I'm assuming that other accessibility software has similar methods for jumping between sections of the page.

## Checklist:

### DIFF
--- a/src/components/Contacts/ContactsMap/ContactsMapPanel.tsx
+++ b/src/components/Contacts/ContactsMap/ContactsMapPanel.tsx
@@ -169,7 +169,9 @@ export const ContactsMapPanel: React.FC<ContactMapsPanelProps> = ({
     <>
       <Box padding={2}>
         <Box display="flex" justifyContent="space-between" alignItems="center">
-          <Typography variant="h6">{t('Partners by Status')}</Typography>
+          <Typography variant="h6" id="left-panel-header">
+            {t('Partners by Status')}
+          </Typography>
           <IconButton onClick={onClose} aria-label={t('Close')}>
             <Close titleAccess={t('Close')} />
           </IconButton>
--- a/src/components/Shared/Filters/FilterPanel.tsx
+++ b/src/components/Shared/Filters/FilterPanel.tsx
@@ -614,9 +614,9 @@ export const FilterPanel: React.FC<FilterPanelProps & BoxProps> = ({
                 justifyContent="space-between"
                 alignItems="center"
               >
-                <Typography variant="h6">
+                <Typography variant="h6" id="left-panel-header">
                   {selectedFilterCount > 0
-                    ? t('Filter ({{count}})', {
+                    ? t('Filter ({{count}} active)', {
                         count: selectedFilterCount,
                       })
                     : t('Filter')}
--- a/pages/accountLists/[accountListId]/reports/financialAccounts/[financialAccountId]/entries.page.test.tsx
+++ b/pages/accountLists/[accountListId]/reports/financialAccounts/[financialAccountId]/entries.page.test.tsx
@@ -69,7 +69,7 @@ describe('Financial Accounts Page', () => {
     const { findByRole } = render(<Components />);
 
     expect(
-      await findByRole('heading', { name: 'Filter (1)' }),
+      await findByRole('heading', { name: 'Filter (1 active)' }),
     ).toBeInTheDocument();
   });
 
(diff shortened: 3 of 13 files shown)


src/components/Shared/Filters/FilterPanel.tsx
```diff
@@ -614,9 +614,9 @@ export const FilterPanel: React.FC<FilterPanelProps & BoxProps> = ({
                 justifyContent="space-between"
                 alignItems="center"
               >
-                <Typography variant="h6">
+                <Typography variant="h6" id="left-panel-header">
                   {selectedFilterCount > 0
-                    ? t('Filter ({{count}})', {
+                    ? t('Filter ({{count}} active)', {
```
wrandall22, line 619: Looks like this change broke a test.

src/components/Contacts/ContactsMap/ContactsMapPanel.tsx
```diff
@@ -169,7 +169,9 @@ export const ContactsMapPanel: React.FC<ContactMapsPanelProps> = ({
     <>
       <Box padding={2}>
         <Box display="flex" justifyContent="space-between" alignItems="center">
-          <Typography variant="h6">{t('Partners by Status')}</Typography>
+          <Typography variant="h6" id="left-panel-header">
```
wrandall22, line 172: Just want to make double check we are certain none of these left-panel-header ids can be on the screen at the same time.

canac: I just fixed the one place where `SidePanelsLayout` was nested inside itself resulting in duplicate ids.

dr-bizz (commented):
Thank you for making these changes; they are fantastic!  

One issue I'm struggling with is initiating the voiceover on the Contacts page. Perhaps it's just me, but it isn't intuitive to navigate to the filter menu, main content, and the contact drawer.  

When I refresh the page, I'm taken to the main navigation, which is great, but when I press `Ctrl+Option+shift+Up` followed by `Ctrl+Option+shift+Right`, I'm directed to the Help Scout icon, which causes me to bypass the filters, main content, and the contact drawer (if it is opened).  

To access the filters, main content, and contact drawer, I have to follow these steps using `Ctrl+Option+shift` to reach them:

1. Up (Only have to press up if VoiceOver was started to read the main navigation)
2. Right,
3. Up
4. Down,
5. Right (Focused on the filters)

canac:
Thanks for testing it out. I'm on Sequoia, and maybe different macOS versions are different. But for me once I get to "web content", down takes me to the navbar, then the right arrow takes me to filters, then the contacts list, then the right panel, then the HelpScout beacon. That feels fairly intuitive. I wonder why it was taking so many extra steps for you?

One thing I learned during the user study with a visually impaired user is that VoiceOver is free but is not the most advances accessibility software. The better ones are JAWS and NVD, and they are Windows-only. There is a way to quickly access the page regions with a spinner selector (Web Rotor) but I can't figure out how to get it back on.

My goal in this PR is to provide semantic roles and labels for the various sections of the pages. Various accessibility software  will do a better or poorer job at surfacing this information to the user in a helpful and actionable way, and VoiceOver may be one that falls short. (Or it could also just be my/our lack of familiarity with the tool.)

canac:
Here's another way to navigate:
* Ctrl+Option+U to get to the quick selector
* Ctrl+Option+Right until it says Landmarks
* Then up and down to select a section of the page or type to filter
* After selecting on Ctrl+Option+U will keep you on landmarks to make it easier to choose one next time

canac:
@dr-bizz Would you be able to look at this again?

dr-bizz:
Will do

dr-bizz (approved):
It's working great now! It's strange, I'm doing the same thing as I was doing, but this time it's working perfectly. The only thing I changed was that I restarted my Mac. I wonder if my VoiceOver was having issues with Chrome.

Great work on this!

rguinee:
In my testing, it describes the aside as a "complementary." Did you experience this @canac or @dr-bizz?

canac:
> In my testing, it describes the aside as a "complementary."

@rguinee `complementary` is the default role for `aside`: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/aside#technical_summary. I believe that `<aside>` has the correct semantics, but let me know if you think a different role would be more appropriate.

rguinee:
@canac I think that's great, no suggested alternatives. And I edited my initial comment to link to something similar from WCAG. I wasn't sure if it would announce it as a "landmark - complementary" but I guess saying "complementary" is obvious enough? Would be a good question for another user with vision impairment (maybe someone other than John since there's already a limitation from computer savviness). I've been wondering if UserTesting.com has people in their panel that we could run some of these tests by.

rguinee (approved):

PR #1296, opened 2025-04-05
[HS-1337456] Fix Notification Count

caleballdrin:
## Description
https://secure.helpscout.net/conversation/2896873270/1337456?viewId=7669074

The notification count is not being updated when a notification is clicked or when 'Mark All as Read' is clicked.

- Remove optimisticResponse checks
- Fix cache changes (CC)

## Checklist:


canac:
Thanks for fixing this! I was digging into why this wasn't working the old way and realized it was related to something I changed with the cache. I pushed a commit that should fix that piece. Can you try reverting your changes so that we're updating the cache manually again and see if it still works for you? Make sure to either test in preview or locally remove ` && process.env.NODE_ENV === 'production'` in `src/lib/apollo/cache.ts` because this bug only appears when we're persisting the cache to local storage.

The manual cache updating code saves some network requests, so that's why I'd like to keep it if possible.

canac (approved):
Looks good and tests great! It would be good to rebase before merging and combine the commits from your first approach with the commits that revert it.

PR #1297, opened 2025-04-08
MPDX-8572 - Removing corepack and adding yarn releases back 

dr-bizz:
## Description
Previously, we removed `.yarn/releases/yarn-{version}.cjs` to rely on Node's corepack for managing the yarn version. However, this has caused several issues for developers configuring their setup, and Node recently announced that `corepack` (which has been experimental) will not be included in future versions of Node as they will no longer support it. 

Given this situation, I have reverted to the `.yarn/releases/yarn-{version}.cjs` file. It does add some size to our repository, but I believe it's worthwhile as it simplifies the setup process for developers and ensures long-term support for future Node versions.

## Checklist:


canac (approved):
Works great! Should we go ahead and use the currently latest yarn version (v4.8.1) while we're at it?

I apologize for the back and forth on this. `corepack` seemed like a great thing before I knew that other devs would have trouble with their environments and before I found out that node would drop support for it. Sorry for recommending it 😬.

dr-bizz:
I've been upgrading all the repos to 4.7.0. I think I'll leave them at 4.7.0 for now until we want to update them all again or there is an update which greatly improves performance and dev experience.

PR #1303, opened 2025-04-30
[no-Jira] Upgrade clsx, formidable, and cross-spawn

canac:
## Description

Upgrade clsx, formidable, and cross-spawn. The Dependabot PRs weren't working for some reason, possibly because we use Git LFS now.

The changes to the upload handlers are because in `formidable` v3, fields and files are always arrays ([changelog](https://github.com/node-formidable/formidable/blob/master/CHANGELOG.md#300)). 

Closes #1200 #1238 #1301

## Checklist:


pages/api/csv-update.page.ts
```diff
@@ -78,9 +78,12 @@ const updateCsvData = async (
       return;
     }
 
-    const {
-      fields: { accountListId, csvFileId, initialData, patchedData, include },
-    } = await parseBody(req);
+    const { fields } = await parseBody(req);
+    const accountListId = fields.accountListId?.[0];
```
dr-bizz, line 82: It's interesting that it has switched to return an array for each item.

dr-bizz (commented):
I've done a first glance at it, but will look again when I have time

pages/api/csv-update.page.ts
canac: I know. It seems that instead of it being either a string/file or an array of strings/files, they decided to make it always be an array.

dr-bizz (approved):
The code looks great! Thank you for doing this. I was able to test the avatar upload. It might be good to get Scott and the team to UAT the CSV, or if you have tested it already, it should be good.

canac:
> It might be good to get Scott and the team to UAT the CSV, or if you have tested it already, it should be good.

Yeah, I tested the CSV upload when I was working on this.

PR #1304, opened 2025-05-02
No-Jira Check to ensure pledgeAmount is not an empty string

caleballdrin:
## Description
Issue:
We get this error when clearing the pledge amount.
Screenshot 2025-04-29 at 10 41 31 AM

Changes:
- Check to ensure pledgeAmount is not an empty string

## Checklist:

### DIFF
--- a/src/components/Contacts/ContactDetails/ContactDonationsTab/PartnershipInfo/EditPartnershipInfoModal/EditPartnershipInfoModal.tsx
+++ b/src/components/Contacts/ContactDetails/ContactDonationsTab/PartnershipInfo/EditPartnershipInfoModal/EditPartnershipInfoModal.tsx
@@ -150,10 +150,10 @@ export const EditPartnershipInfoModal: React.FC<
           pledgeStartDate: attributes.pledgeStartDate?.toISODate() ?? null,
           nextAsk: attributes.nextAsk?.toISODate() ?? null,
           primaryPersonId: attributes.primaryPersonId,
-          pledgeAmount:
-            (attributes.pledgeAmount as any) === ''
-              ? null
-              : Number(attributes.pledgeAmount),
+          // When the field is blank, the value will be an empty string, even though TypeScript says the type is `number | null | undefined`
+          pledgeAmount: !attributes.pledgeAmount
+            ? null
+            : Number(attributes.pledgeAmount),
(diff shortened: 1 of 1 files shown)


src/components/Contacts/ContactDetails/ContactDonationsTab/PartnershipInfo/EditPartnershipInfoModal/EditPartnershipInfoModal.tsx
```diff
@@ -150,6 +150,10 @@ export const EditPartnershipInfoModal: React.FC<
           pledgeStartDate: attributes.pledgeStartDate?.toISODate() ?? null,
           nextAsk: attributes.nextAsk?.toISODate() ?? null,
           primaryPersonId: attributes.primaryPersonId,
+          pledgeAmount:
+            (attributes.pledgeAmount as any) === ''
+              ? null
+              : Number(attributes.pledgeAmount),
```
canac, line 156: What do you think about this to avoid the `any` cast and parsing a number that is already a number?

```suggestion
            typeof attributes.pledgeAmount === 'number'
              ? attributes.pledgeAmount
              // When the field is blank, the value will be an empty string, even though TypeScript says the type is `number | null | undefined`
              : null,
```

canac (approved):
Thanks for the fix! I tested it and it works (although I'm not sure it correctly handles the case where `attributes.pledgeAmount === null`). You don't have to use my suggestion, but can you put in a comment explaining why we have to do this check?

src/components/Contacts/ContactDetails/ContactDonationsTab/PartnershipInfo/EditPartnershipInfoModal/EditPartnershipInfoModal.tsx
dr-bizz: I wonder if you could chnage it to the below as it would catch `empty strings`, `null` and `undefined` which are the types for `attributes.pledgeAmount`
```suggestion
            !attributes.pledgeAmount
              ? null
              : Number(attributes.pledgeAmount),
```

dr-bizz (commented):
Thank you for looking into fixing this

PR #1305, opened 2025-05-02
MPDX-8549 - 14 month report change

wjames111:
## Description

Web - 14 month reports change "Total" to "12 Month Total" and data needs to be from the middle 12 months, not include current incomplete month. Jira ticket available [here](https://jira.cru.org/browse/MPDX-8549)

## Checklist:

### DIFF
--- a/src/lib/dateRangeHelpers.test.ts
+++ b/src/lib/dateRangeHelpers.test.ts
@@ -1,0 +1,19 @@
+import { Settings } from 'luxon';
+import { getTwelveMonthReportDateRange } from './dateRangeHelpers';
+
+describe('getTwelveMonthReportDateRange', () => {
+  it('should return correct dates on the first day of the month', () => {
+    Settings.now = () => new Date(2025, 5, 1).valueOf();
+    expect(getTwelveMonthReportDateRange()).toBe('2024-06-01...2025-05-31');
+  });
+
+  it('should return correct dates in the middle of the month', () => {
+    Settings.now = () => new Date(2025, 5, 15).valueOf();
+    expect(getTwelveMonthReportDateRange()).toBe('2024-06-01...2025-05-31');
+  });
+
+  it('should return correct dates at the end of the month', () => {
+    Settings.now = () => new Date(2025, 5, 30).valueOf();
+    expect(getTwelveMonthReportDateRange()).toBe('2024-06-01...2025-05-31');
+  });
+});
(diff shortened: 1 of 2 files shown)


src/lib/dateRangeHelpers.test.ts
```diff
@@ -0,0 +1,24 @@
+import { DateTime } from 'luxon';
+import { getTwelveMonthReportDateRange } from './dateRangeHelpers';
+
+describe('getTwelveMonthReportDateRange', () => {
+  it('should return startDate and endDate as ISO strings', () => {
+    const { startDate, endDate } = getTwelveMonthReportDateRange();
```
canac, line 6: Can you add tests that check the actual start and end date values? You can do `Settings.now = () => new Date(2024, 7, 31).valueOf();` to mock the current date. I'd like to see at least three tests: one for the first day of the month (e.g. May 1), a middle day of the month (e.g. May 15), and the last day of the month (e.g. May 31). That will help ensure that our logic works for any day.

src/lib/dateRangeHelpers.ts
```diff
@@ -0,0 +1,17 @@
...
+  endDate: string;
+} => {
+  const startDate = DateTime.now()
+    .startOf('month')
+    .minus({ months: 12 })
+    .toISODate();
+  const endDate = DateTime.now()
+    .startOf('month')
+    .minus({ days: 1 })
+    .endOf('day')
```
canac, line 14: Do you think that `.minus({ months: 1 }).endOf('month')` would be clearer? Since what we want is the end of the previous month.

src/lib/dateRangeHelpers.ts
```diff
@@ -0,0 +1,17 @@
...
+    .toISODate();
+  return { startDate, endDate };
```
canac, line 16: Personally, I would have this return the formatted string, but it's not required.

```suggestion
  return `${startDate}...${endDate}`;
```

canac (commented):
Thanks for making the helper function for calculating the date range!

canac:
@wjames111 Can you update the type name and comment in `cache.ts`?

https://github.com/CruGlobal/mpdx-react/blob/18eb5963a36f73b67129fd80977bec920c7da82b/src/lib/apollo/cache.ts#L69-L71

canac (approved):
This looks great! Thanks for the hard work! Just rebase and this is good to merge! 🚀

PR #1307, opened 2025-05-02
MPDX-8524 - Change the order of the newsletter options in the dropdown

wjames111:
## Description

I think it would be more intuitive if None were at the top and Both were at the bottom instead of None being in the middle of the options dropdown. Jira ticket available [here](https://jira.cru.org/browse/MPDX-8524)

## Checklist:

### DIFF
--- a/src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/EditMailingInfoModal/EditMailingInfoModal.test.tsx
+++ b/src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/EditMailingInfoModal/EditMailingInfoModal.test.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { ThemeProvider } from '@mui/material/styles';
-import { render, waitFor } from '@testing-library/react';
+import { render, waitFor, within } from '@testing-library/react';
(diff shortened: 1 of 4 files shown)


src/components/Contacts/ContactDetails/ContactDetailsTab/Mailing/EditMailingInfoModal/EditMailingInfoModal.tsx
```diff
@@ -162,7 +162,15 @@ export const EditMailingInfoModal: React.FC<EditMailingInfoModalProps> = ({
                         )
                       }
                     >
-                      {Object.values(SendNewsletterEnum).map((value) => (
+                      {[
```
canac, line 165: Since I don't expect this enum to change often if ever, I think it would be simpler and clearer to just provide an array with the desired order and map over that array.

canac: All good! What you had would work great but is probably overkill in this particular case.

canac (approved):
Great work on this! I think this makes the dropdown more intuitive to users and makes it easy to find the option they're looking for.

PR #1309, opened 2025-05-08
Mpdx 8463 fix phone numbers validation

wjames111:
## Description

On the fix email addresses page, find a contact with an editable existing phone number. Select the entire number and type "A" to replace the phone number with "A". Note that there is no validation error. Type another "A" and you will see the validation error. It appears that the validation is running on the previous value of the input instead of the current value.

Jira ticket available [here](https://jira.cru.org/browse/MPDX-8463)

**Things to note**
The initial "out of sync" error was caused by the onSubmit function being triggered on every onChange event. This approach was intended to keep the parent component—specifically the one with the "Confirm Individual" button—updated with the latest form state. However, it led to several issues. For one, updates were sometimes being made based on stale state, meaning the "out of sync" error was just a symptom of a deeper problem. Additionally, because submission was decoupled from validation, the "Confirm Individual" action wasn't respecting Formik's validation logic—allowing submissions even when fields like the phone number contained errors.

After discussing this with @canac, he proposed a cleaner solution: introducing a submitAll flag in the parent component, and using a useEffect in the child form component to trigger a submission when that flag is set. Alongside this, we lifted Formik's state up to the parent level. This approach ensures that Formik’s error handling works as expected—blocking submissions when errors are present and reflecting the latest validation messages on the fields. It also guarantees that the state used in mutation calls is current.

This solution works well overall, but there is one notable drawback. When performing a bulk submit, each individual form submits separately. While this effectively prevents invalid data—like errored phone numbers—from being sent, it results in a separate mutation call for each form, which could be inefficient.

## Checklist:

### DIFF
--- a/src/components/Tool/FixPhoneNumbers/Contact.tsx
+++ b/src/components/Tool/FixPhoneNumbers/Contact.tsx
@@ -1,5 +1,5 @@
 import NextLink from 'next/link';
-import React, { ReactElement, useEffect, useState } from 'react';
+import React, { ReactElement, useState } from 'react';
 import styled from '@emotion/styled';
 import { mdiCheckboxMarkedCircle } from '@mdi/js';
 import { Icon } from '@mdi/react';
@@ -15,7 +15,7 @@ import {
   Link,
   Typography,
 } from '@mui/material';
-import { Formik, useFormikContext } from 'formik';
+import { Formik } from 'formik';
 import { useSnackbar } from 'notistack';
 import { useTranslation } from 'react-i18next';
 import { makeStyles } from 'tss-react/mui';
@@ -26,23 +26,12 @@ import useGetAppSettings from 'src/hooks/useGetAppSettings';
 import theme from '../../../theme';
 import { ContactPhoneNumbers } from './ContactPhoneNumbers';
 import { PersonInvalidNumberFragment } from './GetInvalidPhoneNumbers.generated';
+import { ImperativeSubmit } from './ImperativeSubmit';
 import PhoneValidationForm from './PhoneNumberValidationForm';
 import {
   useUpdateInvalidPhoneNumbersMutation,
   useUpdatePhoneNumberMutation,
 } from './UpdateInvalidPhoneNumbers.generated';
-
-const ImperativeSubmit: React.FC<{ submitAll: boolean }> = ({ submitAll }) => {
-  const { submitForm } = useFormikContext();
-
-  useEffect(() => {
-    if (submitAll) {
-      submitForm();
-    }
-  }, [submitAll]);
-
-  return null;
-};
 
 const useStyles = makeStyles()(() => ({
   left: {},
--- a/src/components/Tool/FixPhoneNumbers/ImperativeSubmit.test.tsx
+++ b/src/components/Tool/FixPhoneNumbers/ImperativeSubmit.test.tsx
@@ -1,0 +1,19 @@
+import React from 'react';
+import { Formik } from 'formik';
+import { render, waitFor } from '__tests__/util/testingLibraryReactMock';
+import { ImperativeSubmit } from './ImperativeSubmit';
+
+describe('Fix PhoneNumber Contact', () => {
+  it('default', async () => {
+    const submit = jest.fn();
+    render(
+      <Formik initialValues={{}} onSubmit={submit}>
+        <ImperativeSubmit submitAll={true} />
+      </Formik>,
+    );
+
+    await waitFor(() => {
+      expect(submit).toHaveBeenCalledTimes(1);
+    });
+  });
+});
--- a/src/components/Tool/FixPhoneNumbers/ImperativeSubmit.tsx
+++ b/src/components/Tool/FixPhoneNumbers/ImperativeSubmit.tsx
@@ -1,0 +1,18 @@
+import React, { useEffect } from 'react';
+import { useFormikContext } from 'formik';
+
+interface Props {
+  submitAll: boolean;
+}
+
+export const ImperativeSubmit: React.FC<Props> = ({ submitAll }) => {
+  const { submitForm } = useFormikContext();
+
+  useEffect(() => {
+    if (submitAll) {
+      submitForm();
+    }
+  }, [submitAll]);
+
+  return null;
+};


dr-bizz:
Is this ready for review? It's marked as Draft

.vscode/settings.json
```diff
@@ -19,5 +19,8 @@
   "prettier.prettierPath": ".yarn/sdks/prettier/index.js",
   "typescript.enablePromptUseWorkspaceTsdk": true,
   "jest.jestCommandLine": "yarn jest",
-  "editor.bracketPairColorization.enabled": true
+  "editor.bracketPairColorization.enabled": true,
+  "files.associations": {
+    "*.yaml": "home-assistant"
+  }
```
dr-bizz, line 25: Could you explain this change?

src/components/Tool/FixPhoneNumbers/Contact.tsx
```diff
@@ -143,6 +133,7 @@ interface NumberToDelete {
 }
 
 interface Props {
+  submitAll: boolean;
   person: PersonInvalidNumberFragment;
   handleChange: (
     personId: string,
```
dr-bizz, line 139: You can remove `handleChange` from the props

src/components/Tool/FixPhoneNumbers/Contact.tsx
```diff
@@ -295,7 +311,7 @@ const Contact: React.FC<Props> = ({
                     action={
                       <Button
                         data-testid={`confirmButton-${person.id}`}
-                        onClick={() => handleSingleConfirm(person, numbers)}
+                        onClick={() => handleSubmit()}
```
dr-bizz, line 314: If I'm reading this correctly, this oes the same thing as `handleSingleConfirm(person, numbers)`

src/components/Tool/FixPhoneNumbers/Contact.test.tsx
```diff
@@ -321,16 +324,28 @@ describe('Fix PhoneNumber Contact', () => {
...
+    const { getByTestId, getByRole } = render(<TestComponent />);
+    const textInput = getByTestId('textfield-contactTestId-number2');
+    userEvent.clear(textInput);
+    userEvent.type(textInput, 'p');
+
+    userEvent.click(getByRole('button', { name: 'Confirm' }));
+    await waitFor(() => {
+      expect(handleSingleConfirm).not.toHaveBeenCalled();
+    });
+  });
```
dr-bizz, line 350: It might also be worth writing a test to make sure we don't have stale data. But don't waste too much time on it

src/components/Tool/FixPhoneNumbers/ContactPhoneNumbers.tsx
```diff
@@ -0,0 +1,240 @@
...
+import { useTranslation } from 'react-i18next';
+import { makeStyles } from 'tss-react/mui';
+import { useLocale } from 'src/hooks/useLocale';
+import { dateFormatShort } from 'src/lib/intlFormat';
+import { isEditableSource, sourceToStr } from 'src/utils/sourceHelper';
+import theme from '../../../theme';
+import { PhoneNumber } from './Contact';
+import { PersonInvalidNumberFragment } from './GetInvalidPhoneNumbers.generated';
+
+const useStyles = makeStyles()((theme: Theme) => ({
```
dr-bizz, line 29: `theme ` is already defined.
```suggestion
const useStyles = makeStyles()(() => ({
```

src/components/Tool/FixPhoneNumbers/ContactPhoneNumbers.tsx
```diff
@@ -0,0 +1,240 @@
...
+    person: PersonInvalidNumberFragment,
+    numbers: PhoneNumber[],
+  ) => void;
+  setFieldValue: (field: string, value: string) => void;
+  errors: any;
+}
+
+export const ContactPhoneNumbers: React.FC<Props> = ({
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  key,
```
dr-bizz, line 76: This variable is not used

dr-bizz: you don't need to create a parameter for it. When adding key React knows what it is.

src/components/Tool/FixPhoneNumbers/ContactPhoneNumbers.tsx
```diff
@@ -0,0 +1,240 @@
...
+                <TextField
+                  {...field}
+                  style={{ width: '100%' }}
+                  size="small"
+                  inputProps={{
+                    'data-testid': `textfield-${personId}-${phoneNumber?.id}`,
+                  }}
+                  onChange={(e) => {
+                    setFieldValue(`numbers.${index}.number`, e.target.value);
+                  }}
```
dr-bizz, line 183: you could use the `handleChange` formIk function

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.test.tsx
```diff
@@ -212,9 +212,12 @@ describe('FixPhoneNumbers-Home', () => {
...
     await waitFor(() => {
-      expect(mockEnqueue).toHaveBeenCalledWith(`Phone numbers updated!`, {
-        variant: 'success',
-      });
+      expect(mockEnqueue).toHaveBeenCalledWith(
+        `Successfully updated phone numbers for Simba Lion`,
+        {
+          variant: 'success',
+        },
+      );
```
dr-bizz, line 220: Maybe here is a good place to ensure stale data is not stored and sent to the server.

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -336,6 +281,7 @@ const FixPhoneNumbers: React.FC<Props> = ({ accountListId }: Props) => {
                   (person: PersonInvalidNumberFragment) => (
                     <Contact
                       key={person.id}
+                      submitAll={submitAll}
                       person={person}
                       handleChange={handleChange}
```
dr-bizz, line 286: Delete this line as we no longer need the `handleChange`

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -164,38 +140,7 @@ const FixPhoneNumbers: React.FC<Props> = ({ accountListId }: Props) => {
   };
```
dr-bizz, line 140: I know this is the wrong line, but you can delete the `handleChange` function as it's no longer needed

dr-bizz: This will also help with the CodeScene score.

dr-bizz (changes requested):
Great work on this! Thank you for splitting up the component into two files, making it easier to read and maintain.

I left some comments, nothing crazy.

Not being able to submit all of the phone numbers at once is a bummer. But maybe we will find a way to speed it up if it becomes an issue.

dr-bizz:
I was not able to test this properly on the UI. Can we get some more phone numbers on the test account to fix? you might need to ask Justin or Mark how to manually make a phone number appear in the fix phone numbers tools page

dr-bizz:
Is this ready for another review or are you still working on things?

dr-bizz:
Works great. And what is even greater is that GraphQL grouped the bulk requests as one request by default.

.vscode/settings.json
```diff
@@ -18,6 +18,5 @@
   "eslint.nodePath": ".yarn/sdks",
   "prettier.prettierPath": ".yarn/sdks/prettier/index.js",
   "typescript.enablePromptUseWorkspaceTsdk": true,
-  "jest.jestCommandLine": "yarn jest",
-  "editor.bracketPairColorization.enabled": true
+  "jest.jestCommandLine": "yarn jest"
```
dr-bizz, line 21: Can we keep `editor.bracketPairColorization.enabled` on the file
```suggestion
  "jest.jestCommandLine": "yarn jest",
  "editor.bracketPairColorization.enabled": true,
```

src/components/Tool/FixPhoneNumbers/Contact.tsx
dr-bizz: Sorry, I don't think I was clear. I wasn't requesting for it to change, just confirming that it mostly does the same thing.

src/components/Tool/FixPhoneNumbers/Contact.test.tsx
```diff
@@ -321,16 +322,62 @@ describe('Fix PhoneNumber Contact', () => {
...
+
+  it('should not submit stale data', async () => {
```
dr-bizz, line 350: Do we know if this test would have failed on the previous logic?

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.test.tsx
dr-bizz: Tests written in contact.test.tsx

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -24,11 +24,9 @@ import { ToolsGridContainer } from '../styledComponents';
 import Contact, { PhoneNumber, PhoneNumberData } from './Contact';
 import {
   PersonInvalidNumberFragment,
-  PersonPhoneNumberFragment,
   useGetInvalidPhoneNumbersQuery,
 } from './GetInvalidPhoneNumbers.generated';
 import { useUpdateInvalidPhoneNumbersMutation } from './UpdateInvalidPhoneNumbers.generated';
-import { determineBulkDataToSend } from './helper';
```
dr-bizz, line 31: Since we no longer need `determineBulkDataToSend`, we can go ahead and delete the `./helper.ts` file.

This made me wonder, since we pass down a state value which triggers the submit call. What happens if you have 100 fixPhoneNumbers, will this fire all 100? Will all 100 fire if they are not all rendered on the page?

dr-bizz (approved):

dr-bizz (changes requested):
Awesome! thank you for sharing the video recording with me! That works great. Before you merge, lets make sure the test are passing

src/components/Tool/FixPhoneNumbers/Contact.tsx
```diff
@@ -195,17 +179,43 @@ const Contact: React.FC<Props> = ({
...
+            .test(
+              'is-phone-number',
+              t('This field is not a valid phone number'),
+              (val) => typeof val === 'string' && /\d/.test(val),
+            )
+            .required(t('This field is required')),
+        }),
       )
-      .required(t('This field is required')),
+      .required('Must have at least 1 Phone Number to confirm'),
```
canac, line 197: This label needs to be localized.

src/components/Tool/FixPhoneNumbers/Contact.tsx
```diff
@@ -266,7 +276,7 @@ const Contact: React.FC<Props> = ({
   };
 
   return (
-    <>
+    <FormikProvider value={formik}>
```
canac, line 279: Although this works, I think it would be better to use a normal `<Formik>` component instead of `<FormikProvider>` directly.

src/components/Tool/FixPhoneNumbers/Contact.tsx
```diff
@@ -346,212 +357,26 @@ const Contact: React.FC<Props> = ({
...
-                                    </Box>
-                                  </Box>
-                                )}
-                              </Box>
-                            </Grid>
-                          </>
-                        )}
-                      </Formik>
-                    ))}
+                    <FieldArray
```
canac, line 360: Can you help me understand why this is a `<FieldArray>` since we're not using `arrayHelpers`? I'm wondering if it could just be a regular `values.numbers.map(...)`.

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -91,26 +74,19 @@ export interface ModalState {
...
 
-export interface FormValues {
-  people: FormValuesPerson[];
-}
-
 interface Props {
   accountListId: string;
 }
 
 const FixPhoneNumbers: React.FC<Props> = ({ accountListId }: Props) => {
```
canac, line 86: How hard would it be to move all of the mutation logic inside of `Contact` now? The parent doesn't need to know how to submit changes anymore since `Contact` always decides when to submit changes now.

I don't think we need the logic that copies data state either.

src/components/Tool/FixPhoneNumbers/ContactPhoneNumbers.tsx
```diff
@@ -0,0 +1,234 @@
...
+      </Grid>
+      <Grid item xs={12} sm={6} className={classes.paddingB2}>
+        <Box
+          display="flex"
+          justifyContent="flex-start"
+          className={clsx(classes.responsiveBorder, classes.paddingX)}
+        >
+          <FormControl fullWidth>
+            <Field name={`numbers.${index}.number`}>
+              {({ field }: any) => (
```
canac, line 169: I believe the correct type is `FieldProps`.

```suggestion
              {({ field }: FieldProps) => (
```

canac (commented):
I haven't look at the tests yet, just the main component code. This looks really good and was exactly what I had in mind! I'm thinking that if we move all the mutation logic inside of `Contact`, then `FixPhoneNumbers` can be simplified significantly. Let me know if I missed something and that won't work for some reason.

src/components/Tool/FixPhoneNumbers/ContactPhoneNumbers.tsx
```diff
@@ -0,0 +1,231 @@
...
+        </Box>
+      </Grid>
+      <Grid item xs={6} sm={2} className={classes.paddingB2}>
+        <Box
+          display="flex"
+          justifyContent="center"
+          className={classes.paddingX}
+        >
+          <Typography display="flex" alignItems="center">
+            {phoneNumber?.primary ? (
```
dr-bizz, line 120: Same here

src/components/Tool/FixPhoneNumbers/ContactPhoneNumbers.tsx
```diff
@@ -0,0 +1,231 @@
...
+                    fontWeight="bold"
+                  >
+                    {t('Source')}:
+                  </Typography>
+                </Hidden>
+                <Tooltip title={t('Set as Primary')} placement="left">
+                  <StarOutlineIcon
+                    data-testid={`starOutlineIcon-${personId}-${phoneNumber?.id}`}
+                    className={classes.hoverHighlight}
+                    onClick={() => handleChangePrimary(phoneNumber?.id)}
```
dr-bizz, line 151: Same here as well. It looks like all the properties on phoneNumber have optional chaining.

dr-bizz (commented):
The tests look okay to me. I'll let Caleb take over reviewing this PR, as he had the most recent comments.

src/components/Tool/FixPhoneNumbers/Contact.tsx
canac: I think we have two options:

1. Most most of the logic and JSX in `<Contact>` into a `<ContactForm>` component. Then make `Contact` nest `<ContactForm>` inside of `<Formik>` and put the submit `useEffect` in `ContactForm`.
2. Create an `ImperativeSubmit` component something and stick it inside the `<Formik>` element. It doesn't render anything and all it does is submit the form it is inside of when its `submitAll` prop becomes true.

```tsx
const ImperativeSubmit: React.FC<{ submitAll: boolean }> = ({ submitAll }) => {
  const { submitForm } = useFormikContext();

  useEffect(() => {
    if (submitAll) {
      submitForm();
    }
  }, [submitAll]);

  return null;
};
```

I'm trying to think which approach would be the easiest to maintain and the easiest for new devs to understand, and right now I lean slightly towards the second approach. But what do you think? Does it feel intuitive to you what `ImperativeSubmit` is doing?

src/components/Tool/FixPhoneNumbers/Contact.tsx
```diff
@@ -1,46 +1,50 @@
...
+import { ContactPhoneNumbers } from './ContactPhoneNumbers';
 import { PersonInvalidNumberFragment } from './GetInvalidPhoneNumbers.generated';
 import PhoneValidationForm from './PhoneNumberValidationForm';
-import { useUpdatePhoneNumberMutation } from './UpdateInvalidPhoneNumbers.generated';
+import {
+  useUpdateInvalidPhoneNumbersMutation,
+  useUpdatePhoneNumberMutation,
+} from './UpdateInvalidPhoneNumbers.generated';
+
+const ImperativeSubmit: React.FC<{ submitAll: boolean }> = ({ submitAll }) => {
```
canac, line 35: Can you add a props interface like we do for most of our components? Sorry, when I gave you that code snippet, I was just writing out the MVP.

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -131,137 +99,22 @@ const FixPhoneNumbers: React.FC<Props> = ({ accountListId }: Props) => {
     const existingSources = new Set<string>();
```
canac, line 99: Can you remove the comment `// Create a mutable copy of the query data and store in the state` since it is not longer accurate?

This `useState`/`useEffect` can also just be a `useMemo`.

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.tsx
```diff
@@ -131,137 +99,22 @@ const FixPhoneNumbers: React.FC<Props> = ({ accountListId }: Props) => {
...
-    personId: string,
-    numberIndex: number,
-    newNumber: string,
-  ): void => {
-    const temp = { ...dataState };
-    dataState[personId].phoneNumbers[numberIndex].number = newNumber;
-    setDataState(temp);
-  };
-
   const handleBulkConfirm = async () => {
```
canac, line 116: This handler doesn't need to be `async` anymore.

```suggestion
  const handleBulkConfirm = () => {
```

src/components/Tool/FixPhoneNumbers/Contact.tsx
canac: I would probably also put it its own file and write at least one test for it.

In your test, make the component have a `<Formik>` form that contains a `<ImperativeSubmit submitForm>` (for test simplicity submit the form immediately). You can pass a `jest.fn()` to `onSubmit` and test that it is called.

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.test.tsx
```diff
@@ -212,13 +213,61 @@ describe('FixPhoneNumbers-Home', () => {
...
+
+    const { getByText, getAllByTestId, queryByTestId, getByTestId } = render(
+      <Components data={largeTestData} count={51} />,
+    );
+    await waitFor(() => {
+      expect(queryByTestId('loading')).not.toBeInTheDocument();
+    });
+
+    expect(
+      await getByText('You have 51 phone numbers to confirm.'),
```
canac, line 253: Only `findByText` needs `await`.

```suggestion
      getByText('You have 51 phone numbers to confirm.'),
```

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.test.tsx
```diff
@@ -212,13 +213,61 @@ describe('FixPhoneNumbers-Home', () => {
...
+    ).toBeInTheDocument();
+
+    const phoneNumberElements = await waitFor(() =>
+      getAllByTestId('phoneNumbers'),
+    );
+    await waitFor(() => {
+      expect(phoneNumberElements).toHaveLength(51);
+    });
```
canac, line 261: Instead of waiting twice, try this:


```suggestion
    await waitFor(() =>
      expect(getAllByTestId('phoneNumbers')).toHaveLength(51)
    );
```

src/components/Tool/FixPhoneNumbers/FixPhoneNumbers.test.tsx
```diff
@@ -212,13 +213,61 @@ describe('FixPhoneNumbers-Home', () => {
...
+          variant: 'success',
+        },
+      );
     });
 
     expect(
       getByText('No people with phone numbers need attention'),
     ).toBeVisible();
   });
+  it('should handle 100+ phone numbers gracefully', async () => {
```
canac, line 228:
```suggestion
  it('should handle 50+ phone numbers gracefully', async () => {
```

canac (commented):
Sorry for the delay, Will. Looks like that `ImperativeSubmit` is working great and let us simplify `FixPhoneNumbers` like I was hoping. I think this just needs a little cleanup and some `ImperativeSubmit` tests, and this should be good to go! Let me know if it's not clear how I was suggesting to test `ImperativeSubmit`.

canac:
@wjames111 Great work with keeping this PR moving and fixing this tricky bug! 🎉

PR #1312, opened 2025-05-12
MPDX-8374 - Helpducks filtering

wrandall22:
## Description

[Jira](https://jira.cru.org/browse/MPDX-8374)

This PR creates a proxy that filters out help content to just MPDx. Two new environment variables are being added to do the filtering, but if they are not set, it defaults to the existing functionality.

## Checklist:


wrandall22:
I tried to build out some tests for the new proxy API, but it is complaining about ECMAScript not being available in that context. I don't know why `/pages/api/helpducks` should act differently than `/pages/api/utils` in this regard. If I could get some assistance getting over that hurdle, I can build out better tests.

dr-bizz:
@canac is out this week

wrandall22:
I will also note that I did test this in the app itself manually, and it all seems to work there.

pages/api/helpjuice/api/questions/popular.page.ts
```diff
@@ -0,0 +1,21 @@
+import { NextApiRequest, NextApiResponse } from 'next';
+import fetch from 'node-fetch';
+
+const proxyPopularQuestions = async (
+  req: NextApiRequest,
+  res: NextApiResponse,
+): Promise<void> => {
+  try {
+    const helpjuiceUrl = process.env.HELPJUICE_CATEGORY_ID
```
dr-bizz, line 9: is helpjuice 1 word or 2? Should it be written like `helpJuice`

pages/api/helpjuice/search.page.ts
```diff
@@ -0,0 +1,23 @@
...
+const proxySearchParent = async (
+  req: NextApiRequest,
+  res: NextApiResponse,
+): Promise<void> => {
+  try {
+    const searchTerm = req.query.query;
+
+    const helpjuiceUrl = process.env.HELPJUICE_PARENT_CATEGORY
+      ? `${process.env.HELPJUICE_ORIGIN}/search?query=${searchTerm}&parent_category=${process.env.HELPJUICE_PARENT_CATEGORY}`
+      : `${process.env.HELPJUICE_ORIGIN}/search?query=${searchTerm}`;
```
dr-bizz, line 13:
```suggestion
    const parentCategory = process.env.HELPJUICE_PARENT_CATEGORY;
    const helpjuiceUrl = `${process.env.HELPJUICE_ORIGIN}/search?query=${searchTerm}${
      parentCategory ? `&parent_category=${parentCategory}` : ''
    }`;
```

pages/api/helpjuice/search.page.ts
```diff
@@ -0,0 +1,23 @@
+import { NextApiRequest, NextApiResponse } from 'next';
+import fetch from 'node-fetch';
+
+const proxySearchParent = async (
+  req: NextApiRequest,
+  res: NextApiResponse,
+): Promise<void> => {
+  try {
+    const searchTerm = req.query.query;
```
dr-bizz, line 9: Not required, but you could also write it like this
```suggestion
    const { query: searchTerm } = req.query;
```

pages/api/helpjuice/api/questions/popular.page.ts
```diff
@@ -0,0 +1,21 @@
...
+      ? `${process.env.HELPJUICE_ORIGIN}/api/questions/popular?category_id=${process.env.HELPJUICE_CATEGORY_ID}`
+      : `${process.env.HELPJUICE_ORIGIN}/api/questions/popular`;
```
dr-bizz, line 11: Same suggestion here as search.page.ts to remove duplicated code

pages/api/helpjuice/[...path].page.ts
```diff
@@ -0,0 +1,48 @@
...
+    return originalPath;
+  }
+  let pathArray;
+  if (typeof originalPath === 'string') {
+    pathArray = originalPath.split(',');
+  } else {
+    pathArray = originalPath;
+  }
+  return pathArray.join('/');
+};
```
Comment (dr-bizz) [inline pages/api/helpjuice/[...path].page.ts:15]: You might be able to simplify this to the below.

- Using the optional variable means you don't need to specify undefined.
- Then we can simplify the bottom half

```
export const buildUrlPath = (originalPath?: string | string[]) => {
  if (!originalPath) {
    return originalPath;
  }
  return Array.isArray(originalPath)
    ? originalPath.join('/')
    : originalPath.split(',').join('/');
};
```

pages/api/helpjuice/[...path].page.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const proxyCatchAll = async (
+  req: NextApiRequest,
+  res: NextApiResponse,
+): Promise<void> => {
+  try {
+    const allQueryParams = req.query;
+    const path = req.query.path;
+    delete allQueryParams.path;
```
Comment (dr-bizz) [inline pages/api/helpjuice/[...path].page.ts:24]: Instead of modifying data you can use this technic to get the path and get all the other properties but the path.

```suggestion
    const { path, ...allQueryParams } = req.query;
```

pages/api/helpjuice/[...path].page.ts
```diff
@@ -0,0 +1,48 @@
...
+    );
+
+    Object.entries(allQueryParams).forEach(([key, value]) => {
+      let queryParamValue;
+      if (Array.isArray(value)) {
+        queryParamValue = value.join(',');
+      } else {
+        queryParamValue = value;
+      }
+      helpjuiceUrl.searchParams.append(key, queryParamValue ?? '');
```
Comment (dr-bizz) [inline pages/api/helpjuice/[...path].page.ts:37]:
```suggestion
      const queryParamValue = Array.isArray(value) ? value.join(',') : value ?? '';
      helpjuiceUrl.searchParams.append(key, queryParamValue);
```

dr-bizz (changes requested):
Great work on this! I haven't yet tested the UI, but the code is looking good. I've added some recommendations on how to improve the code.

dr-bizz:
Once this is ready to review for the UI. Can you add the preview label on it?

pages/api/helpjuice/api/questions/popular.page.ts
wrandall22: I've only seen it as one word.

next.config.ts
```diff
@@ -73,6 +73,8 @@ const config: NextConfig = {
     ).toString(),
     HELPJUICE_ORIGIN: process.env.HELPJUICE_ORIGIN,
     HELPJUICE_KNOWLEDGE_BASE_URL: process.env.HELPJUICE_KNOWLEDGE_BASE_URL,
+    HELPJUICE_PARENT_CATEGORY: process.env.HELPJUICE_PARENT_CATEGORY,
+    HELPJUICE_CATEGORY_ID: process.env.HELPJUICE_CATEGORY_ID,
```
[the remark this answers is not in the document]
wrandall22: They will be defaults for our org, but not for non-Cru orgs. I suppose we put defaults in for the API urls which are in the same boat.

dr-bizz:
> I tried to build out some tests for the new proxy API, but it is complaining about ECMAScript not being available in that context. I don't know why `/pages/api/helpducks` should act differently than `/pages/api/utils` in this regard. If I could get some assistance getting over that hurdle, I can build out better tests.

Have you looked at what we've done here `__tests__/pages/api/stopImpersonating.test.ts`?

wrandall22:
Getting the following (unrelated to my PR) error on husky:

```
src/components/Tool/Appeal/Modals/DeleteAppealContact/DeleteAppealContactModal.tsx:97:11 - error TS2353: Object literal may only specify known properties, and 'id' does not exist in type 'AppealContactDeleteMutationInput'.

97           id: appealContactId,
```

Maybe the same one mentioned in another PR?

wrandall22:
Sorry for the force push, but was able to get around the TS issue.

pages/api/helpjuice/api/questions/popular.page.ts
dr-bizz: I've only seen it as one word, but it looks weird. Not something we need to amend in this PR.

pages/api/helpjuice/search.test.ts
```diff
@@ -0,0 +1,83 @@
...
+      res as unknown as NextApiResponse,
+    );
+
+    expect(res._getStatusCode()).toEqual(200);
+    expect(fetch).toHaveBeenCalledWith(
+      `${process.env.HELPJUICE_ORIGIN}/search?query=${searchTerm}`,
+    );
+  });
+
+  it('should return search results', async () => {
```
dr-bizz, line 56: What if there is no search query? I assume there always will be, so it might not matter.

dr-bizz (approved):
Looks great.

pages/api/helpjuice/search.test.ts
wrandall22: In my testing, when there is no search in the UI, it will go to the popular questions instead of here.

PR #1314, opened 2025-05-14
MPDX 8522 address auto complete override

wjames111:
## Description

Address auto-complete doesn't work for some addresses that contain Appt #s - I can't enter unmatched addresses such as like these:
545 N Holliston #1
Pasadena, CA 91106
or
2213 S Conway Rd Apt 1322
Orlando, FL 32812

Best solution we could come up with was to allow users to override autocomplete and enter their own addresses. Which can be achieved by simply removing autoSelect from AutoComplete component. More work may be needed on this depending on how we want to address this issue.

Jira ticket available [here](https://jira.cru.org/browse/MPDX-8522#add-comment)

## Checklist:


caleballdrin (approved):
Great! It was already possible to override the address. But it was just difficult with the autoselect. Let's see what Scott thinks about this solution.

PR #1315, opened 2025-05-16
MPDX-8388 - Suggested task tags show up in tags autocomplete

wjames111:
## Description

Suggested tags seem to save on the server as lowercase. That means that when I edit a task with suggested tags, the suggested tags are correctly highlighted, but they also show up in the "Additional Tags" dropdown.

Jira ticket available [here](https://jira.cru.org/browse/MPDX-8388)

**Things to Note**
So I think the easiest solution to this problem would be to filter out the selected task tags from autocomplete. However this still felt a little awkward. In addition to this I wanted to make the suggested tags and additional tags a little more in sync with each other so when one is selected it's reflected in the other, and the same with removing. The only caveat is when a suggested tag is selected I did not want it to create an entry in additional tags, so they aren't entirely in sync with each other, but more so then before.

## Checklist:

### DIFF
--- a/src/components/Task/Modal/Form/TaskModalForm.test.tsx
+++ b/src/components/Task/Modal/Form/TaskModalForm.test.tsx
@@ -199,6 +199,66 @@ describe('TaskModalForm', () => {
     });
   }, 10000);
 
+  it('should toggle suggested tags and additional tags', async () => {
+    const { findByRole, getByRole, queryByRole } = render(
+      <LocalizationProvider dateAdapter={AdapterLuxon}>
+        <SnackbarProvider>
+          <GqlMockedProvider<{
+            AssigneeOptions: AssigneeOptionsQuery;
+            ContactOptions: ContactOptionsQuery;
+            TagOptions: TagOptionsQuery;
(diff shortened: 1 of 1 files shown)


src/components/Task/Modal/Form/TaskModalForm.test.tsx
```diff
@@ -199,6 +199,66 @@ describe('TaskModalForm', () => {
...
+    userEvent.click(getByRole('combobox', { name: 'Tags' }));
+    userEvent.click(await findByRole('option', { name: 'tag-1' }));
+
+    expect(getByRole('button', { name: 'tag-1' })).toBeInTheDocument();
+
+    userEvent.click(getByRole('combobox', { name: 'Tags' }));
+    await waitFor(() =>
+      expect(queryByRole('option', { name: 'tag-1' })).not.toBeInTheDocument(),
+    );
+  }, 10000);
```
dr-bizz, line 260: Do you need to have this time limit?

dr-bizz (approved):
This is looking great! Awesome job on this.

PR #1316, opened 2025-05-20
[MPDX-8569] Dashed donations goal line

canac:
## Description

Make the donations goal line dashed for users without health indicator data.

MPDX-8569

## Testing

* Log into an account without health indicator dat
* Go to the dashboard
* Verify that the goal line in the donations graph is dashed (previously it was only dashed in the legend)

## Checklist:

### DIFF
--- a/src/components/Dashboard/DonationHistories/DonationHistories.tsx
+++ b/src/components/Dashboard/DonationHistories/DonationHistories.tsx
@@ -116,6 +116,12 @@ const DonationHistories = ({
   const goalColor = palette.graphite.main;
   const averageColor = palette.graphite.main;
   const pledgedColor = palette.yellow.main;
+  const goalLineStyles = {
+    stroke: goalColor,
+    strokeDasharray: '5,8',
+    strokeLinecap: 'round' as const,
+    strokeWidth: 3,
+  };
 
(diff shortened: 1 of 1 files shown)


src/components/Dashboard/DonationHistories/DonationHistories.tsx
```diff
@@ -116,6 +116,12 @@ const DonationHistories = ({
   const goalColor = palette.graphite.main;
   const averageColor = palette.graphite.main;
   const pledgedColor = palette.yellow.main;
+  const goalLineStyles = {
+    stroke: goalColor,
+    strokeDasharray: '5,8',
+    strokeLinecap: 'round' as const,
```
dr-bizz, line 122: Do we need the type casting?

dr-bizz (approved):
Awesome! I only had one comment

src/components/Dashboard/DonationHistories/DonationHistories.tsx
canac: The type for `strokeLinecap` is `"butt" | "round" | "square" | "inherit"`, so we have to use `as const` so that TypeScript interprets `'round'` as a literal string and not just `string`.

PR #1321, opened 2025-05-21
MPDX-8388 - suggested tags autocomplete update

wjames111:
## Description

Reverts code from initial suggested tags autocomplete branch, filters out suggested tags from additional tags.
Jira ticket available [here](https://jira.cru.org/browse/MPDX-8388)

## Checklist:

### DIFF
--- a/src/hooks/usePhaseData.ts
+++ b/src/hooks/usePhaseData.ts
@@ -80,48 +80,6 @@ export const usePhaseData = (phaseEnum?: PhaseEnum | null): GetPhaseData => {
     return allTags;
   }, [constants]);
 
-  // phaseData = {
-  //   contactStatuses: ['CALL_FOR_DECISION'],
-  //   id: 'FOLLOW_UP',
-  //   name: 'Follow Up',
-  //   tasks: [
-  //     'FOLLOW_UP_PHONE_CALL',
-  //     'FOLLOW_UP_EMAIL',
-  //     'FOLLOW_UP_TEXT_MESSAGE',
-  //     'FOLLOW_UP_SOCIAL_MEDIA',
-  //     'FOLLOW_UP_IN_PERSON',
-  //   ],
-  //   results: {
-  //     resultOptions: [
-  //       {
-  //         name: 'INITIATION_RESULT_NO_RESPONSE',
-  //         suggestedContactStatus: null,
-  //         suggestedNextActions: [
-  //           'FOLLOW_UP_PHONE_CALL',
-  //           'FOLLOW_UP_EMAIL',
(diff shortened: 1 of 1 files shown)


src/components/Task/Modal/Form/Complete/TaskModalCompleteForm.tsx
```diff
@@ -397,6 +397,7 @@ const TaskModalCompleteForm = ({
               <Grid item>
                 <TagsAutocomplete
                   accountListId={accountListId}
+                  phaseTags={phaseTags ?? []}
```
caleballdrin, line 400: This looks great! I noticed that it is only filtering out the phase tags/suggested tags of the current phase that it is in. So other suggested tags from other phases are not being filtered out. I think you'll need to map over the phases object to get all the suggested tags from each phase.

caleballdrin (changes requested):
This is great and almost done!

src/components/Task/Modal/Form/Complete/TaskModalCompleteForm.tsx
dr-bizz: Good catch

dr-bizz:
@wjames111 As a general rule of thumb, we should only ask one senior engineer to review a PR. This is to save senior engineers time from reviewing the same thing.

There are times when we might want multiple senior engineers to review code, especially when it's a large PR with numerous changes where more eyes are beneficial. However, I don't believe that is the case with this PR.

src/hooks/usePhaseData.ts
```diff
@@ -67,6 +68,18 @@ export const usePhaseData = (phaseEnum?: PhaseEnum | null): GetPhaseData => {
     [constants],
   );
 
+  const allPhaseTags = useMemo(() => {
+    const allTags: Set<string> = new Set();
+    constants?.phases?.forEach((phase) => {
```
dr-bizz, line 73: Where are you getting the value `constants` from?

You will need to either pass it in or do soemthing like this:
```
const constants = useApiConstants();
```

dr-bizz (changes requested):
Good work on this. I have left a few comments.

Sorry, I started reviewing this on Friday, but got called into a bunch of meetings, so having to finish it off today

src/hooks/usePhaseData.ts
dr-bizz: It should be to prevent it from using an old version of `constants`

dr-bizz (approved):
Awesome work on this! Clean up the commit history and this will be good to UAT or merge into production.

PR #1323, opened 2025-06-06
HS-1380937 Fix Decimal Amount

caleballdrin:
## Description
https://secure.helpscout.net/conversation/2962259146/1380937?viewId=7669074

Formatting the decimal in this PR introduced a bug when formating large numbers without decimals.

The issue was that it was checking the last dot . or comma , and choosing that as the decimal separater. But that doesn't work if the number entered is 1,234

The fix is to use the locale to determine what the decimal separater should be. 

## Checklist:

### DIFF
--- a/src/lib/intlFormat.ts
+++ b/src/lib/intlFormat.ts
@@ -56,28 +56,30 @@ export const amountFormat = (
 
 export const normalizeCurrencyString = (
   input: string | null | undefined,
+  locale: string = 'en-US',
 ): number | null => {
   if (!input) {
     return null;
   }
 
-  const sanitized = input.trim().replace(/\s/g, '');
+  // Use Intl.NumberFormat to determine group and decimal separators for the locale
+  const parts = new Intl.NumberFormat(locale).formatToParts(1234567.89);
+  const group = parts.find((p) => p.type === 'group')?.value || ',';
+  const decimal = parts.find((p) => p.type === 'decimal')?.value || '.';
 
-  // If input contains both . and , assume the last one is the decimal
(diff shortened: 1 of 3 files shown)


src/lib/intlFormat.ts
```diff
@@ -56,28 +56,30 @@ export const amountFormat = (
...
 ): number | null => {
   if (!input) {
     return null;
   }
 
-  const sanitized = input.trim().replace(/\s/g, '');
+  // Use Intl.NumberFormat to determine group and decimal separators for the locale
+  const parts = new Intl.NumberFormat(locale).formatToParts(1234567.89);
+  const group = parts.find((p) => p.type === 'group')?.value || ',';
+  const decimal = parts.find((p) => p.type === 'decimal')?.value || '.';
```
dr-bizz, line 68: Let's make `p` into `part` so it more readable when scanning

dr-bizz: Also, lets add brackets to these functions

dr-bizz (commented):
Could you use something simular to this for for the amount?

```
      style: 'decimal',
      minimumFractionDigits: 2,
      maximumFractionDigits: 2,
      trailingZeroDisplay: 'stripIfInteger',
    })
    ```

dr-bizz (changes requested):
I'm getting an error on it now.
If I type in `30.500,00` it changes it to `30.50` when it should be `30,500.00`.

Even when I switched the locale and language to Spanish-latin

dr-bizz (approved):
Sorry, it was tester error. I was testing spanish-latin, instead of Portuguese.

Great work on this! This fixes the bug.

PR #1326, opened 2025-06-11
MPDX-8620 - Only Show Fill Out Weekly Report button show for those with reports

wjames111:
## Description

FILL OUT WEEKLY REPORT is only available if the account is shared with a coach.  Make this available even if there is no coach.   I (Scott Paros) also believe this report is only set up properly for US staff, but the report link to the questions is available for global staff as well.  We need to hide this link for global staff.

This PR really only addresses the second part of this request as "Weekly Reports" don't seem to be dependent on being shared with a coach. At least as far as I can tell.

Jira ticket available [here](https://jira.cru.org/browse/MPDX-8620)

## Checklist:

### DIFF
--- a/src/components/Dashboard/ThisWeek/WeeklyActivity/WeeklyActivity.test.tsx
+++ b/src/components/Dashboard/ThisWeek/WeeklyActivity/WeeklyActivity.test.tsx
@@ -9,7 +9,16 @@ import {
   GetWeeklyActivityQueryDefaultMocks,
   GetWeeklyActivityQueryLoadingMocks,
 } from './WeeklyActivity.mock';
+import {
+  EmptyCoachingAnswerSetMock,
+  PopulateCoachingAnswerSetMock,
+} from './WeeklyReportModal/WeeklyReport.mock';
 import WeeklyActivity from '.';
+
+// Mock the useOrganizationId hook
+jest.mock('src/hooks/useOrganizationId', () => ({
+  useOrganizationId: () => 'org-123',
+}));
 
 const accountListId = 'abc';
 
@@ -121,7 +130,10 @@ describe('WeeklyActivity', () => {
       <SnackbarProvider>
         <ThemeProvider theme={theme}>
           <MockedProvider
-            mocks={GetWeeklyActivityQueryDefaultMocks()}
+            mocks={[
+              ...GetWeeklyActivityQueryDefaultMocks(),
+              PopulateCoachingAnswerSetMock(),
+            ]}
             addTypename={false}
           >
             <WeeklyActivity accountListId={accountListId} />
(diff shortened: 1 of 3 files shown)


src/components/Dashboard/ThisWeek/WeeklyActivity/WeeklyActivity.test.tsx
```diff
@@ -135,4 +144,32 @@ describe('WeeklyActivity', () => {
...
+          </MockedProvider>
+        </ThemeProvider>
+      </SnackbarProvider>,
+    );
+
+    expect(
+      await findByRole('link', { name: 'View Activity Detail' }),
+    ).toBeInTheDocument();
+
+    await waitFor(() => {
```
canac, line 169: Why is this test passing when we give it the same mocked data as the other test where the button is showing?

src/components/Dashboard/ThisWeek/WeeklyActivity/WeeklyReportModal/WeeklyReport.mock.ts
```diff
@@ -0,0 +1,71 @@
...
+const emptyCoachingAnswerSet: CurrentCoachingAnswerSetQuery = {
+  currentCoachingAnswerSet: {
+    questions: [],
+    answers: [],
+    id: 'answer-set-123',
+    completedAt: null,
+  },
+};
+
+export const CurrentCoachingAnswerSetDefaultMocks = (): MockedResponse[] => {
```
canac, line 44: Thanks for figuring out the weirdness with the mocks! Since a test will never want to use both of these mocks, can you split them into two separate mock exports and give them descriptive names like `populateCoachingAnswerSetMock` and `emptyCoachingAnswerSetMock`? You can make each of them a mock object - no need for them to be an array.

src/components/Dashboard/ThisWeek/WeeklyActivity/WeeklyActivity.tsx
```diff
@@ -252,7 +266,12 @@ const WeeklyActivity = ({ accountListId }: Props): ReactElement => {
             </Table>
           </TableContainer>
         </CardContent>
-        <CardActions sx={{ justifyContent: 'space-between' }}>
+        <CardActions
+          sx={{
+            justifyContent:
+              questions.length > 0 ? 'space-between' : 'flex-start',
```
canac, line 272: Do you think this should be `flex-end` so that the button is on the right like the other dashboard cards?

```suggestion
              questions.length > 0 ? 'space-between' : 'flex-end',
```

canac (commented):
Thanks for fixing the tests! This is almost ready!

src/components/Dashboard/ThisWeek/WeeklyActivity/WeeklyActivity.test.tsx
canac: Thanks for figuring that out!

src/components/Dashboard/ThisWeek/WeeklyActivity/WeeklyReportModal/WeeklyReport.mock.ts
```diff
@@ -41,31 +41,32 @@ const emptyCoachingAnswerSet: CurrentCoachingAnswerSetQuery = {
...
-      request: {
-        query: CurrentCoachingAnswerSetDocument,
-        variables: {
-          accountListId: 'abc',
-          organizationId: 'org-123',
-        },
-      },
-      result: {
-        data: populatedCoachingAnswerSet,
+export const PopulateCoachingAnswerSetMock = (): MockedResponse => {
```
canac, line 44:
```suggestion
export const PopulatedCoachingAnswerSetMock = (): MockedResponse => {
```

canac:
This looks great! Just rebase and you should be good to go!

PR #1331, opened 2025-06-18
MPDX-8593 Confirm mass task deletion

kegrimes:
## Description

In this PR, I updated the mass task deletion preventing users from accidentally deleting all tasks. We got multiple reports from HelpScout highlighting this issue, so I have added a checkbox for users to select when they wish to delete 5 or more tasks. To make the modal more apparent, I have also added an alert box stating the severity of the delete action. 

## Checklist:

### DIFF
--- a/src/components/Task/MassActions/ConfirmationModal/MassActionsTasksConfirmationModal.tsx
+++ b/src/components/Task/MassActions/ConfirmationModal/MassActionsTasksConfirmationModal.tsx
@@ -1,5 +1,12 @@
-import React from 'react';
-import { DialogActions, DialogContent, Typography } from '@mui/material';
+import React, { useState } from 'react';
+import {
+  Alert,
+  Checkbox,
+  DialogActions,
+  DialogContent,
+  FormControlLabel,
+  Typography,
+} from '@mui/material';
 import { useTranslation } from 'react-i18next';
 import {
   CancelButton,
@@ -20,6 +27,51 @@ export const MassActionsTasksConfirmationModal: React.FC<
 > = ({ open, idsCount, action, setOpen, onConfirm }) => {
   const { t } = useTranslation();
 
+  // KG: useState for checkbox
+  const [checked, setChecked] = useState(false);
+
+  const handleChange = (event) => {
+    setChecked(event.target.checked);
+  };
+
+  // KG: get correct task message based on action
+  function renderMessage() {
+    if (action === 'complete' || (action === 'delete' && idsCount < 5)) {
+      return (
+        <Typography>
+          {t(
+            'Are you sure you wish to {{action}} the {{count}} selected tasks?',
+            { action, count: idsCount },
+          )}
+        </Typography>
+      );
+    } else {
+      return (
+        <Typography>
+          <Alert severity="error">
+            {t(
+              'Deleting these tasks is permanent and cannot be undone. Please make sure you want to permanently delete them before proceeding.',
+            )}
+          </Alert>
+          <FormControlLabel
+            control={
+              <Checkbox
+                //color="error"
+                checked={checked}
+                onChange={handleChange}
+                inputProps={{ 'aria-label': 'controlled' }}
+              />
+            }
+            label={t(
+              'Yes, I want to {{action}} the {{count}} selected tasks.',
+              { action, count: idsCount },
+            )}
+          />
+        </Typography>
+      );
+    }
+  }
+
   return (
     <Modal
       title={action === 'complete' ? t('Complete Tasks') : t('Delete Tasks')}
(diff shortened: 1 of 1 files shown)


src/components/Task/MassActions/ConfirmationModal/MassActionsTasksConfirmationModal.tsx
```diff
@@ -20,23 +27,71 @@ export const MassActionsTasksConfirmationModal: React.FC<
 > = ({ open, idsCount, action, setOpen, onConfirm }) => {
   const { t } = useTranslation();
 
+  // KG: useState for checkbox
```
dr-bizz, line 30:
```suggestion
```

src/components/Task/MassActions/ConfirmationModal/MassActionsTasksConfirmationModal.tsx
```diff
@@ -20,23 +27,71 @@ export const MassActionsTasksConfirmationModal: React.FC<
...
+  const [checked, setChecked] = useState(false);
```
dr-bizz, line 31: Can we make this variable more descriptive.

Having a name like `hasConfirmedDeletion` shows that it should have a boolean value by reading it, it also gives the reader an understanding of what it's used for.

```suggestion
  const [hasConfirmedDeletion, setHasConfirmedDeletion] = useState(false);
```

src/components/Task/MassActions/ConfirmationModal/MassActionsTasksConfirmationModal.tsx
```diff
@@ -20,23 +27,71 @@ export const MassActionsTasksConfirmationModal: React.FC<
...
+
+  const shouldConfirmDeletion = action === 'delete' && idsCount > 5;
```
dr-bizz, line 33: Every time the component rerenders, it is recalculating this variable. In this case, it's a pretty simple equation, but if we wanted to Optimize this component, we could wrap it in a useMemo React hook.
https://react.dev/reference/react/useMemo

```suggestion
  const shouldConfirmDeletion = useMemo(() => (
    action === 'delete' && idsCount > 5
  ), [action, idsCount]);
```

src/components/Task/MassActions/ConfirmationModal/MassActionsTasksConfirmationModal.tsx
```diff
@@ -20,23 +27,71 @@ export const MassActionsTasksConfirmationModal: React.FC<
...
+
+  const handleChange = (event) => {
+    setChecked(event.target.checked);
```
dr-bizz, line 36: You could also write it as `(!checked)` since the new value will always be the opposite of the current one.

```suggestion
    setChecked(!checked)
```

src/components/Task/MassActions/ConfirmationModal/MassActionsTasksConfirmationModal.tsx
```diff
@@ -20,23 +27,71 @@ export const MassActionsTasksConfirmationModal: React.FC<
...
+    if (action === 'complete' || !shouldConfirmDeletion) {
+      return (
+        <Typography>
+          {t(
+            'Are you sure you wish to {{action}} the {{count}} selected tasks?',
+            { action, count: idsCount },
+          )}
+        </Typography>
+      );
+    } else {
```
dr-bizz, line 50: I think it would be best to move this render function into the JSX, to make it more readable.

We only use a render function like this in rare cases.

src/components/Task/MassActions/ConfirmationModal/MassActionsTasksConfirmationModal.tsx
```diff
@@ -20,23 +27,71 @@ export const MassActionsTasksConfirmationModal: React.FC<
...
+      return (
+        <Typography>
+          <Alert severity="error">
+            {t(
+              'Deleting these tasks is permanent and cannot be undone. Please make sure you want to permanently delete them before proceeding.',
+            )}
```
dr-bizz, line 56: Same for this. I would move this JSX into the main JSX return for readability.

src/components/Task/MassActions/ConfirmationModal/MassActionsTasksConfirmationModal.tsx
```diff
@@ -20,23 +27,71 @@ export const MassActionsTasksConfirmationModal: React.FC<
...
+          </Alert>
+          <FormControlLabel
+            control={
+              <Checkbox
+                //color="error"
```
dr-bizz, line 61: We can remove this line

src/components/Task/MassActions/ConfirmationModal/MassActionsTasksConfirmationModal.tsx
```diff
@@ -20,23 +27,71 @@ export const MassActionsTasksConfirmationModal: React.FC<
...
+
+  const shouldConfirmDeletion = action === 'delete' && idsCount > 5;
+
+  const handleChange = (event) => {
+    setChecked(event.target.checked);
+  };
+
+  // KG: get correct task message based on action
+  function renderMessage() {
+    if (action === 'complete' || !shouldConfirmDeletion) {
```
dr-bizz, line 41: Can we turn the `action` property into a enum. That way it becomes easier to use in if statements.

Research enum in TS here:  https://www.typescriptlang.org/docs/handbook/enums.html

dr-bizz (changes requested):
Awesome work on this! The code is lookomg good, I just have a few comments on how to improve it.

Again, great work.

src/components/Task/MassActions/ConfirmationModal/MassActionsTasksConfirmationModal.tsx
dr-bizz: The way we write our code, we generally only write comments when needed. Instead, we make variable and function names very descriptive to allow the user to read the code without needing comments.

However, there are cases were comments are needed

PR #1333, opened 2025-06-20
Mpdx 8642 tags autocomplete

wjames111:
## Description

## Checklist:


canac:
You're right that this kind of ends up being a wrapper component. My apologies for not filtering the list in that Google Doc more. I think it might be best to put a pin in this and focus on the other autocompletes. If we do use this PR, I would try to put the options that are used by most of the tag autocompletes (like `filterSelectedOptions`, `popupIcon`, and `textFieldPlaceholder`, for example) in the common component. That way the components that use the tag autocomplete don't have to pass in as many options. But still your point stands that no logic is going into the common tags component, and we're probably better off focusing on other ones, like currency or contact status.

PR #1335, opened 2025-06-24
MPDX-7917 Fixed missing alert message when uploading the same large image

kegrimes:
## Description

Initially, HelpScout reported wanting to upgrade the UI to show an alert message when uploading images larger than 1MB. It was discovered that this ticket was already resolved, however, there was a minor issue missed. When attempting to upload the same file twice, the alert message would not appear. I found out handling a file change would only be triggered if the file input was changed. To fix this, I reset the input to ensure uploading the same file could be attempted more than once. 

## Checklist:

### DIFF
--- a/src/components/Contacts/ContactDetails/ContactDetailsTab/People/Items/PersonModal/PersonName/PersonName.tsx
+++ b/src/components/Contacts/ContactDetails/ContactDetailsTab/People/Items/PersonModal/PersonName/PersonName.tsx
@@ -71,6 +71,9 @@ export const PersonName: React.FC<PersonNameProps> = ({
     if (personId && file) {
       setAvatar(file);
     }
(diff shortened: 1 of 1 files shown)


dr-bizz (changes requested):
This is great! Could we add a test on `src/components/Contacts/ContactDetails/ContactDetailsTab/People/Items/PersonModal/PersonModal.test.tsx` to ensure the `event.target.value` is reset to `''`?

dr-bizz (approved):
Really excellent work on this!

When reviewing, I realised that we don't need the test we created since we mock the `validateAvatar` function and we are checking that it errors with that message. This is the same as the test `should notify the user about validation errors`.

Sorry for sending you on a goose chase. I would remove your edits to the test file and deploy.

PR #1339, opened 2025-06-27
[HS-1394249] Fix contacts count in appeal contacts list

canac:
## Description

A user found that if you click an appeal from the appeals list, then try to select all contacts, the selected count would be higher than the number of contacts in the list. This is because `allExcludedContacts` loaded once because `contactsFilters.appealStatus` wasn't set, then stopped loading because `contactsFilters.appealStatus` changed to `active`, causing that query to be skipped. However, `allContactIds` was still including those excluded contacts because `allExcludedContactsPrevious` contained the excluded contacts from before the query was skipped.

The fix is to ignore excluded contacts if the query is skipped (i.e. if we are in the list view). Excluded contacts should only be added to `allContactIds` in the flows view.

[HelpScout ticket](https://secure.helpscout.net/conversation/2982182920/1394249?viewId=7296729)

## Testing

* Go to the appeals page
* Go to an appeal containing excluded contacts
* Go to the list view
* Click the select all checkbox
* Verify that "X selected" number matches the "Showing X" number and that excluded contacts are not counted as being selected

## Checklist:

### DIFF
--- a/src/components/Tool/Appeal/AppealsContext/AppealsContext.tsx
+++ b/src/components/Tool/Appeal/AppealsContext/AppealsContext.tsx
@@ -186,6 +186,7 @@ export const AppealsProvider: React.FC<AppealsContextProps> = ({
   // In the flows view, we need the total count of all contacts in every column, but the API
   // filters out contacts excluded from an appeal. We have to load excluded contacts also and
   // manually merge them with the other contacts.
+  const inFlowsView = viewMode === TableViewModeEnum.Flows;
   const {
     data: allExcludedContacts,
     previousData: allExcludedContactsPrevious,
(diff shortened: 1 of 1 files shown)


src/components/Tool/Appeal/AppealsContext/AppealsContext.tsx
```diff
@@ -186,6 +186,7 @@ export const AppealsProvider: React.FC<AppealsContextProps> = ({
   // In the flows view, we need the total count of all contacts in every column, but the API
   // filters out contacts excluded from an appeal. We have to load excluded contacts also and
   // manually merge them with the other contacts.
+  const inFlowsView = !contactsFilters.appealStatus;
```
dr-bizz, line 189: Coudl you not check the viewMode against `TableViewModeEnum.Flows`?

dr-bizz (approved):
Awesome work on this!

src/components/Tool/Appeal/AppealsContext/AppealsContext.tsx
canac: I was just blindly copying what was there, but that's a great suggestion!

PR #1341, opened 2025-06-30
MPDX-8657 - Created Likely to Give Select shared component

zweatshirt:
## Description

- Corresponds to Jira ticket MPDX-8657 for OKR 41.
- Created a shared component called LikelyToGiveSelect in `src/common/Selects/LikelyToGiveSelect.tsx`
 
Refactored these files to use the new shared component: 
1. src/components/Contacts/ContactDetails/ContactDonationsTab/PartnershipInfo/EditPartnershipInfoModal/EditPartnershipInfoModal.tsx 
2. src/components/Contacts/MassActions/EditFields/MassActionsEditFieldsModal.tsx

## Checklist:

### DIFF
--- a/src/components/common/LikelyToGiveSelect/LikelyToGiveSelect.test.tsx
+++ b/src/components/common/LikelyToGiveSelect/LikelyToGiveSelect.test.tsx
@@ -1,0 +1,27 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { LikelyToGiveEnum } from 'src/graphql/types.generated';
+import { LikelyToGiveSelect } from './LikelyToGiveSelect';
+
+describe('LikelyToGiveSelect', () => {
+  it('should render a Select with all LikelyToGiveEnum options', async () => {
+    const { getByRole, getByText } = render(
+      <LikelyToGiveSelect value="" onChange={() => {}} />,
+    );
+    await userEvent.click(getByRole('combobox'));
+    expect(getByText('Least Likely')).toBeInTheDocument();
+  });
+
+  it('should pass props to Select and show selected value', () => {
+    const { getByTestId, getByText } = render(
+      <LikelyToGiveSelect
+        value={LikelyToGiveEnum.LeastLikely}
+        onChange={jest.fn()}
+        data-testid="likely-to-give-select"
+      />,
+    );
+    expect(getByTestId('likely-to-give-select')).toBeInTheDocument();
+    expect(getByText('Least Likely')).toBeInTheDocument();
+  });
+});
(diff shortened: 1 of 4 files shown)


zweatshirt:
I was unsure to what extent I should have generalized this component. It seems that the entire ContactInputWrapper surrounding this LikelyToGiveSelect component can be made into a general component. I am wondering as well if the name of the component is inappropriate, since I took the entire FormControl.

Another thing I noticed is that the NewsletterSelect is nearly identical in functionality. Would it make sense to create a ContactInputWrapper component that can accept functionality for both?

src/components/common/LikelyToGiveSelect/LikelyToGiveSelect.test.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { LikelyToGiveEnum } from 'src/graphql/types.generated';
+import { LikelyToGiveSelect } from './LikelyToGiveSelect';
+
+jest.mock('src/utils/functions/getLocalizedLikelyToGive', () => ({
```
canac, line 7: Usually, we don't mock simple utilities like this. Can you get it to pass without mocking?

src/components/common/LikelyToGiveSelect/LikelyToGiveSelect.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+}));
+
+describe('LikelyToGiveSelect', () => {
+  it('should render a Select with all LikelyToGiveEnum options', async () => {
+    render(<LikelyToGiveSelect value="" onChange={() => {}} />);
+
+    const select = screen.getByRole('combobox');
+    await userEvent.click(select);
+
+    Object.values(LikelyToGiveEnum).forEach((val) => {
```
canac, line 18: Usually, we try to put `expect`s in the main body of the test, instead of in an `if` bock or a loop. Instead of this being a loop, can you just test that one or two of the options are in the document?

src/components/common/LikelyToGiveSelect/LikelyToGiveSelect.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+jest.mock('src/utils/functions/getLocalizedLikelyToGive', () => ({
+  getLocalizedLikelyToGive: (_t: any, val: string) => `${val}`,
+}));
+
+describe('LikelyToGiveSelect', () => {
+  it('should render a Select with all LikelyToGiveEnum options', async () => {
+    render(<LikelyToGiveSelect value="" onChange={() => {}} />);
+
+    const select = screen.getByRole('combobox');
+    await userEvent.click(select);
```
canac, line 16: In most of our tests, we pull the `getBy...` functions from the `render` return value.

Also, I usually don't assign elements that it found to a variable (like `select`) unless it's used in multiple places to keep the test a little shorter.

```suggestion
    const { getByRole } = render(<LikelyToGiveSelect value="" onChange={() => {}} />);

    await userEvent.click(getByRole('combobox'));
```

canac (changes requested):
This looks really great, Zach! Thanks for the time you put into it! I just left some style comments about our unwritten norms for writing tests.

PR #1342, opened 2025-06-30
MPDX-8656 Created a shared component for PledgeFrequencyEnum

kegrimes:
## Description

This PR refactors duplicate logic in `EditPartnershipInfoModal.tsx` and `Contact.tsx`, both of which use `PledgeFrequencyEnum` in similar `<Select>` components. To reduce redundancy and improve reusability, I have created a shared `PledgeFrequencySelect` component. It combines the `<Select>` and `<MenuItem>` structure into a single file. 

## Checklist:

### DIFF
--- a/src/common/Selects/PledgeFrequencySelect.test.tsx
+++ b/src/common/Selects/PledgeFrequencySelect.test.tsx
@@ -1,0 +1,33 @@
+import { ThemeProvider } from '@mui/material/styles';
+import { AdapterLuxon } from '@mui/x-date-pickers/AdapterLuxon';
+import { LocalizationProvider } from '@mui/x-date-pickers/LocalizationProvider';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import theme from '../../theme';
+import { PledgeFrequencySelect } from './PledgeFrequencySelect';
+
+const mockOnChange = jest.fn();
+const label = '';
+const value = '';
+
+describe('PledgeFrequencySelect', () => {
+  it('should render the select', () => {
+    const { getByRole } = render(
+      <LocalizationProvider dateAdapter={AdapterLuxon}>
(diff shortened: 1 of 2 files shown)


src/common/Selects/PledgeFrequencySelect.tsx
```diff
@@ -0,0 +1,26 @@
+import { MenuItem, MenuItemProps, Select, SelectProps } from '@mui/material';
+import { PledgeFrequencyEnum } from 'src/graphql/types.generated';
+
+type PledgeFrequencySelectProps = Partial<SelectProps> & {
+  MenuItemProps?: Partial<MenuItemProps> & { [testid: string]: string };
```
canac, line 5: The component sets `data-testid="pledgeFrequencyOptions"`, but I don't see any tests using that test id, so it should be safe to remove. That will let us remove this `MenuItemProps` prop.

src/common/Selects/PledgeFrequencySelect.tsx
```diff
@@ -0,0 +1,26 @@
...
+  getFunctionLabel: (label: string) => React.ReactNode;
```
canac, line 6: Because we don't need to customize the label, we can use `getLocalizedPledgeFrequency` directly instead of passing in `getFunctionLabel` as a prop.

canac (changes requested):
Great work on this! Would you be also able to add some a test for the new `PledgeFrequencySelect` component? Since it's so simple I think that one test checking that the select renders and that you can select an option would be sufficient.

kegrimes:
Thank you Caleb! I kept one intermediate variable because it was used in two separate lines. If you would still prefer the test to contain no intermediate variables, I can remove it easily!

canac (approved):
Great work with this one! Feel free to merge it to production! 🚀

PR #1344, opened 2025-07-02
MPDX-8549 - Update 12 Month Totals.

wjames111:
## Description

Update for ticket [MPDX-8549](https://jira.cru.org/browse/MPDX-8549)

 "I was expecting this to continue showing 14 months but only totaling the middle 12 months.  Now I am seeing only 12 months (July to June).   Today is 7/1 therefore maybe July donations haven't arrived yet. 

The important part is we need the total to be 12 complete months.  If a donor gives $100 /month , then I want to always see a total of $1200.  Never $1100, $1300, $1400. "

This PR fixes the total to show only the last 12 complete months. I left the report at 12 months instead of changing it back to 14 as it sounds like this part was optional. Additionally it feels a little counter intuitive to show a table of 14 months and have a column that only shows the 12 month total.

## Checklist:

### DIFF
--- a/pages/api/Schema/reports/twelveMonth/datahandler.ts
+++ b/pages/api/Schema/reports/twelveMonth/datahandler.ts
@@ -112,7 +112,11 @@ export const mapTwelveMonthReport = (
             return {
               id: contactDonationInfo.contact_id,
               name: contact?.contact_name ?? '',
-              total: Number(contactDonationInfo.total),
+              total: contactDonationInfo.months.reduce(
+                (accumulator, currentValue) =>
+                  accumulator + Number(currentValue.total),
+                0,
+              ),
               average: Number(contactDonationInfo.average),
               minimum: Number(contactDonationInfo.minimum),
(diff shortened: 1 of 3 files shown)


pages/api/Schema/reports/twelveMonth/datahandler.ts
```diff
@@ -112,7 +112,11 @@ export const mapTwelveMonthReport = (
             return {
               id: contactDonationInfo.contact_id,
               name: contact?.contact_name ?? '',
-              total: Number(contactDonationInfo.total),
+              total: contactDonationInfo.months.reduce(
```
canac, line 115: Can you walk me through why this needed changed? If the report contained 12 months of $100, was the total not $1200?

canac (commented):
I could be wrong, but I got the impression that Scott wanted the report to go back to having 14 months. But we can wait for him to respond in the ticket. This PR should also probably be approved by Scott before merging since there's been a lot of confusion with the requirements.

PR #1348, opened 2025-07-03
MPDX 8663 Fix dashboard weekly report completed task count

kegrimes:
## Description

It was brought to our attention in Helpscout that the **Dashboard's weekly report** was displaying the incorrect number of completed tasks. Since this same calculation is used in the **Coaching Report's level of effort**, I compared both and found that the coaching report was correctly showing the total counts. During the investigation, I also noticed that the Follow Up Letter/Card field was missing from the level of effort calculation, which caused a slight discrepancy. 

After the backend team updated the query to include the missing Follow Up field, I added it to the coaching report. Before fixing the dashboard totals, I tested each task field and found that **six fields** were missing from the dashboard count. I then updated the `GetWeeklyActivity.graphql` file to include those missing fields.

## Checklist:


canac:
@kegrimes FYI, you apparently need the dash in the ticket number for Jira to notice it and link it to the ticket. i.e. "MPDX-8663" instead of just "MPDX 8663".

PR #1352, opened 2025-07-03
[MPDX-8667] Eliminate race conditions in appeals filter state

canac:
## Description

In the appeals list, storing the appeal status filter in the URL was a major source of race conditions and the filters and URL getting out of sync. This PR eliminates those issues by storing the appeal status in state instead of in the URL.

This PR is highly-dependent on work in #1343, but I'm putting it in a separate PR for ease of review since it is fairly self-contained.

MPDX-8667

## Checklist:

### DIFF
--- a/src/components/Tool/Appeal/AppealsContext/AppealsContext.test.tsx
+++ b/src/components/Tool/Appeal/AppealsContext/AppealsContext.test.tsx
@@ -14,16 +14,21 @@ import {
   ListHeaderCheckBoxState,
   TableViewModeEnum,
 } from 'src/components/Shared/Header/ListHeader';
-import { useUrlFilters } from 'src/components/common/UrlFiltersProvider/UrlFiltersProvider';
 import { useMassSelection } from 'src/hooks/useMassSelection';
 import theme from 'src/theme';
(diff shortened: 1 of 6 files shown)


dr-bizz (approved):
Thank you for fixing this. Using React state to keep track of the Appeal status is much easier.

Since this is changing how the old MPDX worked, users could get to a certain status by URL. This could have knock-on effects.
For example, if users bookmarked the Given status page. - this will no longer work.
Please confirm with Scott that he's on board with the new functionality before merging it. - It would solve a lot of issues if we didn't need to use the URL

canac:
> Please confirm with Scott that he's on board with the new functionality before merging it. - It would solve a lot of issues if we didn't need to use the URL

@dr-bizz I reached out to Scott, but he's on vacation until August 4, so I don't expect a response until then. If I remember correctly, the migrating appeals to the new context providers will be buggy without these changes. It would be good to get his stamp of approval, but I know we also don't want to delay finishing this OKR and moving on to other things. My gut says that the improvements to robustness is worth losing the ability to bookmark specific sections of the list. But let me know if you think we should wait for Scott's approval before moving forward.

dr-bizz:
Lets move forward and if he dislikes it, we can revert

PR #1355, opened 2025-07-07
MPDX-8666 Combine appeal autocompletes

kegrimes:
## Description

Both `MassActionsAddToAppealModal` and `AddDonation` use Autocomplete components for selecting appeals. To improve reusability and maintain a clean codebase, I have combined the two into a shared component called `AppealAutocomplete`. 

Previously, each Autocomplete used a different query to retrieve appeal data - one queried by `accountList`, and the other by appeals directly - but both returned very similar data. To reduce redundancy, the shared component now uses `useGetAppealsForMassActionQuery` for both cases. As a result of this change, `AddDonation.test.tsx` needed to be updated since it relied on the old query, which has been adjusted accordingly.

I am unsure if the new component requires a test, as both parent components already verify that the Autocomplete renders and gets the correct selected value.

## Checklist:


canac (commented):
Thanks for addressing the things I brought up before in code review! I believe this can be simplified slightly by removing the `inputProps` prop fro `AppealAutocomplete`.

PR #1356, opened 2025-07-08
MPDX-8670 -  HelpScout issue: overdue donations

zweatshirt:
## Description

[HelpScout ticket](https://secure.helpscout.net/conversation/2989324928/1398590)

- Resolve overdue bug on ContactHeaderStatusSeciton as well as in ContactPartnershipStatus which is used in ContactRow(s)
- Compares pledgeStartDate to lateAt date to determine which value is appropriate to use in deciding if a donor's donation is late.

## Checklist:


src/utils/functions/getDonationLateStatus.ts
canac, line 1: We normally write our functions as arrow functions. It's not better than regular functions, just a style choice in this codebase. Do you mind updating your helpers?

src/utils/functions/getDonationLateStatus.ts
```diff
@@ -0,0 +1,49 @@
...
+  OnTime,
+  LateLessThirty,
+  LateMoreThirty,
+  LateMoreSixty,
+}
+
+function selectLaterDate(lateAt: string, pledgeStartDate: string): string {
+  const lateDate = DateTime.fromISO(lateAt);
+  const pledgeDate = DateTime.fromISO(pledgeStartDate);
+  return lateDate > pledgeDate ? lateAt : pledgeStartDate;
```
canac, line 13: One nice property of ISO dates is that they can be lexicographically compared. So the parsing step can be skipped.

```suggestion
  return lateAt > pledgeStartDate ? lateAt : pledgeStartDate;
```

src/utils/functions/getDonationLateStatus.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+  return ContactLateStatusEnum.LateMoreSixty;
+}
+
+export function getDonationLateStatus(
+  lateAt?: string | null,
+  pledgeStartDate?: string | null,
+): ContactLateStatusEnum | undefined {
+  // Determine which date to use
+  const dateToUse =
```
canac, line 34: Can we give this a slightly more descriptive name? What do you think about `laterDate`?

```suggestion
  const laterDate =
```

src/utils/functions/getDonationLateStatus.test.ts
```diff
@@ -0,0 +1,54 @@
+import { DateTime } from 'luxon';
+import { ContactLateStatusEnum } from '../../components/Contacts/ContactPartnershipStatus/ContactLateStatusLabel/ContactLateStatusLabel';
+import { getDonationLateStatus } from './getDonationLateStatus';
+
+const mockNow = DateTime.fromISO('2025-06-15T12:00:00Z');
+jest.spyOn(DateTime, 'now').mockReturnValue(mockNow as DateTime<true>);
```
canac, line 6: By default, "now" is mocked to January 1, 2020 (it's set up in `__tests__/util/setup.ts`). Can you either update your tests to be relative to that "now" or do the mocking like this:

```ts
Settings.now = () => new Date(2025, 5, 15).valueOf();
```

canac (changes requested):
Thanks for the work on this! I like how your broke the logic up into helpers that are really easy to read and test.

Now that this logic is getting more complex, this looks like a great chance to make more use of `ContactLateStatusLabel`. Can you pass the `lateAt` and `pledgeStartDate` to `ContactLateStatusLabel` as props and have that component be responsible for calculating the late status enum and rendering it?

I'd also love some tests that sets just `lateAt`, sets just `pledgeStartDate`, and one that sets both. If we add tests for `ContactLateStatusLabel`, we can do that in one centralized place.

zweatshirt:
@canac or @wjames111 Would it be okay to suppress the CodeScene Cloud Delta Analysis in instances like this? It is complaining about test descriptions as sort of a false positive. That being said, maybe it is also a sign that I should simply the test file. Thanks for the feedback!

src/components/Contacts/ContactPartnershipStatus/ContactLateStatusLabel/ContactLateStatusLabel.test.tsx
```diff
@@ -1,54 +1,121 @@
...
 import { render } from '@testing-library/react';
+import { Settings } from 'luxon';
 import theme from '../../../../theme';
-import {
-  ContactLateStatusEnum,
-  ContactLateStatusLabel,
-} from './ContactLateStatusLabel';
+import { ContactLateStatusLabel } from './ContactLateStatusLabel';
+
+Settings.now = () => new Date(2020, 1, 1).valueOf();
```
canac, line 8: Sorry, I think I confused you in my other comment. The `setup.ts` file runs for every test, so by default "now" is mocked to Jan 1, 2020. So this line isn't needed unless you wanted a different now timestamp.

src/components/Contacts/ContactPartnershipStatus/ContactLateStatusLabel/ContactLateStatusLabel.test.tsx
```diff
@@ -1,54 +1,121 @@
...
 
 describe('ContactLateStatusLabel', () => {
-  it('should render on time', () => {
+  it('should render on time when both dates are in the future', () => {
     const { getByText } = render(
       <ThemeProvider theme={theme}>
-        <ContactLateStatusLabel lateStatusEnum={ContactLateStatusEnum.OnTime} />
+        <ContactLateStatusLabel
+          lateAt="2020-02-06T12:00:00Z"
+          pledgeStartDate="2020-02-04T12:00:00Z"
```
canac, line 16: The GraphQL type of these fields is `ISO8601Date`, which means that on the server, the dates won't have the time or timezone fields. These strings would match the `ISO8601DateTime` type. Do you mind removing the time information from these and all the others in this file?

```suggestion
          lateAt="2020-02-06"
          pledgeStartDate="2020-02-04"
```

src/components/Contacts/ContactPartnershipStatus/ContactLateStatusLabel/ContactLateStatusLabel.test.tsx
```diff
@@ -1,54 +1,121 @@
...
+        />
       </ThemeProvider>,
     );
 
     expect(getByText('(On time)')).toBeInTheDocument();
   });
 
-  it('should render less than 30 days late', () => {
+  // this case should never happen but it is worth testing
```
canac, line 24: Are you sure this can never happen? If the donations are on time, wouldn't `lateAt` be in the future?

src/components/Contacts/ContactPartnershipStatus/ContactLateStatusLabel/ContactLateStatusLabel.test.tsx
```diff
@@ -1,54 +1,121 @@
...
         <ContactLateStatusLabel
-          lateStatusEnum={ContactLateStatusEnum.LateMoreThirty}
+          lateAt="2019-05-017T12:00:00Z"
+          pledgeStartDate="2019-05-17T12:00:00Z"
         />
       </ThemeProvider>,
     );
-
-    expect(getByText('(30-60 days late)')).toBeInTheDocument();
+    // Should use pledgeStartDate since it's later (46 days ago = LateMoreThirty)
```
canac, line 59: This comment doesn't seem accurate

canac (commented):
This is great! Thanks for moving the logic into ContactPartnershipStatus so that it's super easy for other components to use. Just a few minor tweaks, and this should be ready to go!

zweatshirt:
@canac Sorry to re-request a review. I noticed a small discrepancy when calculating for the enums that was not immediately obvious from looking at the UI. When determining the enums, it is necessary to consider the current date when calculating the total time passed since the last donation.

src/components/Contacts/ContactPartnershipStatus/ContactLateStatusLabel/getDonationLateStatus.ts
```diff
@@ -0,0 +1,85 @@
...
+};
+
+const getStatusFromDays = (daysDiff: number): ContactLateStatusEnum => {
+  if (daysDiff < 0) {
+    return ContactLateStatusEnum.OnTime;
+  }
+  if (daysDiff < 30) {
+    return ContactLateStatusEnum.LateLessThirty;
+  }
+  if (daysDiff < 60) {
```
dr-bizz, line 21: As the last enum us called `LateMoreSixty`, I think this if should read less or equal to 60, then `LateMoreThirty`.


```suggestion
  if (daysDiff <= 60) {
```

src/components/Contacts/ContactPartnershipStatus/ContactLateStatusLabel/getDonationLateStatus.ts
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * lateAt is determined by whether a donation has been received
+ * within the expected time frame (Frequency window).
+ * pledgeStartDate is the date the contact started their pledge.
+ * Donations should only be late if the current date exceeds
+ * pledgeStartDate + the frequency window.
+ * Both are in ISO 8601 format (e.g., "2020-01-01").
+ */
+export const getDonationLateStatus = (
+  lateAt?: string | null | undefined,
```
dr-bizz, line 45: You do't normally need to add `undefined` when you define it as a optional parameter. I understand there are times which it is needed, which this could be one of them?

src/components/Contacts/ContactPartnershipStatus/ContactLateStatusLabel/getDonationLateStatus.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  // If lateAt is provided and is in the future, return OnTime
+  if (lateAt && DateTime.fromISO(lateAt) > DateTime.now()) {
+    return ContactLateStatusEnum.OnTime;
+  }
+
+  const laterDate =
+    lateAt && pledgeStartDate
+      ? selectLaterDate(lateAt, pledgeStartDate)
+      : pledgeStartDate || lateAt;
```
dr-bizz, line 61: We normally prefer ?? over ||, as it's normally a safer operator.
```suggestion
      : pledgeStartDate ?? lateAt;
```

dr-bizz (approved):
This is awesome!! Great work Zach! I left a few comments, but it's looking great.

src/components/Contacts/ContactPartnershipStatus/ContactLateStatusLabel/getDonationLateStatus.ts
zweatshirt: I've removed it here, I don't think it is strictly necessary.

PR #1364, opened 2025-07-10
MPDX-8676 Expose clear filters

kegrimes:
## Description

In this PR, I replaced the `clearSearchTerm()` function with a new `clearFilters()` function in `UrlFiltersProvider.tsx`. This new function resets both the filters and search term stored in context, allowing multiple components to easily clear all search-related values with a single function call. I also updated all references to `clearSearchTerm()` in other files to use `clearFilters()` instead.

## Checklist:

### DIFF
--- a/src/components/Shared/Filters/FilterPanel.tsx
+++ b/src/components/Shared/Filters/FilterPanel.tsx
@@ -134,7 +134,7 @@ export const FilterPanel: React.FC<FilterPanelProps & BoxProps> = ({
   const {
     activeFilters: selectedFilters,
     setActiveFilters,
-    clearSearchTerm,
+    clearFilters,
   } = useUrlFilters();
 
   const updateSelectedFilter = (name: FilterKey, value?: FilterValue) => {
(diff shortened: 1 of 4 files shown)


src/components/Shared/Filters/FilterPanel.tsx
```diff
@@ -559,8 +559,7 @@ export const FilterPanel: React.FC<FilterPanelProps & BoxProps> = ({
   };
 
   const handleClearAllClick = () => {
```
canac, line 561: Could we remove this function entirely and just pass `clearFilters` directly to the link component?

canac (changes requested):
Thanks for working on this Katelyn! I have one small suggestion, but thanks for adding this!

Can you also remove `handleClearAll` from all of the context objects? It shouldn't be needed anymore.

canac:
Thanks Katelyn! Can you rebase this PR on top of the latest commits from `8646-extract-url-filters-context` and fix the update the tests to match the changes you made? I added tests to that branch this morning.

canac (approved):
This looks great! Can you tweak the test name and rebase down to a single commit? I'll let you know when the PRs under this have merged.

src/components/common/UrlFiltersProvider/UrlFiltersProvider.test.tsx
```diff
@@ -129,13 +129,13 @@ describe('useUrlFilters', () => {
...
     const { result } = renderHook(() => useUrlFilters(), {
       wrapper: Wrapper,
     });
 
     act(() => {
-      result.current.clearSearchTerm();
+      result.current.clearFilters();
     });
 
     expect(result.current.searchTerm).toBe('');
```
canac, line 141: Sorry, I missed this before. But we should also test that the filters are reset.

```suggestion
    expect(result.current.activeFilters).toEqual({});
    expect(result.current.searchTerm).toBe('');
```

PR #1383, opened 2025-07-29
MPDX MPD Goal calculator move sidepanel

wjames111:
## Description

I moved sidePanelsLayout directly into GoalCalculator as some significant changes needed to be made to allow for the side icon row, the left step (or anchor) panel and the right information panel to all work correctly. This PR should address these issues. If you would like to see the right panel working you will need to add `rightPanelComponent` to one of the "Steps" and a component that you want rendered there.

Side note, initially this PR aimed to address some of UI changes needed for "Mileage" but it sounds like this will need to be discussed a little more before going in a specific direction.

## Checklist:

### DIFF
--- a/src/components/Reports/GoalCalculator/GoalCalculator.tsx
+++ b/src/components/Reports/GoalCalculator/GoalCalculator.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import CircleIcon from '@mui/icons-material/Circle';
+import InfoIcon from '@mui/icons-material/Info';
 import RadioButtonUncheckedIcon from '@mui/icons-material/RadioButtonUnchecked';
 import {
   Box,
@@ -18,20 +19,29 @@ import { useTranslation } from 'react-i18next';
 import {
   HeaderTypeEnum,
   MultiPageHeader,
+  multiPageHeaderHeight,
 } from 'src/components/Shared/MultiPageLayout/MultiPageHeader';
+import theme from 'src/theme';
 import { useGoalCalculator } from './Shared/GoalCalculatorContext';
 
 const StyledCategoryTitle = styled(Typography)(({ theme }) => ({
   marginBottom: 0,
   marginTop: theme.spacing(1),
   paddingLeft: theme.spacing(2),
+  whiteSpace: 'nowrap',
 }));
+
+const StyledTitle = styled(Typography)({
+  flex: '1 1 100%',
+});
 
 const StyledStepListItemButton = styled(ListItemButton)(({ theme }) => ({
   paddingLeft: theme.spacing(3),
   paddingRight: 0,
   paddingTop: 0,
   paddingBottom: 0,
+  whiteSpace: 'nowrap',
+  overflow: 'hidden',
 }));
 
 const StyledStepListItemIcon = styled(ListItemIcon)(({ theme }) => ({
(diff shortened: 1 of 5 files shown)


rguinee:
@wjames111 Generally yeah, looks good. I like that they can create these line items of mileage. Did you set this up where filling in the form below the table adds a new row? We'd also need some text somewhere that explains that $0.70 is for ministry and $0.21 is for medical. Like if the rate was a dropdown we could have the options label be "$0.70 - Ministry" and "$0.21 - Medical". Thoughts?

dbenton9:
> @dbenton9 is having something like this for mileage doable on the backend or do we need to simplify this into a single field that is saved?

@wjames111 I've read the comments but I'm unclear on the question. Help me understand what you mean when you say "something like this".

canac:
@wjames111 This is great work! You're right that I redesigned and simplified the layout. Do you think you can rebase this onto `main` and get that working with the new layout that doesn't use `SidePanelLayout`?

Also, seems like it would be easier to directly use the `rightPanel` in the `SidePanelLayout` in the goal calculator page instead of reimplementing it in `GoalCalculator.tsx`. I'm thinking that you could move `<GoalCalculatorProvider>` up so that it is outside of `<SidePanelLayout>` and move the right panel you built into a component that can be rendered in the `rightPanel` slot. I don't know if I'm explaining that well. Does that make sense?

src/components/Reports/GoalCalculator/GoalCalculator.tsx
```diff
@@ -55,14 +55,26 @@ const StyledDefaultContent = styled(Box)(({ theme }) => ({
   padding: theme.spacing(2),
 }));
 
-const StyledDrawer = styled(Box)<{ open: boolean }>(({ theme, open }) => ({
+const StyledDrawer = styled(Box)<{
```
canac, line 58: When you add custom props to a styled component, you also need to add a `shouldForwardProp` method to exclude those custom props. You can search the codebase for examples.

src/components/Reports/GoalCalculator/GoalCalculator.tsx
```diff
@@ -66,41 +109,85 @@ export const GoalCalculator: React.FC<GoalCalculatorProps> = ({
...
+    headerRef,
     handleCategoryChange,
     handleStepChange,
+    toggleRightPanel,
+    toggleDrawer,
+    setDrawerOpen,
+    headerHeight,
   } = useGoalCalculator();
   const { t } = useTranslation();
+  const [iconContainerWidth, setIconContainerWidth] = useState<string>('54px');
```
canac, line 123: Is there any way to get this to work without having to calculate the icon width? I'm probably looking at the wrong thing, but it looks fine to me without that `left: calc(${iconContainerWidth} + 1px)` prop.

canac: OK, so you're trying to make the right panel not cover up the icons? This is what I'm seeing with your latest commit.

Screenshot 2025-07-31 at 4 51 55 PM

Another solution would be to define a constant for the icon stack width: `const iconPanelWidth = theme.spacing(4)`, use it to set the width of the stack, and use it in the `calc`. Does that make sense? That way we know the width instead of having to measure it. I'd like to avoid the measurement logic there's not another way to do this because it add complexity.

canac: Thanks for refactoring it to be simpler!

src/components/Reports/GoalCalculator/GoalCalculator.tsx
```diff
@@ -66,13 +114,32 @@ export const GoalCalculator: React.FC<GoalCalculatorProps> = ({
...
     handleCategoryChange,
     handleStepChange,
+    toggleRightPanel,
+    toggleDrawer,
+    setDrawerOpen,
   } = useGoalCalculator();
   const { t } = useTranslation();
+  const iconPanelWidth = theme.spacing(5);
+
+  const handleCategoryIconClick = (categoryId: typeof selectedCategoryId) => {
```
canac, line 128: In a follow-up PR at some point, I'd love to see this logic moved to the context. But it's not crucial and can wait.

PR #1384, opened 2025-07-29
[no-Jira] Goal calculator personal info dropdowns

canac:
## Description

Make the personal info inputs dropdowns instead of text inputs. There may need to be some changes to the enums, and the geographic locations may end up being a server constant. But at least this gets us closer.

## Checklist:


canac:
> the only difference I noticed what not including the first and last name

@wjames111 I thought someone else was doing this, but since Zach hasn't yet, I added those fields in this PR. Thanks for pointing that out!

PR #1397, opened 2025-08-05
MPDX-8717 - Subcategory Table

wjames111:
## Description
[MPDX-8717](https://jira.cru.org/browse/MPDX-8717)
This PR does the following:
- Implements new DataGrid editable table for input subcategory line items or for directly inputting a total.
- Makes table reusable in shared components
- Category components now default to using the new DataGrid component when no component argument is provided
- Renames the remaining folder and files that still contain the step wording to use category
- Removes many of the category folders/files as they now default to using the DataGrid component
- Adds missing Enums and categories to match the Figma design
- Reorders Step list to match Figma Design

## Checklist:

### DIFF
--- a/src/components/Reports/GoalCalculator/CalculatorSettings/Categories/OneTimeGoalsCategory/OneTimeGoalsCategoryForm/OneTimeGoalsCategoryForm.test.tsx
+++ b/src/components/Reports/GoalCalculator/CalculatorSettings/Categories/OneTimeGoalsCategory/OneTimeGoalsCategoryForm/OneTimeGoalsCategoryForm.test.tsx
@@ -1,0 +1,111 @@
+import React from 'react';
+import { fireEvent, render } from '@testing-library/react';
+import { Formik } from 'formik';
+import TestWrapper from '__tests__/util/TestWrapper';
+import { OneTimeGoalsCategoryForm } from './OneTimeGoalsCategoryForm';
+
+const defaultValues = {
+  additionalGoals: [
+    { label: 'Emergency Fund', amount: 5000 },
+    { label: 'Equipment', amount: 2000 },
+  ],
+};
+
+const renderWithFormik = (initialValues = defaultValues) => {
+  return render(
+    <TestWrapper>
+      <Formik initialValues={initialValues} onSubmit={() => {}}>
+        <OneTimeGoalsCategoryForm />
+      </Formik>
+    </TestWrapper>,
+  );
+};
+
+describe('OneTimeGoalsCategoryForm', () => {
+  it('renders existing goals', () => {
+    const { getByDisplayValue } = renderWithFormik();
+
+    expect(getByDisplayValue('Emergency Fund')).toBeInTheDocument();
+    expect(getByDisplayValue('5000')).toBeInTheDocument();
+    expect(getByDisplayValue('Equipment')).toBeInTheDocument();
+    expect(getByDisplayValue('2000')).toBeInTheDocument();
+  });
+
+  it('renders add goal button', () => {
+    const { getByRole } = renderWithFormik();
+
+    expect(getByRole('button', { name: '+ Add Goal' })).toBeInTheDocument();
+  });
+
+  it('adds a new goal when add button is clicked', () => {
+    const { getByRole, getAllByLabelText } = renderWithFormik();
+
+    const addButton = getByRole('button', { name: '+ Add Goal' });
+    fireEvent.click(addButton);
+
+    expect(getAllByLabelText('Goal Label')).toHaveLength(3);
+    expect(getAllByLabelText('Amount')).toHaveLength(3);
+  });
+
+  it('removes a goal when delete button is clicked', () => {
+    const { getAllByLabelText } = renderWithFormik();
+
+    expect(getAllByLabelText('Goal Label')).toHaveLength(2);
+    expect(getAllByLabelText('Delete goal')).toHaveLength(2);
+
+    const deleteButtons = getAllByLabelText('Delete goal');
+    fireEvent.click(deleteButtons[0]);
+
+    expect(getAllByLabelText('Goal Label')).toHaveLength(1);
+    expect(getAllByLabelText('Delete goal')).toHaveLength(1);
+  });
+
+  it('updates goal label when input changes', () => {
+    const { getByDisplayValue } = renderWithFormik();
+
+    const labelInput = getByDisplayValue('Emergency Fund');
+    fireEvent.change(labelInput, { target: { value: 'Updated Fund' } });
+
+    expect(getByDisplayValue('Updated Fund')).toBeInTheDocument();
+  });
+
+  it('updates goal amount when input changes', () => {
+    const { getByDisplayValue } = renderWithFormik();
+
+    const amountInput = getByDisplayValue('5000');
+    fireEvent.change(amountInput, { target: { value: '7500' } });
+
+    expect(getByDisplayValue('7500')).toBeInTheDocument();
+  });
+
+  it('renders with empty goals list', () => {
+    const { getByRole, queryAllByLabelText } = renderWithFormik({
+      additionalGoals: [],
+    });
+
+    expect(queryAllByLabelText('Goal Label')).toHaveLength(0);
+    expect(queryAllByLabelText('Amount')).toHaveLength(0);
+
+    expect(getByRole('button', { name: '+ Add Goal' })).toBeInTheDocument();
+  });
+
+  it('renders currency adornment for amount fields', () => {
+    const { getAllByLabelText } = renderWithFormik();
+
+    const amountFields = getAllByLabelText('Amount');
+    amountFields.forEach((field) => {
+      expect(field.closest('.MuiInputBase-root')).toBeInTheDocument();
+    });
+  });
+
+  it('sets correct input attributes for amount fields', () => {
+    const { getAllByLabelText } = renderWithFormik();
+
+    const amountFields = getAllByLabelText('Amount');
+    amountFields.forEach((field) => {
+      expect(field).toHaveAttribute('type', 'number');
+      expect(field).toHaveAttribute('min', '0');
+      expect(field).toHaveAttribute('step', '0.01');
+    });
+  });
+});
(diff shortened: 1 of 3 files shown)


src/components/Reports/GoalCalculator/CalculatorSettings/Categories/InformationCategory/InformationCategoryForm/InformationCategoryFinancialForm.tsx
canac, line 1: Thank you for renaming these files and components!

The personal and financial information sections are kind of special cases. They're not categories in the API. I wonder if we could just call them FinancialInformationForm and PersonalInformationForm?

src/components/Reports/GoalCalculator/Shared/GoalCalculatorContext.tsx
```diff
@@ -54,8 +54,8 @@ export const GoalCalculatorProvider: React.FC<Props> = ({ children }) => {
   // Static categories - no memoization to avoid React queue issues
   const steps = [
     useCalculatorSettings(),
-    useHouseholdExpenses(),
     useMinistryExpenses(),
```
canac, line 57: Great catch!

src/components/Reports/GoalCalculator/SharedComponents/GoalCalculatorGrid/GoalCalculatorGrid.tsx
```diff
@@ -0,0 +1,276 @@
...
+  GridValidRowModel,
+} from '@mui/x-data-grid';
+import { Form, Formik, FormikProps } from 'formik';
+import { useTranslation } from 'react-i18next';
+import * as yup from 'yup';
+import { useGoalCalculator } from '../../Shared/GoalCalculatorContext';
+import { StyledGrid } from './StyledGrid';
+
+const StyledCard = styled(Card)({
+  borderRadius: '4px',
```
canac, line 17: We can pull the border radius from the theme.

```suggestion
  borderRadius: theme.shape.borderRadius
```

src/components/Reports/GoalCalculator/SharedComponents/GoalCalculatorGrid/GoalCalculatorGrid.tsx
```diff
@@ -0,0 +1,276 @@
...
+    {
+      field: 'amount',
+      headerName: t('Amount'),
+      flex: 1,
+      minWidth: 150,
+      editable: true,
+      type: 'number',
+      align: 'center',
+      headerAlign: 'center',
+      renderCell: (params) => `$${params.value.toLocaleString()}`,
```
canac, line 215: We want to use a currency formatter method here with USD and the user's locale.

src/components/Reports/GoalCalculator/GoalCalculator.tsx
```diff
@@ -202,7 +203,12 @@ export const GoalCalculator: React.FC<GoalCalculatorProps> = ({
                     </IconButton>
                   )}
                 </Typography>
-                {category.component}
+                {category.component || (
+                  <GoalCalculatorGrid
+                    headerName="Expense Name"
```
canac, line 208: Can we pass in the category's title here? I think we should let categories specify the call to action text as well. Unless it's OK to always use the text "Add Expense", in which case we should remove this prop and inline the label in `GoalCalculatorGrid`.

canac (commented):
Amazing work on the table! I love how many boilerplate components you were able to remove. The table is really nice with inline editing!

A couple overall thoughts:
* If you sort the table, the total row is sorted as well. We need to find a way to always keep it at the bottom or disable sorting.
* Can you add some tests for the table editing? If there's anything that will go away when the backend API is ready, you can skip those. But at least adding rows, removing rows, editing rows, and summing the total should have tests.
* Are we working on the Direct Input input in a future PR?

canac:
> I'll also see if I can get the total to stay on the last row (if that's not possible would you prefer disabling the sorting or pulling the total outside the table?)

@wjames111 Disabling sorting would be easiest. [Row pinning](https://mui.com/x/react-data-grid/row-pinning/) is a paid feature in DataGrid. If we decide we really need sorting, we can put the total outside of the table, like Donations. But it's finicky getting the columns to line up.

canac:
> I'm thinking the Direct Input's would be in a separate PR as well.

@wjames111 I'm totally fine with that. It's probably worth making another Jira ticket to track that work.

src/components/Reports/GoalCalculator/SharedComponents/GoalCalculatorGrid/GoalCalculatorGrid.tsx
canac: Everything in the goal calculator is in USD. We don't need to take the user's currency preference into account.

src/components/Reports/GoalCalculator/GoalCalculator.tsx
canac: Sounds great! It sounds like `headerName` is being used for more than just the header now, so I'd suggest renaming it to something like `categoryName`.

src/components/Reports/GoalCalculator/SharedComponents/GoalCalculatorGrid/GoalCalculatorGrid.tsx
```diff
@@ -196,7 +196,7 @@ const GoalCalculatorGridForm: React.FC<GoalCalculatorGridFormProps> = ({
   const columns: GridColDef[] = [
     {
       field: 'name',
-      headerName: t(headerName),
+      headerName: t(categoryName),
```
canac, line 199: `t` needs to be passed a static string, not a variable so that `yarn:extract` can search the codebase for all of our labels. We should make sure that `categoryName` is localized where it is passed in and then remove this `t()` call.

src/components/Reports/GoalCalculator/CalculatorSettings/CalculatorSettings.tsx
```diff
@@ -21,20 +17,15 @@ export const useCalculatorSettings = (): GoalCalculatorStep => {
       {
         id: GoalCalculatorCategoryEnum.Information,
         title: t('Information'),
-        subtitle: t('Take a moment to verify your information.'),
-        component: <InformationStep />,
+        component: <InformationCategory />,
```
canac, line 20: Did you mean to remove the `subtitle`?

src/components/Reports/GoalCalculator/SharedComponents/GoalCalculatorGrid/StyledGrid.tsx
```diff
@@ -0,0 +1,61 @@
...
+    fontWeight: 'bold',
+  },
+  '.MuiDataGrid-columnSeparator': {
+    display: 'none',
+  },
+  '.MuiDataGrid-cell[data-field="actions"]': {
+    '&:focus': {
+      outline: 'none',
+    },
+    '& .MuiDataGrid-actionsCell': {
```
canac, line 17:
```suggestion
    '.MuiDataGrid-actionsCell': {
```

src/components/Reports/GoalCalculator/SharedComponents/GoalCalculatorGrid/GoalCalculatorGrid.tsx
```diff
@@ -0,0 +1,274 @@
...
+  );
+};
+
+interface GoalCalculatorGridFormProps {
+  formikProps: FormikProps<GoalCalculatorGridFormValues>;
+  categoryName: string;
+}
+
+const GoalCalculatorGridForm: React.FC<GoalCalculatorGridFormProps> = ({
+  formikProps,
```
canac, line 139: Can we delete this prop and get the formik context with `useFormikContext()`?

src/components/Reports/GoalCalculator/SharedComponents/GoalCalculatorGrid/GoalCalculatorGrid.tsx
```diff
@@ -0,0 +1,274 @@
...
+            onClick={() => handleDelete(params.id as number)}
+            showInMenu={false}
+          />,
+        ];
+      },
+    },
+  ];
+
+  return (
+    <Box sx={{ pt: 2 }}>
```
canac, line 240: Is it possible to remove the `Box` and put this style on the `StyledCard`?

src/components/Reports/GoalCalculator/SharedComponents/GoalCalculatorGrid/GoalCalculatorGrid.test.tsx
```diff
@@ -0,0 +1,232 @@
...
+      </TestWrapper>,
+    );
+    const freelanceRow = getByText('Freelance Work').closest('[role="row"]');
+    userEvent.hover(freelanceRow!);
+    const deleteButtons = freelanceRow?.querySelectorAll(
+      '[aria-label="Delete"]',
+    );
+    if (deleteButtons && deleteButtons.length > 0) {
+      userEvent.click(deleteButtons[0] as Element);
+    }
```
canac, line 78: You should be able to just find the first button with `querySelector`.

```suggestion
    const deleteButton = freelanceRow?.querySelector(
      '[aria-label="Delete"]',
    );
    if (deleteButton) {
      userEvent.click(deleteButton as Element);
    }
```

src/components/Reports/GoalCalculator/SharedComponents/GoalCalculatorGrid/GoalCalculatorGrid.tsx
```diff
@@ -0,0 +1,274 @@
...
+      <StyledCard>
+        <StyledControlsBox>
+          <StyledAddButton
+            variant="text"
+            onClick={addExpense}
+            size="small"
+            startIcon={<AddIcon />}
+          >
+            {t(`Add ${categoryName}`)}
```
canac, line 249: This label needs to be a static string.

```suggestion
            {t('Add {{category}}', { category: categoryName })}
```

canac (commented):
This is amazing! Just a couple more tweaks, and this should be good to go. Thanks for making this so nice!

canac (approved):

PR #1399, opened 2025-08-07
[no-Jira] Add initial GitHub Copilot instructions

canac:
## Description

Adds basic GitHub Copilot instructions. From my research, these instructions are included with every prompt, so they need to be as lean as possible to avoid filling up the context window. There's probably more we could add, but for now I'm adding instructions to correct behavior I've seen multiple times from GitHub Copilot:

* Using `npm` instead of `yarn`
* Using the wrong command to run `tsc`
* Creating components with a default export
* Adding `await` to `userEvent.click()` (that is correct in v14, but we are on v13)

## Checklist:

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -8,6 +8,7 @@ MPDX is a Next.js 15 React application built with TypeScript using pages router.
 
 - **Frontend**: Next.js v15, React, Material UI v5, TypeScript
 - **Data Loading**: Apollo Client GraphQL
+- **Forms**: Formik and yup for validation
 - **Testing**: Jest, React Testing Library v13
 - **Internationalization**: react-i18next
 - **Code Quality**: ESLint, Prettier
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -14,7 +14,7 @@ MPDX is a Next.js 15 React application built with TypeScript using pages router.
 
 ## File Naming Conventions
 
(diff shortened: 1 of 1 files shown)


.github/copilot-instructions.md
```diff
@@ -0,0 +1,34 @@
...
+
+MPDX is a Next.js 15 React application built with TypeScript using pages router.
+
+## Technology Stack
+
+- **Frontend**: Next.js v15, React, Material UI v5, TypeScript
+- **Data Loading**: Apollo Client GraphQL
+- **Testing**: Jest, React Testing Library v13
+- **Internationalization**: react-i18next
+- **Code Quality**: ESLint, Prettier
```
dr-bizz, line 13: Should you also add a part about forms and that it should use Formik and yup for validation

.github/copilot-instructions.md
```diff
@@ -0,0 +1,34 @@
...
+- `yarn test` - Run tests
+- `yarn test [file]` - Run tests for a single file
+- `yarn lint` - Run ESLint linter
+- `yarn lint:ts` - Run TypeScript typechecking
+- `yarn prettier:write` - Run Prettier formatter
+
+## Guidelines
+
+- Always use named exports instead of default exports.
+- Always write the least amount of code possible. Avoid using unnecessary styles, variables, and test providers.
```
dr-bizz, line 34: Maybe also add, create Styled components where possible to avoid duplication

dr-bizz (approved):
Going to approve as I don't see anything wrong with it, I just had some coments

PR #1417, opened 2025-08-22
MPDX-8784 Create shared styledComponents

kegrimes:
## Description

When migrating reports from Staff Web to MPDX, we noticed many duplicate/redefined styled components throughout the codebase. To reduce redundancy, I have created a single file called `styledComponents.ts` where common components like `LoadingIndicator` are defined. If more are styled components are found, this PR will address them.

## Checklist:


canac:
It's possible that now any page that uses any styled component in that file is pulling in all of the MUI components it imports. I'm not entirely sure. The best way to find out would be to be each styled component in its own file and see if that brings down the bundle size.

PR #1420, opened 2025-08-22
[no-Jira] Upgrade prettier to v3

canac:
## Description

I was trying to use the newish `satisfies` TypeScript operator on another PR and noticed that it messed up prettier. It appears to be because prettier wasn't new enough. This PR simply upgrades to prettier 3 and removes the `trailingComma` configuration because it defaults to `all` now. It seems that the main breaking change is how it formats nested ternaries.

## Checklist:


dr-bizz (approved):
The code looks good. I have a question with updating the tests, but will approve as it's not blocking.

PR #1424, opened 2025-08-27
MPDX-8563: Allow anniversary delete and input on enter

zweatshirt:
## Description
[Jira ticket](https://jira.cru.org/browse/MPDX-8563)
Issue:
- People could already technically delete anniversary values, they were just unable to do it by pressing enter.
     - This is true of any other date fields that would exist in the Contact Details modal (namely, the Birthday field).
- People were also unable to save a valid date by pressing enter.
- Both behaviors, being unable to save data on enter and being unable to  clear data on enter, is inconsistent with other fields in the form.

Solution:
- Extend upon DesktopTimeField and DesktopDateField to have an Enter onKeyDown handler.

## Checklist:

### DIFF
--- a/src/components/common/DateTimePickers/DesktopDateField.tsx
+++ b/src/components/common/DateTimePickers/DesktopDateField.tsx
@@ -84,7 +84,11 @@ export const DesktopDateField: React.FC<DesktopDateFieldProps> = ({
       }}
       onKeyDown={(event) => {
         if (event.key === 'Enter') {
-          parsedDate?.isValid ? onChange(parsedDate) : onChange(null);
+          if (parsedDate?.isValid) {
+            onChange(parsedDate);
+          } else {
+            onChange(null);
+          }
         }
         props.onKeyDown?.(event);
       }}
--- a/src/components/common/DateTimePickers/DesktopTimeField.test.tsx
+++ b/src/components/common/DateTimePickers/DesktopTimeField.test.tsx
@@ -146,4 +146,30 @@ describe('DesktopTimeField', () => {
 
     expect(getByRole('dialog')).toBeInTheDocument();
   });
+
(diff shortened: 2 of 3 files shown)


zweatshirt:
> @zweatshirt Thanks so much for improving this! The way it worked before is definitely a bad user experience.
> 
> I have an idea for how to fix this at the source. The root cause seems to be that `DesktopDateField.tsx` only calls `onChange` in `onBlur`. I'd like for us to update `onChange` in `DesktopDateField` to try to parse the date that the user is starting to type. If It's an invalid date, do nothing because they probably haven't finished typing (e.g. "1/1/"). But if the input is empty or the date is a valid date, go ahead and call `onChange`. That way if they hit enter without tabbing away from the field, the form will have the date they just typed and be able to type it. Does that make sense?
> 
> I'm not saying you need to change it now, but `DesktopTimeField` has the same issue.

Thanks for the detailed response Caleb. I noticed the issue in DesktopDateField but tried to keep the PR in the scope of this Contact Details modal. I'll work on making that change now.

src/components/common/DateTimePickers/DesktopDateField.tsx
```diff
@@ -82,6 +82,12 @@ export const DesktopDateField: React.FC<DesktopDateFieldProps> = ({
       onChange={(event) => {
         setRawDate(event.target.value);
       }}
+      onKeyDown={(event) => {
+        if (event.key === 'Enter') {
+          parsedDate?.isValid ? onChange(parsedDate) : onChange(null);
```
canac, line 87: Can we use an `if` here instead of a ternary operator?

canac (approved):
This looks great! Should we add this logic to `DesktopTimeField` also?

PR #1433, opened 2025-09-09
Mpdx 8573 gql support

zweatshirt:
## Description

## Checklist:


zweatshirt:
@canac This is ready for review and all tests pass locally. The API changes haven't been merged into `master` of `mpdx_api` yet so tests are failing here still. Thanks for considering a review!

src/components/Reports/FourteenMonthReports/GetFourteenMonthReport.graphql
```diff
@@ -1,55 +1,58 @@
...
+  status
+}
+
+fragment accountList on AccountList {
+  id
+  createdAt
+  updatedAt
+}
+
+query GetFourteenMonthReport(
```
canac, line 19: This query could be optimized because it loads salary and donor currency donations, even though only one is needed. Can you look into using `@include` and `@skip` to only load one or the other? You'll have to add a boolean argument for whether to load salary or donor currency donations. Look at `src/components/Task/Modal/Form/Inputs/TagsAutocomplete/TagsAutocomplete.graphql` for an example.

src/components/Reports/FourteenMonthReports/useFourteenMonthReport.ts
```diff
@@ -0,0 +1,127 @@
...
+                months: contactDonationInfo.months.map((month, index) => {
+                  const salaryCurrencyTotal = month.donations.reduce(
+                    (convertedTotal, donation) =>
+                      convertedTotal + Number(donation.converted_amount),
+                    0,
+                  );
+                  return {
+                    month: data.months[index],
+                    total: isSalaryType
+                      ? salaryCurrencyTotal
```
canac, line 98: Why do we have to calculate the month's total manually for salary currency donations? Does the API not do that for us?

src/components/Reports/FourteenMonthReports/useFourteenMonthReport.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    const isSalaryType =
+      currencyType === FourteenMonthReportCurrencyType.Salary;
+    const currencyGroups: CurrencyGroups = data.currencyGroups;
+
+    return {
+      currencyType,
+      salaryCurrency: data.defaultCurrency ?? 'USD',
+      currencyGroups: Object.entries(currencyGroups ?? {}).map(
```
canac, line 51: Since the `currencyGroups` type isn't nullable (it is `JSON!`), I don't think we need to handle missing values.

```suggestion
      currencyGroups: Object.entries(currencyGroups).map(
```

src/components/Reports/FourteenMonthReports/useFourteenMonthReport.ts
```diff
@@ -0,0 +1,127 @@
...
+            ),
+            minimum: currencyGroup.donation_infos.reduce(
+              (minimumTotal, contactDonationInfo) =>
+                minimumTotal + Number(contactDonationInfo.minimum),
+              0,
+            ),
+          },
+          contacts: currencyGroup.donation_infos
+            .map((contactDonationInfo) => {
+              const contact = data.donorInfos.find(
```
canac, line 77: I don't think we would want to show rows where we couldn't find a contact, right? Since most of the fields will be empty. Can we return `null` if we don't find a contact and then filter out `null` values before sorting?

canac (commented):
This is great work, Zach! Sorry you had to do so much work to massage the API's data, but your migration looks really good! I like that you made a hook to handle that data transformation.

canac:
@zweatshirt This one may be a good one to get UAT from Scott on to make sure that nothing broke accidentally. I didn't see anything obviously wrong, but it would be good to double-check. Especially since we lose typesafety when the GraphQL gives us raw JSON data. I wouldn't want to accidentally have the wrong name of a field or something.

src/components/Reports/FourteenMonthReports/GetFourteenMonthReport.graphql
zweatshirt: For sure. I was considering the best approach for a while. I like this idea so I'll do that.

zweatshirt:
@canac Coincidentally I'm meeting with Scott tomorrow for categories filtering in the Staff Expense Report, so I'll bring this work up to him then as well.

src/components/Reports/FourteenMonthReports/Layout/Table/Table.test.tsx
```diff
@@ -1,14 +1,163 @@
...
+          pledgeCurrency: 'CAD',
+          pledgeFrequency: 'Monthly',
+        },
+      ],
+    },
+  },
+});
+
+// Mock the GraphQL hook to return our test data
+jest.mock('../../GetFourteenMonthReport.generated', () => ({
```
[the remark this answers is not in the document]
canac: Absolutely! Just sent you a meeting invite.

src/components/Reports/FourteenMonthReports/useFourteenMonthReport.ts
```diff
@@ -0,0 +1,152 @@
...
+                (donor) => donor.contactId === contactDonationInfo.contact_id,
+              );
+              if (!contact) {
```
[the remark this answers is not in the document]
zweatshirt: `Staging API` label applied due to work unrelated to this PR

src/components/Reports/FourteenMonthReports/FourteenMonthReport.test.tsx
```diff
@@ -198,7 +256,7 @@ const TestComponent: React.FC<TestComponentProps> = ({
             onNavListToggle={onNavListToggle}
             currencyType={currencyType}
             isNavListOpen={isNavListOpen}
-            designationAccounts={designationAccounts}
+            designationAccounts={designationAccounts ?? []}
```
canac, line 259: Was this an intentional change? I don't think it's necessary.

src/components/Reports/FourteenMonthReports/FourteenMonthReport.test.tsx
```diff
@@ -175,7 +229,7 @@ const TestComponent: React.FC<TestComponentProps> = ({
   currencyType,
   isNavListOpen = true,
   designationAccounts,
-  mutationSpy,
+  isSalary,
```
canac, line 232: Can we remove this prop and use `currencyType === FourteenMonthReportCurrencyType.Salary`?

src/components/Reports/FourteenMonthReports/Layout/Table/Table.test.tsx
```diff
@@ -169,57 +102,86 @@ const router = {
...
 
-interface ComponentsProps {
+interface TestComponentProps {
   orderBy?: OrderBy | null;
 }
-const Components: React.FC<ComponentsProps> = ({ orderBy = 'name' }) => {
-  const mocks = transformedData();
-  const getTotals = () => mocks?.currencyGroups[0].totals.months || [];
 
+const InnerComponent: React.FC<Pick<TestComponentProps, 'orderBy'>> = ({
```
canac, line 109: This `Pick` doesn't do anything because `TestComponentProps` only has an `orderBy` key.

```suggestion
const InnerComponent: React.FC<TestComponentProps> = ({
```

src/components/Reports/FourteenMonthReports/Layout/Table/Table.test.tsx
```diff
@@ -169,57 +102,86 @@ const router = {
...
+      orderBy={orderBy ?? null}
+      orderedContacts={currencyGroup.contacts}
+      salaryCurrency={currencyGroup.currency}
+      onRequestSort={onRequestSort}
+      totals={currencyGroup.totals.months}
+    />
+  );
+};
+
+const Providers: React.FC<{ children: React.ReactElement }> = ({
```
canac, line 136: Since it's only used once, you can combine this `Providers` component with `TestComponent`. Just have `TestComponent` render all of the providers and `InnerComponent`.

canac (commented):
Thanks for addressing everything from the last review and refactoring the test mocking code!

src/components/Reports/FourteenMonthReports/Layout/Table/Table.test.tsx
zweatshirt: I guess since TestComponentProps is used twice in this file it's best to use the props interface

canac: `Pick` is a great choice when you only want some of the props. But since you want all of the props here (even though it's just one), it seems good to keep things simple and use `TestComponentProps`.

zweatshirt:
@canac Scott is reporting partial data in the preview environment for his account compared to his account in prod, but I am wondering if this is simply due to the Preview Environment being run against the staging DB?

PR #1436, opened 2025-09-11
[no-Jira] Reduce `yarn gql` retry attempts

canac:
## Description

Reduce the number of attempts to retry `gql codegen`. My experience is that if it fails once, it will usually fail all 5 times for some reason. Retrying 5 times make it take longer before the job fails and I can rerun it.

## Checklist:


dr-bizz (approved):
Thank you for maintaining these

PR #1442, opened 2025-09-18
MPDX-8799 Add expense categories to MPGA report

kegrimes:
## Description

After receiving the expense categories and test data for a given staff account, I was able to implement them in the MPGA report. This required updating the pie chart, totals context, and other files as well as creating a custom hook to filter through expense categories. This list is **_tentative_** and are subject to change. 

Expense categories:
- Ministry reimbursement
- Healthcare reimbursement
- Salary
- Additional salary
- Benefits
- Assessment
- Donation
- Transfer
- Other

## Checklist:

### DIFF
--- a/src/components/Reports/MPGAIncomeExpensesReport/Helper/MPGAReportEnum.tsx
+++ b/src/components/Reports/MPGAIncomeExpensesReport/Helper/MPGAReportEnum.tsx
@@ -6,3 +6,14 @@ export enum ReportTypeEnum {
 export enum FundTypes {
   Primary = 'Primary',
 }
+
+export enum ExpenseCategoriesEnum {
+  Transfer = 'transfer',
+  Ministry = 'ministry reimbursement',
+  Healthcare = 'healthcare reimbursement',
+  AdditionalSalary = 'additional salary',
+  Salary = 'salary',
+  Benefits = 'benefits',
+  Other = 'other',
+  Assessment = 'assessment',
+}
--- a/src/components/Reports/MPGAIncomeExpensesReport/TotalsContext/TotalsContext.tsx
+++ b/src/components/Reports/MPGAIncomeExpensesReport/TotalsContext/TotalsContext.tsx
@@ -1,4 +1,5 @@
 import React, { useMemo } from 'react';
+import { useExpenseCategories } from 'src/hooks/useExpenseCategories';
 import { AllData, DataFields } from '../mockData';
 
 export type TotalsType = {
@@ -6,7 +7,7 @@ export type TotalsType = {
   expensesTotal: number;
   ministryTotal: number;
   healthcareTotal: number;
-  miscTotal: number;
+  assessmentTotal: number;
   otherTotal: number;
 };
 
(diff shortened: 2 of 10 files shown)


src/hooks/useExpenseCategories.test.ts
```diff
@@ -0,0 +1,16 @@
+import { renderHook } from '@testing-library/react';
+import { mockData } from 'src/components/Reports/MPGAIncomeExpensesReport/mockData';
+import { useExpenseCategories } from './useExpenseCategories';
+
+describe('useExpenseCategories', () => {
+  it('should categorize expenses correctly', () => {
```
dr-bizz, line 6: Can you write a test that on change of data it updates the value?

src/components/Reports/MPGAIncomeExpensesReport/TotalsContext/TotalsContext.tsx
```diff
@@ -35,56 +36,40 @@ export const TotalsProvider: React.FC<TotalsContextProps> = ({
...
 
-  // const healthcareTotal = useMemo(() => {
-  //   return sum(data.healthcareExpenses);
-  // }, [data.healthcareExpenses]);
+  const incomeTotal = useMemo(() => sum(data.income), [data.income]);
 
-  // const miscTotal = useMemo(() => {
-  //   return sum(data.misc);
-  // }, [data.misc]);
+  const ministryTotal = sum(ministry);
```
dr-bizz, line 45: Would it be worth moving these sums into `useExpenseCategories`, as they will stay the same unless data changes?

src/components/Reports/MPGAIncomeExpensesReport/TotalsContext/TotalsContext.tsx
```diff
@@ -35,56 +36,40 @@ export const TotalsProvider: React.FC<TotalsContextProps> = ({
...
-  const ministryTotal = 2_456.78;
-  const healthcareTotal = 1_799.26;
-  const miscTotal = 1_732.11;
-  const otherTotal = 980.25;
+  const otherTotal = sum(other);
 
-  const expensesTotal = useMemo(() => {
-    return sum(data.expenses);
-  }, [data.expenses]);
+  const expensesTotal = useMemo(
```
dr-bizz, line 53: Same here

src/components/Reports/MPGAIncomeExpensesReport/TotalsContext/TotalsContext.tsx
dr-bizz: Also it looks like the total is on the mock data, I assume this won't be the case when we get the real data?

src/components/Reports/MPGAIncomeExpensesReport/Helper/MPGAReportEnum.tsx
```diff
@@ -6,3 +6,14 @@ export enum ReportTypeEnum {
...
+export enum ExpenseCategoriesEnum {
+  Transfer = 'transfer',
+  Ministry = 'ministry reimbursement',
+  Healthcare = 'healthcare reimbursement',
+  AdditionalSalary = 'additional salary',
+  Salary = 'salary',
+  Benefits = 'benefits',
+  Other = 'other',
+  Assessment = 'assessment',
+}
```
dr-bizz, line 19: What are your thoughts about putting this Enum on the API? Since the data is coming from there, in my mind, it would make sense that is also where the Enum is created.

dr-bizz: this could be a question for Shelby cc @dbenton9

dr-bizz (commented):
Really good work on this. I've left some comments, but not many at all and none which are big issues.

Thank you for owning the MPGA report.

src/components/Reports/MPGAIncomeExpensesReport/Helper/MPGAReportEnum.tsx
kegrimes: Is this something that is in progress already? I think it would be a good idea.

PR #1682, opened 2026-04-06
Add spouseCalculations to ASR context

dr-bizz:
## Summary
This PR is a follow-up for https://github.com/CruGlobal/mpdx-react/pull/1680 to address Will's comments

### Changes
- Added `spouseCalculations` to `AdditionalSalaryRequestContext` (mirroring existing `calculations`)
- Replaced all direct `requestData?.latestAdditionalSalaryRequest?.calculations` and `requestData?.latestAdditionalSalaryRequest?.spouseCalculations` access with context values
- Updated tests to provide calculations at the top level of mock context values

## Test plan
- [x] All 318 AdditionalSalaryRequest tests pass
- [x] TypeScript compiles cleanly
- [x] ESLint passes

### DIFF
--- a/src/components/Reports/AdditionalSalaryRequest/Shared/useAdditionalSalaryRequestForm.ts
+++ b/src/components/Reports/AdditionalSalaryRequest/Shared/useAdditionalSalaryRequestForm.ts
@@ -9,7 +9,6 @@ import { currencyFormat } from 'src/lib/intlFormat';
 import { amount, phoneNumber } from 'src/lib/yupHelpers';
 import { CompleteFormValues } from '../AdditionalSalaryRequest';
 import {
-  useAdditionalSalaryRequestQuery,
   useSubmitAdditionalSalaryRequestMutation,
   useUpdateAdditionalSalaryRequestMutation,
 } from '../AdditionalSalaryRequest.generated';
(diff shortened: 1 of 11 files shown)


src/components/Reports/AdditionalSalaryRequest/Shared/useAdditionalSalaryRequestForm.ts
```diff
@@ -77,17 +76,13 @@ export const useAdditionalSalaryRequestForm = (
...
 
   const { primaryAccountBalance } = useFormUserInfo();
 
-  const { data: requestData } = useAdditionalSalaryRequestQuery({
-    variables: { isSpouse },
-  });
-  const individualCap =
-    requestData?.latestAdditionalSalaryRequest?.calculations.currentSalaryCap ??
-    0;
+  const individualCap = calculations?.currentSalaryCap ?? 0;
```
dr-bizz: We are doing this in the context, which is why we don't need to call `useAdditionalSalaryRequestQuery` here

wjames111 (approved):
Looks great thanks for the refactor! Sorry I didn't realize how many test files would need to change. I did notice there's a few mocks with `requestData` in them that I don't think are needed anymore.

dr-bizz:
@wjames111 Do you know if this is still relevant? I started it before I went out on leave.

dr-bizz:
This is no longer relevant

PR #1707, opened 2026-04-10
MPDX-9448 - Add step navigation to PDS Calc, and lift up PdsGoalCalculatorLayout (DRY refactor)

zweatshirt:
## Description

- Adds step navigation for the calculator (including back functionality which isn't in GoalCalculator, but I personally prefer this. We can remove it if desired). 
- Lifts up PdsGoalCalculatorLayout into PdsGoalCalculator.tsx to avoid redundancy. The steps can now focus on their own content.
- Remove SetupForm which was wrapped in SetupStep unnecessarily after simplifying structure.

## Testing

- Go to`reports/pdsGoalCalculator`
- Check the step navigation, and that it works well

## Checklist:


zweatshirt:
@wjames111 Sorry this seems like a lot of changes but it's not, most of the diff is caused by moving SetupForm to be in SetupStep.tsx

src/components/Reports/PdsGoalCalculator/Setup/SetupStep.tsx
```diff
@@ -1,18 +1,198 @@
...
+          {t('What do you want to call this goal?')}
+        </Typography>
+      </Box>
+      <AutosaveTextField
+        fieldName="name"
+        schema={schema}
+        label={t('Goal Name')}
+      />
+
+      <Divider sx={{ my: 4, mx: -4 }} />
```
wjames111, line 82: Question: Do we need a negative left and right margin here?

src/components/Reports/Shared/CalculationReports/DirectionButtons/DirectionButtons.tsx
```diff
@@ -94,9 +96,9 @@ export const DirectionButtons: React.FC<DirectionButtonsProps> = ({
         mt: 5,
         display: 'flex',
         justifyContent:
-          !handleDiscard && !showBackButton && !isSubmission
-            ? 'flex-end'
-            : 'space-between',
+          alignStart || handleDiscard || showBackButton || isSubmission
+            ? 'space-between'
+            : 'flex-end',
```
wjames111: Optional: This is showing on the left hand side, I think the other reports are right by default. I'd probably opt for continuity here over what works/looks best.

wjames111 (approved):
Thanks for lifting up the layout, I think it's a lot cleaner now!

src/components/Reports/Shared/CalculationReports/DirectionButtons/DirectionButtons.tsx
zweatshirt: I was going by the Goal Calculator, I realized this as well. The Goal Calculator is left/start aligned, other reports right/end aligned. What do you think? I'm fine with either. If you think end aligned, I'm cool with that. Do we eventually want to change the goal calculator for consistency?

src/components/Reports/PdsGoalCalculator/Setup/SetupStep.tsx
zweatshirt: Without a significant refactor I think so. This is to allow the divider to span the full width of `MainContent`. The alternative is to add props for adjustable margin/padding to `MainContent` or do something similar.
This divider is something I took from GoalCalculator as well.

src/components/Reports/Shared/CalculationReports/DirectionButtons/DirectionButtons.tsx
zweatshirt: I changed it to be end aligned. We can change Goal Calculator eventually.

zweatshirt: Also removed the back button since GoalCalculator doesn't have it.. Maybe I need to make a PR to make all the forms consistent at this point.

PR #1708, opened 2026-04-13
MPDX-9444 - Designation Support Calc: Pull 403b contribution from HCM

zweatshirt:
## Description

Pulls 403b contribution directly from HCM, sets it as an uneditable field. 
The drawback to pulling this value frontend side is an extra HCM call. The upside is this value isn't meant to be persisted, so it's not entirely appropriate to pull it API side and apply it to the DesignationSupportCalculation model.

While in this file, I also added the name and avatar display, to more closely match the GoalCalculator and help users understand that they are working with their own information.

## Testing

- Go to `reports/pdsGoalCalculator`
- Open a goal or create a new one.
- Check that the 403b contribution is correctly pulled from HCM. 
- Check that the Name + Avatar box looks good.

## Checklist:

### DIFF
--- a/src/components/Reports/PdsGoalCalculator/Setup/SetupStep.test.tsx
+++ b/src/components/Reports/PdsGoalCalculator/Setup/SetupStep.test.tsx
@@ -126,6 +126,71 @@ describe('SetupStep', () => {
     ).toBeDisabled();
   });
 
+  it('displays the sum of tax-deferred and Roth 403b contribution percentages', async () => {
+    const { findByRole } = render(
+      <PdsGoalCalculatorTestWrapper
+        calculationMock={fullTimeSalariedMock}
+        hcmUserMock={{
+          fourOThreeB: {
+            currentTaxDeferredContributionPercentage: 5,
+            currentRothContributionPercentage: 3,
+          },
+        }}
+      >
+        <SetupStep />
+      </PdsGoalCalculatorTestWrapper>,
+    );
+
+    await waitFor(async () =>
+      expect(
+        await findByRole('textbox', {
+          name: '403b Contribution Percentage',
+        }),
+      ).toHaveValue('8'),
+    );
+  });
+
+  it('403b field is empty when hcm data has no 403b entry', async () => {
+    const { findByRole } = render(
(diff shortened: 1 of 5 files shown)


src/components/Reports/PdsGoalCalculator/Setup/SetupStep.test.tsx
```diff
@@ -126,6 +126,73 @@ describe('SetupStep', () => {
...
+            avatar: 'https://example.com/jane.png',
+          },
+        }}
+      >
+        <SetupStep />
+      </PdsGoalCalculatorTestWrapper>,
+    );
+
+    await waitFor(async () =>
+      expect(await findByTestId('info-name-typography')).toHaveTextContent(
```
wjames111: Question: is the `await findByTestId` inside the `waitFor` needed?

wjames111 (approved):
Looks great!

src/components/Reports/PdsGoalCalculator/Setup/SetupStep.test.tsx
zweatshirt: I couldn't fix it to work any other way, I've had this issue in the past with test id finds. I'll look to see if I can find something better.

I had a question for you: will having a potentially 0% 403b contrib throw off our calculations? If there is a situation where the user simply doesn't have a 403b what should we do? Just not include it in any calculations?

canac: Yeah, you have to do `waitFor` if the test-id exists initially but doesn't have the expected content. If you're inside `waitFor`, you can do `getByTestId` to simplify it a little bit.

zweatshirt: Thanks @canac I'll do that!

wjames111: @zweatshirt not sure I was thinking just not including it but maybe I'm overlooking something. Might be a good question for Daniel, or Caleb might know.

PR #1714, opened 2026-04-14
MPDX-9350, MPDX-9357 Hide certain features for Cru US staff groups

kegrimes:
## Description

Need these groups to differentiate groups part of Cru US staff:
- New Staff = `supported_rmo`,  `raising_initial_support_payroll_eligible` & `raising_initial_support_no_payroll`
- Senior Staff (SS) =  `supported_rmo`,  `active_payroll_eligible`, `full_time_regular`
- Part-time Field Staff (PTFS) =  `employee_ptfs`
- Paid with designation =  `designation`
- Interns =  `employee_us_intern` & `employee_international_intern`

ASR --> RMO staff only (senior staff & new staff); light version for PTFS; paid with designation & interns should NOT have access
Salary Calculator --> Senior staff only
MHA Calculator --> Senior staff only

_The backend should already be..._

- ASR = validating payroll eligible, interns as ineligible, and RMO staff or PTFS staff
- Salary = validating senior staff only
- MHA = validating senior staff only

_This PR will..._

- Remove duplicate HCM query
- (ASR) Ensure paid with designation & interns cannot access ASR - hide tabs and form
- (Salary) Ensure new staff, PTFS, paid with designation, & interns cannot access salary calc - hide tabs and form
- (MHA) Ensure new staff, PTFS, paid with designation, & interns cannot access MHA - hide tabs and form
- Hide reports that require staff account id when user does not have one
- Add `user_type_verified` check. If not verified, show existing MPDX navbar/pages (hide new reports). If verified, show reports based on user group (this is meant to replace the `DISABLE_NEW_REPORTS` environment variable when it is removed)
- Design change for modal

**Getting clarification from Scott on a limited ASR form for part-time field staff**

## Testing

**Impersonate user: "courtney.campbell@cru.org"**
ASR - Senior Staff: ✅
- Go to `/hrTools/additionalSalaryRequest`
- Check that ASR is visible to user

Salary Calc - Senior Staff: ✅
- Go to `/hrTools/salaryCalculator`
- Check that Salary Calc is visible to user

MHA - Senior Staff: ✅
- Go to `/hrTools/mhaCalculator`
- - Check that MHA is visible to user

**Impersonate user: "timothy.ethington@cru.org"**
ASR - New Staff: ✅
- Go to `/hrTools/additionalSalaryRequest`
- Check that ASR is visible to user

Salary Calc - New Staff: ✅
- Go to `/hrTools/salaryCalculator`
- Check that Salary Calc is NOT visible to user

MHA - New Staff: ✅
- Go to `/hrTools/mhaCalculator`
- - Check that MHA is NOT visible to user

**Impersonate user: "christina.garber@athletesinaction.org"**
ASR - PTFS:
- Go to `/hrTools/additionalSalaryRequest`
- Check that ASR is visible to user

Salary Calc - PTFS: 
- Go to `/hrTools/salaryCalculator`
- Check that Salary Calc is NOT visible to user

MHA - PTFS :
- Go to `/hrTools/mhaCalculator`
- - Check that MHA is NOT visible to user

**Impersonate user: "jacqueline.cortez@cru.org"**
ASR - Paid with Designation: ✅
- Go to `/hrTools/additionalSalaryRequest`
- Check that ASR is NOT visible to user

Salary Calc - Paid with Designation: ✅
- Go to `/hrTools/salaryCalculator`
- Check that Salary Calc is NOT visible to user

MHA - Paid with Designation: ✅
- Go to `/hrTools/mhaCalculator`
- - Check that MHA is NOT visible to user

**Impersonate user: "nataley.guetherman@cru.org"**
ASR - Intern:
- Go to `/hrTools/additionalSalaryRequest`
- Check that ASR is NOT visible to user

Salary Calc - Intern:
- Go to `/hrTools/salaryCalculator`
- Check that Salary Calc is NOT visible to user

MHA - Intern: 
- Go to `/hrTools/mhaCalculator`
- - Check that MHA is NOT visible to user

Hardcode Test:
- Senior Staff ✅
- New Staff ✅
- PTFS ✅
- Paid with designation ✅
- Interns ✅

## Checklist:


kegrimes:
# 🤖 Multi-Agent Code Review Report

**PR**: #1714 — MPDX-9350, MPDX-9357 Hide certain features for Cru US staff groups
**Branch**: `add-cru-us-groups` → `main`
**Agents**: 6 of 7 (Financial skipped — no calculation changes)
**Risk Score**: 8/15 — **HIGH**

---

## 🚫 HIGH PRIORITY ISSUES (Severity 7-9)

### 1. Implicit `any` type on `usStaffGroups` parameter

**Severity**: 8/10 (Consensus from 6 agents)
**File**: `src/hooks/useUsStaffGroups.ts:10`

```typescript
const usStaffGroups = (staff) => ({  // ← implicit any
```

The `staff` parameter has no type annotation, violating the project's "no `any` types in new code" standard. If HCM field names change in the schema, this function would silently return `false` for all group checks, making users appear eligible when they shouldn't be.

**Fix**:
```typescript
import { HcmQuery } from './Hcm.generated';
type StaffInfo = HcmQuery['hcm'][number]['staffInfo'];

const usStaffGroups = (staff: StaffInfo | undefined) => ({
```

---

### 2. Duplicate `mhaEit` and `asrEit` selections in Hcm.graphql

**Severity**: 7.5/10 (Consensus from 4 agents)
**File**: `src/components/Reports/Shared/HcmData/Hcm.graphql:32-54`

Both `mhaEit` and `asrEit` are selected twice. The second `asrEit` drops `ineligibilityReason`. This is a merge artifact from consolidating the two `.graphql` files. While GraphQL merges duplicates at runtime, this is a maintenance hazard.

**Fix**: Remove lines 49-54 (the duplicate selections).

---

### 3. `useUsStaffGroups` fires HCM query unconditionally

**Severity**: 6.5/10 (Consensus from 3 agents)
**File**: `src/components/Shared/UserTypeAccess/UserTypeAccess.tsx:35-36`

`useUsStaffGroups` is called on every `UserTypeAccess` render, even when `isAsr` and `isSalaryCalc` are both falsy. This fires an unnecessary HCM network request on every page using `UserTypeAccess` (e.g., Financial Accounts pages that only check `allowedUserType`).

**Recommendation**: Add a `skip` parameter to `useUsStaffGroups` or conditionally skip the inner `useHcmQuery` when neither flag is set.

---

### 4. `UserTypeEnum` imported from wrong generated file

**Severity**: 6.5/10 (Consensus from 2 agents)
**File**: `src/hooks/useHrToolsNavItems.ts:3`

```typescript
import { UserTypeEnum } from 'pages/api/graphql-rest.page.generated';  // ← wrong source
```

Every other file imports `UserTypeEnum` from `src/graphql/types.generated`. This inconsistency could cause issues if the REST-proxy enum diverges.

**Fix**: Change to `import { UserTypeEnum } from 'src/graphql/types.generated';`

---

## ⚠️ IMPORTANT ISSUES (Severity 5-7)

### 5. Missing test coverage for `useUsStaffGroups` edge cases

**Severity**: 6.5/10

Missing tests:
- HCM query error/failure state
- Empty HCM data (`data.hcm = []`)
- `RaisingInitialSupportNoPayroll` assignment status (only `PayrollEligible` tested)
- `EmployeeInternationalIntern` user type (only `EmployeeUsIntern` tested)
- `effectiveDate` prop not tested at `UserTypeAccess` integration level

### 6. `queryByRole` misuse in NavMenu test

**Severity**: 7/10
**File**: `src/components/Layouts/Primary/TopBar/Items/NavMenu/NavMenu.test.tsx`

`queryByRole` is synchronous — `await` on it is misleading and the test may pass vacuously before the component finishes rendering. Should use `waitFor` wrapping the assertion.

### 7. No dedicated tests for modified navigation hooks

**Severity**: 7/10

`useHrToolsNavItems`, `useReportNavItems`, and `useNavPages` were modified to add group-based filtering but lack colocated test files. Per CLAUDE.md standards, every modified hook should have a colocated `.test.{ts,tsx}`.

### 8. Flash of content for ineligible users during HCM loading

**Severity**: 6/10

When HCM data is loading, `useUsStaffGroups` returns `false` for both ineligible flags (user appears eligible). If the query resolves and the user is ineligible, children briefly render before being replaced by `LimitedAccess`.

### 9. Client-side only access control

**Severity**: 7/10

All restrictions are UI-level feature hiding. Confirm with the backend team that the server-side API enforces the same group-based eligibility rules for ASR submissions and salary calculator mutations.

---

## 💡 SUGGESTIONS (Severity 3-5)

| # | Suggestion | Severity |
|---|-----------|----------|
| 1 | Extract `'user_type_verified'` into a shared `useIsUserTypeVerified()` hook (repeated in 5+ locations) | 4/10 |
| 2 | Remove unused `seniorStaff` computation from `usStaffGroups` helper | 3/10 |
| 3 | Consider `featureType?: 'asr' \| 'salaryCalc'` enum prop on `UserTypeAccess` instead of accumulating boolean flags | 4/10 |
| 4 | `useMemo` in `useReportNavItems` depends on `userOptionData` object reference (unstable) — should depend on `.value` | 5/10 |
| 5 | `LimitedAccess` component lacks semantic `<main>` landmark for accessibility | 5/10 |
| 6 | Add code comment on `alwaysAllow` bypass noting it intentionally skips group checks | 4/10 |

---

## 📝 SUMMARY

| Agent | High | Important | Suggestions | Confidence |
|-------|------|-----------|-------------|------------|
| 🔒 Security | 1 | 2 | 3 | Medium-High |
| 🏗️ Architecture | 2 | 2 | 3 | High |
| 💾 Data Integrity | 2 | 3 | 3 | High |
| 🧪 Testing | 3 | 4 | 2 | Medium |
| 👤 UX | 1 | 2 | 4 | High |
| 📋 Standards | 2 | 4 | 0 | High |

## 🎯 RECOMMENDED NEXT STEPS

**Must fix before merge:**
- [x] Add type annotation to `usStaffGroups` parameter (`src/hooks/useUsStaffGroups.ts:10`)
- [x] Remove duplicate `mhaEit`/`asrEit` selections in `Hcm.graphql`
- [x] Fix `UserTypeEnum` import in `useHrToolsNavItems.ts:3`

**Should address:**
- [x] Add `skip` logic to `useUsStaffGroups` when neither `isAsr` nor `isSalaryCalc` is set
- [ ] Add missing test cases for `useUsStaffGroups` (error state, empty data, both enum variants)
- [x] Fix `queryByRole` → `waitFor` in NavMenu test

**Verify externally:**
- [ ] Confirm with backend team that server-side API enforces same group-based eligibility rules

---

_🤖 Generated by MPDX Multi-Agent Review System v3.0 | 6 agents | Risk: HIGH (8/15)_

kegrimes:
@canac I've been thinking about what we should do for PDS staff and thought instead of isolating that one case, maybe we should broaden it. What if we hide the HCM reports (ASR, salary calc, and MHA) from anyone who doesn't have a staff account id? PDS staff members wouldn't see them at all since their Cru US group can't be determined from HCM anyway (HCM needs a staff account id to query). This fixes the tab visibility issue in the nav bar. The trade-off is that if someone navigates directly to the URL, they'd see a "no staff account id" error instead of a "wrong user group" error — but that feels small since most users wouldn't know the URL exists if the tab is hidden in the first place.

Now, if someone doesn't have a staff account id but should, that's a separate issue — though it might still be worth it. I'm not sure how often that comes up. 

Also, Randy is talking to Scott about maybe changing HCM's API so that at least PTFS and interns can be fixed! We would only need to worry about PDS if this is the case. Right now, it is filtering out everything besides `Employee - Staff` and `Employee - Staff - nonRMO Spouse`. I suggested adding `Employee - PTFS`, `Employee - International Intern`, and `Employee - US Intern`

I think it is really important to fix the tab visibility issue but, let me know your thoughts!

canac:
> What if we hide the HCM reports (ASR, salary calc, and MHA) from anyone who doesn't have a staff account id?

I'm all for that change! If they see a "no staff account" error, I think that's OK because the error is correct: they don't have a staff account, and they need one to submit ASR, salary requests, or MHA requests.

> Now, if someone doesn't have a staff account id but should, that's a separate issue — though it might still be worth it.

I think that's also OK. Someone without a staff account id who should will see the "no staff account" error and contact support. We'll know that the issue is in MDPX and not in HCM.

> I think it is really important to fix the tab visibility issue

What exactly is the tab visibility issue? (I haven't looked at the code yet but am just about to. Maybe that will answer my question.)

pages/accountLists/[accountListId]/hrTools/salaryCalculator/[calculationId]/index.page.tsx
```diff
@@ -32,10 +32,19 @@ const SalaryCalculatorSavingStatus: React.FC = () => {
...
-export const SalaryCalculatorEditPage: React.FC = () => {
+export const SalaryCalculatorEditOuterPage: React.FC = () => {
+  return (
+    <SalaryCalculatorProvider>
+      <SalaryCalculatorEditPage />
+    </SalaryCalculatorProvider>
+  );
+};
+
+const SalaryCalculatorEditPage: React.FC = () => {
```
Comment (canac) [inline pages/accountLists/[accountListId]/hrTools/salaryCalculator/[calculationId]/index.page.tsx]: I'd consider naming them `SalaryCalculatorEditPage` and `SalaryCalculatorEditInnerPage` so that the exported component has a cleaner name. The fact that we have to have multiple components in this file is an implementation detail that we can hide from the components that use it.

This feedback applies to the index page as well.

src/hooks/useReportsDisabled.ts
```diff
@@ -0,0 +1,13 @@
+import { useUserOptionQuery } from './UserPreference.generated';
+
+export function useReportsDisabled() {
```
canac, line 3: question: It looks like we're always taking the output of this hook and passing it into `UserTypeAccess`. Can we make `UserTypeAccess` smarter and more self-contained and move this logic into the component?

kegrimes:
Sorry! The issue was that since we can't classify PDS users from HCM, they will see all the reports in the nav bar, even though ASR and salary calc should be hidden from them. I was suggesting we should hide all HCM reports from users without a staff account id because that would fix the tabs showing for PDS and then we don't need to care about what they see on the actual page since they shouldn't know the URL.

src/components/Reports/Shared/HcmData/Hcm.graphql
```diff
@@ -1,36 +1,34 @@
...
+      userPersonType
+      assignmentStatus
+      assignmentCategory
     }
     currentSalary {
       grossSalaryAmount
       lastRegularPaymentDate
       lastUpdated
     }
-    mhaEit {
```
canac: praise: Thanks for removing duplicate fields!

src/components/Shared/ConfirmUserGroupModal/ConfirmUserGroupModal.tsx
```diff
@@ -43,6 +43,17 @@ export const ConfirmUserGroupModal: React.FC<ConfirmUserGroupModalProps> = ({
...
   const userTypeLabel = getUserType(userType, t);
 
+  const description =
+    userType === UserTypeEnum.UsStaff
+      ? t(
+          'Users in this group receive (mostly) US donations and are paid through our US HR system.',
+        )
+      : userType === UserTypeEnum.GlobalStaff
+        ? t(
+            'Users in this group receive (mostly) non-US donations and are paid through our Global Netsuite system.',
```
canac:
```suggestion
            'Users in this group receive (mostly) non-US donations and are paid through our Global NetSuite system.',
```

src/components/Shared/ConfirmUserGroupModal/ConfirmUserGroupModal.tsx
```diff
@@ -76,10 +87,25 @@ export const ConfirmUserGroupModal: React.FC<ConfirmUserGroupModalProps> = ({
       <DialogTitle>{t('Is this your user group?')}</DialogTitle>
       <DialogContent>
         <Alert severity="info" sx={{ mb: 2 }}>
-          <Box sx={{ display: 'flex', gap: 0.5, mb: 2 }}>
-            <Typography>{t('The user group for your account is:')}</Typography>
-            <Typography fontWeight="bold">{userTypeLabel}</Typography>
+          <Box sx={{ display: 'flex', gap: 0.5, mb: description ? 0 : 2 }}>
+            {userType !== UserTypeEnum.NonCru ? (
```
canac: optional: Can we flip this ternary (check the `=== NonCru` case first)? Right now we're checking if they're not non-Cru (double negative).

src/components/Shared/UserTypeAccess/UserTypeAccess.tsx
```diff
@@ -2,54 +2,76 @@ import Loading from 'src/components/Loading';
...
   }
 
-  if (error) {
+  if (userError) {
     return <LimitedAccess userGroupError />;
   }
 
-  if (userLoading && !userType) {
-    return null;
+  if ((userLoading && !userType) || hcmLoading) {
```
canac, line 68: This will render the loading state while HCM data is loading, even if it's cached.

src/components/Shared/UserTypeAccess/UserTypeAccess.tsx
```diff
@@ -2,54 +2,76 @@ import Loading from 'src/components/Loading';
 import { useStaffAccountQuery } from 'src/components/Reports/StaffAccount.generated';
 import { useGetUserQuery } from 'src/components/User/GetUser.generated';
 import { UserTypeEnum } from 'src/graphql/types.generated';
+import { useUsStaffGroups } from 'src/hooks/useUsStaffGroups';
 import { LimitedAccess } from '../LimitedAccess/LimitedAccess';
 
 interface UserTypeAccessProps {
   allowedUserType?: UserTypeEnum;
   requireStaffAccount?: boolean;
+  requireUserGroups?: 'asr' | 'salaryCalc';
```
canac: I personally like string literals, but the repo convention is to use enums for things like this.

src/components/Shared/UserTypeAccess/UserTypeAccess.tsx
```diff
@@ -2,54 +2,76 @@ import Loading from 'src/components/Loading';
...
+  effectiveDate?: string | null;
   children: React.ReactElement;
   alwaysAllow?: boolean;
 }
 
 export const UserTypeAccess: React.FC<UserTypeAccessProps> = ({
   allowedUserType = UserTypeEnum.UsStaff,
```
canac: optional: For consistency with the other prop names.

```suggestion
  requiredUserType = UserTypeEnum.UsStaff,
```

src/components/Shared/UserTypeAccess/UserTypeAccess.tsx
```diff
@@ -2,54 +2,76 @@ import Loading from 'src/components/Loading';
...
+    inSalaryCalcIneligibleGroup,
+    loading: hcmLoading,
+  } = useUsStaffGroups({ effectiveDate: date, skip });
+
   const userType = data?.user.userType;
+  const cruUsStaff = userType === UserTypeEnum.UsStaff;
+
+  const limitedAccess =
+    (userType && userType !== allowedUserType) ||
+    (isAsr && cruUsStaff && inAsrIneligibleGroup) ||
```
canac: What is the `cruUsStaff` check for? It's redundant for ASRs and salary calcs, right?

src/hooks/useHrToolsNavItems.ts
```diff
@@ -1,14 +1,24 @@
...
+
+  const { inAsrIneligibleGroup, inSalaryCalcIneligibleGroup } =
+    useUsStaffGroups();
+  const usStaff = userType === UserTypeEnum.UsStaff;
 
   const hrToolsNavItems: NavItems[] = [
     {
       id: 'salaryCalculator',
       title: t('Salary Calculator'),
+      hideItem: usStaff && inSalaryCalcIneligibleGroup,
```
canac, line 24: This shows salary calc to non-US staff? Is this so that we don't hide it form people without a confirmed user type?

src/hooks/useUsStaffGroups.ts
```diff
@@ -0,0 +1,77 @@
...
+ */
+export function useUsStaffGroups({
+  effectiveDate,
+  skip,
+}: {
+  effectiveDate?: string;
+  skip?: boolean;
+} = {}) {
+  const { data, loading } = useHcmQuery({
+    variables: effectiveDate ? { effectiveDate } : undefined,
```
canac: If you pass undefined for a variable, it will be omitted.

```suggestion
    variables: { effectiveDate },
```

src/hooks/useUsStaffGroups.ts
```diff
@@ -0,0 +1,77 @@
...
+    skip,
+  });
+  const [user, spouse] = data?.hcm ?? [];
+
+  // When no effectiveDate is provided (is ASR), select the eligible user
```
canac: I don't love detecting the request form type based on whether or not it has an effective date. Salary requests could also have one user eligible and the other spouse not eligible.

src/hooks/useUsStaffGroups.ts
```diff
@@ -0,0 +1,77 @@
...
+    : spouse?.asrEit?.asrEligibility
+      ? spouse
+      : undefined;
+
+  const staff = effectiveDate ? user : (asrEligiblePerson ?? user);
+
+  const { newStaff, partTimeFieldStaff, paidWithDesignation, intern } =
+    usStaffGroups(staff?.staffInfo);
+
+  const inAsrIneligibleGroup = paidWithDesignation || intern;
```
canac: Can we not use `asrEligibility` and `salaryRequestEligible` here?

pages/accountLists/[accountListId]/hrTools/salaryCalculator/[calculationId]/index.page.tsx
```diff
@@ -46,7 +55,11 @@ export const SalaryCalculatorEditPage: React.FC = () => {
       <Head>
         <title>{`${appName} | ${t('Salary Calculator')}`}</title>
       </Head>
-      <UserTypeAccess requireStaffAccount>
+      <UserTypeAccess
+        requireStaffAccount
+        requireUserGroups="salaryCalc"
+        effectiveDate={calculation?.effectiveDate}
```
Comment (canac) [inline pages/accountLists/[accountListId]/hrTools/salaryCalculator/[calculationId]/index.page.tsx]: Do we need to pass in the effective date here? Or should we just check whether the user is eligible as of today?

canac (commented):
Here are my thoughts from the first round of review. I'm going to test the UI now!

src/components/Shared/UserTypeAccess/UserTypeAccess.tsx
kegrimes: You are right, thanks for pointing that out!

pages/accountLists/[accountListId]/hrTools/salaryCalculator/[calculationId]/index.page.tsx
Comment (kegrimes) [reply pages/accountLists/[accountListId]/hrTools/salaryCalculator/[calculationId]/index.page.tsx]: Ah okay! I think I was confused on what effectiveDate meant, but if that doesn't matter, then I will gladly get rid of that prop.

src/hooks/useUsStaffGroups.ts
kegrimes: I agree. I changed it around a little where we call `usStaffGroups` twice for ASR and salary calc - this hook no longer has parameters. I don't believe this is too much of a concern since there are only 2 reports we are worrying about here. But, let me know if you think otherwise!

src/components/Shared/UserTypeAccess/UserTypeAccess.tsx
kegrimes: Okay, I changed `hcmLoading` to use `loading && !data` instead of `loading && !skip`. I think this will only be true when there is no data truly present

src/hooks/useHrToolsNavItems.ts
kegrimes: That check will hide salary calculator to users who are US staff and are in an ineligible group. So this would be true for new staff. Is that what you were asking?

src/hooks/useUsStaffGroups.ts
kegrimes: You are correct! We can do it based on eligibility, I forgot I cleaned up the ASR backend check and salary calc already checks for only senior staff

src/hooks/useHrToolsNavItems.ts
canac: I'm just saying that if they aren't US Staff, we won't hide the item. Should it be `hideItem: !usStaff || inSalaryCalcIneligibleGroup`?

src/hooks/useReportsDisabled.ts
kegrimes: The hook is also being used in the useNavPages and useReportNavItems hooks which we need to determine if we need to hide certain reports.

src/lib/apollo/client.ts
```diff
@@ -52,14 +54,16 @@ const makeClient = (apiToken: string) => {
...
-          } else {
+          } else if (!suppressErrors) {
             snackNotifications.error(graphQLError.message);
           }
         });
 
         if (networkError) {
           dispatch('mpdx-api-error');
-          snackNotifications.error(networkError.message);
+          if (!suppressErrors) {
```
canac: `networkError` is when they aren't connected to the internet or DNS fails or something like that. So I think we want to keep the snackbar notification then.

kegrimes: Oops, didn't realize I deleted that, thanks!

src/hooks/useHrToolsNavItems.ts
kegrimes: Oh I see what you are saying! The HR Tools tab should already be covered when a user is not Cru US staff (`hideTab: !showTab`). So, in this file, we only need to check if a user is a US staff member and inside an ineligible group.

kegrimes:
@canac I tested each US group locally and they all are working! Although, there was a couple weird scenarios that we should probably sort out:

1. matthew.pfaff@cru.org is a senior staff who is ineligible for MHA. This might be a data issue because he should be eligible with the new MHA changes in the backend. Although, maybe he didn't meet criteria for the IBS courses. In this case, I feel like we would want MHA to be visible, right? What are your thoughts?

```
"peopleGroupSupportType": "SUPPORTED_RMO",
"userPersonType": "EMPLOYEE_STAFF",
"assignmentStatus": "ACTIVE_PAYROLL_ELIGIBLE",
"assignmentCategory": "FULL_TIME_REGULAR",

"mhaEligibility": "false"
``` 

2. lindsey.williams@athletesinaction.org is a PDS with no staff account, but Savings Fund Transfer and Ministry Partner Reminders are still visible saying there is no staff account. This is the same error as the HCM reports: `NO_STAFF_ACCOUNT`

I only applied the error handling to ASR, salary calculator, and MHA because those are the ones that require US group checks. Should we be hiding all HCM reports if they do not have a staff account id? It seems like maybe it would be beneficial to hide those reports.

src/hooks/useNavPages.tsx
```diff
@@ -121,7 +121,7 @@ export function useNavPages(coachingAccountCount: boolean, isSearch = false) {
                 showInSearchDialog: true,
               })),
               showInNav: true,
-              hideTab: !showTab,
+              hideTab: !!data && !showTab,
```
canac, line 124: I think we need to add `data` to the dependencies now.

src/hooks/useUsStaffGroups.ts
```diff
@@ -0,0 +1,79 @@
...
+  useHcmQuery,
+} from 'src/components/Reports/Shared/HcmData/Hcm.generated';
+import {
+  AssignmentStatusEnum,
+  PeopleGroupSupportTypeEnum,
+  UserPersonTypeEnum,
+} from 'src/graphql/types.generated';
+
+// Hide MHA tab by HR classification only; ignore mhaEligibility (IBS courses are resolvable).
+function mhaIneligible(user: HcmQuery['hcm'][number]['staffInfo']) {
```
canac, line 13: question: Do we need to check the spouse here too? So that ineligible users can still create MHA requests as long as their spouse is eligible?

src/hooks/useUsStaffGroups.ts
```diff
@@ -0,0 +1,79 @@
...
+ * The Salary Calculator and MHA check always runs against the logged-in user.
+ */
+export function useUsStaffGroups(skip?: boolean) {
+  const { data, loading, error } = useHcmQuery({
+    skip,
+    context: { suppressErrors: true },
+  });
+  const [user, spouse] = data?.hcm ?? [];
+
+  const eligibleUser = user?.asrEit?.asrEligibility
```
canac: optional: This is to check whether either user is eligible for ASRs? You could do `[user, spouse].some(person => person?.asrEit?.asrEligibility)`.

src/hooks/useHrToolsNavItems.ts
```diff
@@ -1,18 +1,33 @@
...
+  const { data } = useGetUserQuery();
+  const userType = data?.user.userType;
+  const usStaff = userType === UserTypeEnum.UsStaff;
+
+  const {
+    inAsrIneligibleGroup,
+    inSalaryCalcIneligibleGroup,
+    inMhaIneligibleGroup,
+    hasNoStaffAccount,
+  } = useUsStaffGroups(!usStaff);
```
canac: Now that we suppress errors, I don't think we ever want to skip this query. If we skip it until the user finishes loading, that creates a waterfall for US Staff.

The HCM query is pretty slow, so I think we also want to add `doNotBatch: true` to the context in `useUsStaffGroups`. Normally Apollo sends several GraphQL requests to the server at the same time. The response takes as long as the slowest response. To optimize, we add `doNotBatch` to requests we know are much slower than average so that they can run in parallel and not slow down everything else.

src/hooks/useHrToolsNavItems.ts
canac: So in that case, can we simplify this condition to `hideItem: inSalaryCalcIneligibleGroup`?

src/hooks/useReportNavItems.ts
```diff
@@ -13,10 +15,14 @@ export type NavItems = {
 export function useReportNavItems(): NavItems[] {
   const { t } = useTranslation();
   const { data } = useGetUserQuery();
+  const { reportsDisabled } = useReportsDisabled();
+
   const userType = data?.user.userType;
   const usStaff = userType === UserTypeEnum.UsStaff;
   const globalStaff = userType === UserTypeEnum.GlobalStaff;
 
+  const { hasNoStaffAccount } = useUsStaffGroups(!usStaff);
```
canac: This is a faster way to check for no staff account because it doesn't require an HCM call.

```suggestion
  const hasNoStaffAccount = data && typeof data.user.staffAccountId !== 'string'
```

canac (commented):
This is getting really close! Just had a couple more things. Sorry for all the performance nit-picks. I just want to make sure that this is as fast as possible so that users aren't waiting for menus to be populated.

src/hooks/useUsStaffGroups.ts
kegrimes: I like this! I can maybe the same for MHA

src/hooks/useUsStaffGroups.ts
kegrimes: Yes, I will add it!

src/hooks/useHrToolsNavItems.ts
kegrimes: Yes!

src/hooks/useUsStaffGroups.ts
```diff
@@ -30,21 +30,21 @@ function mhaIneligible(user: HcmQuery['hcm'][number]['staffInfo']) {
...
 
-  const eligibleUser = user?.asrEit?.asrEligibility
-    ? user
-    : spouse?.asrEit?.asrEligibility
-      ? spouse
-      : user;
+  const asrEligible = [user, spouse].some(
+    (person) => person?.asrEit?.asrEligibility,
+  );
+  const mhaEligible =
```
canac: Don't we need to check `mhaIneligible` instead of `mhaEligibility`? What if neither spouse has done their IBS courses?

kegrimes: We still are! If no eligible person is found between the user and spouse, it should be defaulting to the user. Then, we pass that user into `mhaIneligible` to check if the user is eligible, but has not done IBS courses. 

mhaEligible --> Hcm type
```
const mhaEligible =
    [user, spouse].find((person) => person?.mhaEit?.mhaEligibility) ?? user;
```

Check if user exists and pass in user's staff info
```
(!!mhaEligible && mhaIneligible(mhaEligible.staffInfo));
```

So, it will default to user if both have not done IBS courses.

canac: I was just thinking we could simplify to something like this. Is there a reason to check `mhaEligibility`? Or can we just loop through the people and see if any of them are ineligible according to `mhaIneligible`?

```ts
const people = data?.hcm ?? [];
const [user] = people;

// If there is no staff account, we want to hide all reports that require it
const hasNoStaffAccount =
  error?.graphQLErrors.some(
    (graphQLError) => graphQLError.extensions?.code === 'NO_STAFF_ACCOUNT',
  ) ?? false;

const inAsrIneligibleGroup =
  hasNoStaffAccount ||
  people.some((person) => person.asrEit.asrEligibility === false);
const inSalaryCalcIneligibleGroup =
  hasNoStaffAccount || user?.salaryRequestEligible === false;
const inMhaIneligibleGroup =
  hasNoStaffAccount ||
  people.some((person) => mhaIneligible(person.staffInfo));
```

kegrimes: I think that could work but we need to use `every` instead of `some`. If user is ineligible, `inAsrIneligibleGroup` would be `true` and hide the form from the spouse.

canac:
~~Sorry, I probably mistakenly told you to change this, but the HR Tools menu is hidden from non-US Staff. I think they need to be able to access the regular goal calculator, right?~~

~~So I think we need to remove the `hideTab` on HR Tools and instead hide all of the tools (except for the goal calculator) if the person isn't US Staff. (This makes me wonder if we really need to check the `userType` for HR Tools and whether we should just rely on the HCM data. But I don't want to hold this up any further.)~~

EDIT: The goal calculator is intentionally not available to non-US Staff.

src/hooks/useUsStaffGroups.ts
canac: Ah, you're right. We definitely want `every`. And `every` returns true if the array is empty, when we want it to be `false`. Probably not worth using my code lol. But hopefully it shows what I was saying that we can just check `mhaIneligible` instead of checking `mhaEligibility` too.

kegrimes: Honestly it might be cleaner like this because I was already checking if the hcm data was empty. What do you think? It reads better in my opinion!

```
const allAsrIneligible =
    !!people &&
    people.every((person) => person?.asrEit?.asrEligibility === false);
  const allMhaIneligible =
    !!people && people.every((person) => mhaIneligible(person.staffInfo));

  // If there is no staff account, we want to hide all reports that require it
  const hasNoStaffAccount =
    error?.graphQLErrors.some(
      (graphQLError) => graphQLError.extensions?.code === 'NO_STAFF_ACCOUNT',
    ) ?? false;

  const inAsrIneligibleGroup = hasNoStaffAccount || allAsrIneligible;
  const inSalaryCalcIneligibleGroup =
    hasNoStaffAccount || user?.salaryRequestEligible === false;
  const inMhaIneligibleGroup = hasNoStaffAccount || allMhaIneligible;
```

canac: `people` is truthy even if the array is empty. To check for empty array, you probably want `!!people.length && ...`

Otherwise, looks great!

canac (approved):
It looks really good! Thanks for iterating with me on this to get it perfect! I think the end result is really clear and easy to reason about, which means we have more confidence that the logic is correct.

PR #1722, opened 2026-04-16
No-Jira - Disable Continue on Setup Step until values filled by user

zweatshirt:
## Description

- Since future steps depend on these values in SetupStep, it is important to enforce input from the user before they can continue.

## Testing

- Go to `reports/goalCalculator`
- Ensure all user input fields must be filled before continuing the calculator

## Checklist:

### DIFF
--- a/src/components/Reports/PdsGoalCalculator/PdsGoalCalculator.test.tsx
+++ b/src/components/Reports/PdsGoalCalculator/PdsGoalCalculator.test.tsx
@@ -1,7 +1,22 @@
 import React from 'react';
-import { render } from '@testing-library/react';
+import { render, waitFor } from '@testing-library/react';
+import {
+  DesignationSupportSalaryType,
+  DesignationSupportStatus,
+} from 'src/graphql/types.generated';
 import { PdsGoalCalculator } from './PdsGoalCalculator';
 import { PdsGoalCalculatorTestWrapper } from './PdsGoalCalculatorTestWrapper';
+
+const completeSetupMock = {
+  id: 'goal-1',
+  name: 'Test Goal',
+  status: DesignationSupportStatus.FullTime,
+  salaryOrHourly: DesignationSupportSalaryType.Salaried,
+  payRate: 50000,
+  hoursWorkedPerWeek: null,
+  benefits: 1500,
+  geographicLocation: 'Orlando, FL',
+};
 
 describe('PdsGoalCalculator', () => {
   it('renders the setup step by default', () => {
@@ -14,5 +29,42 @@ describe('PdsGoalCalculator', () => {
     expect(
       getByRole('heading', { level: 6, name: 'Calculator Setup' }),
     ).toBeInTheDocument();
+  });
+
+  it('disables Continue on Setup step when a required field is missing', async () => {
+    const { findByRole } = render(
+      <PdsGoalCalculatorTestWrapper
+        calculationMock={{ ...completeSetupMock, name: '' }}
+      >
+        <PdsGoalCalculator />
+      </PdsGoalCalculatorTestWrapper>,
+    );
+
+    const continueButton = await findByRole('button', { name: /continue/i });
+    await waitFor(() => expect(continueButton).toBeDisabled());
+  });
+
+  it('disables Continue on Setup step when geographicLocation is not set', async () => {
+    const { findByRole } = render(
+      <PdsGoalCalculatorTestWrapper
+        calculationMock={{ ...completeSetupMock, geographicLocation: null }}
+      >
+        <PdsGoalCalculator />
+      </PdsGoalCalculatorTestWrapper>,
+    );
+
+    const continueButton = await findByRole('button', { name: /continue/i });
+    await waitFor(() => expect(continueButton).toBeDisabled());
+  });
+
+  it('enables Continue on Setup step when all required fields are filled', async () => {
+    const { findByRole } = render(
+      <PdsGoalCalculatorTestWrapper calculationMock={completeSetupMock}>
+        <PdsGoalCalculator />
+      </PdsGoalCalculatorTestWrapper>,
+    );
+
+    const continueButton = await findByRole('button', { name: /continue/i });
+    await waitFor(() => expect(continueButton).toBeEnabled());
   });
 });
--- a/src/components/Reports/PdsGoalCalculator/PdsGoalCalculator.tsx
+++ b/src/components/Reports/PdsGoalCalculator/PdsGoalCalculator.tsx
@@ -42,7 +42,7 @@ const StepContent: React.FC = () => {
           formTitle={currentStep.title}
           handleNextStep={handleContinue}
           handlePreviousStep={handlePreviousStep}
-          isValid={allValid}
+          disableNext={!allValid}
         />
       )}
     </>
--- a/src/components/Reports/PdsGoalCalculator/PdsGoalCalculator.tsx
+++ b/src/components/Reports/PdsGoalCalculator/PdsGoalCalculator.tsx
@@ -1,6 +1,10 @@
 import React from 'react';
 import { SectionList } from 'src/components/Reports/GoalCalculator/SharedComponents/SectionList';
 import { DirectionButtons } from 'src/components/Reports/Shared/CalculationReports/DirectionButtons/DirectionButtons';
+import {
+  AutosaveForm,
+  useAutosaveForm,
+} from 'src/components/Shared/Autosave/AutosaveForm';
 import { PdsGoalCalculatorStepEnum } from './PdsGoalCalculatorHelper';
 import { ReimbursableExpensesStep } from './ReimbursableExpenses/ReimbursableExpensesStep';
 import { SalaryStep } from './Salary/SalaryStep';
@@ -24,27 +28,38 @@ const CurrentStep: React.FC = () => {
   }
 };
 
-export const PdsGoalCalculator: React.FC = () => {
+const StepContent: React.FC = () => {
   const { currentStep, stepIndex, steps, handleContinue, handlePreviousStep } =
     usePdsGoalCalculator();
-  const sections = currentStep.sections;
-
+  const { allValid } = useAutosaveForm();
   const isLastStep = stepIndex === steps.length - 1;
+
+  return (
+    <>
+      <CurrentStep />
+      {!isLastStep && (
+        <DirectionButtons
+          formTitle={currentStep.title}
+          handleNextStep={handleContinue}
+          handlePreviousStep={handlePreviousStep}
+          isValid={allValid}
+        />
+      )}
+    </>
+  );
+};
+
+export const PdsGoalCalculator: React.FC = () => {
+  const { currentStep } = usePdsGoalCalculator();
+  const sections = currentStep.sections;
 
   return (
     <PdsGoalCalculatorLayout
       sectionListPanel={<SectionList sections={sections} />}
       mainContent={
-        <>
-          <CurrentStep />
-          {!isLastStep && (
-            <DirectionButtons
-              formTitle={currentStep.title}
-              handleNextStep={handleContinue}
-              handlePreviousStep={handlePreviousStep}
-            />
-          )}
-        </>
+        <AutosaveForm>
+          <StepContent />
+        </AutosaveForm>
       }
     />
   );
--- a/src/components/Reports/PdsGoalCalculator/Setup/SetupStep.tsx
+++ b/src/components/Reports/PdsGoalCalculator/Setup/SetupStep.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react';
+import React, { useEffect, useMemo } from 'react';
 import CalculateIcon from '@mui/icons-material/Calculate';
 import {
   Autocomplete,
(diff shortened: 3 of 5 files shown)


src/components/Reports/PdsGoalCalculator/Setup/SetupStep.tsx
```diff
@@ -72,6 +80,20 @@ export const SetupStep: React.FC = () => {
     [goalGeographicConstantMap],
   );
 
+  const autosaveForm = useOptionalAutosaveForm();
+  const geographicLocationValue = calculation?.geographicLocation;
+  useEffect(() => {
```
zweatshirt, line 85: [Important] **Geographic Multiplier UX trap.** The Autocomplete (line 259 in this file) uses `value={calculation?.geographicLocation ?? 'None'}`, which displays `'None'` (a real option in `mpdGoalGeographicConstants`) whenever the underlying value is `null`. After this PR makes the field required, the user sees `'None'` as if selected, but this `useEffect` fires `markInvalid` and Continue is disabled — with no error state on the Autocomplete. Other fields show red helperText via AutosaveTextField, but this one bypasses that path.

Suggested fix:

```tsx
<Autocomplete
  options={locations}
  value={calculation?.geographicLocation ?? null}
  onChange={(_, newValue) => saveField({ geographicLocation: newValue })}
  disabled={!calculation}
  size="small"
  renderInput={(params) => (
    <TextField
      {...params}
      label={t('Geographic Multiplier')}
      required
      error={!!calculation && !calculation.geographicLocation}
      helperText={
        !!calculation && !calculation.geographicLocation
          ? t('Geographic Multiplier is a required field')
          : t('If not applicable, select "None"')
      }
    />
  )}
/>
```

Passing `null` lets MUI render the field empty with floating label. `required` + `error` + `helperText` match the other fields' UX and add `aria-invalid` for screen readers.

src/components/Reports/PdsGoalCalculator/Setup/SetupStep.tsx
```diff
@@ -47,20 +48,27 @@ export const SetupStep: React.FC = () => {
...
           .number()
-          .nullable()
+          .required(t('Hours Worked is a required field'))
           .min(0, t('Hours Worked must be a positive number')),
         benefits: yup
           .number()
-          .nullable()
+          .required(t('Benefits is a required field'))
           .min(0, t('Benefits must be a positive number')),
+        geographicLocation: yup
```
zweatshirt, line 69: [Medium] **Dead schema rule + unprecedented pattern.** The `geographicLocation: yup.string().required(...)` rule you're adding here is never validated against — the runtime check is a manual `useEffect` at line 85 that only tests truthiness. And calling `markValid`/`markInvalid` from a component's useEffect is a new pattern (no other callers in the repo outside `useAutosave.ts`). SalaryCalculator has an `AutosaveAutocomplete` wrapper (`src/components/Reports/SalaryCalculator/Autosave/AutosaveAutocomplete.tsx`) that would be the cleaner precedent.

Options:
- **Cheap:** delete the yup rule with a comment (validation is driven by the effect).
- **Preferred (likely follow-up PR):** introduce `AutosaveAutocomplete` mirroring `AutosaveTextField` so validity flows through `useAutoSave`. That also addresses the IMP-1 UX symptom because the wrapper would consume `error`/`helperText` from the schema.

Flagged by Architecture and Standards agents.

src/components/Reports/PdsGoalCalculator/Setup/SetupStep.tsx
```diff
@@ -47,20 +48,27 @@ export const SetupStep: React.FC = () => {
...
+        status: yup
+          .string()
+          .required(t('Employment Status is a required field')),
+        salaryOrHourly: yup
+          .string()
+          .required(t('Pay Type is a required field')),
         payRate: yup
           .number()
-          .nullable()
+          .required(t('Pay Rate is a required field'))
```
zweatshirt, line 59: [Suggestion] `yup.number().required().min(0)` accepts literal `0`. For a staff salary, `$0/yr` isn't meaningful. Consider `.moreThan(0, t('Pay Rate must be greater than zero'))` for `payRate` and also for `hoursWorkedPerWeek` (line 63). `benefits` can stay `.min(0)` since zero benefits is legitimate. Out of scope for the Continue-button work — file for follow-up if desired.

src/components/Reports/PdsGoalCalculator/Setup/SetupStep.tsx
```diff
@@ -47,20 +48,27 @@ export const SetupStep: React.FC = () => {
     () =>
       yup.object({
         name: yup.string().required(t('Goal Name is a required field')),
-        status: yup.string().nullable(),
-        salaryOrHourly: yup.string().nullable(),
+        status: yup
```
zweatshirt, line 51: [Suggestion] Verify server-side parity: the new `required()` rules are client-only enforcement. If `DesignationSupportCalculationUpdateInput` (or the REST proxy) doesn't also reject missing values, a stale client or alternate entry path could still submit incomplete data. Not blocking — the same fields were `.nullable()` before this PR, so no new risk — but good to confirm.

src/components/Reports/Shared/CalculationReports/DirectionButtons/DirectionButtons.tsx
```diff
@@ -141,6 +143,7 @@ export const DirectionButtons: React.FC<DirectionButtonsProps> = ({
             variant="contained"
             color="primary"
             onClick={overrideNext ?? handleNextStep}
+            disabled={disableNext}
```
zweatshirt, line 146: [Medium] **No feedback on why Continue is disabled.** `disabled={disableNext}` alone gives users no explanation once they've noticed they can't advance. After IMP-1 is fixed, field-level errors will explain most cases, but a tooltip on the button is still the clearest affordance.

Suggested:

```tsx
<Tooltip title={disableNext ? t('Complete all required fields to continue') : ''}>
  <span>
    <Button
      variant="contained"
      color="primary"
      onClick={overrideNext ?? handleNextStep}
      disabled={disableNext}
      aria-describedby={disableNext ? 'continue-disabled-reason' : undefined}
    >
      {buttonTitle ?? t('Continue')}
      <ChevronRight sx={{ ml: 1 }} />
    </Button>
  </span>
</Tooltip>
```

(The `<span>` wrapper is required because disabled buttons don't fire pointer events — MUI tooltip gotcha.)

Note: SalaryCalculator's ContinueButton has the same gap — fixing it here improves both.

src/components/Reports/Shared/CalculationReports/DirectionButtons/DirectionButtons.tsx
```diff
@@ -21,6 +21,7 @@ interface DirectionButtonsProps {
   additionalApproval?: boolean;
   splitAsr?: boolean;
   disableSubmit?: boolean;
+  disableNext?: boolean;
```
zweatshirt, line 24: [Medium] **Prop bloat.** This shared component now carries 24 props covering three distinct use cases (MHA formik submit, ASR formik submit, PDS autosave continue). Continue uses `disableNext`, Submit uses `submitCount ? !isValid : false`, and `disableSubmit` affects the SubmitModal. The next contributor who wants PDS to also show validity on a submission step, or ASR to disable Continue while invalid, will have to add yet another combination.

Follow-up refactor idea (not blocking this PR): consolidate to a single `disabled` prop computed by callers. The Formik callers already have `isValid`/`submitCount` locally. Shrinks the contract from 'know the branch + pick right prop' to 'compute disabled once.'

zweatshirt (commented):
## Multi-Agent Code Review

**Verdict:** APPROVED_WITH_SUGGESTIONS
**Risk:** MEDIUM (5/10)
**Mode:** standard (5 agents: Architecture, Testing, Standards, UX, Financial Reporting)

### Summary

The PR correctly wires `AutosaveForm` + `disableNext` to disable the Continue button when Setup step fields are empty. The shared-component change (`DirectionButtons`) was carefully scoped — the new `disableNext` prop defaults to `undefined`, so existing callers (MHA StepOne/Two/Three, AdditionalSalaryRequest) are unaffected. Verified across all 5 callsites.

### What landed well
- Mirrors the SalaryCalculator pattern (`AutosaveForm` + `StepContent` extraction).
- New `disableNext` prop is purposefully separate from `isValid` to avoid disabling Continue in the ASR RequestPage flow.
- Conditional fields (`hoursWorkedPerWeek`, `benefits`) re-enable Continue correctly on unmount via the existing `useAutosave` cleanup.
- i18n: all new required-error messages wrapped in `t()`.
- Financial Reporting agent: **no financial calculation code in this PR** — all checks pass or N/A.

### Top findings

1. **[Important 8.0]** `SetupStep.tsx:259` — The Geographic Multiplier Autocomplete displays `'None'` when the underlying value is `null` (`value={calculation?.geographicLocation ?? 'None'}`), which now looks like a filled field even though the required check fails. Users see `'None'` selected but Continue disabled with no error state on the control. See the inline comment on the new useEffect for the suggested fix.

2. **[Medium 7.0]** `DirectionButtons.tsx:146` — Disabled Continue has no tooltip/aria explanation. Once the Autocomplete error state is fixed, field-level messages cover most cases, but a `<Tooltip>` on the button itself still helps.

3. **[Medium 6.5]** `SetupStep.tsx:69` — The new `geographicLocation` yup rule is effectively dead (the runtime check is a manual `useEffect` testing truthiness). Either delete the rule or introduce `AutosaveAutocomplete` so validation flows through `useAutoSave` consistently.

### Important-tier note

> WARNING: 1 Important finding (severity 8.0) — strongly recommended fix that doesn't block merge but can't be dismissed via `/dismiss`. Address in this PR or create a follow-up ticket.

### Agents

| Agent | Findings |
|---|---|
| Architecture | 3 medium + 2 suggestions (useEffect pattern, prop bloat, duplicate hook call) |
| Testing | 2 medium (Submit-branch leak test, schema-invalid value test) + 2 suggestions |
| Standards | 2 concerns (useEffect pattern, server-side validation parity WARN); checklist otherwise PASS |
| UX | 1 important + 2 medium + 3 suggestions (Autocomplete, tooltip, loading flicker) |
| Financial Reporting | No financial calculation code in this PR |

No agent challenged another's findings. No severity spreads ≥4 points. All 5 changed files received coverage from 2+ agents — no gap review needed.

src/components/Reports/PdsGoalCalculator/PdsGoalCalculator.tsx
```diff
@@ -24,27 +28,38 @@ const CurrentStep: React.FC = () => {
...
 
+  return (
+    <>
+      <CurrentStep />
+      {!isLastStep && (
+        <DirectionButtons
+          formTitle={currentStep.title}
+          handleNextStep={handleContinue}
+          handlePreviousStep={handlePreviousStep}
+          disableNext={!allValid}
```
zweatshirt, line 45: [Medium] **Loading-state flicker.** `useAutosaveForm()` starts with `invalidFields = []` → `allValid = true`. Child fields then mount, run their validity effects, and `allValid` flips to `false`. On first render (or slow Apollo load) Continue briefly shows enabled, then disabled.

Suggested: include the loading gate in the disabled check.

```tsx
const { currentStep, stepIndex, steps, handleContinue, handlePreviousStep, calculation } =
  usePdsGoalCalculator();
...
disableNext={!calculation || !allValid}
```

The Autocomplete at `SetupStep.tsx:263` already gates itself on `!calculation` — the Continue button should follow suit.

src/components/Reports/PdsGoalCalculator/PdsGoalCalculator.test.tsx
```diff
@@ -15,4 +30,41 @@ describe('PdsGoalCalculator', () => {
       getByRole('heading', { level: 6, name: 'Calculator Setup' }),
     ).toBeInTheDocument();
   });
+
+  it('disables Continue on Setup step when a required field is missing', async () => {
```
zweatshirt, line 34: [Medium] **Missing edge-case coverage.** The three new tests cover empty/null values, but not 'field is filled but schema-invalid' — e.g., `payRate: -5` violates `.min(0)`. Without this, a future change that silently drops `.min(0)` from the schema won't be caught at the feature level.

```tsx
it('disables Continue on Setup step when a required field is invalid', async () => {
    <PdsGoalCalculatorTestWrapper
      calculationMock={{ ...completeSetupMock, payRate: -5 }}
    >
      <PdsGoalCalculator />
  );
  const continueButton = await findByRole('button', { name: /continue/i });
  await waitFor(() => expect(continueButton).toBeDisabled());
});
```

src/components/Reports/Shared/CalculationReports/DirectionButtons/DirectionButtons.test.tsx
```diff
@@ -114,6 +117,24 @@ describe('DirectionButtons', () => {
     expect(await findByRole('button', { name: title })).toBeInTheDocument();
   });
 
+  it('disables Continue when disableNext is true', async () => {
```
zweatshirt, line 120: [Medium] **Missing regression test for Submit branch.** All three new `disableNext` tests exercise the Continue branch (default `isSubmission=false`). The code is correct — `disabled={disableNext}` lives only in the `!isSubmission` branch — but a test would pin that contract. A refactor that accidentally applies `disabled={disableNext}` to the Submit branch wouldn't be caught here.

```tsx
it('does not disable Submit even when disableNext is true', async () => {
    <TestComponent isSubmission={true} disableNext={true} />,
  );
  expect(await findByRole('button', { name: /submit/i })).toBeEnabled();
});
```

zweatshirt:
Closing in favor of https://github.com/CruGlobal/mpdx-react/pull/1756

PR #1728, opened 2026-04-17
MPDX-9416 - Display paycheck process date associated with effective date to provide context for users

zweatshirt:
## Description

- Implements a hook for displaying the paycheck process date associated with the given effective date. This is to provide users context. It might be overkill. 

[Jira ticket](https://jira.cru.org/browse/MPDX-9370)

## Testing

- Go to the SalaryCalculator
- Continue or create a salary request.
- Determine if the change of displaying the process date is appropriate or not.

## Checklist:

### DIFF
--- a/src/components/Reports/SalaryCalculator/Shared/EffectiveDateNote.tsx
+++ b/src/components/Reports/SalaryCalculator/Shared/EffectiveDateNote.tsx
@@ -1,0 +1,27 @@
+import React from 'react';
+import { Typography } from '@mui/material';
+import { useTranslation } from 'react-i18next';
+import { useEffectivePaycheckDate } from './useEffectivePaycheckDate';
+
+export const EffectiveDateNote: React.FC = () => {
+  const { t } = useTranslation();
+  const paycheckDate = useEffectivePaycheckDate();
+
+  if (!paycheckDate) {
+    return null;
+  }
+
+  return (
+    <Typography
+      variant="body2"
+      color="textSecondary"
+      role="note"
+      component="span"
+    >
+      {t('Values shown reflect the paycheck dated {{date}}.', {
+        date: paycheckDate,
+        interpolation: { escapeValue: false },
+      })}
+    </Typography>
+  );
+};
--- a/src/components/Reports/SalaryCalculator/Shared/EffectiveDateNote.tsx
+++ b/src/components/Reports/SalaryCalculator/Shared/EffectiveDateNote.tsx
@@ -20,7 +20,6 @@ export const EffectiveDateNote: React.FC = () => {
     >
       {t('Values shown reflect the paycheck dated {{date}}.', {
         date: paycheckDate,
-        interpolation: { escapeValue: false },
       })}
     </Typography>
   );
--- a/src/components/Reports/SalaryCalculator/Shared/useEffectivePaycheckDate.test.tsx
+++ b/src/components/Reports/SalaryCalculator/Shared/useEffectivePaycheckDate.test.tsx
@@ -1,0 +1,98 @@
+import React from 'react';
+import { renderHook, waitFor } from '@testing-library/react';
+import { SalaryRequestStatusEnum } from 'src/graphql/types.generated';
+import {
+  SalaryCalculatorTestWrapper,
+  SalaryCalculatorTestWrapperProps,
+} from '../SalaryCalculatorTestWrapper';
+import { useEffectivePaycheckDate } from './useEffectivePaycheckDate';
+
+const wrapperWith = (
+  props: Partial<SalaryCalculatorTestWrapperProps> = {},
+): React.FC<{ children: React.ReactNode }> => {
+  const Wrapper: React.FC<{ children: React.ReactNode }> = ({ children }) => (
+    <SalaryCalculatorTestWrapper {...props}>
+      {children}
+    </SalaryCalculatorTestWrapper>
+  );
+  return Wrapper;
+};
+
+const mutationSpy = jest.fn();
+
+describe('useEffectivePaycheckDate', () => {
+  it('returns the formatted regularProcessDate when the effective date matches a payroll date', async () => {
+    const { result } = renderHook(() => useEffectivePaycheckDate(), {
+      wrapper: wrapperWith({
+        salaryRequestMock: {
+          status: SalaryRequestStatusEnum.InProgress,
+          effectiveDate: '2026-06-01',
+        },
+        payrollDates: [
+          { startDate: '2026-06-01', regularProcessDate: '2026-06-10' },
+          { startDate: '2026-12-01', regularProcessDate: '2026-12-10' },
+        ],
+      }),
+    });
+
+    await waitFor(() => expect(result.current).toBe('6/10/2026'));
+  });
+
+  it('returns null when the effective date is not in the payroll dates list', async () => {
+    const { result } = renderHook(() => useEffectivePaycheckDate(), {
+      wrapper: wrapperWith({
+        salaryRequestMock: {
+          status: SalaryRequestStatusEnum.InProgress,
+          effectiveDate: '2026-06-01',
+        },
+        payrollDates: [
+          { startDate: '2026-12-01', regularProcessDate: '2026-12-10' },
+        ],
+        onCall: mutationSpy,
+      }),
+    });
+
+    await waitFor(() =>
+      expect(mutationSpy).toHaveGraphqlOperation('PayrollDates'),
+    );
+    expect(result.current).toBeNull();
+  });
+
+  it('returns null when payrollDates is empty', async () => {
+    const { result } = renderHook(() => useEffectivePaycheckDate(), {
+      wrapper: wrapperWith({
+        salaryRequestMock: {
+          status: SalaryRequestStatusEnum.InProgress,
+          effectiveDate: '2026-06-01',
+        },
+        payrollDates: [],
+        onCall: mutationSpy,
+      }),
+    });
+
+    await waitFor(() =>
+      expect(mutationSpy).toHaveGraphqlOperation('PayrollDates'),
+    );
+    expect(result.current).toBeNull();
+  });
+
+  it('returns null when the calculation has no effective date', async () => {
+    const { result } = renderHook(() => useEffectivePaycheckDate(), {
+      wrapper: wrapperWith({
+        salaryRequestMock: {
+          status: SalaryRequestStatusEnum.InProgress,
+          effectiveDate: null,
+        },
+        payrollDates: [
+          { startDate: '2026-06-01', regularProcessDate: '2026-06-10' },
+        ],
+        onCall: mutationSpy,
+      }),
+    });
+
+    await waitFor(() =>
+      expect(mutationSpy).toHaveGraphqlOperation('PayrollDates'),
+    );
+    expect(result.current).toBeNull();
+  });
+});
--- a/src/components/Reports/SalaryCalculator/Shared/useEffectivePaycheckDate.ts
+++ b/src/components/Reports/SalaryCalculator/Shared/useEffectivePaycheckDate.ts
@@ -1,0 +1,29 @@
+import { useMemo } from 'react';
+import { DateTime } from 'luxon';
+import { useLocale } from 'src/hooks/useLocale';
+import { dateFormatShort } from 'src/lib/intlFormat';
+import { usePayrollDatesQuery } from '../EffectiveDateStep/PayrollDates.generated';
+import { useSalaryCalculator } from '../SalaryCalculatorContext/SalaryCalculatorContext';
+
+export const useEffectivePaycheckDate = (): string | null => {
+  const { calculation } = useSalaryCalculator();
+  const { data } = usePayrollDatesQuery();
+  const locale = useLocale();
+
+  return useMemo(() => {
+    if (!calculation?.effectiveDate || !data?.payrollDates) {
+      return null;
+    }
+
+    const payrollDate = data.payrollDates.find(
+      (entry) => entry.startDate === calculation.effectiveDate,
+    );
+
+    return payrollDate
+      ? dateFormatShort(
+          DateTime.fromISO(payrollDate.regularProcessDate),
+          locale,
+        )
+      : null;
+  }, [calculation?.effectiveDate, data?.payrollDates, locale]);
+};
--- a/src/components/Reports/SalaryCalculator/Shared/useEffectivePaycheckDate.ts
+++ b/src/components/Reports/SalaryCalculator/Shared/useEffectivePaycheckDate.ts
@@ -1,29 +1,12 @@
-import { useMemo } from 'react';
-import { DateTime } from 'luxon';
-import { useLocale } from 'src/hooks/useLocale';
-import { dateFormatShort } from 'src/lib/intlFormat';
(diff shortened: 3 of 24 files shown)


src/components/Reports/SalaryCalculator/Shared/useEffectivePaycheckDate.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+    });
+
+    await waitFor(() =>
+      expect(mutationSpy).toHaveGraphqlOperation('PayrollDates'),
+    );
+    expect(result.current).toBeNull();
+  });
+
+  it('returns null when the calculation has no effective date', async () => {
+    const mutationSpy = jest.fn();
```
zweatshirt: [Medium] **Shadowed `mutationSpy`.** This local `const mutationSpy = jest.fn()` shadows the module-scoped spy at line 21. The three earlier tests share the module-scoped spy without a `mockClear` between them; this fourth test uses a fresh one. The inconsistency is a test-isolation smell and an `no-shadow` lint hit.

Recommended fix — drop the shadow and add a `beforeEach` reset:

```diff
 describe('useEffectivePaycheckDate', () => {
+  beforeEach(() => {
+    mutationSpy.mockClear();
+  });

   it('returns null when the calculation has no effective date', async () => {
-    const mutationSpy = jest.fn();
     const { result } = renderHook(() => useEffectivePaycheckDate(), {
```

_Note: your local branch already includes a fix (`1db211b88 Remove redundant mutationSpy declaration ...`) that is not yet pushed — pushing will resolve this._

Flagged by: Testing (4.0), Standards (5.0).

zweatshirt (commented):

**Mode:** standard · **Agents:** Architecture, Testing, Standards, UX, Financial Reporting

### Verdict
No blockers. One Medium-tier finding and a handful of lower-severity suggestions. Refactor is behavior-preserving; tests pass; i18n/TypeScript/hook-placement conventions all honored.

### Risk Assessment
- **Score:** 8/10 (HIGH) — driven by file count (8 non-test files), though non-test line delta is ~70 lines of mechanical `subheader={<EffectiveDateNote />}` wiring.
- **Reviewer level:** mid-level/senior.
- Friday deploy caution.

### Dependency Impact
- `EffectiveDateNote` — 5 expected dependents
- `useEffectivePaycheckDate` — 2 dependents
- No breaking changes; no renamed/removed exports.

### Summary by Agent
| Agent | Critical | High | Important | Medium | Suggestions |
|-------|:-:|:-:|:-:|:-:|:-:|
| Architecture | 0 | 0 | 0 | 0 | 3 |
| Testing & Quality | 0 | 0 | 0 | 1 | 4 |
| Standards | 0 | 0 | 0 | 1 | 0 |
| UX | 0 | 0 | 0 | 2 | 8 |
| Financial Reporting | 0 | 0 | 0 | 0 | 3 |
| **Total (deduped)** | **0** | **0** | **0** | **1** | **~16** |

### Highlights
- **Medium:** shadowed `mutationSpy` at `useEffectivePaycheckDate.test.tsx:80` — agreed by Testing and Standards agents. Note: your local branch already has an unpushed fix (`1db211b88 Remove redundant mutationSpy declaration in useEffectivePaycheckDate tests`); pushing it will resolve this finding.
- **Medium:** `usePayrollDatesQuery` errors are swallowed (`error` not destructured); note silently vanishes on API failure.
- **Medium:** layout-shift / pop-in as the note mounts on five cards once the query resolves — consider a Skeleton or accept the shift deliberately.
- **Medium:** `role="note"` is the first use of this ARIA role in `src/`. Valid, but screen-reader support varies — consider decoupling test discovery via `data-testid`.
- **Suggestions:** drop redundant `interpolation: { escapeValue: false }`; fix `{...props}` spread ordering to protect `onCall={mutationSpy}`; consider a degraded fallback when `effectiveDate` is set but no `payrollDates` entry matches; weigh visual repetition of the same note on 5 stacked cards vs a page-level banner.

### No Blockers / No Dismissed Findings
Review posted against PR HEAD `53df1b694`. CI must still pass before merging.

src/components/Reports/SalaryCalculator/Shared/useEffectivePaycheckDate.ts
```diff
@@ -0,0 +1,26 @@
+import { useMemo } from 'react';
+import { DateTime } from 'luxon';
+import { useLocale } from 'src/hooks/useLocale';
+import { dateFormatShort } from 'src/lib/intlFormat';
+import { usePayrollDatesQuery } from '../EffectiveDateStep/PayrollDates.generated';
+import { useSalaryCalculator } from '../SalaryCalculatorContext/SalaryCalculatorContext';
+
+export const useEffectivePaycheckDate = (): string | null => {
+  const { calculation } = useSalaryCalculator();
+  const { data } = usePayrollDatesQuery();
```
zweatshirt, line 10: [Medium] **Silent error state on `usePayrollDatesQuery`.** Only `data` is destructured; `error` is ignored. If the query fails, the note silently never appears — users lose the "values apply to paycheck dated X" context across five cards with zero indication that something went wrong.

Because the note is auxiliary context (not the primary data for any card), a full `<Alert>` would be overkill. Reasonable options:
- Document the silent-degrade choice inline so future maintainers don't assume it's an oversight.
- Surface query errors centrally in the `SalaryCalculatorContext` provider so every consumer benefits.

Flagged by: UX.

src/components/Reports/SalaryCalculator/Shared/EffectiveDateNote.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { Typography } from '@mui/material';
+import { useTranslation } from 'react-i18next';
+import { useEffectivePaycheckDate } from './useEffectivePaycheckDate';
+
+export const EffectiveDateNote: React.FC = () => {
+  const { t } = useTranslation();
+  const paycheckDate = useEffectivePaycheckDate();
+
+  if (!paycheckDate) {
```
zweatshirt, line 10: [Medium] **Layout-shift on async pop-in.** `EffectiveDateNote` returns `null` while `usePayrollDatesQuery` is loading, then mounts ~20px of `body2` into five CardHeader subheader slots simultaneously when the query resolves. Minor visually but happens across five cards at once.

Options:
- Render a `Skeleton variant="text" width={260}` while loading to reserve space.
- Reserve space with a `&nbsp;` placeholder.
- Accept the shift deliberately (lowest-effort) — but make it a conscious decision.

Flagged by: UX.

src/components/Reports/SalaryCalculator/Shared/EffectiveDateNote.tsx
```diff
@@ -0,0 +1,27 @@
...
+    return null;
+  }
+
+  return (
+    <Typography
+      variant="body2"
+      color="textSecondary"
+      role="note"
```
zweatshirt, line 18: [Medium] **First use of `role="note"` in `src/`.** Valid WAI-ARIA 1.2, semantically appropriate for supplementary info. Two caveats:

1. Screen-reader announcement is inconsistent (NVDA announces, JAWS/VoiceOver often don't).
2. The tests rely on `findByRole('note')`, coupling the a11y role to test discovery. If you ever change the role for UX reasons, five test files break.

Consider keeping the role only if its a11y intent is deliberate, and using `data-testid` for test queries.

Flagged by: UX (5.0), Architecture (3.5).

src/components/Reports/SalaryCalculator/Shared/EffectiveDateNote.tsx
```diff
@@ -0,0 +1,27 @@
...
+      component="span"
+    >
+      {t('Values shown reflect the paycheck dated {{date}}.', {
+        date: paycheckDate,
+        interpolation: { escapeValue: false },
```
zweatshirt, line 23: [Suggestion] **`interpolation: { escapeValue: false }` is unnecessary here.** `paycheckDate` is the output of `Intl.DateTimeFormat` (digits, slashes, dashes only) — nothing for i18next to HTML-escape. The option is typically used only when interpolating HTML into a `<Trans>`.

```diff
-      {t('Values shown reflect the paycheck dated {{date}}.', {
-        date: paycheckDate,
-        interpolation: { escapeValue: false },
-      })}
+      {t('Values shown reflect the paycheck dated {{date}}.', { date: paycheckDate })}
```

Flagged by: Architecture (3.0), UX (3.0).

src/components/Reports/SalaryCalculator/Shared/EffectiveDateNote.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { Typography } from '@mui/material';
+import { useTranslation } from 'react-i18next';
+import { useEffectivePaycheckDate } from './useEffectivePaycheckDate';
+
+export const EffectiveDateNote: React.FC = () => {
+  const { t } = useTranslation();
+  const paycheckDate = useEffectivePaycheckDate();
+
+  if (!paycheckDate) {
+    return null;
```
zweatshirt, line 11: [Suggestion] **Consider a degraded fallback when no payroll date matches.** If `calculation.effectiveDate` is set but no `payrollDates` entry matches (backend drift, stale calculation, or an effective date outside the known pay periods), the note silently disappears across five cards. Users then see financial values with no indication of which paycheck they apply to.

One option:

```tsx
const { calculation } = useSalaryCalculator();
const paycheckDate = useEffectivePaycheckDate();
if (!calculation?.effectiveDate) return null;
return (
  <Typography /* ... */>
    {paycheckDate
      ? t('Values shown reflect the paycheck dated {{date}}.', { date: paycheckDate })
      : t('Values shown reflect the selected effective date.')}
  </Typography>
);
```

Flagged by: Financial Reporting (4.5), UX (4.0).

src/components/Reports/SalaryCalculator/YourInformation/MaxAllowableSection/MaxAllowableSection.test.tsx
```diff
@@ -25,18 +25,35 @@ const defaultSalaryRequestMock: DeepPartial<
...
-const TestComponent: React.FC<
-  Pick<SalaryCalculatorTestWrapperProps, 'salaryRequestMock'>
-> = ({ salaryRequestMock }) => (
+const TestComponent: React.FC<SalaryCalculatorTestWrapperProps> = ({
+  salaryRequestMock,
+  ...props
+}) => (
   <SalaryCalculatorTestWrapper
-    salaryRequestMock={merge({}, defaultSalaryRequestMock, salaryRequestMock)}
     onCall={mutationSpy}
```
zweatshirt: [Suggestion] **Spread order lets callers accidentally override `mutationSpy`.** `{...props}` is spread AFTER `onCall={mutationSpy}`, so any caller passing `onCall` via props would silently replace the default spy — breaking the `mutationSpy` assertions elsewhere in this file. Move `onCall={mutationSpy}` after the spread to preserve the intent.

```diff
 <SalaryCalculatorTestWrapper
-  onCall={mutationSpy}
   {...props}
+  onCall={mutationSpy}
   salaryRequestMock={merge({}, defaultSalaryRequestMock, salaryRequestMock)}
 >
```

No current caller exploits this, so latent footgun rather than a bug today.

Flagged by: Testing.

src/components/Reports/SalaryCalculator/Shared/EffectiveDateNote.tsx
zweatshirt: **Retraction:** this suggestion was wrong — I applied the fix locally and the tests failed. Without `interpolation: { escapeValue: false }`, i18next HTML-escapes the `/` in `6/10/2026` into `&#x2F;`, producing `6&#x2F;10&#x2F;2026` in the rendered output. The option is necessary here. Please disregard this comment.

zweatshirt:
@canac Mostly curious what your opinion is of this in the UI. I think it's kind of overkill, but it seemed better than adding it inline everywhere. I could also just add it to the top of the page. This is the "paycheck as of" UAT request.

src/components/Reports/SalaryCalculator/Shared/useEffectivePaycheckDate.ts
```diff
@@ -0,0 +1,29 @@
+import { useMemo } from 'react';
+import { DateTime } from 'luxon';
+import { useLocale } from 'src/hooks/useLocale';
+import { dateFormatShort } from 'src/lib/intlFormat';
+import { usePayrollDatesQuery } from '../EffectiveDateStep/PayrollDates.generated';
+import { useSalaryCalculator } from '../SalaryCalculatorContext/SalaryCalculatorContext';
+
+export const useEffectivePaycheckDate = (): string | null => {
```
canac, line 4: question: Would it make sense to use `useEffectiveDateOptions` instead of this hook? I know that would change the formatting of the date, so we could consider adding another field to the items returned by `useEffectiveDateOptions`.

src/components/Reports/SalaryCalculator/Shared/EffectiveDateNote.tsx
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  return (
+    <Typography
+      variant="body2"
+      color="textSecondary"
+      role="note"
+      component="span"
+    >
+      {t('Values shown reflect the paycheck dated {{date}}.', {
```
canac, line 21: I did some research, and by default i18next escapes values like / and <. But since React already escapes them for us, that leads to the double-escaping issue. Can you add `interpolation: { escapeValue: false }` to the global `src/lib/i18n.ts` config so that we don't have to worry about this anymore? Just add a comment saying that we're relying on React to handle escaping, which it does. We're completely safe as long as we don't use `dangerouslySetInnerHTML`. You can remove any other occurrences of `interpolation: { escapeValue: false }` in the codebase too.

src/components/Reports/SalaryCalculator/Shared/useEffectivePaycheckDate.ts
```diff
@@ -0,0 +1,29 @@
...
+  const locale = useLocale();
+
+  return useMemo(() => {
+    if (!calculation?.effectiveDate || !data?.payrollDates) {
+      return null;
+    }
+
+    const payrollDate = data.payrollDates.find(
+      (entry) => entry.startDate === calculation.effectiveDate,
+    );
```
canac: optional: This works without an early return.

```suggestion
    const payrollDate = data?.payrollDates.find(
      (entry) => entry.startDate === calculation?.effectiveDate,
    );
```

canac (approved):
Looks really good, thanks for adding it!

src/components/Reports/SalaryCalculator/Shared/useEffectivePaycheckDate.ts
zweatshirt: Nice, good idea

src/components/Reports/SalaryCalculator/Shared/useEffectivePaycheckDate.ts
zweatshirt: I think that would work but we would still need to find by the effective date right since it returns an array? I'm thinking we could make `useEffectivePaycheckDate` a simple wrapper over useEffectiveDateOptions, and do what you mentioned (return an additional `shortLabel` property from `useEffectiveDateOptions`)

canac: Good point, maybe it's not even worth it.

PR #1737, opened 2026-04-21
[MPDX-9369] - Wire up Hours Per Week calculator to back-end with autosave

wjames111:
## Description

- Wire up the Hours Per Week calculator grid to persist data via `designationSupportHoursItems` GraphQL mutations
- Add `CreateDesignationSupportHoursItem` and `UpdateDesignationSupportHoursItem` mutations to `PdsGoalCalculations.graphql`
- Initialize the grid from server data (`calculation.designationSupportHoursItems`) when available, falling back to hardcoded defaults for new calculations
- Autosave individual hours items on cell edit (create for new/default items, update for existing)
- Autosave the recalculated `averageHoursPerWeek` to the calculation on every edit
- Add an "Apply to Hours Worked" button so users can explicitly push the average into the `hoursWorkedPerWeek` field, decoupling the calculator from the hours worked input
- Update tests to use `PdsGoalCalculatorTestWrapper` with proper GraphQL mocks

Related Jira ticket: [MPDX-9369](https://jira.cru.org/browse/MPDX-9369)

## Testing

- Go to a PDS Goal Calculator
- Open the Hours Per Week calculator from the Setup step (click the calculator icon next to "Hours Worked")
- Edit hours/weeks values in the grid and verify the average updates
- Reload the page and reopen the calculator — verify the edited values persisted
- Add a new entry, edit it, and verify it persists after reload
- Click "Apply to Hours Worked" and verify the Hours Worked field updates
- Verify that editing the Hours Worked field directly still works independently of the calculator

## Checklist:

### DIFF
--- a/src/components/Reports/PdsGoalCalculator/GoalsList/PdsGoalCalculations.graphql
+++ b/src/components/Reports/PdsGoalCalculator/GoalsList/PdsGoalCalculations.graphql
@@ -18,6 +18,15 @@ fragment PdsGoalCalculationFields on DesignationSupportCalculation {
   conferenceRetreatCosts
   ministryTravelMeals
   otherAnnualReimbursements
+  designationSupportHoursItems {
+    id
+    hoursPerWeek
+    numberOfWeeks
+    name
+    label
+    position
+    predefined
+  }
 }
 
 query PdsGoalCalculations($after: String) {
@@ -60,6 +69,38 @@ mutation UpdatePdsGoalCalculation(
   updateDesignationSupportCalculation(input: { attributes: $attributes }) {
     designationSupportCalculation {
       ...PdsGoalCalculationFields
+    }
+  }
+}
+
+mutation CreateDesignationSupportHoursItem(
+  $attributes: DesignationSupportHoursItemCreateInput!
+) {
+  createDesignationSupportHoursItem(input: { attributes: $attributes }) {
+    designationSupportHoursItem {
+      id
+      hoursPerWeek
+      numberOfWeeks
+      name
+      label
+      position
+      predefined
+    }
+  }
+}
+
+mutation UpdateDesignationSupportHoursItem(
+  $attributes: DesignationSupportHoursItemUpdateInput!
+) {
+  updateDesignationSupportHoursItem(input: { attributes: $attributes }) {
+    designationSupportHoursItem {
+      id
+      hoursPerWeek
+      numberOfWeeks
+      name
+      label
+      position
+      predefined
     }
   }
 }
--- a/src/components/Reports/PdsGoalCalculator/GoalsList/PdsGoalCalculations.graphql
+++ b/src/components/Reports/PdsGoalCalculator/GoalsList/PdsGoalCalculations.graphql
@@ -104,3 +104,10 @@ mutation UpdateDesignationSupportHoursItem(
     }
   }
 }
+
+mutation DeleteDesignationSupportHoursItem($id: ID!) {
+  deleteDesignationSupportHoursItem(input: { id: $id }) {
+    id
+  }
+}
+
--- a/src/components/Reports/PdsGoalCalculator/GoalsList/PdsGoalCalculations.graphql
+++ b/src/components/Reports/PdsGoalCalculator/GoalsList/PdsGoalCalculations.graphql
@@ -110,4 +110,3 @@ mutation DeleteDesignationSupportHoursItem($id: ID!) {
     id
   }
 }
-
--- a/src/components/Reports/PdsGoalCalculator/PdsGoalCalculatorTestWrapper.tsx
+++ b/src/components/Reports/PdsGoalCalculator/PdsGoalCalculatorTestWrapper.tsx
@@ -80,6 +80,7 @@ const calculationDefault = gqlMock<
       geographicLocation: null,
     },
   },
+  variables: { id: 'goal-1' },
 }).designationSupportCalculation;
 
 const hcmUserDefault = gqlMock<HcmUserQuery, HcmUserQueryVariables>(
--- a/src/components/Reports/PdsGoalCalculator/PdsGoalCalculatorTestWrapper.tsx
+++ b/src/components/Reports/PdsGoalCalculator/PdsGoalCalculatorTestWrapper.tsx
@@ -137,7 +137,8 @@ export const PdsGoalCalculatorTestWrapper: React.FC<
               ...(calculationMock
                 ? {
                     PdsGoalCalculation: {
-                      designationSupportCalculation: calculationMock,
+                      designationSupportCalculation:
+                        calculationMock as PdsGoalCalculationQuery['designationSupportCalculation'],
                     },
                   }
                 : {}),
--- a/src/components/Reports/PdsGoalCalculator/PdsGoalCalculatorTestWrapper.tsx
+++ b/src/components/Reports/PdsGoalCalculator/PdsGoalCalculatorTestWrapper.tsx
@@ -16,7 +16,9 @@ import {
 import { GoalCalculatorConstantsQuery } from 'src/hooks/goalCalculatorConstants.generated';
 import theme from 'src/theme';
 import {
+  PdsGoalCalculationDocument,
   PdsGoalCalculationQuery,
+  PdsGoalCalculationQueryVariables,
   PdsGoalCalculationsDocument,
   PdsGoalCalculationsQuery,
   PdsGoalCalculationsQueryVariables,
@@ -61,6 +63,24 @@ export type PdsGoalCalculationsMock = DeepPartial<
 export type PdsGoalCalculationMock = DeepPartial<
   PdsGoalCalculationQuery['designationSupportCalculation']
 >;
+
+const calculationDefault = gqlMock<
+  PdsGoalCalculationQuery,
+  PdsGoalCalculationQueryVariables
+>(PdsGoalCalculationDocument, {
+  mocks: {
+    designationSupportCalculation: {
+      id: 'goal-1',
+      name: 'Test Goal',
+      status: DesignationSupportStatus.FullTime,
+      salaryOrHourly: DesignationSupportSalaryType.Salaried,
+      payRate: 50000,
+      hoursWorkedPerWeek: null,
+      benefits: 1500,
+      geographicLocation: null,
+    },
+  },
+}).designationSupportCalculation;
 
 const hcmUserDefault = gqlMock<HcmUserQuery, HcmUserQueryVariables>(
   HcmUserDocument,
@@ -134,14 +154,13 @@ export const PdsGoalCalculatorTestWrapper: React.FC<
                   calculationsMock,
                 ),
               },
-              ...(calculationMock
-                ? {
-                    PdsGoalCalculation: {
-                      designationSupportCalculation:
-                        calculationMock as PdsGoalCalculationQuery['designationSupportCalculation'],
-                    },
-                  }
-                : {}),
+              PdsGoalCalculation: {
+                designationSupportCalculation: merge(
+                  {},
+                  calculationDefault,
+                  calculationMock,
+                ),
+              },
               HcmUser: {
                 hcm:
                   hcmUserMock === null
--- a/src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.test.tsx
+++ b/src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.test.tsx
@@ -173,9 +173,7 @@ describe('HoursPerWeekGrid', () => {
     const regularRow = (await findByText('Regular Week')).closest(
       '[role="row"]',
     );
-    const hoursCell = regularRow?.querySelector(
-      '[data-field="hoursPerWeek"]',
-    );
+    const hoursCell = regularRow?.querySelector('[data-field="hoursPerWeek"]');
     userEvent.dblClick(hoursCell!);
 
     await waitFor(() => {
@@ -219,9 +217,7 @@ describe('HoursPerWeekGrid', () => {
     const regularRow = (await findByText('Regular Week')).closest(
       '[role="row"]',
     );
-    const hoursCell = regularRow?.querySelector(
-      '[data-field="hoursPerWeek"]',
-    );
+    const hoursCell = regularRow?.querySelector('[data-field="hoursPerWeek"]');
     userEvent.dblClick(hoursCell!);
 
     await waitFor(() => {
@@ -233,14 +229,11 @@ describe('HoursPerWeekGrid', () => {
 
     // Average = 20 hrs * 48 wks / 48 wks = 20.0
     await waitFor(() =>
-      expect(mutationSpy).toHaveGraphqlOperation(
-        'UpdatePdsGoalCalculation',
-        {
-          attributes: {
-            averageHoursPerWeek: 20,
-          },
+      expect(mutationSpy).toHaveGraphqlOperation('UpdatePdsGoalCalculation', {
+        attributes: {
+          averageHoursPerWeek: 20,
         },
-      ),
+      }),
     );
   });
 
@@ -271,9 +264,7 @@ describe('HoursPerWeekGrid', () => {
 
     // Warning should disappear and Apply should be enabled
     await waitFor(() => {
-      expect(
-        queryByRole('alert'),
-      ).not.toBeInTheDocument();
+      expect(queryByRole('alert')).not.toBeInTheDocument();
     });
 
     const applyButton = getByRole('button', { name: 'Apply to Hours Worked' });
--- a/src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.test.tsx
+++ b/src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.test.tsx
@@ -65,7 +65,7 @@ describe('HoursPerWeekGrid', () => {
 
     expect(await findByText('Hours Per Week Calculator')).toBeInTheDocument();
     await waitForDataToLoad();
-    expect(getByText('Regular Week')).toBeInTheDocument();
+    expect(await findByText('Regular Week')).toBeInTheDocument();
     expect(getByText('Travel')).toBeInTheDocument();
     expect(getByText('Unpaid Vacation')).toBeInTheDocument();
     expect(getByText('Total')).toBeInTheDocument();
@@ -183,18 +183,16 @@ describe('HoursPerWeekGrid', () => {
     });
     userEvent.tab();
 
-    // Default entries have "default-" IDs since calculation loads after
-    // initial render, so cell edits trigger create (not update)
     await waitFor(() =>
       expect(mutationSpy).toHaveGraphqlOperation(
-        'CreateDesignationSupportHoursItem',
+        'UpdateDesignationSupportHoursItem',
         {
           attributes: {
+            id: 'item-regular',
             designationSupportCalculationId: 'goal-1',
             label: 'Regular Week',
             hoursPerWeek: 20,
             numberOfWeeks: 48,
-            position: 0,
           },
         },
       ),
--- a/src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.test.tsx
+++ b/src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.test.tsx
@@ -119,9 +119,32 @@ describe('HoursPerWeekGrid', () => {
     expect(queryByText('Apply to Hours Worked')).not.toBeInTheDocument();
   });
 
-  it('calls onApply with the rounded average when Apply button is clicked', async () => {
+  it('shows warning and disables Apply when weeks do not add up to 52', async () => {
     const onApply = jest.fn();
-    const { findByText, getByText, getByDisplayValue } = render(
+    const { findByText, getByText, getByRole } = render(
+      <PdsGoalCalculatorTestWrapper
+        calculationMock={defaultCalculationMock}
+        onCall={mutationSpy}
+      >
+        <HoursPerWeekGrid onApply={onApply} />
+      </PdsGoalCalculatorTestWrapper>,
+    );
+
+    await waitForDataToLoad();
+    await findByText('Regular Week');
+
+    // Default entries have 48 weeks total, 4 remaining
+    expect(
+      getByText(/Weeks must add up to 52. 4 week\(s\) remaining./),
+    ).toBeInTheDocument();
+
+    const applyButton = getByRole('button', { name: 'Apply to Hours Worked' });
+    expect(applyButton).toBeDisabled();
+  });
+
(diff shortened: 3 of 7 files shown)


wjames111 (commented):
## Multi-Agent Code Review — PR #1737

**Risk Level:** HIGH (8/10) | **Agents:** 6 | **Mode:** Standard with debate rounds

### Verdict: BLOCKERS FOUND

**2 blockers must be resolved before merge:**
1. `deleteEntry` is local-only — no server delete mutation exists; deleted rows reappear on reload
2. `saveHoursItem` sends `position: -1` on every row edit due to `indexOf` referential equality bug

---

### Important Issues (Severity 7.0-7.9)

These are strongly recommended fixes that don't block merge but can't be dismissed:

- **No test verification of mutation variables** (7.5) — `mutationSpy` checks operation names but never asserts `CreateDesignationSupportHoursItem` or `UpdateDesignationSupportHoursItem` variables. The indexOf bug is invisible to the test suite.
- **No error feedback when mutations fail** (7.5) — `saveHoursItem`, `addEntry`, and `saveField` calls have no error handling. Failed mutations leave local state diverged from server.
- **No test for autosave of `averageHoursPerWeek`** (7.0) — The core feature of this PR (autosave on cell edit) is not verified in tests.

### Medium Priority (Severity 5.0-6.9)

- Dual autosave fields (`averageHoursPerWeek` vs `hoursWorkedPerWeek`) may diverge — verify this is intentional (6.5)
- No loading state while server data loads (6.5)
- Panel doesn't close after Apply — no success feedback to user (6.0)
- Default entries not persisted until first edit (5.5)
- Type assertion `as` in test wrapper bypasses type checking (5.5)
- Stale `entries.length` in `addEntry` under rapid clicks (5.0)

### Suggestions

- Extract rounding `Math.round(x * 10) / 10` into a shared helper — it appears in 3 places
- `refetchQueries` on mutations is redundant since the grid uses local `useState`, not Apollo cache
- Consider disabling "Apply" button during mutations via `isMutating`

### Review Summary

| Architecture | 0 | 1 | 0 | 3 | High |
| Testing | 0 | 0 | 2 | 3 | High |
| Standards | 1 | 0 | 0 | 1 | High |
| Data Integrity | 1 | 0 | 0 | 3 | High |
| UX | 0 | 0 | 1 | 4 | High |
| Financial | 0 | 0 | 0 | 2 | High |
| **Total** | **1** | **1** | **3** | **4** | |

src/components/Reports/PdsGoalCalculator/GoalsList/PdsGoalCalculations.graphql
```diff
@@ -63,3 +72,35 @@ mutation UpdatePdsGoalCalculation(
...
+      id
+      hoursPerWeek
+      numberOfWeeks
+      name
+      label
+      position
+      predefined
+    }
+  }
+}
```
wjames111, line 106: [Critical] **Missing `DeleteDesignationSupportHoursItem` mutation — deleted rows reappear on reload**

Severity: 9.5/10 — Consensus from 6/6 agents

`deleteEntry` in `HoursPerWeekGrid.tsx` (line 226) only removes the row from local React state (`setEntries`). There is no server-side delete mutation. When any subsequent `refetchQueries: ['PdsGoalCalculation']` fires (from editing another row) or on page reload, "deleted" rows reappear from the server.

Additionally, `deleteEntry` does not call `saveField` to autosave the recalculated average, so the server-side `averageHoursPerWeek` diverges from the displayed value after deletion.

**Required action:** Add a `DeleteDesignationSupportHoursItem` mutation here (alongside Create and Update), wire it into `deleteEntry`, and autosave the recalculated average after deletion.

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
```diff
@@ -83,39 +120,108 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
...
+      if (entry.id.startsWith('temp-') || entry.id.startsWith('default-')) {
+        const result = await trackMutation(
+          createHoursItem({
+            variables: {
+              attributes: {
+                designationSupportCalculationId: calculation.id,
+                label: entry.label,
+                hoursPerWeek: entry.hoursPerWeek,
+                numberOfWeeks: entry.weeks,
+                position: currentEntries.indexOf(entry),
```
wjames111: [High] **`indexOf` referential equality bug — sends `position: -1` on every row edit**

Severity: 8.0/10 — Flagged by Architecture, confirmed by 3 other agents in debate

`currentEntries.indexOf(entry)` uses `===` referential equality. The `entry` parameter is always a freshly constructed object (built in `processRowUpdate` at lines 242-249), so it is never the same reference as any element in `currentEntries`. `indexOf` always returns `-1`, meaning every create and update mutation sends `position: -1` to the server.

This same bug exists on line 164 (update branch).

**Fix:**
```diff
- position: currentEntries.indexOf(entry),
+ position: currentEntries.findIndex((e) => e.id === entry.id),
```
Apply in both the create branch (line 140) and update branch (line 164).

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
```diff
@@ -127,15 +233,50 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
...
+        predefined: newRow.predefined as boolean,
+      };
+
+      updateEntry(newRow.id as string, {
+        label: updatedEntry.label,
+        hoursPerWeek: updatedEntry.hoursPerWeek,
+        weeks: updatedEntry.weeks,
+      });
+
+      saveHoursItem(updatedEntry, entries);
```
wjames111: [Important] **No error handling for mutation failures**

Neither `saveHoursItem` (called here without `await`) nor `saveField` (line 273) have error handling. If either mutation fails:
- The user receives no feedback
- Local state has already been optimistically updated and will be out of sync with the server
- If `saveHoursItem` fails but `saveField` succeeds, the server has an `averageHoursPerWeek` that doesn't match actual hours items

Consider: (1) `await`ing `saveHoursItem` before calling `saveField`, and (2) adding `.catch()` handlers that surface errors via snackbar/`enqueueSnackbar`.

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.test.tsx
```diff
@@ -1,103 +1,215 @@
...
+      label: 'Unpaid Vacation',
+      hoursPerWeek: 0,
+      numberOfWeeks: 0,
+      name: 'vacation',
+      position: 2,
+      predefined: true,
+    },
+  ],
+};
+
```
wjames111, line 42: [Important] **`mutationSpy` never verifies create/update mutation variables**

The spy is used only to detect when `PdsGoalCalculation` query fires (for data loading). None of the tests assert that `CreateDesignationSupportHoursItem` or `UpdateDesignationSupportHoursItem` are called with correct variables (`position`, `designationSupportCalculationId`, `hoursPerWeek`, `numberOfWeeks`, `label`).

This means the `indexOf` bug (which sends `position: -1` on every edit) is invisible to the test suite.

Also missing: a test verifying `UpdatePdsGoalCalculation` is called with the autosaved `averageHoursPerWeek` after cell edits — the core feature of this PR.

Suggested additions:
```ts
await waitFor(() =>
  expect(mutationSpy).toHaveGraphqlOperation('UpdateDesignationSupportHoursItem', {
    attributes: expect.objectContaining({
      hoursPerWeek: 20,
      numberOfWeeks: 48,
    }),
  }),
);
```

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
```diff
@@ -83,39 +120,108 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
...
+      result.data?.createDesignationSupportHoursItem
+        ?.designationSupportHoursItem;
+    if (created) {
+      setEntries((prev) =>
+        prev.map((e) => (e.id === tempId ? { ...e, id: created.id } : e)),
+      );
+    }
+  }, [t, calculation, createHoursItem, trackMutation, entries.length]);
 
   const deleteEntry = useCallback((id: string | number) => {
```
wjames111: [Medium] **Delete does not autosave the recalculated average**

When a row is deleted, `averageHoursPerWeek` recalculates correctly in the UI (via `useMemo`), but `saveField({ averageHoursPerWeek: ... })` is never called here — it's only called inside `processRowUpdate`. The server-side average remains stale until the next row edit.

Also note: once a `DeleteDesignationSupportHoursItem` mutation is added (per the critical blocker), this callback should call `saveField` to persist the new average.

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
```diff
@@ -28,43 +44,73 @@ export interface HoursPerWeekEntry {
...
+  predefined: boolean;
+  position: number;
 }
 
 interface HoursPerWeekGridProps {
-  onAverageHoursChange?: (averageHoursPerWeek: number) => void;
+  onApply?: (averageHoursPerWeek: number) => void;
 }
 
+const MAX_TOTAL_WEEKS = 52;
```
zweatshirt, line 61: Probably not really necessary because there will always be 52 weeks in a year, but we can make this into an API constant if needed

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
```diff
@@ -28,43 +44,73 @@ export interface HoursPerWeekEntry {
...
+  const [updateHoursItem] = useUpdateDesignationSupportHoursItemMutation();
+  const [deleteHoursItem] = useDeleteDesignationSupportHoursItemMutation();
+  const saveField = useSaveField();
+
+  const [entries, setEntries] = useState<HoursPerWeekEntry[]>(() => {
+    const items = calculation?.designationSupportHoursItems;
+    if (items && items.length > 0) {
+      return items
+        .slice()
+        .sort((a, b) => (a.position ?? 0) - (b.position ?? 0))
```
zweatshirt: Optional: 
```suggestion
        .sort((hoursItemA, hoursItemB) => (hoursItemA.position ?? 0) - (hoursItemB.position ?? 0))
```

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
```diff
@@ -28,43 +44,73 @@ export interface HoursPerWeekEntry {
...
         canDelete: false,
+        predefined: true,
+        position: 2,
       },
-    ],
-    [t],
-  );
-  const [entries, setEntries] = useState<HoursPerWeekEntry[]>(defaultEntries);
+    ];
+  });
```
zweatshirt: I'm thinking this fallback isn't actually necessary since these are created API side. What do you think? Is the API not working for this well? I can change it if needed.

zweatshirt (commented):
Wasn't able to get through review before the end of the day so I'll pick this back up tomorrow

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
```diff
@@ -83,67 +129,239 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
...
+        hoursPerWeek: updatedEntry.hoursPerWeek,
+        weeks: updatedEntry.weeks,
+      });
+
+      await saveHoursItem(updatedEntry, entries);
+
+      // Autosave the recalculated average to the calculation
+      const updatedEntries = entries.map((e) =>
+        e.id === updatedEntry.id ? updatedEntry : e,
+      );
```
zweatshirt: Can we use full variable names here and below as well?

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
```diff
@@ -225,8 +443,9 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
 
   return (
     <Box>
-      <Typography variant="h6" sx={{ mb: 2 }}>
-        {t('Hours Per Week Calculator')}
+      <Typography variant="h6">{t('Hours Per Week Calculator')}</Typography>
+      <Typography variant="body2" color="text.secondary" sx={{ mb: 2 }}>
+        {t('Weeks must add up to 52 to reflect a full calendar year.')}
       </Typography>
```
zweatshirt, line 437: Optional: I might remove this or change the description here to something more general about the calculator, since we show the same information in the Alert below
```suggestion
```

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
```diff
@@ -83,67 +129,239 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
...
+          await trackMutation(
+            deleteHoursItem({
+              variables: { id: entryId },
+              refetchQueries: ['PdsGoalCalculation'],
+            }),
+          );
+        }
+        saveField({
+          averageHoursPerWeek: Math.round(newAverage * 10) / 10,
+        });
```
zweatshirt, line 282: Question: Do we want to round only for display, and instead send the raw average to the API?

src/components/Reports/PdsGoalCalculator/PdsGoalCalculatorTestWrapper.tsx
```diff
@@ -137,7 +137,8 @@ export const PdsGoalCalculatorTestWrapper: React.FC<
               ...(calculationMock
                 ? {
                     PdsGoalCalculation: {
-                      designationSupportCalculation: calculationMock,
+                      designationSupportCalculation:
+                        calculationMock as PdsGoalCalculationQuery['designationSupportCalculation'],
```
zweatshirt: Thinking it may be best to avoid typecasting here. We can create a default with gqlMock and then merge the default and the passed in `calculationMock` prop. This is how we managed this in SalaryCalculator

zweatshirt (commented):
More items. It's looking great Will!

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
wjames111: nice this is clearer!

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
wjames111: Yeah I mean its always going to be 52, but that number is used in quite a few places, seems reasonable to make it a variable, unless I'm missing something.

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
wjames111: Your right we shouldn't need these anymore.

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
wjames111: True yeah, open to suggestions. the Alert shows conditionally though, and I feel like it should be upfront information. Especially considering the calculator literally won't let you go over 52.

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
wjames111: Yeah that seems like a good idea.

src/components/Reports/PdsGoalCalculator/GoalsList/PdsGoalCalculations.graphql
```diff
@@ -63,3 +72,41 @@ mutation UpdatePdsGoalCalculation(
...
+) {
+  createDesignationSupportHoursItem(input: { attributes: $attributes }) {
+    designationSupportHoursItem {
+      id
+      hoursPerWeek
+      numberOfWeeks
+      name
+      label
+      position
+      predefined
```
zweatshirt, line 87: What do you think of making this into a fragment since it's duplicated multiple times in this file?

src/components/Reports/PdsGoalCalculator/Shared/Autosave/useSaveField.ts
```diff
@@ -20,28 +24,35 @@ export const useSaveField = () => {
...
+                  ...calculation,
+                  ...attributes,
+                },
+              },
+            },
+          }),
+        );
+      } catch {
+        enqueueSnackbar(t('Failed to save changes. Please try again.'), {
+          variant: 'error',
```
zweatshirt, line 51: Question: Do we strictly need the Snackbar here? I think `apollo/client.ts` will handle this for us, and this would produce a double snackbar. Then we can also remove the try catch. This seems to be the pattern across the app

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
```diff
@@ -83,67 +115,239 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
...
+  const weeksRemaining = MAX_TOTAL_WEEKS - totalWeeks;
+
+  const saveHoursItem = useCallback(
+    async (entry: HoursPerWeekEntry, currentEntries: HoursPerWeekEntry[]) => {
+      if (!calculation) {
+        return;
+      }
+
+      try {
+        if (entry.id.startsWith('default-')) {
```
zweatshirt, line 127: Will entry ids ever now start with 'default-'? I don't think they will now that we have gotten rid of the hardcoded fallbacks but I could be wrong

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
```diff
@@ -83,67 +115,239 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
...
+            }),
+          );
+        }
+        saveField({
+          averageHoursPerWeek: newAverage,
+        });
+      } catch {
+        setEntries(previousEntries);
+        enqueueSnackbar(t('Failed to delete entry. Please try again.'), {
+          variant: 'error',
```
zweatshirt, line 286: Same issue here. I don't think these custom snackbars are needed. Let me know if that's not the case

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
```diff
@@ -83,67 +115,239 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
...
+          ?.designationSupportHoursItem;
+      if (created) {
+        setEntries((prev) =>
+          prev.map((e) => (e.id === tempId ? { ...e, id: created.id } : e)),
+        );
+      }
+    } catch {
+      setEntries((prev) => prev.filter((e) => e.id !== tempId));
+      enqueueSnackbar(t('Failed to add entry. Please try again.'), {
+        variant: 'error',
```
zweatshirt, line 241: Same here

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
zweatshirt, line 1: There is a lot of logic and lines in this file in general making it harder to read. Wondering if it's worth making a hook to separate view logic and business logic concerns for readability? Can be a follow up PR

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
```diff
@@ -83,67 +115,239 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
...
+              variables: {
+                attributes: {
+                  designationSupportCalculationId: calculation.id,
+                  label: entry.label,
+                  hoursPerWeek: entry.hoursPerWeek,
+                  numberOfWeeks: entry.weeks,
+                  position: currentEntries.findIndex((e) => e.id === entry.id),
+                },
+              },
+              refetchQueries: ['PdsGoalCalculation'],
```
zweatshirt, line 139: Do we want to do an optimistic response here instead? Not critical now, but an improvement

zweatshirt (dismissed):
Sorry to hold you up Will, I still think there are some improvements we can make. Thanks for all the work on this!! Feel free to disregard the ones you disagree with.

canac (approved):
Blind approval to unblock Katelyn. ✅ I'm trusting Zach's previous review.

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
wjames111: Your right I think this whole block can be removed now. The problem it was originally meant to solve was keeping users from editing an entry on a DesignationSupportHoursItem before it had been created yet.

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
wjames111: Good suggestion, not sure if it's needed here though. We're already tracking local state with `updateEntry` so it should be pretty immediate feedback for the user.

src/components/Reports/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
wjames111: Yeah I agree there's a lot going on now.

PR #1756, opened 2026-05-01
[MPDX-9481] - Disable Continue button in PDS Goal Calculator until required fields are filled

wjames111:
## Description

- Disable the Continue button on the PDS Goal Calculator Setup step until all required fields are filled
- Make Setup form fields (`status`, `salaryOrHourly`, `payRate`, `hoursWorkedPerWeek`, `benefits`) required via Yup validation (previously nullable)
- Wire `AutosaveForm` validation state into the `DirectionButtons` component via a new `disableNext` prop
- Add localized validation messages for each newly required field
- Related ticket: [MPDX-9481](https://jira.cru.org/browse/MPDX-9481)

## Testing

- Go to the PDS Goal Calculator (HR Tools → PDS Goal Calculator)
- On the Setup step, clear a required field (e.g. Pay Rate or Benefits)
- Check that the Continue button becomes disabled
- Fill in all required fields
- Check that the Continue button becomes enabled and navigates to the next step

## Checklist:


src/components/Shared/Autosave/AutosaveForm.tsx
```diff
@@ -40,30 +40,39 @@ interface AutosaveFormProps {
...
 
   // Mark a field as invalid
   const markInvalid = useCallback((fieldName: string) => {
+    setHasRegisteredFields(true);
     setInvalidFields((prev) =>
       prev.includes(fieldName) ? prev : [...prev, fieldName],
     );
   }, []);
 
+  const allValid = hasRegisteredFields ? invalidFields.length === 0 : false;
```
wjames111, line 67: [Medium] **AutosaveForm `allValid` default change may affect other consumers** (6.5/10)

`allValid` previously started as `true` (empty `invalidFields` = valid). It now starts as `false` until at least one field calls `markValid`/`markInvalid`. This is correct for PDS Goal Calculator, but `AutosaveForm` has 5 dependents including **SalaryCalculator** and **StepNavigation**.

If those consumers gate navigation on `allValid` and their fields don't register promptly on mount, their Continue/Submit buttons could be permanently disabled.

**Recommended:** Verify SalaryCalculator and StepNavigation flows still work correctly with this change. Run their test suites and manually test.

src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.test.tsx
```diff
@@ -39,6 +39,58 @@ const fullTimeHourlyMock = {
 };
 
 describe('SetupStep', () => {
+  it('disables fields while calculation data is loading', async () => {
+    const { findByRole } = render(
+      <PdsGoalCalculatorTestWrapper calculationMock={undefined as never}>
```
wjames111: [Medium] **`undefined as never` type cast bypasses TypeScript** (6.5/10)

`calculationMock={undefined as never}` silences the type checker to simulate a loading state. This is the only usage of `as never` in the codebase — it's not an established pattern.

If the wrapper's prop type tightens further, this cast silently suppresses a type error that would flag a real problem.

**Recommended:** Widen `PdsGoalCalculatorTestWrapper`'s `calculationMock` prop type to explicitly accept `undefined`, then change this to `calculationMock={undefined}`.

src/components/Shared/Autosave/AutosaveForm.tsx
```diff
@@ -40,30 +40,39 @@ interface AutosaveFormProps {
...
  * fields are invalid.
+ *
+ * allValid starts as false until at least one field registers its status, preventing a brief
+ * flash of enabled submit/continue buttons before fields validate.
  */
 export const AutosaveForm: React.FC<AutosaveFormProps> = ({ children }) => {
   // Track the list of fields that are currently invalid
   const [invalidFields, setInvalidFields] = useState<string[]>([]);
+  // Track whether any field has reported its status yet
+  const [hasRegisteredFields, setHasRegisteredFields] = useState(false);
```
wjames111, line 51: [Medium] **Missing unit test for `hasRegisteredFields` initial state** (6.0/10)

The `hasRegisteredFields` flag is load-bearing — it prevents a flash of enabled Continue button before fields validate. However, there's no direct unit test in `AutosaveForm.test.tsx` verifying that `allValid` is `false` before any field registers.

Integration tests cover this indirectly, but a direct unit test would guard against regressions.

**Suggested test:**
```tsx
it('allValid is false before any field registers', () => {
  const StatusOnly: React.FC = () => {
    const { allValid } = useAutosaveForm();
    return <h1>{allValid ? 'Valid' : 'Invalid'}</h1>;
  };
  const { getByRole } = render(
    <AutosaveForm><StatusOnly /></AutosaveForm>,
  );
  expect(getByRole('heading')).toHaveTextContent('Invalid');
});
```

src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.tsx
```diff
@@ -47,20 +47,32 @@ export const SetupStep: React.FC = () => {
...
           .min(0, t('Pay Rate must be a positive number')),
-        hoursWorkedPerWeek: yup
-          .number()
-          .nullable()
-          .min(0, t('Hours Worked must be a positive number')),
-        benefits: yup
-          .number()
-          .nullable()
-          .min(0, t('Benefits must be a positive number')),
+        hoursWorkedPerWeek: yup.number().when('salaryOrHourly', {
```
wjames111, line 60: [Medium] **No test for conditional validation field-switch scenario** (5.5/10)

The `yup.when` rules make `hoursWorkedPerWeek` required only when not Salaried, and `benefits` required only when not Part-time. No test exercises switching employment type (e.g., Hourly → Salaried) and verifying the validation/button state adapts correctly.

The first PdsGoalCalculator test (Salaried with `hoursWorkedPerWeek: null`) implicitly validates the Salaried case, which is adequate but indirect.

src/components/HrTools/Shared/CalculationReports/DirectionButtons/DirectionButtons.tsx
```diff
@@ -137,14 +139,25 @@ export const DirectionButtons: React.FC<DirectionButtonsProps> = ({
...
+            title={
+              disableNext
+                ? t('Complete all required fields to continue')
+                : ''
+            }
           >
-            {buttonTitle ?? t('Continue')}
-            <ChevronRight sx={{ ml: 1 }} />
-          </Button>
+            <span>
```
wjames111: [Medium] **Tooltip not keyboard-accessible on disabled button** (5.0/10)

The `<span>` wrapper correctly intercepts hover events for the tooltip on the disabled button (standard MUI pattern). However, it has no `tabIndex`, so keyboard-only users cannot focus it and never see "Complete all required fields to continue." This affects WCAG 2.1 SC 1.4.13 (Content on Hover or Focus).

**Suggested fix:**
```tsx
// Before
<span>
// After
<span tabIndex={disableNext ? 0 : undefined}>
```

Alternatively, consider `aria-disabled="true"` with a click guard instead of native `disabled`, which keeps the button in tab order.

src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.tsx
wjames111, line 60: [Medium] **`yup.when` receives incomplete context from `useAutoSave`** (5.0/10)

`useAutoSave` calls `schema.validateSyncAt(fieldName, { [fieldName]: value })` with only the single field's value. The `yup.when('salaryOrHourly', ...)` condition receives `undefined` for the sibling field, so `undefined !== 'Salaried'` is always `true` — the `then` branch always executes.

This works today only because conditional rendering hides the field when it shouldn't be required. If someone later shows the field as disabled instead of hidden, validation will silently break.

**Recommended:** Add a code comment documenting this coupling between conditional rendering and the `yup.when` validation logic.

wjames111 (commented):
## Multi-Agent Code Review Report

**Agents:** 4 specialized reviewers (Architecture, Testing, Standards, UX) with debate rounds
**Mode:** Standard
**Risk Score:** 7/10 (HIGH)

### Verdict: APPROVED WITH SUGGESTIONS

No blockers found. 6 medium-priority improvements recommended (severity 5.0-6.9).

### Dependency Impact

**HIGH IMPACT:** `AutosaveForm.tsx` — 5 dependents. `allValid` now defaults to `false` (was `true`). Verify SalaryCalculator and StepNavigation work correctly with this change.

### Standards Compliance

All checklist items PASS: named exports, i18n coverage, TypeScript types, Formik+Yup, colocated tests, no debug output.

### Review Summary

| Agent | Critical | High | Medium | Suggestions | Confidence |
|-------|----------|------|--------|-------------|------------|
| Architecture | 0 | 0 | 2 | 2 | High |
| Testing | 0 | 0 | 3 | 1 | High |
| UX | 0 | 0 | 1 | 2 | High |
| **Total** | **0** | **0** | **6** | **4** | |

---

**SalaryCalculator's StepNavigation has no tooltip on disabled Continue/Submit** (5.5/10) — This PR establishes a good tooltip pattern for disabled buttons in `DirectionButtons`. The SalaryCalculator's equivalent buttons lack this guidance. Consider addressing in a follow-up PR.

zweatshirt:
Do we also want to disable the navigation from the sidebar icons as well? We can do that in a separate PR though. This is looking great. Going through the code now.

src/components/HrTools/PdsGoalCalculator/PdsGoalCalculator.test.tsx
```diff
@@ -15,4 +16,43 @@ describe('PdsGoalCalculator', () => {
...
+
+  it('enables the Continue button when all required Setup fields are filled', async () => {
+    const { findByRole } = render(
+      <PdsGoalCalculatorTestWrapper>
+        <PdsGoalCalculator />
+      </PdsGoalCalculatorTestWrapper>,
+    );
+
+    const continueButton = await findByRole('button', { name: /continue/i });
+    await waitFor(() => expect(continueButton).not.toBeDisabled());
```
zweatshirt, line 32: We can probably remove the continueButton variable, a little win
```suggestion
    await waitFor(() => expect(await findByRole('button', { name: /continue/i })).not.toBeDisabled());
```

src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.tsx
```diff
@@ -47,20 +47,32 @@ export const SetupStep: React.FC = () => {
...
+          is: (val: string) => val !== DesignationSupportSalaryType.Salaried,
+          then: (s) =>
+            s
+              .required(t('Hours Worked is a required field'))
+              .min(0, t('Hours Worked must be a positive number')),
```
zweatshirt: Now may be a good time to change the minimums on these to be greater than $0 (that is, require a positive amount). What do you think?

src/components/HrTools/PdsGoalCalculator/Shared/Autosave/AutosaveTextField.tsx
```diff
@@ -32,6 +32,7 @@ export const AutosaveTextField: React.FC<AutosaveTextFieldProps> = ({
       {...fieldProps}
       {...props}
       disabled={fieldProps.disabled || props.disabled}
+      helperText={fieldProps.helperText ?? props.helperText}
```
zweatshirt, line 35: Do we want to add a test for this?

src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.test.tsx
```diff
@@ -38,7 +38,61 @@ const fullTimeHourlyMock = {
...
+      </PdsGoalCalculatorTestWrapper>,
+    );
+
+    const goalNameInput = await findByRole('textbox', { name: 'Goal Name' });
+    await waitFor(() => expect(goalNameInput).toHaveValue('Test Goal'));
+
+    userEvent.clear(goalNameInput);
+
+    const payRateInput = await findByRole('spinbutton', { name: 'Pay Rate' });
+    fireEvent.change(payRateInput, { target: { value: '' } });
```
zweatshirt: From what I understand, we usually avoid using `fireEvent` in this codebase.

src/components/HrTools/PdsGoalCalculator/PdsGoalCalculator.test.tsx
```diff
@@ -15,4 +16,43 @@ describe('PdsGoalCalculator', () => {
...
+    const goalNameInput = await findByRole('textbox', { name: 'Goal Name' });
+    await waitFor(() => expect(goalNameInput).not.toBeDisabled());
+
+    const continueButton = await findByRole('button', { name: /continue/i });
+    expect(continueButton).toBeDisabled();
+
+    userEvent.type(goalNameInput, 'My Goal');
+
+    await waitFor(() => expect(continueButton).not.toBeDisabled());
+  });
```
zweatshirt, line 134:
```suggestion
    expect(continueButton).not.toBeDisabled();
```

src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.test.tsx
zweatshirt, line 1: Optional Claude suggestion: SetupStep.test.tsx switching tests assert visibility but not that Continue re-enables when the now-hidden field was the only invalid one. Add an end-to-end test from PdsGoalCalculator for that boundary.

zweatshirt (changes requested):
Thanks Will, I have some small change suggestions but this is looking good! As long as there was no regression for the regular salary calculator due to the change to `allValid`'s default, this is great! I was hoping we could get the patch coverage up a bit before merging. I think a couple of the test suggestions I mentioned should help with that.

src/components/HrTools/PdsGoalCalculator/PdsGoalCalculator.test.tsx
wjames111: I think it would have to be:
```
 await waitFor(async () => expect(await findByRole('button', { name: /continue/i })).not.toBeDisabled());
```
While it works I think I'd opt for the intermediate variable here.

zweatshirt: I'm definitely ok with that, it's a small preference so no issues there either way we go

zweatshirt:
I didn't realize you were right on the cusp of the test coverage job passing. I think that's definitely fine, I wasn't paying attention I'm sorry about that. I'll approve this!

zweatshirt (approved):

PR #1760, opened 2026-05-04
MPDX-9530 - PDS and MPD Goal Calculators: Shared Goal Card

zweatshirt:
## Description

Reduce duplication by making GoalCard a shared component between MPD Goal Calculator and PDS Goal Calculator
[Jira ticket](https://jira.cru.org/browse/MPDX-9530)

## Testing

- Go to PDS Goal Calculator and MPD Goal Calculator and test that View, Delete behavior still behave, and that the cards stylistically look the same as before creating the shared component.

## Checklist:


zweatshirt (commented):
## Multi-Agent Code Review — PR #1760

**Verdict: APPROVED WITH SUGGESTIONS** — No blockers, no important issues. One legitimate Medium-priority i18n concern (orphaned translation key) and a few stylistic / follow-up suggestions.

**Risk:** 6/10 (MEDIUM) — refactor of a top-level Shared component affecting two feature trees.

**Agents:** Architecture, Testing, Standards, UX, Data Integrity (5 specialized reviewers)

### Severity-spread investigation

During debate, Standards (8.5 BLOCKING) and Testing (7.0 Important) flagged a missing `beforeEach(mutationSpy.mockClear())` in the new shared test, claiming the cancel test would fail with `expected 0 calls, received 1`. Architecture rated this 3.0 (cosmetic) noting the project has `clearMocks: true`. Direct verification of `jest.config.js` confirms `clearMocks: true` — Jest auto-clears all mocks between tests project-wide, so the explicit `beforeEach` is redundant and the shared test is correct as-is. The Standards/Testing BLOCKING flags are **invalidated** by this fact.

### Highlights

**What this PR does well**
- Net −196 source lines via extraction; ~150 lines of duplicate layout/structure consolidated into one shared component
- MPD `GoalCard` gains ellipsis + tooltip behavior — visual parity with PDS card
- Removes the unused star/`primary` toggle feature (only `GoalsList.tsx:76` calls `<GoalCard>`, never with `renderStar`)
- Test reorganization: presentational behavior covered by 8 shared-component tests; wrappers retain only data-plumbing tests (delete mutation, view URL, calculated total)
- Type system enforces parity — both wrappers must satisfy `GoalCardProps`
- All new `t()` strings are existing translation keys (no new entries needed in `translation.json`)

**Acknowledged trade-offs**
- Tooltip-on-hover always shows (even for non-truncated names) — explicitly chosen by author over runtime DOM-measurement complexity
- MPD card width changed from `minWidth: 350` (flexible) to `width: 350` (fixed) — intentional parity with PDS
- `clearMocks: true` makes the existing `beforeEach(mutationSpy.mockClear())` in the MPD wrapper test redundant — leaving the inconsistency as a follow-up suggestion

### Medium Priority (only one substantive finding)

**M1. `<Trans>` variable rename will orphan the existing translation key (Severity 6.0)**

See inline comment on `Shared/GoalCard/GoalCard.tsx:93`. The existing key in `public/locales/en/translation.json:543` references `goal.name ?? t('Unnamed Goal')` literally; the new code uses `displayName`, so `yarn extract` produces a new key and the old one is pruned. ~10 locales lose this confirmation string until re-translated. Recoverable but worth deciding before merge.

- `Tooltip` on a non-focusable `Typography` is keyboard-inaccessible. Same pattern in `MPGAIncomeExpensesReport`, `SavingsFundTransfer`. Codebase-wide concern, not a regression. Worth a separate ticket.

### Review Summary

|-------|---------:|-----:|----------:|-------:|------------:|-----------:|
| Architecture | 0 | 0 | 0 | 2 | 2 | High |
| Testing | 0 | 0 | 0 | 0 | 4 | Medium (after `clearMocks` correction) |
| Standards | 0 | 0 | 0 | 1 | 1 | Medium (after `clearMocks` correction) |
| UX | 0 | 0 | 0 | 0 | 4 | High |
| Data Integrity | 0 | 0 | 0 | 0 | 2 | High |
| **Total** | **0** | **0** | **0** | **3** | **13** | High overall |

src/components/Shared/GoalCard/GoalCard.tsx
```diff
@@ -0,0 +1,172 @@
...
+      <Confirmation
+        title={t('Delete Goal')}
+        isOpen={deleting}
+        mutation={handleConfirmDelete}
+        handleClose={() => setDeleting(false)}
+        confirmLabel={t('Delete Goal')}
+        cancelLabel={t('Cancel')}
+        message={
+          <Trans t={t}>
+            Are you sure you want to delete <strong>{displayName}</strong>?
```
zweatshirt: [Medium] **`<Trans>` variable rename will orphan the existing translation key.**

The existing key in `public/locales/en/translation.json:543` is:
```
"Are you sure you want to delete <2>{goal.name ?? t('Unnamed Goal')}</2>? Deleting this goal will remove it permanently."
```

This new code uses `<strong>{displayName}</strong>`, so `yarn extract` will produce a new key and the existing translations (~10 locales) become orphaned and pruned by Crowdin sync until re-translated.

Mitigation options:
- **Accept the key change**, let Crowdin re-translate (cheapest)
- **Inline the original expression** to keep the key closer to the original (still produces a new key — `{name ?? t('Unnamed Goal')}` ≠ `{goal.name ?? t('Unnamed Goal')}` — but more recognizable to translators):
  ```tsx
  <Trans t={t}>
    Are you sure you want to delete{' '}
    <strong>{name ?? t('Unnamed Goal')}</strong>? Deleting this goal
    will remove it permanently.
  </Trans>
  ```
- **Preserve the exact key** by giving the shared component a `goal: { name: string | null }` prop instead of `name` — not worth the API contortion.

src/components/Shared/GoalCard/GoalCard.tsx
```diff
@@ -0,0 +1,172 @@
...
+  const { t } = useTranslation();
+  const locale = useLocale();
+  const [deleting, setDeleting] = useState(false);
+
+  const displayName = name ?? t('Unnamed Goal');
+
+  const handleConfirmDelete = async () => {
+    await onDelete();
+    setDeleting(false);
+  };
```
zweatshirt: [Suggestion] **Redundant `setDeleting(false)`.**

`Confirmation.tsx:51-57` already calls `handleClose()` (which is `() => setDeleting(false)` here) inside its `.finally()` block, so this explicit call is redundant. It's pre-existing in both source wrappers and harmless, but the simpler form is:
```tsx
const handleConfirmDelete = () => onDelete();
```
Flagging as a small cleanup; not blocking.

src/components/HrTools/GoalCalculator/GoalCard/GoalCard.tsx
```diff
@@ -1,129 +1,31 @@
...
-  Divider,
-  IconButton,
-  Typography,
-  styled,
-} from '@mui/material';
-import { DateTime } from 'luxon';
-import { Trans, useTranslation } from 'react-i18next';
-import { Confirmation } from 'src/components/common/Modal/Confirmation/Confirmation';
+import React, { useMemo } from 'react';
+import { GoalCard as SharedGoalCard } from 'src/components/Shared/GoalCard/GoalCard';
```
zweatshirt: [Suggestion] **`SharedGoalCard` import alias is novel for this repo.**

Grep shows zero other instances of importing from `src/components/Shared/` with `as`. Two options to consider:
- **Rename this wrapper file** to `MpdGoalCard.tsx` (mirrors `PdsGoalCard.tsx`), which eliminates the name collision entirely and matches the bifurcation already present in the tree. Requires updating `GoalsList.tsx:76`.
- **Keep current pattern** — the alias is well-handled and rename has callsite churn.

Non-blocking; raised because if there's any naming consistency the team prefers, this is the moment.

src/components/HrTools/GoalCalculator/GoalCard/GoalCard.test.tsx
```diff
@@ -31,50 +27,18 @@ const TestComponent: React.FC<TestComponentProps> = ({ primary = false }) => (
...
-  it('shows filled star when primary', () => {
-    const { getByTestId } = render(<TestComponent primary={true} />);
-    expect(getByTestId('StarIcon')).toBeInTheDocument();
-  });
-
-  it('shows outlined star when not primary', () => {
-    const { getByTestId } = render(<TestComponent primary={false} />);
-    expect(getByTestId('StarBorderOutlinedIcon')).toBeInTheDocument();
+  beforeEach(() => {
+    mutationSpy.mockClear();
```
zweatshirt: [Suggestion] **`beforeEach(mutationSpy.mockClear())` is redundant given `clearMocks: true` in `jest.config.js`.**

The project's jest config auto-clears all mocks before every test, so this explicit clear is a no-op. Removing it would resolve the visible inconsistency between this test and the new `Shared/GoalCard/GoalCard.test.tsx` (which correctly omits it). Cosmetic — not blocking.

src/components/HrTools/GoalCalculator/GoalCard/MpdGoalCard.tsx
```diff
@@ -1,129 +1,31 @@
...
   const accountListId = useAccountListId() ?? '';
   const constants = useGoalCalculatorConstants();
-  const [updateGoalCalculation] = useUpdateGoalCalculationMutation();
   const [deleteGoalCalculation] = useDeleteGoalCalculationMutation();
-  const [deleting, setDeleting] = useState(false);
 
   const overallTotal = useMemo(
     () => calculateGoalTotals(goal, constants).overallTotal,
     [goal, constants],
   );
```
zweatshirt, line 24: [Suggestion] **Dead UI consumer of `goal.primary`.**

The star/primary toggle was removed from this card, so `goal.primary` is no longer rendered anywhere in the UI. The `primary` field is still selected by `ListGoalCalculation` fragment (and `UpdateGoalCalculation` mutation still exists in `GoalCalculations.graphql:48-59`) but has no consumer. Candidates for follow-up cleanup:
- Remove `primary` from `ListGoalCalculation` fragment
- Remove the `UpdateGoalCalculation` mutation (the only remaining `useUpdateGoalCalculationMutation` import resolves to a different generated module — `useSaveField.ts` — so this app-layer operation is genuinely dead)

Leave as-is for this PR; flag for a future cleanup ticket.

zweatshirt:
@wjames111 Was just hoping for a quick look at this at this refactor, especially a double check on user behavior (e.g. viewing and deleting). Thanks for considering Will! 

There is an existing bug I am going to fix shortly after merging this in a separate PR. The steps to reproduce the bug are:
1. Select 'View' on a goal card.
2. Rename the goal in the calculator settings.
3. Go back to the goal list dashboard
4. Delete the goal you just renamed
5. A query will occur on a deleted goal since it doesn't get removed from the cache properly.

src/components/Shared/GoalCard/GoalCard.tsx
```diff
@@ -0,0 +1,170 @@
...
+  justifyContent: 'space-between',
+  alignItems: 'center',
+  marginBottom: theme.spacing(1),
+  paddingRight: theme.spacing(1),
+  paddingLeft: theme.spacing(1),
+}));
+
+export interface GoalCardProps {
+  name: string | null | undefined;
+  goalAmount: ReactNode;
```
wjames111: Question: Would it be easier if we just passed in the number to goal amount, then use the currencyFormat and locale from this shared component?

src/components/Reports/Shared/GoalCard/GoalCard.tsx
wjames111, line 1: Thoughts on putting this component into Reports/Shared rather then the global shared folder?

src/components/HrTools/PdsGoalCalculator/GoalCard/PdsGoalCard.tsx
wjames111, line 1: Question: Thoughts on moving the Delete mutation directly into `PdsGoalCard` rather then passing it in?

wjames111 (approved):
Thanks for consolidating this, looks great!

PR #1755, opened 2026-05-01
[MPDX-9455] - Display calculated goal amount on PDS Goal Card

wjames111:
## Description

- Display the calculated total goal amount on each PDS Goal Card in the goals list, so users don't have to open the calculator to see their goal
- Add a `calculatePdsGoalTotal()` helper that composes the three existing calculation functions (salary, reimbursable expenses, other expenses) into a single call returning the final assessment amount
- Wire up goal calculator constants and HCM user data from `PdsGoalsList` to each `PdsGoalCard`
- Matches the layout and behavior of the regular MPD Goal Calculator's GoalCard

Related Jira ticket: [MPDX-9455](https://jira.cru.org/browse/MPDX-9455)

## Testing

- Go to the PDS Goal Calculator goals list page
- Verify each goal card now shows a "Goal Amount" row with a currency-formatted value below the goal name
- Verify the "Last Updated" row still appears below the "Goal Amount" row
- Create a new goal with salary/benefits data, return to the list, and verify the calculated amount updates
- Verify a goal with no pay rate set shows $0.00

## Checklist:

- [x] I have given my PR a title with the format "MPDX-(JIRA#) (summary sentence max 80 chars)"
- [x] I have applied the appropriate labels (_Add the label "Preview" to automatically create a preview environment_)
- [x] I have run the Claude Code `/pr-review` command locally and fixed any relevant suggestions
- [ ] I have requested a review from another person on the project
- [ ] I have tested my changes in preview or in staging
- [ ] I have cleaned up my commit history

PR #1765, opened 2026-05-08
[MPDX-9475] - Add Default and Simple view to PDS Goal Calculator

wjames111:
## Description

Wires the existing API field `DesignationSupportCalculation.formType` (values `DETAILED` and `SIMPLE`) through the PDS Goal Calculator UI so users can choose between a full Default view and a streamlined Simple view at goal creation time and switch between them later from the Settings step.

- New `CreateGoalDialog` opens when the user clicks **Create a New Goal** on the goals list — the user must explicitly pick **Default** or **Simple** before the goal is created. The chosen form type is sent as `formType` on the create mutation.
- Each `PdsGoalCard` shows a small **Default** / **Simple** chip near the goal name so users can identify form types from the list. Legacy goals with a null `formType` show no chip.
- `SetupStep` has a new **Form Type** select (right under **Goal Name**) that lets the user toggle between Default and Simple at any time. Switching is silent — saved field values are preserved on the record so toggling back restores the original calculation instantly.
- When `formType === Simple`:
  - The **Reimbursable Expenses** step is filtered out of the side-panel step list
  - The disabled **403b Contribution Percentage** display field is hidden in the Setup step
  - Lines **2A** (Reimbursable Expenses) and **2B** (403b Contributions) are hidden in the summary table
  - The summary calculation (`usePdsSummaryData`) and per-card total calculation (`calculatePdsGoalTotal`) zero out `reimbursableTotal` and `fourOThreeBPercentage` when building the constants for `calculateOtherExpenses` (the pure helper itself is unchanged)
- Defensive `useEffect` clamps `stepIndex` to 0 if the steps array shrinks after a `formType` change, preventing a `currentStep === undefined` crash.
- Create-mutation error path: failures show an error snackbar, the dialog stays open, and the Create button re-enables so the user can retry.

Related to [MPDX-9475](https://jira.cru.org/browse/MPDX-9475).

### Goal creation

- Go to the PDS Goal Calculator goals list
- Click **Create a New Goal** — a dialog opens with two radio cards (**Default** and **Simple**) and short descriptions of each
- Confirm the **Create** button is disabled until a form type is picked
- Pick **Simple** and click **Create** — you land on a new goal with **3 steps** in the side panel (no Reimbursable Expenses)
- Repeat with **Default** — you land on a goal with **4 steps**

### Form type toggle

- On a Simple goal, go to **Settings** — confirm the **403b Contribution Percentage** field is not shown
- Change **Form Type** to **Default** — confirm the side panel updates to include **Reimbursable Expenses** and the 403b field reappears
- Switch back to **Simple** — confirm everything hides again. If you previously entered reimbursable values, they remain on the record (toggle back to Default to verify they are restored).

### Summary report

- On a **Simple** goal, go to **Summary Report** — confirm rows **2A** (Reimbursable Expenses) and **2B** (403b Contributions) are not present
- Confirm the **Total Goal** is correspondingly lower than for the same inputs in Default mode

### Goal card

- Return to the goals list — confirm each card shows a **Default** or **Simple** chip according to its form type
- Existing goals created before this feature shipped show no chip (legacy `formType: null` behavior)

## Checklist:

- [x] I have given my PR a title with the format \"MPDX-(JIRA#) (summary sentence max 80 chars)\"
- [x] I have applied the appropriate labels (_Add the label \"Preview\" to automatically create a preview environment_)
- [x] I have run the Claude Code \`/pr-review\` command locally and fixed any relevant suggestions

### DIFF
--- a/src/components/HrTools/PdsGoalCalculator/GoalCard/PdsGoalCard.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/GoalCard/PdsGoalCard.tsx
@@ -1,5 +1,8 @@
 import React, { useMemo } from 'react';
+import { Chip } from '@mui/material';
+import { useTranslation } from 'react-i18next';
 import { GoalCard } from 'src/components/Reports/Shared/GoalCard/GoalCard';
+import { DesignationSupportFormType } from 'src/graphql/types.generated';
 import { useAccountListId } from 'src/hooks/useAccountListId';
 import { useGoalCalculatorConstants } from 'src/hooks/useGoalCalculatorConstants';
 import {
@@ -17,6 +20,7 @@ export interface PdsGoalCardProps {
 }
 
 export const PdsGoalCard: React.FC<PdsGoalCardProps> = ({ goal }) => {
+  const { t } = useTranslation();
   const accountListId = useAccountListId() ?? '';
   const [deletePdsGoalCalculation] = useDeletePdsGoalCalculationMutation();
 
@@ -38,6 +42,16 @@ export const PdsGoalCard: React.FC<PdsGoalCardProps> = ({ goal }) => {
     return constants ? calculatePdsGoalTotal(goal, constants) : 0;
   }, [goal, goalMiscConstants, goalGeographicConstantMap, hcmUser]);
 
+  const formTypeBadge = useMemo(() => {
+    if (goal.formType === DesignationSupportFormType.Simple) {
+      return <Chip label={t('Simple')} size="small" />;
+    }
+    if (goal.formType === DesignationSupportFormType.Detailed) {
+      return <Chip label={t('Default')} size="small" />;
+    }
+    return null;
+  }, [goal.formType, t]);
+
   const handleDelete = async () => {
     await deletePdsGoalCalculation({
       variables: { id: goal.id },
@@ -57,6 +71,7 @@ export const PdsGoalCard: React.FC<PdsGoalCardProps> = ({ goal }) => {
       updatedAt={goal.updatedAt}
       viewHref={`/accountLists/${accountListId}/hrTools/pdsGoalCalculator/${goal.id}`}
       onDelete={handleDelete}
+      badge={formTypeBadge}
     />
   );
 };
--- a/src/components/HrTools/PdsGoalCalculator/GoalCard/PdsGoalCard.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/GoalCard/PdsGoalCard.tsx
@@ -42,12 +42,13 @@ export const PdsGoalCard: React.FC<PdsGoalCardProps> = ({ goal }) => {
     return constants ? calculatePdsGoalTotal(goal, constants) : 0;
   }, [goal, goalMiscConstants, goalGeographicConstantMap, hcmUser]);
 
-  let formTypeBadge: React.ReactElement | null = null;
-  if (goal.formType === DesignationSupportFormType.Simple) {
-    formTypeBadge = <Chip label={t('Simple')} size="small" />;
-  } else if (goal.formType === DesignationSupportFormType.Detailed) {
-    formTypeBadge = <Chip label={t('Default')} size="small" />;
-  }
+  const formType = goal.formType ?? DesignationSupportFormType.Detailed;
+  const formTypeBadge =
+    formType === DesignationSupportFormType.Simple ? (
+      <Chip label={t('Simple')} size="small" variant="outlined" />
+    ) : (
+      <Chip label={t('Default')} size="small" />
+    );
 
   const handleDelete = async () => {
     await deletePdsGoalCalculation({
--- a/src/components/HrTools/PdsGoalCalculator/GoalCard/PdsGoalCard.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/GoalCard/PdsGoalCard.tsx
@@ -42,15 +42,12 @@ export const PdsGoalCard: React.FC<PdsGoalCardProps> = ({ goal }) => {
     return constants ? calculatePdsGoalTotal(goal, constants) : 0;
   }, [goal, goalMiscConstants, goalGeographicConstantMap, hcmUser]);
 
-  const formTypeBadge = useMemo(() => {
-    if (goal.formType === DesignationSupportFormType.Simple) {
-      return <Chip label={t('Simple')} size="small" />;
-    }
-    if (goal.formType === DesignationSupportFormType.Detailed) {
-      return <Chip label={t('Default')} size="small" />;
-    }
-    return null;
-  }, [goal.formType, t]);
+  let formTypeBadge: React.ReactElement | null = null;
+  if (goal.formType === DesignationSupportFormType.Simple) {
+    formTypeBadge = <Chip label={t('Simple')} size="small" />;
+  } else if (goal.formType === DesignationSupportFormType.Detailed) {
+    formTypeBadge = <Chip label={t('Default')} size="small" />;
+  }
 
   const handleDelete = async () => {
     await deletePdsGoalCalculation({
--- a/src/components/HrTools/PdsGoalCalculator/GoalsList/CreateGoalDialog.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/GoalsList/CreateGoalDialog.tsx
@@ -1,6 +1,7 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import {
   Button,
+  CircularProgress,
   Dialog,
   DialogActions,
   DialogContent,
@@ -12,8 +13,10 @@ import {
   RadioGroup,
   Typography,
 } from '@mui/material';
+import { visuallyHidden } from '@mui/utils';
 import { useTranslation } from 'react-i18next';
 import { DesignationSupportFormType } from 'src/graphql/types.generated';
+import { isDesignationSupportFormType } from '../Shared/formType';
 
 export interface CreateGoalDialogProps {
   open: boolean;
@@ -33,10 +36,11 @@ export const CreateGoalDialog: React.FC<CreateGoalDialogProps> = ({
     null,
   );
 
-  const handleClose = () => {
-    setSelected(null);
-    onClose();
-  };
+  useEffect(() => {
+    if (open) {
+      setSelected(null);
+    }
+  }, [open]);
 
   const handleCreate = async () => {
     if (selected) {
@@ -44,10 +48,31 @@ export const CreateGoalDialog: React.FC<CreateGoalDialogProps> = ({
     }
   };
 
+  const formTypeOptions: Array<{
+    value: DesignationSupportFormType;
+    title: string;
+    description: string;
+  }> = [
+    {
+      value: DesignationSupportFormType.Detailed,
+      title: t('Default'),
+      description: t(
+        'Full calculator with reimbursable expenses and 403b contributions.',
+      ),
+    },
+    {
+      value: DesignationSupportFormType.Simple,
+      title: t('Simple'),
+      description: t(
+        'Streamlined calculator without reimbursable expenses or 403b contributions.',
+      ),
+    },
+  ];
+
   return (
     <Dialog
       open={open}
-      onClose={handleClose}
+      onClose={onClose}
       maxWidth="sm"
       fullWidth
       aria-labelledby="create-goal-dialog-title"
@@ -57,56 +82,48 @@ export const CreateGoalDialog: React.FC<CreateGoalDialogProps> = ({
       </DialogTitle>
       <DialogContent>
         <FormControl component="fieldset">
-          <FormLabel sx={{ position: 'absolute', width: 1, height: 1, padding: 0, margin: -1, overflow: 'hidden', clip: 'rect(0,0,0,0)', whiteSpace: 'nowrap', border: 0 }}>
-            {t('Select a form type')}
-          </FormLabel>
+          <FormLabel sx={visuallyHidden}>{t('Select a form type')}</FormLabel>
           <RadioGroup
             value={selected ?? ''}
-            onChange={(_, value) =>
-              setSelected(value as DesignationSupportFormType)
-            }
+            onChange={(_, value) => {
+              if (isDesignationSupportFormType(value)) {
+                setSelected(value);
+              }
+            }}
           >
-            <FormControlLabel
-              value={DesignationSupportFormType.Detailed}
-              control={<Radio />}
-              label={
-                <>
-                  <Typography variant="subtitle1">{t('Default')}</Typography>
-                  <Typography variant="body2" color="text.secondary">
-                    {t(
-                      'Full calculator with reimbursable expenses and 403b contributions.',
-                    )}
-                  </Typography>
-                </>
-              }
-              sx={{ alignItems: 'flex-start', mb: 2 }}
-            />
-            <FormControlLabel
-              value={DesignationSupportFormType.Simple}
-              control={<Radio />}
-              label={
-                <>
-                  <Typography variant="subtitle1">{t('Simple')}</Typography>
-                  <Typography variant="body2" color="text.secondary">
-                    {t(
-                      'Streamlined calculator without reimbursable expenses or 403b contributions.',
-                    )}
-                  </Typography>
-                </>
-              }
-              sx={{ alignItems: 'flex-start' }}
-            />
+            {formTypeOptions.map(({ value, title, description }, index) => (
+              <FormControlLabel
+                key={value}
+                value={value}
+                control={<Radio />}
+                label={
+                  <>
+                    <Typography variant="subtitle1">{title}</Typography>
+                    <Typography variant="body2" color="text.secondary">
+                      {description}
+                    </Typography>
+                  </>
+                }
+                sx={{
+                  alignItems: 'flex-start',
+                  mb: index < formTypeOptions.length - 1 ? 2 : 0,
+                }}
+              />
+            ))}
           </RadioGroup>
         </FormControl>
       </DialogContent>
       <DialogActions>
-        <Button onClick={handleClose} disabled={creating}>
+        <Button onClick={onClose} disabled={creating}>
           {t('Cancel')}
         </Button>
         <Button
           onClick={handleCreate}
           variant="contained"
           disabled={!selected || creating}
+          startIcon={
+            creating ? <CircularProgress size={16} color="inherit" /> : null
+          }
         >
           {t('Create')}
         </Button>
--- a/src/components/HrTools/PdsGoalCalculator/GoalsList/CreateGoalDialog.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/GoalsList/CreateGoalDialog.tsx
@@ -1,0 +1,99 @@
+import React, { useState } from 'react';
+import {
+  Button,
+  Dialog,
+  DialogActions,
+  DialogContent,
+  DialogTitle,
+  FormControl,
+  FormControlLabel,
+  Radio,
+  RadioGroup,
+  Typography,
+} from '@mui/material';
+import { useTranslation } from 'react-i18next';
+import { DesignationSupportFormType } from 'src/graphql/types.generated';
+
+export interface CreateGoalDialogProps {
+  open: boolean;
+  onClose: () => void;
+  onCreate: (formType: DesignationSupportFormType) => Promise<void>;
+  creating: boolean;
+}
+
+export const CreateGoalDialog: React.FC<CreateGoalDialogProps> = ({
+  open,
+  onClose,
+  onCreate,
+  creating,
+}) => {
+  const { t } = useTranslation();
+  const [selected, setSelected] = useState<DesignationSupportFormType | null>(
+    null,
+  );
+
+  const handleCreate = async () => {
+    if (selected) {
+      await onCreate(selected);
+    }
+  };
+
+  return (
+    <Dialog open={open} onClose={onClose} maxWidth="sm" fullWidth>
+      <DialogTitle>{t('Create a New Goal')}</DialogTitle>
+      <DialogContent>
+        <FormControl component="fieldset">
+          <RadioGroup
+            value={selected ?? ''}
+            onChange={(_, value) =>
+              setSelected(value as DesignationSupportFormType)
+            }
+          >
+            <FormControlLabel
+              value={DesignationSupportFormType.Detailed}
+              control={<Radio />}
+              label={
+                <>
+                  <Typography variant="subtitle1">{t('Default')}</Typography>
+                  <Typography variant="body2" color="text.secondary">
+                    {t(
+                      'Full calculator with reimbursable expenses and 403b contributions.',
+                    )}
+                  </Typography>
+                </>
+              }
+              sx={{ alignItems: 'flex-start', mb: 2 }}
+            />
+            <FormControlLabel
+              value={DesignationSupportFormType.Simple}
+              control={<Radio />}
+              label={
+                <>
+                  <Typography variant="subtitle1">{t('Simple')}</Typography>
+                  <Typography variant="body2" color="text.secondary">
+                    {t(
+                      'Streamlined calculator without reimbursable expenses or 403b contributions.',
+                    )}
+                  </Typography>
+                </>
+              }
+              sx={{ alignItems: 'flex-start' }}
+            />
+          </RadioGroup>
+        </FormControl>
+      </DialogContent>
+      <DialogActions>
+        <Button onClick={onClose} disabled={creating}>
+          {t('Cancel')}
+        </Button>
+        <Button
+          onClick={handleCreate}
+          variant="contained"
+          disabled={!selected || creating}
+        >
+          {t('Create')}
+        </Button>
+      </DialogActions>
+    </Dialog>
+  );
+};
--- a/src/components/HrTools/PdsGoalCalculator/GoalsList/CreateGoalDialog.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/GoalsList/CreateGoalDialog.tsx
@@ -33,6 +33,11 @@ export const CreateGoalDialog: React.FC<CreateGoalDialogProps> = ({
     null,
   );
 
+  const handleClose = () => {
+    setSelected(null);
+    onClose();
+  };
+
   const handleCreate = async () => {
     if (selected) {
       await onCreate(selected);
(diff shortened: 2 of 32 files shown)


src/components/HrTools/PdsGoalCalculator/GoalsList/CreateGoalDialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+                setSelected(value);
+              }
+            }}
+          >
+            {formTypeOptions.map(({ value, title, description }, index) => (
+              <FormControlLabel
+                key={value}
+                value={value}
+                control={<Radio />}
+                label={
```
wjames111: [Important] **Radio accessible name includes the full description.** MUI's `FormControlLabel` concatenates all text in `label` into the radio's accessible name, so screen readers announce: *"Default Full calculator with reimbursable expenses and 403b contributions."* That's verbose, and the description text changes between PR revisions/translations would silently change the a11y label.

Keep only the title in `label`; render the description as a sibling and reference it via `aria-describedby`:

```tsx
<FormControlLabel
  key={value}
  value={value}
  control={<Radio inputProps={{ 'aria-describedby': `${value}-desc` }} />}
  label={<Typography variant="subtitle1" component="span">{title}</Typography>}
  sx={{ alignItems: 'flex-start', mb: index < formTypeOptions.length - 1 ? 2 : 0 }}
/>
<Typography
  id={`${value}-desc`}
  variant="body2"
  color="text.secondary"
  sx={{ pl: 4, mb: index < formTypeOptions.length - 1 ? 2 : 0 }}
>
  {description}
</Typography>
```

Note: `subtitle1` defaults to `<h6>` via MUI's `variantMapping` — placing it inside a `<label>` is also invalid HTML. Adding `component="span"` fixes both issues.

src/components/HrTools/PdsGoalCalculator/GoalsList/CreateGoalDialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+                  alignItems: 'flex-start',
+                  mb: index < formTypeOptions.length - 1 ? 2 : 0,
+                }}
+              />
+            ))}
+          </RadioGroup>
+        </FormControl>
+      </DialogContent>
+      <DialogActions>
+        <Button onClick={onClose} disabled={creating}>
```
wjames111: [Medium] **Cancel disabled while `creating` removes the visible escape route, but Escape and backdrop-click still close the dialog.** Mismatch between mouse vs keyboard UX: the visible Cancel button looks blocked while Escape silently works. The mutation also continues in the background regardless.

Pick one of:
- Drop `disabled={creating}` on Cancel — let the user back out; the in-flight mutation result is harmless
- Add `disableEscapeKeyDown={creating}` on the `Dialog` and a no-op `onClose` while creating, for parity

The first is friendlier on slow networks.

src/components/HrTools/PdsGoalCalculator/GoalsList/CreateGoalDialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+}
+
+export const CreateGoalDialog: React.FC<CreateGoalDialogProps> = ({
+  open,
+  onClose,
+  onCreate,
+  creating,
+}) => {
+  const { t } = useTranslation();
+  const [selected, setSelected] = useState<DesignationSupportFormType | null>(
```
wjames111: [Medium] **Manual `useState` form state where the rest of the calculator uses Formik + Yup.** Standards rule: "Forms use Formik + Yup — no manual `useState` form state for anything beyond trivial single-field inputs."

Single-radio is borderline trivial, but the dialog already adds reset-on-open effect, manual disable gate, and a custom narrowing helper (`isDesignationSupportFormType`). A small Formik wrapper with a shared `formType: yup.string().oneOf(Object.values(DesignationSupportFormType)).required()` schema (already used in `SetupStep.tsx`) would reduce code and unify the submission model.

Not blocking — flag for explicit decision.

src/components/HrTools/PdsGoalCalculator/Shared/PdsGoalCalculatorContext.tsx
```diff
@@ -78,20 +83,30 @@ export const PdsGoalCalculatorProvider: React.FC<Props> = ({ children }) => {
...
+    PdsGoalCalculatorStepEnum.Setup,
+  );
   const [rightPanelContent, setRightPanelContent] =
     useState<React.ReactNode>(null);
   const [isDrawerOpen, setIsDrawerOpen] = useState<boolean>(true);
   const { trackMutation, isMutating } = useTrackMutation();
 
+  const stepIndex = useMemo(() => {
+    const idx = steps.findIndex((s) => s.step === activeStep);
+    return idx === -1 ? 0 : idx;
```
wjames111, line 121: [Medium] **Silent step fallback when `activeStep` is dropped from `steps`.** When `formType` flips Detailed → Simple while `activeStep === ReimbursableExpenses`, this `useMemo` returns `0` (Setup) but `activeStep` state is **not updated** — the user is silently teleported to Setup with no UI feedback.

Three consequences:
1. UX: no snackbar / toast tells the user why they moved (the existing `handleStepChange` already shows `'The selected step does not exist.'` for invalid navigation, but this implicit path is silent).
2. State desync: `activeStep` retains the orphaned enum. If the user toggles back to Detailed without navigating, they jump back to ReimbursableExpenses — surprising teleportation.
3. Tests: confirmed by `'falls back to Setup when current step does not exist in new form'` but no test asserts the absence of a snackbar in this path.

Fix — promote reconciliation from a derived index to an effect that mutates state:

```tsx
useEffect(() => {
  if (!steps.some((s) => s.step === activeStep)) {
    setActiveStep(steps[0]?.step ?? PdsGoalCalculatorStepEnum.Setup);
    enqueueSnackbar(
      t('Returned to Setup because that step is not available in this form type.'),
      { variant: 'info' },
    );
  }
}, [steps, activeStep, enqueueSnackbar, t]);
```

This keeps `activeStep` in sync with the displayed step and informs the user.

src/components/HrTools/PdsGoalCalculator/Shared/PdsGoalCalculatorContext.tsx
```diff
@@ -78,20 +83,30 @@ export const PdsGoalCalculatorProvider: React.FC<Props> = ({ children }) => {
...
+  }, [steps, activeStep]);
+
   const currentStep = steps[stepIndex];
```
wjames111: [Medium] **`currentStep = steps[stepIndex]` can be `undefined`, but the context type at `PdsGoalCalculatorContext.tsx:25` declares it `PdsGoalCalculatorStep` (non-nullable).** The type lies. `useSteps()` always returns ≥ 3 entries today, but the type doesn't enforce that. The new `StepProbe` test fixture even uses `currentStep?.step ?? 'none'`, suggesting the author already treats it as nullable.

A future change to `useSteps` (e.g., a new form type with no steps yet) would crash any consumer that dereferences `currentStep.step` / `currentStep.title` without optional chaining.

Fix: `currentStep: PdsGoalCalculatorStep | undefined;` (line 25) and update consumers, OR assert non-empty steps in the provider before assigning.

src/components/HrTools/PdsGoalCalculator/GoalsList/PdsGoalsList.tsx
```diff
@@ -38,30 +42,58 @@ export const PdsGoalsList: React.FC = () => {
...
+      const reimbursements = goalMiscConstants.REIMBURSEMENTS_WITH_MAXIMUM;
+      const phoneFee = reimbursements?.PHONE?.fee;
+      const internetFee = reimbursements?.INTERNET?.fee;
+      if (phoneFee === undefined || internetFee === undefined) {
+        enqueueSnackbar(t('Failed to create goal. Please try again.'), {
+          variant: 'error',
+        });
+        return;
+      }
+      detailedDefaults = {
```
wjames111, line 72: [Medium] **Variable shadowing.** Line 70's `const { data } = await ...` shadows the outer `data` (line 39, used for `goals = data?.designationSupportCalculations.nodes`). Line 95's `catch (error)` shadows the outer `error` (used by `useFetchAllPages`). Today the scopes don't cross, but this is a refactor-time landmine and a `no-shadow` smell.

```tsx
const { data: createData } = await createPdsGoalCalculation({ /* ... */ });
const calculation = createData?.createDesignationSupportCalculation?.designationSupportCalculation;
// ...
} catch (mutationError) {
  enqueueSnackbar(t('Failed to create goal. Please try again.'), { variant: 'error' });
}
```

src/components/HrTools/PdsGoalCalculator/GoalsList/PdsGoalsList.tsx
```diff
@@ -38,30 +42,58 @@ export const PdsGoalsList: React.FC = () => {
...
+        ministryCellPhone: phoneFee,
+        ministryInternet: internetFee,
+      };
+    }
+    try {
+      const { data } = await createPdsGoalCalculation({
+        variables: {
+          attributes: {
```
wjames111: [Medium] **`CreatePdsGoalCalculation` mutation lacks `update` / `refetchQueries` / `cache.modify`.** The new goal is created and normalized in Apollo's cache by `id`, but the paginated `designationSupportCalculations.nodes` list field isn't updated. Today the user is redirected immediately so the staleness is invisible — but on `router.back()`, an in-flight `router.push` failure, or any future caller that doesn't navigate away, the list will show stale data.

CLAUDE.md rule: *"Mutations that change cached data include either `update`, `refetchQueries`, or `cache.evict`."*

Simplest fix:

```tsx
const { data: createData } = await createPdsGoalCalculation({
  variables: { /* ... */ },
  refetchQueries: ['PdsGoalCalculations'],
});
```

Note: this is a [Pre-existing] pattern (also missing in `useCreateGoalCalculationMutation` for the MPD calculator), but the new dialog makes this PR a good time to address it.

src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.tsx
```diff
@@ -138,6 +140,25 @@ export const SetupStep: React.FC = () => {
           </Typography>
         </Box>
         <Grid container spacing={3}>
+          <Grid item xs={12}>
+            <AutosaveTextField
+              fieldName="formType"
```
wjames111, line 145: [Medium] **Form Type select renders empty for legacy goals (`formType: null`).** `useAutoSave` derives `value?.toString() ?? ''` (see `useAutosave.ts:26`), so for a legacy `null` goal the MUI Select displays no `MenuItem` and triggers an `Material-UI: You have provided an out-of-range value ''` console warning. Worse — the user sees a blank field with helper text and can't tell which mode they're currently in (the runtime treats null as Detailed via `isSimpleFormType(null) === false`).

Fix options:
- (preferred) Display the runtime-effective value: render with `value={calculation?.formType ?? DesignationSupportFormType.Detailed}` and update on first interaction. This requires either reading the value differently in this one field or adding a placeholder `MenuItem`.
- Add `<MenuItem value="">{t('—')}</MenuItem>` as a placeholder so MUI doesn't warn.

The existing `useSteps.test.tsx:30-37` test confirms `null` formType uses Detailed step semantics — the UI should match that.

src/components/HrTools/PdsGoalCalculator/SummaryReport/PdsSummaryTable.tsx
```diff
@@ -104,20 +106,24 @@ export const PdsSummaryTable: React.FC<PdsSummaryTableProps> = ({
...
+              line: '2B',
+              category: t('403b Contributions'),
+              amount: otherTotals.fourOThreeBContributions,
+            },
+          ]),
       ...(isPartTime
         ? [
             {
-              line: '2C',
+              line: isSimple ? '2A' : '2C',
```
wjames111: [Medium] **Line label `2A` is reused for different content between Simple and Default views.** In Default view, line `2A` = "Reimbursable Expenses". In Simple view, line `2A` = "Work Comp" (PartTime) or "Benefits" (FullTime). Staff familiar with Default's line numbering will misread the same label on a Simple report.

Options:
- Keep a stable schema: leave `2C` for Work Comp/Benefits regardless of variant; `2A`/`2B` simply don't render in Simple
- Drop the letter suffix on Work Comp/Benefits in Simple view (`'2'` only)

Worth confirming with finance / report stakeholders that re-numbering is intentional before merging.

src/components/HrTools/PdsGoalCalculator/GoalCard/PdsGoalCard.tsx
```diff
@@ -38,6 +42,13 @@ export const PdsGoalCard: React.FC<PdsGoalCardProps> = ({ goal }) => {
     return constants ? calculatePdsGoalTotal(goal, constants) : 0;
   }, [goal, goalMiscConstants, goalGeographicConstantMap, hcmUser]);
```
wjames111, line 43: [Important] [Pre-existing] **`PdsGoalCard.goalAmount` shows `assessment` (~$1,038) while `PdsSummaryTable` "Total Goal" shows `overallTotal` (~$9,816)** — a ~9× discrepancy on the same goal.

- `calculatePdsGoalTotal()` returns `otherExpenses.assessment` only (the assessment line: `(subtotal + creditCardFees + attrition) * adminRate`).
- `usePdsSummaryData.overallTotal = otherTotals.subtotal + otherTotals.attrition + otherTotals.creditCardFees + otherTotals.assessment`.
- The unit tests pin these distinct values: `expect(result).toBeCloseTo(1038.21, 1)` for `calculatePdsGoalTotal` vs `expect(overallTotal).toBeCloseTo(9815.77, 0)`.

Pre-dates this PR (introduced in commit `edfa70868` per the goal-card landing PR #1755), but **this PR makes the inconsistency more visible** by adding the Default/Simple badge directly next to the card amount. Staff comparing card-vs-summary will see two very different numbers labeled "goal."

Recommended follow-up (not blocking this PR):
- Either change `PdsGoalCard.goalTotal` to use `usePdsSummaryData(goal, hcmUser).overallTotal`, or rename `calculatePdsGoalTotal` to `calculatePdsAssessment` and add a `calculatePdsGoalOverallTotal` helper. Current naming is semantically misleading.

Flag for a separate ticket — does not block this PR.

src/components/HrTools/PdsGoalCalculator/calculations/calculatePdsGoalTotal.test.ts
```diff
@@ -74,4 +75,25 @@ describe('calculatePdsGoalTotal', () => {
     const withoutGeo = calculatePdsGoalTotal(goal, defaultConstants);
     expect(withGeo).toBeGreaterThan(withoutGeo);
   });
+
+  it('excludes reimbursable expenses and 403b when formType is Simple', () => {
```
wjames111, line 79: [Medium] **Simple-formType test pins only an inequality (`simpleTotal < detailedTotal && simpleTotal > 0`).** A regression that, e.g., accidentally zeroed an unrelated field in the Simple branch could still satisfy these. The Detailed test pins `1038.21`; the Simple test should pin a corresponding number computable from the same `makeGoal` inputs (with `reimbursableTotal=0`, `fourOThreeBPercentage=0`).

Also recommend adding a *consistency invariant* test:

```ts
it('matches usePdsSummaryData.otherTotals.assessment for the same goal+constants', () => {
  // For both Detailed and Simple, expect calculatePdsGoalTotal(goal, constants)
  // === usePdsSummaryData(goal, hcmUser).otherTotals.assessment
});
```

This is the load-bearing invariant the new shared `buildOtherExpensesConstants` is supposed to guarantee — without a test, future drift is undetectable.

wjames111 (commented):
## Multi-Agent Code Review

**Verdict: APPROVED_WITH_SUGGESTIONS** — No blockers. 3 Important findings (sev 7.0–7.9), several medium findings, and many polish suggestions.

**Risk Score:** 10/10 (CRITICAL by formula — capped). Pragmatically HIGH: feature-scoped to PDS Goal Calculator, well-tested, ~487 non-test lines, but touches financial calculations + a new GraphQL field + nested wizard state.

**Mode:** standard | **Agents:** 6 (Architecture, Testing, Standards, Data Integrity, UX, Financial Reporting) + Dependency Impact

---

### Top-Line Summary

Good feature work. The refactor of `calculatePdsGoalTotal` and `usePdsSummaryData` to share `buildPdsGoalConstants` + `buildOtherExpensesConstants` is a clear win for consistency. Tests cover the new behavior thoroughly with the project's standard `GqlMockedProvider` + `mutationSpy + toHaveGraphqlOperation` patterns. Localization, named exports, generated types, and cache `id` selection are all clean.

The two themes worth attention before merge:
1. **Step-state UX when toggling Form Type** — the `activeStep` ↔ `stepIndex` derivation can desync silently when steps shrink (Detailed → Simple while on `ReimbursableExpenses`). Consensus from Architecture, UX, and Data Integrity agents.
2. **A pre-existing display inconsistency** between `PdsGoalCard` (shows `assessment` ~$1k) and `PdsSummaryTable` (shows `overallTotal` ~$10k) is now made more visible by the new badge. Worth a separate ticket — not introduced by this PR but amplified by it.

---

### Important Findings (Severity 7.0–7.9)

These are strongly recommended fixes that don't block merge but **cannot be dismissed via `/dismiss`** (severity ≥ 7). Address them in this PR or create follow-up tickets.

1. **[Sev 7.0] Radio button accessible names include the full description** — `CreateGoalDialog.tsx:99` (UX)
2. **[Sev 7.0] `PdsGoalCard` goalAmount diverges from `PdsSummaryTable` Total Goal** — `PdsGoalCard.tsx:43` (Financial Reporting, [Pre-existing])
3. **[Sev 7.0] `handleStepChange` snackbar path and silent-fallback no-snackbar invariant are untested** — `PdsGoalCalculatorContext.test.tsx` (Testing)

### Medium Priority (Severity 5.0–6.9)

4. [Sev 6.5] Silent fallback to Setup when `activeStep` is dropped from `steps` — multi-agent consensus (Architecture, UX, Data Integrity)
5. [Sev 6.5] `CreatePdsGoalCalculation` mutation lacks `update` / `refetchQueries` / `cache.modify` — Standards [Pre-existing pattern, but worth fixing]
6. [Sev 6.5] Cancel button disabled while creating removes only visible escape route — UX
7. [Sev 6.5] Mocking `useSteps` in context test weakens 4 pre-existing integration tests — Testing
8. [Sev 6.0] `activeStep` retains orphaned step value while `stepIndex` falls back to 0 — Architecture + Data Integrity
9. [Sev 6.0] Variable shadowing of `data` and `error` inside `handleCreateGoal` — Architecture
10. [Sev 6.0] Form Type select renders empty for legacy `formType: null` goals — Data Integrity
11. [Sev 6.0] `currentStep: PdsGoalCalculatorStep` is typed non-nullable but `steps[stepIndex]` can be `undefined` — Data Integrity
12. [Sev 6.0] No test pins the consistency invariant `calculatePdsGoalTotal === usePdsSummaryData.otherTotals.assessment` — Financial Reporting
13. [Sev 5.5] `CreateGoalDialog` uses raw `useState` form state where the rest of the calculator is Formik-driven — Standards (judgment call)
14. [Sev 5.0] `PdsSummaryTable` reuses line label `2A` for different content between Simple/Detailed — UX
15. [Sev 5.0] No defaults are seeded for `ministryCellPhone`/`ministryInternet` when creating Simple — switching back to Detailed leaves them un-seeded — Data Integrity
16. [Sev 5.0] Simple `formType` test asserts only inequality, not a pinned dollar amount — Financial Reporting

### Suggestions (Severity < 5.0, informational only)

- [4.5] No user feedback when toggling Form Type silently removes 403b row + Reimbursable Expenses step — UX
- [4.0] Form Type select default for legacy `null` goals could display "Default" — UX
- [4.0] Add a unit test for `formType.ts` helpers (`isSimpleFormType`, `isDesignationSupportFormType`) — Testing
- [4.0] `useSteps` test description says `null/undefined` but only tests `null` — Testing
- [4.0] `CreateGoalDialog` `onCreate` test only covers Simple branch — also assert Detailed — Testing
- [3.5] `<Typography variant="subtitle1">` defaults to `<h6>` (block element) inside `<label>` — invalid HTML — Architecture
- [3.5] `CircularProgress size={16}` deviates from project's `size={20}` convention — UX
- [3.5] Chip badge has no color/variant differentiation between Default and Simple — UX
- [3.5] Inline 403b sum logic was previously extracted; PR re-inlines it into JSX — Architecture
- [3.0] `PdsGoalsList` error test should also assert the dialog stays open — Testing
- [3.0] Spurious multi-line reformat of the React import block — Architecture
- [3.0] Redundant `getByRole` after `findByRole` in `PdsSummaryTable.test.tsx` — Testing
- [3.0] `isDesignationSupportFormType` parameter could be `unknown` for safer reuse — Data Integrity

---

### Standards Checklist (from `.claude/rules/code-review.md`)

| Area | Status | Note |
|---|---|---|
| Named exports only | PASS | All new exports named |
| File naming | PASS | PascalCase components, `.test.tsx` colocated, `.graphql` PascalCase |
| GraphQL operation names | PASS | No `Get`/`Load` prefixes |
| Hook names start with `use` | PASS | `useSteps`, `usePdsSummaryData`, etc. |
| All user-visible strings via `t()` | PASS | No hardcoded JSX text, no string interpolation, no dynamic keys |
| `id` in selection sets | PASS | Fragment includes `id` and `formType` |
| Pagination handled | PASS | `useFetchAllPages` unaffected by additive field |
| Mutation cache updates | WARN | `CreatePdsGoalCalculation` missing — flagged above |
| No `any` in new code | PASS | Generated types used throughout |
| Forms use Formik + Yup | WARN | `CreateGoalDialog` raw `useState` — judgment call |
| Submit disabled while submitting | PASS | `disabled={!selected || creating}` |
| Tests colocated `*.test.{ts,tsx}` | PASS | Every new file has tests |
| `GqlMockedProvider` typed generics | PASS | `mocksOverride: ApolloErgonoMockMap` properly typed |
| `findBy*` over `waitFor(() => getBy*)` | PASS | Idiomatic |
| No `console.log`/`debugger`/orphan `TODO` | PASS | Clean |
| No `new Date()` (use Luxon) | N/A | No date logic added |

### Financial Checklist

| Item | Status |
|---|---|
| Arithmetic on money values safe | Yes (no new compounding float math) |
| Currency mixing prevented | N/A (USD only) |
| Rounding at display boundary only | Yes |
| Null/undefined amounts handled | Yes (`isSimpleFormType(null)` correctly returns false) |
| Luxon used for dates | N/A |
| Server-provided aggregations preferred | N/A (client-projected goal) |
| Two-path consistency (`calculatePdsGoalTotal` vs `usePdsSummaryData`) | Partial — paths share a helper, but display semantics still diverge (pre-existing) |

### Dependency Impact

- `PdsGoalCalculatorContext.tsx` has 14+ direct dependents (provider). High blast radius for the step-state refactor.
- `useSteps()` signature change: 1 production caller, updated. `GoalCard` `badge?` prop: optional, both callers compatible. No runtime breaks detected.
- The `formType` field added to the `PdsGoalCalculationFields` fragment must be returned by the REST proxy datahandler — if missing, the field is `null`, which all consumers handle.

---

### Files on Auto-Approval Eligibility

This PR scores LOW/MEDIUM in pragmatic risk terms but the formula caps at 10. Verdict logic uses **non-dismissed, non-pre-existing findings only** — no blockers, so verdict is APPROVED_WITH_SUGGESTIONS.

src/components/HrTools/PdsGoalCalculator/Shared/PdsGoalCalculatorContext.test.tsx
```diff
@@ -1,26 +1,103 @@
...
+import SettingsIcon from '@mui/icons-material/Settings';
+import { act, render, renderHook } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { PdsGoalCalculatorStepEnum } from '../PdsGoalCalculatorHelper';
 import { PdsGoalCalculatorTestWrapper } from '../PdsGoalCalculatorTestWrapper';
 import { usePdsGoalCalculator } from './PdsGoalCalculatorContext';
+import { useSteps } from './useSteps';
+import type { PdsGoalCalculatorStep } from './useSteps';
+
+jest.mock('./useSteps', () => ({
```
wjames111, line 12: [Medium] **Mocking `useSteps` weakens the four pre-existing tests** (`'provides steps and current step'`, `'handleContinue advances...'`, `'handlePreviousStep...'`, `'handlePreviousStep does nothing on the first step'`). Before this PR they exercised the real `useSteps` integration; now they all run against `mockedUseSteps.mockReturnValue(detailedSteps)` set in `beforeEach`, so they no longer prove that the context wires `calculation?.formType` into `useSteps`.

Fix options:
- (a) Add at least one `'integrates real useSteps with calculation.formType'` test using `jest.requireActual` and `calculationMock={{ formType: Simple }}` to verify `steps.length === 3`.
- (b) Move the mock under a nested `describe('with mocked useSteps', ...)` so the four pre-existing tests can stay on the real hook.

Also missing: a test that asserts **no snackbar is shown** when the implicit `stepIndex` fallback fires (vs. `handleStepChange` which DOES show a snackbar). The new code intentionally distinguishes these paths — it should be tested:

```tsx
it('does not show a snackbar when the current step disappears after a steps change', async () => {
  /* render Detailed, click into ReimbursableExpenses, rerender with Simple steps,
     expect queryByText('The selected step does not exist.') === null */
});

it('shows a snackbar when handleStepChange targets a non-existent step', async () => {
  /* render with Simple steps, click 'go to reimbursable',
     expect findByText('The selected step does not exist.') */
});
```

src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.tsx
wjames111: /dismiss We don't need backward compatible PDS goals.

src/components/HrTools/PdsGoalCalculator/GoalCard/PdsGoalCard.tsx
wjames111: Bump

src/components/HrTools/PdsGoalCalculator/GoalsList/PdsGoalsList.tsx
wjames111: /dismiss

src/components/Reports/Shared/GoalCard/GoalCard.tsx
```diff
@@ -105,17 +107,21 @@ export const GoalCard: React.FC<GoalCardProps> = ({
       />
 
       <StyledCard>
-        <StyledHeaderBox>
+        <StyledHeaderBox sx={badge ? { gap: 1 } : undefined}>
```
zweatshirt, line 110: This is opinion, but I'm wondering if it might look better to make Simple/Default a row in the card? I like the badge/chip but I'm just not sure I like where it's placed (right next to the name). What do you think of a 'Goal Type' row or similar? Either way, that's not blocking since it's just opinion

zweatshirt (approved):
This is great Will. This PR is big so there's a possibility I missed something, but from what I can tell, this is solid.

src/components/Reports/Shared/GoalCard/GoalCard.tsx
wjames111: I think that's a good call. I've got a few PR's out on against this one, so I'll create a ticket for this and circle back.

PR #1770, opened 2026-05-11
MPDX-9544 Frontend - Add validation to ensure the Requested Gross Salary for SOSA staff doesn't exceed their cap

zweatshirt:
## Description

- We don't want SOSA staff to be able to continue their form if their gross requested salary exceeds their cap.
- This needs to be done against their gross requested salary, and not their base requested salary.
    - Because of this, I chose not to do Formik field validation on the user input for their requested salary, as this can easily cause drift between the frontend and backend values and lock the user into an error state. 
    - The alternative solution if we want to allow input field validation is doing the calculations in the frontend for validation instead of depending on API state.

## Testing

- Use one of the test data emails that is SOSA and ensure that they cannot continue past step 3 if their gross requested salary exceeds their cap.
- Determine if the `Alert` is sufficient in notifying the user.

## Checklist:

### DIFF
--- a/src/components/HrTools/SalaryCalculator/SalaryCalculation/ApprovalProcessCard/ApprovalProcessCard.tsx
+++ b/src/components/HrTools/SalaryCalculator/SalaryCalculation/ApprovalProcessCard/ApprovalProcessCard.tsx
@@ -9,11 +9,13 @@ import { useSalaryCalculator } from '../../SalaryCalculatorContext/SalaryCalcula
 import { StepCard } from '../../Shared/StepCard';
 import { StyledCardHeader } from '../StyledCardHeader';
 import { useCaps } from '../useCaps';
+import { useSosaBlockOverCap } from '../useSosaBlockOverCap';
 
 export const ApprovalProcessCard: React.FC = () => {
   const { t } = useTranslation();
   const { calculation, hcmSpouse } = useSalaryCalculator();
   const { overCapPerson } = useCaps();
+  const { blockOnCap } = useSosaBlockOverCap();
 
   const schema = useMemo(
     () =>
@@ -26,7 +28,7 @@ export const ApprovalProcessCard: React.FC = () => {
   const spouseName = hcmSpouse?.staffInfo.preferredName;
 
   const tier = calculation?.progressiveApprovalTier?.tier;
-  if (!tier) {
+  if (!tier || blockOnCap) {
     return null;
   }
 
--- a/src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/RequestedSalaryCard.test.tsx
+++ b/src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/RequestedSalaryCard.test.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import { render, waitFor } from '@testing-library/react';
 import { DeepPartial } from 'ts-essentials';
+import { UserPersonTypeEnum } from 'pages/api/graphql-rest.page.generated';
+import { AutosaveForm } from 'src/components/Shared/Autosave/AutosaveForm';
 import { SalaryCalculationQuery } from '../../SalaryCalculatorContext/SalaryCalculation.generated';
 import {
   SalaryCalculatorTestWrapper,
@@ -33,7 +35,9 @@ const TestComponent: React.FC<SalaryCalculatorTestWrapperProps> = (props) => (
     }}
     {...props}
   >
-    <RequestedSalaryCard />
+    <AutosaveForm>
+      <RequestedSalaryCard />
+    </AutosaveForm>
   </SalaryCalculatorTestWrapper>
 );
 
@@ -162,6 +166,58 @@ As you set your salary level, the amount you receive should reflect the amount o
             .map((cell) => cell.textContent),
         ).toEqual(expectedCells),
       );
+    });
+  });
+
+  describe('SOSA over-cap Alert', () => {
+    const sosaUser = {
+      staffInfo: {
+        userPersonType: UserPersonTypeEnum.EmployeeStaffNonRmoSpouse,
+      },
+    };
+    const overCapMock: DeepPartial<SalaryCalculationQuery['salaryRequest']> = {
+      calculations: {
+        minimumRequiredSalary: 10002,
+        minimumRequestedSalary: 10003,
+        effectiveCap: 10004,
+        requestedGross: 15000,
+      },
+    };
+
+    it('renders when the SOSA user is over their effective cap', async () => {
+      const { findByRole } = render(
+        <TestComponent
+          hasSpouse={false}
+          hcmUser={{ ...sosaUser, currentSalary: { grossSalaryAmount: 10001 } }}
+          salaryRequestMock={overCapMock}
+        />,
+      );
+
+      expect(await findByRole('alert')).toHaveTextContent(
+        'Your request requires additional approvals and cannot be submitted online. SOSA staff can have requests exceeding the $10,004.00 cap approved for certain geographic locations with the appropriate levels of approval.Please contact payroll@cru.org for further assistance.',
+      );
+      expect(
+        await findByRole('link', { name: 'payroll@cru.org' }),
+      ).toHaveAttribute('href', 'mailto:payroll@cru.org');
+    });
+
+    it('does not render when the user is SOSA but under their cap', async () => {
+      const { queryByRole } = render(
+        <TestComponent
+          hasSpouse={false}
+          hcmUser={{ ...sosaUser, currentSalary: { grossSalaryAmount: 10001 } }}
+        />,
+      );
+
+      await waitFor(() => expect(queryByRole('alert')).not.toBeInTheDocument());
+    });
+
+    it('does not render when the user is not SOSA, even if over cap', async () => {
+      const { queryByRole } = render(
+        <TestComponent hasSpouse={false} salaryRequestMock={overCapMock} />,
+      );
+
+      await waitFor(() => expect(queryByRole('alert')).not.toBeInTheDocument());
     });
   });
 });
--- a/src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/RequestedSalaryCard.tsx
+++ b/src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/RequestedSalaryCard.tsx
@@ -1,7 +1,9 @@
-import React, { useMemo } from 'react';
+import React, { useEffect, useMemo } from 'react';
 import {
+  Alert,
   CardContent,
   CardHeader,
+  Link,
   Table,
   TableBody,
   TableCell,
@@ -10,6 +12,7 @@ import {
 } from '@mui/material';
 import { Trans, useTranslation } from 'react-i18next';
 import * as yup from 'yup';
+import { useAutosaveForm } from 'src/components/Shared/Autosave/AutosaveForm';
 import { amount } from 'src/lib/yupHelpers';
 import { AutosaveTextField } from '../../Autosave/AutosaveTextField';
 import { CalculationFieldsFragment } from '../../SalaryCalculatorContext/SalaryCalculation.generated';
@@ -17,6 +20,8 @@ import { useSalaryCalculator } from '../../SalaryCalculatorContext/SalaryCalcula
 import { EffectiveDateNote } from '../../Shared/EffectiveDateNote';
 import { StepCard, StepTableHead } from '../../Shared/StepCard';
 import { useFormatters } from '../../Shared/useFormatters';
+import { useCaps } from '../useCaps';
+import { useSosaBlockOverCap } from '../useSosaBlockOverCap';
 
 export const RequestedSalaryCard: React.FC = () => {
   const { t } = useTranslation();
@@ -26,6 +31,19 @@ export const RequestedSalaryCard: React.FC = () => {
     hcmSpouse,
   } = useSalaryCalculator();
   const { formatCurrency } = useFormatters();
+  const { overCapPerson } = useCaps();
+  const { isUserSosa, blockOnCap } = useSosaBlockOverCap();
+  const { markValid, markInvalid } = useAutosaveForm();
+
+  // Disable the Continue button while the saved gross exceeds the SOSA cap.
+  useEffect(() => {
+    if (blockOnCap) {
+      markInvalid('over-cap');
+    } else {
+      markValid('over-cap');
+    }
+    return () => markValid('over-cap');
+  }, [blockOnCap, markValid, markInvalid]);
 
   const minimumSalaryValue =
     salaryCalculation?.calculations.minimumRequestedSalary;
@@ -160,6 +178,8 @@ export const RequestedSalaryCard: React.FC = () => {
                   schema={schema}
                   label={t('Requested salary')}
                   required
+                  error={blockOnCap}
+                  saveOnChange={isUserSosa}
                 />
               </TableCell>
               {hcmSpouse && (
@@ -175,6 +195,22 @@ export const RequestedSalaryCard: React.FC = () => {
             </TableRow>
           </TableBody>
         </Table>
+
+        {blockOnCap && (
+          <Alert severity="error" sx={{ mt: 2 }}>
+            <Trans t={t}>
+              Your request requires additional approvals and cannot be submitted
+              online. SOSA staff can have requests exceeding the{' '}
+              {{ cap: overCapPerson?.effectiveCap }} cap approved for certain
+              geographic locations with the appropriate levels of approval.
+              <br />
+              <br />
+              Please contact{' '}
+              <Link href="mailto:payroll@cru.org">payroll@cru.org</Link> for
+              further assistance.
+            </Trans>
+          </Alert>
+        )}
       </CardContent>
     </StepCard>
   );
--- a/src/components/HrTools/SalaryCalculator/SalaryCalculation/useCaps.ts
+++ b/src/components/HrTools/SalaryCalculator/SalaryCalculation/useCaps.ts
@@ -1,7 +1,7 @@
 import { useSalaryCalculator } from '../SalaryCalculatorContext/SalaryCalculatorContext';
 import { useFormatters } from '../Shared/useFormatters';
 
-interface OverCapPerson {
+export interface OverCapPerson {
   /** The name of the person whose salary is over their effective cap */
   name: string | null;
 
@@ -12,6 +12,9 @@ interface OverCapPerson {
 interface UseCapsResult {
   /** The sum of the users' requested gross salaries */
   combinedGross: number;
+
+  /** Whether the user is over their effective cap */
+  overUserCap: boolean;
 
   /** The person whose salary is over their effective cap */
   overCapPerson: OverCapPerson | null;
@@ -44,6 +47,7 @@ export const useCaps = (): UseCapsResult => {
 
   return {
     combinedGross,
+    overUserCap,
     overCapPerson,
   };
 };
--- a/src/components/HrTools/SalaryCalculator/Autosave/AutosaveTextField.tsx
+++ b/src/components/HrTools/SalaryCalculator/Autosave/AutosaveTextField.tsx
@@ -13,11 +13,21 @@ export interface AutosaveTextFieldProps
   > {
   fieldName: Exclude<keyof SalaryRequestUpdateInput, 'manuallySplitCap'>;
   schema: yup.Schema;
+  /** Additional error flag from the parent; OR'd with the field's own validation error */
+  error?: boolean;
+
+  /**
+   * Save on every keystroke instead of on blur. Defaults to true for select
+   * boxes; opt-in for text inputs where immediate feedback is required.
+   */
+  saveOnChange?: boolean;
 }
 
 export const AutosaveTextField: React.FC<AutosaveTextFieldProps> = ({
   fieldName,
   schema,
+  error: externalError,
+  saveOnChange,
   ...props
 }) => {
   const saveField = useSaveField();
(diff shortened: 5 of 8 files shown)


src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/useSosaBlockOnCap.ts
```diff
@@ -0,0 +1,41 @@
...
+  const { markValid, markInvalid } = useAutosaveForm();
+
+  const isUserSosa =
+    hcmUser?.staffInfo.userPersonType ===
+    UserPersonTypeEnum.EmployeeStaffNonRmoSpouse;
+  const blockOnCap = isUserSosa && overUserCap;
+
+  useEffect(() => {
+    if (blockOnCap) {
+      markInvalid('over-cap');
```
zweatshirt: [Medium] **Continue-button gating contract is untested.** This `markInvalid('over-cap')` is the entire stated purpose of the hook (per the JSDoc at line 18), but no test asserts that calling it actually disables the Continue button or that the cleanup `markValid('over-cap')` fires on unmount. The Alert-rendering tests in `RequestedSalaryCard.test.tsx` cover visible UI only.

**Cheapest fix (~15 lines):** Render `<RequestedSalaryCard />` + `<StepNavigation />` together under one `<AutosaveForm>` (without the StepNavigation mock) and assert `getByRole('button', { name: 'Continue' })` is disabled when SOSA+over-cap and enabled when not. This pins the wiring end-to-end and locks down the cleanup contract.

_Flagged by Testing, Architecture, Standards._

src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/RequestedSalaryCard.test.tsx
```diff
@@ -164,4 +168,53 @@ As you set your salary level, the amount you receive should reflect the amount o
       );
     });
   });
+
+  describe('SOSA over-cap Alert', () => {
```
zweatshirt, line 172: [Medium] **SOSA + spouse branch combinations untested.** All three SOSA tests below use `hasSpouse={false}`. The invariant `blockOnCap = isUserSosa && overUserCap` — not `&& overCapPerson !== null` — is the most refactor-fragile line in the diff and has no regression net.

Two branches worth pinning:
- **SOSA user under cap, spouse over cap** — `overCapPerson` is non-null (points to spouse), but `blockOnCap` should be `false`. A future refactor that flips to `overCapPerson !== null` would silently start blocking SOSA users for their spouse's over-cap.
- **Married SOSA user over cap** — Alert should name the *user's* numbers, not the spouse's.

_Flagged by Testing, Financial._

src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/useSosaBlockOnCap.ts
```diff
@@ -0,0 +1,41 @@
...
+  overCapPerson: OverCapPerson | null;
+}
+
+/**
+ * Encapsulates the SOSA-over-cap blocking behavior for the Requested Salary step:
+ * - Derives `blockOnCap` from the saved calculation (single source of truth with the server).
+ * - Registers an `over-cap` invalid field with the autosave form so the Continue button is gated.
+ * - Exposes `overCapPerson` for the Alert message.
+ */
+export const useSosaBlockOnCap = (): UseSosaBlockOnCapResult => {
```
zweatshirt: [Suggestion] **Hook couples derivation with side-effect registration.** The hook returns derived state (`blockOnCap`, `overCapPerson`) AND registers an `over-cap` invalid field on `<AutosaveForm>` (effect at L31–38). The implicit contract is that calling this hook anywhere outside an `<AutosaveForm>` throws, and that calling it from a second consumer collides on the `'over-cap'` field name.

Single consumer today, so cost > benefit. Flagging only as a watch-item: if a second consumer ever needs `blockOnCap`/`overCapPerson` for display-only purposes, split into a pure `useSosaOverCap()` + a separate registration hook.

_Flagged by Architecture._

src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/useSosaBlockOnCap.ts
```diff
@@ -0,0 +1,41 @@
...
+  const isUserSosa =
+    hcmUser?.staffInfo.userPersonType ===
+    UserPersonTypeEnum.EmployeeStaffNonRmoSpouse;
+  const blockOnCap = isUserSosa && overUserCap;
+
+  useEffect(() => {
+    if (blockOnCap) {
+      markInvalid('over-cap');
+    } else {
+      markValid('over-cap');
```
zweatshirt: [Suggestion] **`'over-cap'` magic string repeated 3×.** Lines 33, 35, and 37 each use the literal `'over-cap'`. Lift to a module-level `const OVER_CAP_FIELD = 'over-cap'` to centralize the literal and prevent drift if renamed. Also makes the namespace ownership obvious (this field key shares a namespace with all real Formik fields registered to the same `<AutosaveForm>`).

_Flagged by Architecture._

src/components/HrTools/SalaryCalculator/SalaryCalculation/useCaps.ts
```diff
@@ -1,25 +1,29 @@
...
 
+  /** Whether the user is over their effective cap */
+  overUserCap: boolean;
+
   /** The person whose salary is over their effective cap */
   overCapPerson: OverCapPerson | null;
 }
 
 export const useCaps = (): UseCapsResult => {
   const { calculation, hcmUser, hcmSpouse } = useSalaryCalculator();
```
zweatshirt, line 24: [Suggestion] **`useCaps` still has no dedicated unit test.** Now that the hook exports a public `OverCapPerson` interface and is consumed by 5 callers, a direct `useCaps.test.ts` would pin down: `combinedGross` math, `overUserCap` truth table, `overCapPerson` precedence (user wins over spouse when both over cap), and the null short-circuit when neither is over cap. Indirect coverage exists through three consumer tests, but the precedence rule isn't directly exercised.

Estimated ~40 lines via `renderHook`.

_Flagged by Testing._

src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/useSosaBlockOnCap.ts
```diff
@@ -0,0 +1,41 @@
+import { useEffect } from 'react';
```
zweatshirt: [Suggestion] **No colocated `useSosaBlockOnCap.test.ts(x)`.** The repo Standards Checklist says "Every new component, hook, and lib function has a colocated `*.test.{ts,tsx}`". Sibling hooks in the same area (`useEffectivePaycheckDate.test.tsx`, `useDeleteSalaryCalculation.test.tsx`, `useEffectiveDateOptions.test.ts`, `useSubmitDialogContent.test.tsx`) all have dedicated tests. Behavior is covered via `RequestedSalaryCard.test.tsx`, so this is a WARN, not a FAIL. Adding the dedicated test would also resolve the Medium-priority gating-test gap.

_Flagged by Standards._

src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestSummaryCard/RequestSummaryCard.tsx
```diff
@@ -120,7 +120,7 @@ export const RequestSummaryCard: React.FC = () => {
       &apos;s Gross Requested Salary exceeds their individual Maximum Allowable
       Salary. If this is correct, please provide reasoning for why{' '}
       {{ name: overCapPerson?.name }}&apos;s Requested Salary should exceed{' '}
-      {{ cap: overCapPerson?.effectiveCap }} in the Additional Information
+      {{ cap: formatCurrency(overCapPerson?.effectiveCap) }} in the Additional Information
```
zweatshirt: [Suggestion] **Consider extracting `formatCurrency(...)` into a local variable before the `<Trans>`.** Sibling `MaxAllowableSection.tsx` does:

```tsx
const formattedSingleCap = formatCurrency(...);
// ...
<Trans>... {{ singleCap: formattedSingleCap }} ...</Trans>
```

The current inline form works (i18next handles the object expression) but the extracted form is friendlier for `i18next-extract` to parse and easier for translators to inspect. Apply the same pattern in `ApprovalProcessCard.tsx:62`.

_Flagged by UX._

zweatshirt (commented):
## Multi-Agent Code Review — PR #1770 (v2)

**Verdict:** APPROVED WITH SUGGESTIONS — No blockers. The v2 refactor cleanly addressed the previous round's two medium findings (Alert variant deviation + `OverCapPerson` field-name mismatch). Remaining items are non-blocking test-coverage gaps and minor stylistic polish.

**Risk:** 6/10 (MEDIUM) — 5 source files, 74 non-test lines, single-domain scope.

**Agents:** Architecture, Testing, Standards, UX, Financial Reporting.

### Summary

| Tier | Count |
|------|-------|
| Critical (9.0+) | 0 |
| High Priority (8.0–8.9) | 0 |
| Important (7.0–7.9) | 0 |
| Medium (5.0–6.9) | 2 |
| Suggestions (<5.0) | 7 |
| Pre-existing | 1 |

### Top medium-priority items

1. **Continue-button gating contract untested** — `useSosaBlockOnCap`'s `markInvalid('over-cap')` side effect (the entire reason the hook exists) is not asserted by any test. Carried over from round 1.
2. **SOSA + spouse branch combinations untested** — All new tests use `hasSpouse={false}`. The invariant `blockOnCap = isUserSosa && overUserCap` (not `&& overCapPerson !== null`) is the most refactor-fragile line in the diff and has no regression net.

Both are non-blocking (severity 5) but would be quick wins.

### What's confirmed good

- All 5 `useCaps` consumers verified aligned with the new raw-number interface (`Receipt`, `useSubmitDialogContent` use only `combinedGross`; the 3 over-cap consumers correctly wrap fields in `formatCurrency()` at the display boundary).
- Financial checklist: pass on every item (single source of truth = server, formatting at display boundary, USD-only, server-provided aggregations preferred, safe handling of `undefined` via `formatCurrency`'s `?? 0`).
- `<Trans>` `{{ var: expr }}` pattern matches 5+ precedents in the codebase.
- All four affected test files were observed to pass after the refactor.

src/components/HrTools/SalaryCalculator/SalaryCalculation/ApprovalProcessCard/ApprovalProcessCard.tsx
```diff
@@ -57,7 +59,7 @@ export const ApprovalProcessCard: React.FC = () => {
               exceeds their individual Maximum Allowable Salary. If this is
               correct, please provide reasoning for why{' '}
               {{ name: overCapPerson?.name }}&apos;s Salary should exceed{' '}
-              {{ cap: overCapPerson?.effectiveCap }} for division head approval
+              {{ cap: formatCurrency(overCapPerson?.effectiveCap) }} for division head approval
```
zweatshirt: [Suggestion] **Same as the `RequestSummaryCard` finding** — consider extracting `formatCurrency(overCapPerson?.effectiveCap)` into a local variable before the `<Trans>` for translator/extraction friendliness.

_Flagged by UX._

src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/RequestedSalaryCard.test.tsx
```diff
@@ -164,4 +168,53 @@ As you set your salary level, the amount you receive should reflect the amount o
...
+
+    it('does not render when the user is SOSA but under their cap', async () => {
+      const { queryByRole } = render(
+        <TestComponent
+          hasSpouse={false}
+          hcmUser={{ ...sosaUser, currentSalary: { grossSalaryAmount: 10001 } }}
+        />,
+      );
+
+      await waitFor(() => expect(queryByRole('alert')).not.toBeInTheDocument());
```
zweatshirt, line 212: [Suggestion] **Negative-Alert tests assert absence without anchoring to a positive render first.** The `waitFor(() => expect(queryByRole('alert')).not.toBeInTheDocument())` will pass on the first tick before the salary-calculation query resolves. A regression that delays the Alert past the initial render would slip through. Awaiting some other rendered element (e.g., a rowheader cell) before the absence assertion would catch that.

_Flagged by Standards._

src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/useSosaBlockOnCap.ts
```diff
@@ -0,0 +1,41 @@
...
+    UserPersonTypeEnum.EmployeeStaffNonRmoSpouse;
+  const blockOnCap = isUserSosa && overUserCap;
+
+  useEffect(() => {
+    if (blockOnCap) {
+      markInvalid('over-cap');
+    } else {
+      markValid('over-cap');
+    }
+    return () => markValid('over-cap');
```
zweatshirt: [Suggestion] **Cleanup behavior is correct but undocumented.** `return () => markValid('over-cap')` always marks the field valid on effect re-run or unmount. This is the right call (don't leave the form stuck invalid after navigating away), but a one-line comment explaining the intent would help future contributors understand why the cleanup unconditionally marks valid.

_Flagged by Standards._

src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/RequestedSalaryCard.test.tsx
zweatshirt: @canac If I had to guess, a test like this would not only be unnecessary but inaccurate. There should never be a case where a user is SOSA and their spouse is also able to input in the form right? Since SOSA always means non-RMO spouse? Thanks Caleb

zweatshirt:
@canac Waiting for Grace or CD to respond with the proper alert message. I am curious what you think about this in terms of UI design. I didn't choose to go with input field validation (see PR description). Thank you Caleb

src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/RequestedSalaryCard.tsx
```diff
@@ -175,6 +193,20 @@ export const RequestedSalaryCard: React.FC = () => {
             </TableRow>
           </TableBody>
         </Table>
+
+        {blockOnCap && (
+          <Alert severity="error" sx={{ mt: 2 }}>
+            <Trans
+              t={t}
+              defaults="Your request requires additional approvals and cannot be submitted online. SOSA staff can have requests exceeding the {{cap}} cap approved for certain geographic locations with the appropriate levels of approval.<br /><br />Please contact <email>payroll@cru.org</email> for further assistance."
```
canac: Instead of `defaults`, let's use the form of `<Trans>` where you put the text inside the element.

```tsx
<Trans ...>
  Your request requires additional approvals ...
</Trans>
```

src/components/HrTools/SalaryCalculator/SalaryCalculation/useCaps.ts
```diff
@@ -1,25 +1,29 @@
...
   /** The name of the person whose salary is over their effective cap */
   name: string | null;
 
-  /** The formatted effective cap of the person whose salary is over their effective cap */
-  effectiveCap: string;
+  /** The requested gross salary of the person whose salary is over their effective cap */
+  requestedGross: number;
+
+  /** The effective cap of the person whose salary is over their effective cap */
+  effectiveCap: number;
```
canac: question: Why did this change? I might have missed something, but everywhere I see this being used, we format the number as currency.

src/components/HrTools/SalaryCalculator/SalaryCalculation/useCaps.ts
```diff
@@ -33,17 +37,20 @@ export const useCaps = (): UseCapsResult => {
   const overCapPerson = overUserCap
     ? {
         name: hcmUser?.staffInfo.preferredName ?? null,
-        effectiveCap: formatCurrency(calcs.effectiveCap),
+        requestedGross: calcs.requestedGross,
```
canac: I don't see `requestedGross` used anywhere.

```suggestion
```

src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/useSosaBlockOnCap.ts
```diff
@@ -0,0 +1,31 @@
...
+export const useSosaBlockOnCap = (): UseSosaBlockOnCapResult => {
+  const { hcmUser } = useSalaryCalculator();
+  const { overUserCap, overCapPerson } = useCaps();
+
+  const isUserSosa =
+    hcmUser?.staffInfo.userPersonType ===
+    UserPersonTypeEnum.EmployeeStaffNonRmoSpouse;
+  const blockOnCap = isUserSosa && overUserCap;
+
+  return { isUserSosa, blockOnCap, overCapPerson };
```
canac: subjective: Maybe it's just me, but hooks that return partial data from other hooks are hard for me to follow because the same data can be returned from multiple hooks. Could we not return `overCapPerson` and require callers to get `overCapPerson` from `useCaps` themselves? Feel free to disagree.

```suggestion
  return { isUserSosa, blockOnCap };
```

src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/useSosaBlockOnCap.ts
```diff
@@ -0,0 +1,31 @@
...
+  blockOnCap: boolean;
+
+  /** The person whose salary is over their effective cap, or null */
+  overCapPerson: OverCapPerson | null;
+}
+
+/**
+ * Derives whether a SOSA-class user is blocked because their saved gross exceeds their
+ * effective cap. Pure read; the form-context wiring (Continue-button gating via
+ * `markInvalid`) lives in the consuming component.
```
canac: This comment feels very AI-written with the caveman plugin. I also like to include the business rules behind the code, something like this.

```suggestion
 * Determine whether a SOSA user's requested gross salary exceeds their effective cap. Requests over their cap may not be submitted online.
```

src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/useSosaBlockOnCap.ts
canac: I kind of agree with this suggestion. It would just be a couple lines to inline it into the component.

src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/useSosaBlockOnCap.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
```
canac: optional: What do you think of this hook being called `useSosaOverCap`? Or `useSosaBlockOverCap`?

src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/RequestedSalaryCard.tsx
```diff
@@ -26,6 +30,18 @@ export const RequestedSalaryCard: React.FC = () => {
     hcmSpouse,
   } = useSalaryCalculator();
   const { formatCurrency } = useFormatters();
+  const { isUserSosa, blockOnCap, overCapPerson } = useSosaBlockOnCap();
+  const { markValid, markInvalid } = useAutosaveForm();
+
+  // Gate the Continue button while the saved gross exceeds the SOSA cap.
```
canac:
```suggestion
  // Disable the Continue button while the saved gross exceeds the SOSA cap
```

src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/RequestedSalaryCard.test.tsx
canac: Correct, SOSA staff should not have a spouse. The server and client both ignore the ineligible spouse (even if they are in HCM).

canac (approved):
I have a few suggestions, but the core logic looks good! Do you think we should hide the "Approval Process" card when the error message displays since they can't submit anyway?

src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/useSosaBlockOnCap.ts
zweatshirt: Oh I thought the part about `markInvalid` was actually kind of helpful to give context to the purpose of the hook, but I can change this to be more useful and general, since it could potentially be used elsehwere. If we remove overCapPerson here, the hook may not be necessary.

src/components/HrTools/SalaryCalculator/SalaryCalculation/useCaps.ts
zweatshirt: It was brought up by Claude as a suggestion. I was torn between it. I understood the argument to be `formatCurrency` being purely at the presentation layer allows this hook to be more reusable by returning the raw values. That being said, if we using `formatCurrency` everywhere this hook is called anyway, you are right that it's best to keep the formatting here. I'll change it back

canac: To achieve the same effect, we could update the comment to say that it is the formatted cap. Or even rename the field to `formattedEffectiveCap`. To me, the fact that it is a `string` instead of a `number` is a strong hint that it is a formatted cap.

src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/useSosaBlockOnCap.ts
canac: You can keep that explanation about `markInvalid` if you think it's helpful! If you rewrite it to not be terse caveman style, I'm fine with it!

It's unfortunate that caveman mode affects code comments like this too.

zweatshirt: @canac I don't have caveman mode :') just a sloppy AI generated comment...

canac: Oh sorry, this text sounds like how Claude talks since I added https://github.com/JuliusBrussee/caveman

PR #1772, opened 2026-05-12
SAA-645 Require note on savings fund transfer form

jeff-clay:
## Description

- Makes the Note field required on one-time fund transfers (previously labeled "Note (Optional)", which produced a confusing SAA 422 when left blank).
- Defaults the Note to `<lastName> Savings Fund Transfer in MPDX` using the current user's `lastName`. Editable — clearing it triggers the required-field error.
- Front-end enforcement only; recurring transfers are unaffected (their description is server-synthesized).

**Related:** [SAA-645](https://jira.cru.org/browse/SAA-645)

## Testing

- Open **HR Tools → Savings Fund Transfer → New Transfer**, **One Time** schedule.
- Confirm Note is pre-filled with `<your-last-name> Savings Fund Transfer in MPDX`.
- Submit unchanged → succeeds.
- Open another new transfer, clear Note (or enter only whitespace), submit → "Note is required" inline error, no submission.
- Switch to **Monthly** schedule — no Note field shown, recurring submits normally.

## Checklist:

- [x] I have given my PR a title with the format "MPDX-(JIRA#) (summary sentence max 80 chars)" — using `SAA-645` since this work is tracked in the SAA project's Jira

### DIFF
--- a/src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
@@ -170,61 +170,41 @@ describe('TransferModal', () => {
       ).toBeInTheDocument();
     });
 
-    it('should require a note for one-time transfers', async () => {
-      const { getByRole, findByText } = render(<Components />);
+    it.each([
+      { label: 'empty', input: '' },
+      { label: 'whitespace-only', input: '   ' },
+    ])(
+      'should reject $label notes on one-time transfers',
+      async ({ input }) => {
+        const { getByRole, findByText } = render(<Components />);
 
-      userEvent.click(getByRole('combobox', { name: /from account/i }));
-      userEvent.click(getByRole('option', { name: /staff account/i }));
-      userEvent.click(getByRole('combobox', { name: /to account/i }));
-      userEvent.click(getByRole('option', { name: /staff savings/i }));
+        userEvent.click(getByRole('combobox', { name: /from account/i }));
+        userEvent.click(getByRole('option', { name: /staff account/i }));
+        userEvent.click(getByRole('combobox', { name: /to account/i }));
+        userEvent.click(getByRole('option', { name: /staff savings/i }));
 
-      const amountField = getByRole('spinbutton', { name: /amount/i });
-      userEvent.clear(amountField);
-      userEvent.type(amountField, '100');
+        const amountField = getByRole('spinbutton', { name: /amount/i });
+        userEvent.clear(amountField);
+        userEvent.type(amountField, '100');
 
-      // User clears the prefilled note before submitting.
-      const noteField = getByRole('textbox', { name: /note/i });
-      userEvent.clear(noteField);
+        const noteField = getByRole('textbox', { name: /note/i });
+        userEvent.clear(noteField);
+        if (input) {
+          userEvent.type(noteField, input);
+        }
 
-      userEvent.click(getByRole('button', { name: /submit/i }));
+        userEvent.click(getByRole('button', { name: /submit/i }));
 
-      expect(await findByText('Note is required')).toBeInTheDocument();
-      expect(mutationSpy).not.toHaveBeenCalledWith(
-        expect.objectContaining({
-          operation: expect.objectContaining({
-            operationName: 'CreateTransfer',
+        expect(await findByText('Note is required')).toBeInTheDocument();
+        expect(mutationSpy).not.toHaveBeenCalledWith(
+          expect.objectContaining({
+            operation: expect.objectContaining({
+              operationName: 'CreateTransfer',
+            }),
           }),
-        }),
-      );
-    });
-
-    it('should reject whitespace-only notes on one-time transfers', async () => {
-      const { getByRole, findByText } = render(<Components />);
-
-      userEvent.click(getByRole('combobox', { name: /from account/i }));
-      userEvent.click(getByRole('option', { name: /staff account/i }));
-      userEvent.click(getByRole('combobox', { name: /to account/i }));
-      userEvent.click(getByRole('option', { name: /staff savings/i }));
-
-      const amountField = getByRole('spinbutton', { name: /amount/i });
-      userEvent.clear(amountField);
-      userEvent.type(amountField, '100');
-
-      const noteField = getByRole('textbox', { name: /note/i });
-      userEvent.clear(noteField);
-      userEvent.type(noteField, '   ');
-
-      userEvent.click(getByRole('button', { name: /submit/i }));
-
-      expect(await findByText('Note is required')).toBeInTheDocument();
-      expect(mutationSpy).not.toHaveBeenCalledWith(
-        expect.objectContaining({
-          operation: expect.objectContaining({
-            operationName: 'CreateTransfer',
-          }),
-        }),
-      );
-    });
+        );
+      },
+    );
 
     it('should validate end date is after transfer date for recurring transfers', async () => {
       const { getByRole, findByLabelText, getByLabelText, findByText } = render(
--- a/src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
@@ -65,11 +65,13 @@ const transferDefaultData: TransferModalData['transfer'] = {
 interface ComponentsProps {
   transfer?: TransferModalData['transfer'];
   type?: TransferTypeEnum;
+  lastName?: string;
 }
 
 const Components = ({
   transfer = transferDefaultData,
   type,
+  lastName = 'Doe',
 }: ComponentsProps) => (
   <SnackbarProvider>
     <ThemeProvider theme={theme}>
@@ -90,6 +92,7 @@ const Components = ({
                 }}
                 funds={fundsMock}
                 handleClose={handleClose}
+                lastName={lastName}
               />
             </StaffSavingFundProvider>
           </GqlMockedProvider>
@@ -179,7 +182,10 @@ describe('TransferModal', () => {
       userEvent.clear(amountField);
       userEvent.type(amountField, '100');
 
-      // Note intentionally left blank.
+      // User clears the prefilled note before submitting.
+      const noteField = getByRole('textbox', { name: /note/i });
+      userEvent.clear(noteField);
+
       userEvent.click(getByRole('button', { name: /submit/i }));
 
       expect(await findByText('Note is required')).toBeInTheDocument();
@@ -286,6 +292,20 @@ describe('TransferModal', () => {
   });
 
   describe('Inputs', () => {
+    it('should default the note to "<lastName> Savings Fund Transfer in MPDX" on new one-time transfers', () => {
+      const { getByRole } = render(<Components lastName="Sleight" />);
+
+      const noteField = getByRole('textbox', { name: /note/i });
+      expect(noteField).toHaveValue('Sleight Savings Fund Transfer in MPDX');
+    });
+
+    it('should not default the note when lastName is missing', () => {
+      const { getByRole } = render(<Components lastName="" />);
+
+      const noteField = getByRole('textbox', { name: /note/i });
+      expect(noteField).toHaveValue('');
+    });
+
     it('should populate initial values from data prop', () => {
       const dataWithValues: TransferModalData['transfer'] = {
         transferFrom: '70056dcb-1a0f-4279-b710-928bcdff811a',
@@ -482,6 +502,7 @@ describe('TransferModal', () => {
       userEvent.clear(amountField);
       userEvent.type(amountField, '100');
 
+      userEvent.clear(noteField);
       userEvent.type(noteField, 'Test note');
 
       userEvent.click(getByRole('button', { name: /submit/i }));
--- a/src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
@@ -8,12 +8,7 @@ import { DateTime } from 'luxon';
 import { SnackbarProvider } from 'notistack';
 import TestRouter from '__tests__/util/TestRouter';
 import { GqlMockedProvider } from '__tests__/util/graphqlMocking';
-import {
-  GetUserDocument,
-  GetUserQuery,
-} from 'src/components/User/GetUser.generated';
-import { UserTypeEnum } from 'src/graphql/types.generated';
-import { createCache } from 'src/lib/apollo/cache';
+import { GetUserQuery } from 'src/components/User/GetUser.generated';
 import theme from 'src/theme';
 import { StaffSavingFundProvider } from '../../StaffSavingFund/StaffSavingFundContext';
 import {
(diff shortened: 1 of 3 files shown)


src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
```diff
@@ -169,6 +179,10 @@ export const TransferModal: React.FC<TransferModalProps> = ({
   const type = data.type || TransferTypeEnum.New;
   const isNew = type === TransferTypeEnum.New;
 
+  const defaultNote = lastName
```
jeff-clay: [Nice-to-have] Consider trimming `lastName` before interpolating, so an upstream value like `'  Sleight  '` doesn't produce a default with leading/trailing whitespace. `const trimmed = lastName?.trim(); const defaultNote = trimmed ? t(...) : '';`

jeff-clay (commented):
Small, well-scoped form/validation change. All SAA-645 acceptance criteria met (required field with visual asterisk, whitespace-only rejected via `.trim().min(1)`, mutation never fires when invalid). Default-note feature is a documented scope addition per product owner. Refactor to `useGetUserQuery` over parsing `staffAccount.name` is the right architectural call — Apollo's cache makes the duplicate hook call free. No must-fix issues; a few nice-to-have hardening suggestions and one optional whitespace test below.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
jeff-clay: [Nice-to-have] `defaultNote` only runs through Formik `initialValues` at mount. If `useGetUserQuery` is still pending on first open (cold Apollo cache), the modal opens with no default. `UserPreferenceProvider` warms the cache at app root so it's unlikely, but if you see it in QA, `enableReinitialize={true}` on Formik or a render-guard in TransfersPage are the two fixes.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
```diff
@@ -135,18 +135,28 @@ const transferSchema = (locale: string) =>
       .number()
       .required(i18n.t('Amount is required'))
       .min(0.01, i18n.t('Amount must be at least $0.01')),
-    note: yup.string().nullable(),
+    note: yup.string().when('schedule', {
```
jeff-clay, line 139: [Mentoring] `.trim().min(1).required()` is the canonical Yup pattern for non-blank — `.required()` alone would still let `'   '` through, because Yup treats any non-undefined string as 'present'. The trim transformation strips whitespace, then min(1) catches the empty result.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
```diff
@@ -256,7 +270,7 @@ export const TransferModal: React.FC<TransferModalProps> = ({
           status: data.transfer.status ?? '',
           transferDate: data.transfer.transferDate ?? getToday(),
           endDate: data.transfer.endDate ?? null,
-          note: data.transfer.note ?? '',
+          note: data.transfer.note || defaultNote,
```
jeff-clay, line 277: [Mentoring] `||` vs `??` matters here. With `||`, empty strings fall through to `defaultNote` (what we want, since `transferDefaultData.note = ''` for new transfers). With `??`, only nullish values fall through and empty stays. The original `?? ''` was just normalizing nullish to empty; the new `|| defaultNote` is an actual upgrade-to-default.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
```diff
@@ -167,6 +170,34 @@ describe('TransferModal', () => {
       ).toBeInTheDocument();
     });
 
+    it('should require a note for one-time transfers', async () => {
```
jeff-clay: [Suggested tests] This test covers the 'user cleared the prefilled note' path. Consider adding a sibling test for 'user typed whitespace only' (`userEvent.type(noteField, '   ')`) to explicitly lock in the `.trim().min(1)` branch.

src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.tsx
```diff
@@ -84,6 +85,7 @@ export const TransfersPage: React.FC<TransfersPageProps> = ({ title }) => {
 
   const { data: staffAccountData, error: staffAccountError } =
     useStaffAccountQuery();
+  const { data: userData } = useGetUserQuery();
```
jeff-clay: [Mentoring] Calling `useGetUserQuery()` here is free in practice — `UserPreferenceProvider` calls the same operation at app root, so Apollo's normalized cache serves this read from memory. No need to thread `lastName` down from a higher ancestor.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
jeff-clay: https://github.com/CruGlobal/mpdx-react/pull/1772/commits/417b0af642d7eb5bb6cebae3add0e815b46a051d

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
jeff-clay: https://github.com/CruGlobal/mpdx-react/pull/1772/commits/417b0af642d7eb5bb6cebae3add0e815b46a051d

jeff-clay: From Claude: Implementing enableReinitialize on Formik would also re-run initialValues on any prop change, which could clobber user input in some edge cases — the cure is potentially worse than the rare race.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
```diff
@@ -135,18 +135,28 @@ const transferSchema = (locale: string) =>
       .number()
       .required(i18n.t('Amount is required'))
       .min(0.01, i18n.t('Amount must be at least $0.01')),
-    note: yup.string().nullable(),
+    note: yup.string().when('schedule', {
+      is: ScheduleEnum.OneTime,
+      then: (schema) =>
+        schema
+          .trim()
+          .min(1, i18n.t('Note is required'))
```
dr-bizz: WThe min is redundant as you are trimming it and it is a required field. Unless you need more than 1 character on the backend.

src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.tsx
dr-bizz: Lets move this hook to the Model. There is no reason to do it here unless we are using `userData` on the component too.

It is best to just add it to the model, so it runs once it is open rather then it running when the page is loaded.

dr-bizz (approved):
This is looking good, but you don't need to add the `min` onto the `yup` logic as required does that already.

src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.tsx
jeff-clay: https://github.com/CruGlobal/mpdx-react/pull/1772/commits/6f5c546d9e52bc69e670b348d272c11ec0a8f48b

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
jeff-clay: https://github.com/CruGlobal/mpdx-react/pull/1772/commits/6f5c546d9e52bc69e670b348d272c11ec0a8f48b

jeff-clay:
@dr-bizz Changes made. I know you approved already, but not sure if you want to do a review of changes. Not sure of protocol.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
```diff
@@ -68,6 +74,27 @@ interface ComponentsProps {
...
+        staffAccountId: null,
+        primaryDesignation: null,
+        userType: UserTypeEnum.UsStaff,
+        preferences: null,
+      },
+    },
+  });
+  return cache;
+};
+
```
dr-bizz: There is a different way to do this. Let me type it up

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
```diff
@@ -65,11 +71,34 @@ const transferDefaultData: TransferModalData['transfer'] = {
 interface ComponentsProps {
   transfer?: TransferModalData['transfer'];
   type?: TransferTypeEnum;
+  lastName?: string;
 }
 
+const buildCacheWithUser = (lastName: string) => {
```
dr-bizz: We don't need to build the cache. We can do the same as we've done here: src/components/DonationTable/DonationTable.test.tsx

We create the mocks on the GqlMockedProvider, which will mock the GQL queries

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
```diff
@@ -135,7 +136,11 @@ const transferSchema = (locale: string) =>
       .number()
       .required(i18n.t('Amount is required'))
       .min(0.01, i18n.t('Amount must be at least $0.01')),
-    note: yup.string().nullable(),
+    note: yup.string().when('schedule', {
+      is: ScheduleEnum.OneTime,
```
dr-bizz, line 140: Are we only adding the required note if it is a one time payment?

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
```diff
@@ -80,6 +109,7 @@ const Components = ({
             createTransfer: CreateTransferMutation;
             updateRecurringTransfer: UpdateRecurringTransferMutation;
           }>
+            cache={buildCacheWithUser(lastName)}
```
dr-bizz: We can mock the response of the GQL query here so we don't need to store it in the cache. (I don't know if this will work out the box, as I haven't tested it, you might need to edit the GetUser)

We only need to enter the values we want to mock as we are randomly generating everything else.
```suggestion
            mocks={{
              GetUser {
                user {
                  id
                  firstName
                  lastName
                }
              }
            }}
```

dr-bizz (changes requested):
Thank you for fixing that issue. We have a few more issues that have arisen.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
jeff-clay: Yeah, a few reasons:
- The note field currently only renders for one time transfers.
- The recurring GraphQL mutation doesn't accept a description
- SAA's recurring transfer synthesizes its own description server-side from the fund/account name.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
jeff-clay: https://github.com/CruGlobal/mpdx-react/pull/1772/commits/fcda99171e02ee31e4351fd20173f391ea72806c

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
jeff-clay: https://github.com/CruGlobal/mpdx-react/pull/1772/commits/fcda99171e02ee31e4351fd20173f391ea72806c

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
jeff-clay: Tried your `mocks` suggestion. It works for the mocks setup, but with `useGetUserQuery` inside the modal the data resolves async, and Formik captures `initialValues` at mount, so the default note ends up empty in that brief window.

Added a render-guard (`if (userLoading) return null`) as the smallest fix. In production it never fires (cache is warmed at app boot by `UserPreferenceProvider`). In tests it lets the mock resolve before assertions run, and let me drop the cache builder entirely.

If you'd prefer a different shape (e.g. compute the default via `useEffect` + `setFieldValue`, or not handle the loading case in the modal at all) happy to change. Just let me know.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
```diff
@@ -169,6 +174,19 @@ export const TransferModal: React.FC<TransferModalProps> = ({
   const type = data.type || TransferTypeEnum.New;
   const isNew = type === TransferTypeEnum.New;
 
+  const { data: userData, loading: userLoading } = useGetUserQuery();
+
+  if (userLoading) {
+    return null;
+  }
```
dr-bizz, line 181: We don't want to do this, as it can refetch data, and then we would no longer render the component.

See the next comment

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
```diff
@@ -169,6 +174,19 @@ export const TransferModal: React.FC<TransferModalProps> = ({
...
+
+  const trimmedLastName = userData?.user?.lastName?.trim();
+  const defaultNote = trimmedLastName
+    ? t('{{lastName}} Savings Fund Transfer in MPDX', {
+        lastName: trimmedLastName,
+      })
+    : '';
+
   const title =
     type === TransferTypeEnum.New
```
dr-bizz, line 191: We can use useMemo to reinitialise the variable `defaultNote` upon `userData` changing.
```suggestion
  const { data: userData } = useGetUserQuery();

 const defaultNote = useMemo(() => {
    const trimmedLastName = userData?.user?.lastName?.trim();
    return trimmedLastName
      ? t('{{lastName}} Savings Fund Transfer in MPDX', {
          lastName: trimmedLastName,
        })
      : '';
  }, [userData, t]);

  const title =
    type === TransferTypeEnum.New
```

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
```diff
@@ -99,9 +108,15 @@ const Components = ({
   </SnackbarProvider>
 );
 
+const renderModal = async (props: ComponentsProps = {}) => {
+  const result = render(<Components {...props} />);
+  await result.findByText(/Fund Transfer/);
+  return result;
+};
```
dr-bizz, line 115: With now we're using useMemo, I'm unsure if you need this.

dr-bizz (changes requested):
Thank you for making those edits. MPDX React is a large app, which means we need to keep it consistent throughout.

I noticed some more issues and have added my suggestions for you to implement.

The loading from GraphQL gets triggered every time it loads, both during the initial load and subsequent sequential loads.

I believe we can resolve this issue with a useMemo, which updates the default value when userData changes.

BradleySchlather:
A new PR was opened for this ticket. That [PR](https://github.com/CruGlobal/mpdx-react/pull/1825) was reviewed and merged to main.

PR #1768, opened 2026-05-08
[MPDX-9539] Show MCC approval level from API

canac:
## Description

Trust the approval level calculated by the API instead of calculating it ourselves.

Needed so that the approval level from https://github.com/CruGlobal/mpdx_api/pull/3315 will show in the UI.

MPDX-9539

## Testing

- Impersonate caleb.cox@cru.org
- Open the in-progress ASR
- Submit the request
- Check that it says you need MCC approval

## Checklist:

- [x] I have given my PR a title with the format "MPDX-(JIRA#) (summary sentence max 80 chars)"
- [x] I have applied the appropriate labels (_Add the label "Preview" to automatically create a preview environment_)
- [ ] I have run the Claude Code `/pr-review` command locally and fixed any relevant suggestions
- [x] I have requested a review from another person on the project
- [x] I have tested my changes in preview or in staging
- [ ] I have cleaned up my commit history

PR #1773, opened 2026-05-13
[MPDX-9539] Use approval reason from server

canac:
## Description

Builds on #1768 and explains why the overlapping requests require MCC approval, even if the request doesn't exceed the user's caps. The API provides the UI with the reason so that approval logic is centralized on the server.

Depends on https://github.com/CruGlobal/mpdx_api/pull/3325

MPDX-9539

## Testing

- ASRs
  - Impersonate caleb.cox@cru.org
  - Open the in-progress ASR
  - Submit the request
  - Check that it says you need MCC approval because you or your spouse as a pending salary request or ASR
Screenshot 2026-05-13 at 12 23 06 PM

- Salary requests
  - Login to the test account
  - Open the in-progress salary request
  - Go to step 3
  - Check that it says you need MCC approval because you or your spouse as a pending salary request or ASR
Screenshot 2026-05-13 at 12 49 07 PM

## Checklist:

### DIFF
--- a/src/components/HrTools/AdditionalSalaryRequest/RequestPage/Helper/CapSubContent.test.tsx
+++ b/src/components/HrTools/AdditionalSalaryRequest/RequestPage/Helper/CapSubContent.test.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { ThemeProvider } from '@mui/material/styles';
 import { render } from '@testing-library/react';
 import { FormikProvider, useFormik } from 'formik';
-import { ProgressiveApprovalTierReasonEnum } from 'pages/api/graphql-rest.page.generated';
+import { ProgressiveApprovalTierReasonEnum } from 'src/graphql/types.generated';
 import theme from 'src/theme';
 import { CompleteFormValues } from '../../AdditionalSalaryRequest';
 import { useAdditionalSalaryRequest } from '../../Shared/AdditionalSalaryRequestContext';
--- a/src/components/HrTools/AdditionalSalaryRequest/RequestPage/Helper/CapSubContent.test.tsx
+++ b/src/components/HrTools/AdditionalSalaryRequest/RequestPage/Helper/CapSubContent.test.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { ThemeProvider } from '@mui/material/styles';
 import { render } from '@testing-library/react';
 import { FormikProvider, useFormik } from 'formik';
+import { ProgressiveApprovalTierReasonEnum } from 'pages/api/graphql-rest.page.generated';
 import theme from 'src/theme';
 import { CompleteFormValues } from '../../AdditionalSalaryRequest';
 import { useAdditionalSalaryRequest } from '../../Shared/AdditionalSalaryRequestContext';
@@ -44,15 +45,16 @@ describe('CapSubContent', () => {
     jest.clearAllMocks();
   });
 
-  it('renders Progressive Approvals messaging when hasBoardCapException is false', () => {
+  it('renders Progressive Approvals messaging when reason is not board cap exception', () => {
     mockUseAdditionalSalaryRequest.mockReturnValue({
-      hasBoardCapException: false,
       requestData: {
         latestAdditionalSalaryRequest: {
           progressiveApprovalTier: {
             approver: 'Division Head',
             approvalTimeframe: '1-2 weeks',
           },
+          progressiveApprovalTierReason:
+            ProgressiveApprovalTierReasonEnum.OverUserCap,
         },
       },
     } as unknown as ReturnType<typeof useAdditionalSalaryRequest>);
@@ -67,12 +69,13 @@ describe('CapSubContent', () => {
     ).not.toBeInTheDocument();
   });
 
-  it('renders nothing when hasBoardCapException is true (message is carried by getCapOverrides)', () => {
+  it('renders nothing when reason is board cap exception (message is carried by getCapOverrides)', () => {
     mockUseAdditionalSalaryRequest.mockReturnValue({
-      hasBoardCapException: true,
       requestData: {
         latestAdditionalSalaryRequest: {
           progressiveApprovalTier: null,
+          progressiveApprovalTierReason:
+            ProgressiveApprovalTierReasonEnum.BoardCapException,
         },
       },
     } as unknown as ReturnType<typeof useAdditionalSalaryRequest>);
--- a/src/components/HrTools/AdditionalSalaryRequest/RequestPage/Helper/getCapOverrides.ts
+++ b/src/components/HrTools/AdditionalSalaryRequest/RequestPage/Helper/getCapOverrides.ts
@@ -1,13 +1,13 @@
 import { TFunction } from 'i18next';
+import { ProgressiveApprovalTierReasonEnum } from 'src/graphql/types.generated';
 
 interface CapOverridesOptions {
   splitAsr: boolean;
-  additionalApproval: boolean;
-  hasBoardCapException: boolean;
+  reason: ProgressiveApprovalTierReasonEnum | null | undefined;
 }
 
 export const getCapOverrides = (
-  { splitAsr, additionalApproval, hasBoardCapException }: CapOverridesOptions,
+  { splitAsr, reason }: CapOverridesOptions,
   t: TFunction,
 ) => {
   if (splitAsr) {
@@ -19,27 +19,49 @@ export const getCapOverrides = (
     };
   }
 
-  if (additionalApproval) {
-    if (hasBoardCapException) {
-      return {
-        title: t(
-          'Your request requires Board approval. Please review the information below to continue.',
-        ),
-        content: t(
-          "You have a Board approved Maximum Allowable Salary (CAP) and your Additional Salary Request exceeds that amount. As a result we need to get their approval for this request. We'll forward your request to them and get back to you with their decision.",
-        ),
-      };
-    }
+  if (!reason) {
+    return { title: undefined, content: undefined };
+  }
 
+  if (reason === ProgressiveApprovalTierReasonEnum.BoardCapException) {
+    return {
+      title: t(
+        'Your request requires Board approval. Please review the information below to continue.',
+      ),
+      content: t(
+        "You have a Board approved Maximum Allowable Salary (CAP) and your Additional Salary Request exceeds that amount. As a result we need to get their approval for this request. We'll forward your request to them and get back to you with their decision.",
+      ),
+    };
+  }
+
+  if (reason === ProgressiveApprovalTierReasonEnum.OverlappingRequests) {
     return {
       title: t(
         'Your request requires additional approval. Please fill in the information below to continue.',
       ),
       content: t(
-        'Your request causes your Total Requested Salary to exceed your Maximum Allowable Salary.',
+        'Your spouse has a pending Additional Salary Request or Salary Request, so this request needs additional approval.',
       ),
     };
   }
 
-  return { title: undefined, content: undefined };
+  const title = t(
+    'Your request requires additional approval. Please fill in the information below to continue.',
+  );
+
+  if (reason === ProgressiveApprovalTierReasonEnum.OverCombinedCap) {
+    return {
+      title,
+      content: t(
+        'Your request causes your combined Total Requested Salary to exceed your combined Maximum Allowable Salary.',
+      ),
+    };
+  }
+
+  return {
+    title,
+    content: t(
+      'Your request causes your Total Requested Salary to exceed your Maximum Allowable Salary.',
+    ),
+  };
 };
--- a/src/components/HrTools/AdditionalSalaryRequest/RequestPage/RequestPage.test.tsx
+++ b/src/components/HrTools/AdditionalSalaryRequest/RequestPage/RequestPage.test.tsx
@@ -105,7 +105,6 @@ const defaultMockContextValue = {
   setIsNewAsr: jest.fn(),
   isSpouse: false,
   hasSpouse: false,
-  hasBoardCapException: false,
   fieldConfig: getFieldConfig(i18n.t),
 };
 
--- a/src/components/HrTools/AdditionalSalaryRequest/RequestPage/RequestPage.test.tsx
+++ b/src/components/HrTools/AdditionalSalaryRequest/RequestPage/RequestPage.test.tsx
@@ -10,6 +10,7 @@ import { GqlMockedProvider } from '__tests__/util/graphqlMocking';
 import {
   AsrStatusEnum,
   ElectionType403bEnum,
+  ProgressiveApprovalTierReasonEnum,
 } from 'src/graphql/types.generated';
 import i18n from 'src/lib/i18n';
 import { amount, phoneNumber } from 'src/lib/yupHelpers';
@@ -408,10 +409,10 @@ describe('RequestPage', () => {
       pageType: PageEnum.New,
       requestData: {
         latestAdditionalSalaryRequest: {
-          calculations: {
-            currentSalaryCap: 50,
-          },
+          calculations: {},
           progressiveApprovalTier: { id: 'tier-1' },
+          progressiveApprovalTierReason:
+            ProgressiveApprovalTierReasonEnum.OverUserCap,
         },
       },
     } as unknown as ReturnType<typeof useAdditionalSalaryRequest>);
@@ -452,22 +453,9 @@ describe('RequestPage', () => {
       currentStep: AdditionalSalaryRequestSectionEnum.CompleteForm,
       pageType: PageEnum.New,
       spouse: {
-        currentSalary: { grossSalaryAmount: 40000 },
         staffInfo: {
           preferredName: 'Jane',
           lastName: 'Doe',
-        },
-      },
-      requestData: {
-        latestAdditionalSalaryRequest: {
-          calculations: {
-            currentSalaryCap: 500,
-            combinedCap: 100000,
-          },
-          spouseCalculations: {
-            currentSalaryCap: 50000,
-            pendingAsrAmount: 100,
-          },
         },
       },
     } as unknown as ReturnType<typeof useAdditionalSalaryRequest>);
@@ -526,6 +514,8 @@ describe('RequestPage', () => {
             pendingAsrAmount: 600,
           },
           progressiveApprovalTier: { id: 'tier-1' },
+          progressiveApprovalTierReason:
+            ProgressiveApprovalTierReasonEnum.OverUserCap,
         },
       },
     } as unknown as ReturnType<typeof useAdditionalSalaryRequest>);
@@ -718,7 +708,6 @@ describe('RequestPage', () => {
       ...defaultMockContextValue,
       currentIndex: mockSteps.length - 2,
       currentStep: AdditionalSalaryRequestSectionEnum.CompleteForm,
-      hasBoardCapException: true,
       requestData: {
         latestAdditionalSalaryRequest: {
           id: 'asr-1',
@@ -730,6 +719,8 @@ describe('RequestPage', () => {
             pendingAsrAmount: 0,
           },
           progressiveApprovalTier: { id: 'tier-1' },
+          progressiveApprovalTierReason:
+            ProgressiveApprovalTierReasonEnum.BoardCapException,
         },
       },
     };
@@ -792,6 +783,7 @@ describe('RequestPage', () => {
               currentSalaryCap: 100000,
             },
             progressiveApprovalTier: null,
+            progressiveApprovalTierReason: null,
           },
         },
       } as unknown as ReturnType<typeof useAdditionalSalaryRequest>);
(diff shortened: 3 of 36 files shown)


canac (commented):
## Multi-Agent Code Review — PR #1773

**Mode:** standard (6 specialized agents + gap review)
**Verdict:** APPROVED_WITH_SUGGESTIONS
**Risk Score:** 10/10 — formula-capped; actual blast radius is small (display-logic refactor)

No critical or high-priority blockers. 5 Important findings (severity 7.0–7.9) are strongly recommended before merge — these cannot be dismissed via `/dismiss` (severity ≥ 7), so please either address them in this PR or open follow-up tickets.

### Summary by agent

| Agent | Critical | High | Important | Medium / Suggestions | Confidence |
|-------|----------|------|-----------|----------------------|------------|
| Architecture     | 0 | 0 | 2 | 3 | M |
| Data Integrity   | 0 | 0 | 2 | 3 | H |
| Testing          | 0 | 0 | 2 | 4 | H |
| UX               | 0 | 0 | 2 | 4 | H/M |
| Standards        | 0 | 0 | 0 | 0 | H — all checklist items PASS |
| Financial        | 0 | 0 | 1 | 2 | M/H |
| Gap Review       | 0 | 0 | 0 | 2 | — |

### Cross-cutting observations

- **`{{approvers}}` vs `{{approver}}` (plural vs singular):** `useSubmitDialogContent.ts:43` uses `{{approvers}}` while `Receipt.tsx`, `RequestSummaryCard.tsx`, and `getCapOverrides.ts` use `{{approver}}` (singular, mapped from `progressiveApprovalTier.approver`). Likely an inadvertent typo in the new branch.
- **Copy drift across 4 surfaces for the OverlappingRequests message:** four subtly different wordings — "Your spouse" vs "You or your spouse", "This will take" vs "This may take", and only `getCapOverrides` mentions "or Salary Request". Each variant is its own i18n key (cost) and a future drift risk. Consider consolidating into a single canonical sentence (or a small `useApprovalReasonCopy()` hook) shared across all four surfaces.
- **Server contract dependency:** SOSA online-submission block (`useSosaBlockOverCap.ts`) and several display fallbacks now depend on the server returning the canonical `progressiveApprovalTierReason` for each scenario. The previous code computed `requestedGross > effectiveCap` locally and was deterministic. Confirm with the backend team that the enum values are mutually exclusive, that `OverUserCap` is the canonical reason whenever a user's individual gross exceeds their individual cap (regardless of combined/overlap status), and that the field is always populated when `progressiveApprovalTier` is set.
- **Dependency impact:** clean — no external callers of removed APIs (`overUserCap`, the old `hasBoardCapException` shape, `additionalApproval` param on `getCapOverrides`).

- **`src/components/HrTools/AdditionalSalaryRequest/Shared/HcmData/Hcm.graphql:34` — `boardCapException` field is now orphaned.** Severity 3.5/10 — [Pre-existing]
  - Flagged by: Architecture Agent
  - Recommended action: Consider dropping `boardCapException` from the Hcm fragment in a follow-up PR; this PR removes the only production consumer (only test mocks still reference it).

This is informational only — it does not block this PR or count toward the verdict.

src/components/HrTools/SalaryCalculator/StepNavigation/useSubmitDialogContent.ts
```diff
@@ -32,10 +34,18 @@ export const useSubmitDialogContent = (): DialogContent => {
   }
 
   let subContent: string;
-  if (hasBoardCapException) {
+  if (reason === ProgressiveApprovalTierReasonEnum.BoardCapException) {
     subContent = t(
       "You have a Board approved Maximum Allowable Salary (CAP) and your salary request exceeds that amount. As a result we need to get their approval for this request. We'll forward your request to them and get back to you with their decision.",
     );
+  } else if (reason === ProgressiveApprovalTierReasonEnum.OverlappingRequests) {
```
canac, line 44: [Important] **Title/body mismatch + plural typo + untested branch.**

Three concerns on this new `OverlappingRequests` branch:

1. **Title doesn't match the branch.** The function returns the same `title` and `content` for every non-default case (lines 61–66): *"Your request requires additional approval because your Gross Salary exceeds your Maximum Allowable Salary…"*. That's false for the `OverlappingRequests` case — the user's salary may be fully within cap; the issue is an overlapping pending request. Users will see contradictory copy in the title vs the `subContent`.
2. **`{{approvers}}` vs `{{approver}}`.** This branch uses template var `approvers` (plural) while the analogous templates in `Receipt.tsx`, `RequestSummaryCard.tsx`, and `getCapOverrides.ts` use `approver` (singular, mapped from `progressiveApprovalTier.approver`). Likely a typo; would silently substitute the same singular value either way, but the inconsistency complicates translation.
3. **Untested.** `useSubmitDialogContent.test.tsx` only adds a test for `BoardCapException`. The new `OverlappingRequests` branch (and its template substitutions) has no coverage.

Suggested fix: restructure to compute `title`/`content`/`subContent` together inside each `if (reason === …)` branch (a small lookup or factored helper). Add a colocated test for the `OverlappingRequests` case asserting both `subContent` substitution and `title` correctness.

src/components/HrTools/SalaryCalculator/SalaryCalculation/useSosaBlockOverCap.ts
```diff
@@ -15,12 +15,14 @@ interface UseSosaBlockOverCapResult {
...
  */
 export const useSosaBlockOverCap = (): UseSosaBlockOverCapResult => {
-  const { hcmUser } = useSalaryCalculator();
-  const { overUserCap } = useCaps();
+  const { hcmUser, calculation } = useSalaryCalculator();
 
   const isUserSosa =
     hcmUser?.staffInfo.userPersonType ===
     UserPersonTypeEnum.EmployeeStaffNonRmoSpouse;
+  const overUserCap =
```
canac, line 25: [Important] **SOSA online-submission safety net now depends on server contract.**

Previously `blockOnCap = isUserSosa && (calcs.requestedGross > calcs.effectiveCap)` — a deterministic client-side guard that fired whenever the saved gross exceeded the effective cap, by direct comparison. Now it fires only when the server returns `progressiveApprovalTierReason === OverUserCap`.

This is a behavioral contract change with a real failure mode: if the server classifies a SOSA user's over-cap request with a different reason (e.g. `OverCombinedCap` for a married SOSA where both individual and combined caps are exceeded, or `OverlappingRequests`, or returns `null` transiently), the block disengages and a SOSA user can submit an over-cap request online — exactly the case the block was designed to prevent. The docstring on line 14 still promises *"Determine whether a SOSA user's requested gross salary exceeds their effective cap,"* which the implementation no longer answers.

**Recommend one of:**
- Confirm with the backend team that `OverUserCap` is the canonical reason whenever an individual gross exceeds an individual cap (regardless of other concurrent conditions) — and update the docstring to make the contract dependency explicit.
- Keep the local `requestedGross > effectiveCap` check as a defensive belt-and-suspenders fallback (cheap to maintain, prevents silent regression).
- Treat `null` reason for a SOSA user with a tier set as "block, unknown state" rather than "allow submission."

Also add a negative-case regression test to `useSosaBlockOverCap.test.tsx`: SOSA user, numerically over cap, reason is `OverCombinedCap` (or null) → assert behavior is what you intend (block or pass).

src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestSummaryCard/RequestSummaryCard.tsx
```diff
@@ -95,14 +98,22 @@ export const RequestSummaryCard: React.FC = () => {
...
-  ) : approvalRequired ? (
+  ) : reason === ProgressiveApprovalTierReasonEnum.OverlappingRequests ? (
+    <Trans t={t}>
+      You or your spouse has a pending Additional Salary Request, so this
+      request needs additional approval. This may take{' '}
+      {{ timeframe: progressiveApprovalTier.approvalTimeframe }} as it needs to
+      be signed off by the {{ approver: progressiveApprovalTier.approver }}.
+      This may affect your selected effective date.
+    </Trans>
+  ) : reason === ProgressiveApprovalTierReasonEnum.OverCombinedCap ? (
```
canac, line 116: [Important] **Fallback branch can render `undefined`'s name and cap.**

The final ternary branch (lines 127–138, just below this one) uses `{{ name: overCapPerson?.name }}` and `{{ cap: overCapPerson?.effectiveCap }}`. It is reached whenever `progressiveApprovalTier` is truthy AND `reason` is not `BoardCapException` / `OverlappingRequests` / `OverCombinedCap` — today that means `OverUserCap`, `OverSpouseCap`, or anything else (including `null` or a future enum value).

`overCapPerson` is only populated when `reason` is exactly `OverUserCap` or `OverSpouseCap` (per `useCaps.ts`). So the alignment holds today, but:
- If the server returns `progressiveApprovalTier` non-null with `progressiveApprovalTierReason` null (the schema marks it `Maybe<…>`), or if a new enum value is added later, the user sees a sentence with `undefined`'s name and a missing `{cap}` interpolation.
- Even within `OverUserCap`/`OverSpouseCap`, `overCapPerson.name` is typed `string | null` (falls back to `null` if `staffInfo.preferredName` is missing) — would render "null's Gross Requested Salary…".

**Fix:** tighten this to an explicit `reason === OverUserCap || reason === OverSpouseCap` check before this branch, plus a final defensive fallback to a generic message (or render nothing) for unknown reasons. Also add a regression test covering "tier present, reason null/unknown".

src/components/HrTools/AdditionalSalaryRequest/RequestPage/Helper/getCapOverrides.ts
```diff
@@ -19,27 +19,49 @@ export const getCapOverrides = (
...
+      title: t(
+        'Your request requires Board approval. Please review the information below to continue.',
+      ),
+      content: t(
+        "You have a Board approved Maximum Allowable Salary (CAP) and your Additional Salary Request exceeds that amount. As a result we need to get their approval for this request. We'll forward your request to them and get back to you with their decision.",
+      ),
+    };
+  }
 
+  if (reason === ProgressiveApprovalTierReasonEnum.OverlappingRequests) {
```
canac, line 37: [Important] **New `OverlappingRequests` branch has no test, and `OverSpouseCap` silently falls through.**

1. **Untested:** `getCapOverrides.test.ts` covers `null`, `OverUserCap`, `OverCombinedCap`, and `BoardCapException`, but not `OverlappingRequests`. Without a test, a typo or copy drift here ships unnoticed.
2. **`OverSpouseCap` fallthrough:** after handling `BoardCapException`, `OverlappingRequests`, and `OverCombinedCap` explicitly, the function returns the generic *"Your request causes your Total Requested Salary to exceed your Maximum Allowable Salary"* content for any other reason — which now includes `OverSpouseCap`. Reads as if it's about the user, not the spouse.
3. **Copy divergence:** this surface says *"Your spouse has a pending Additional Salary Request **or Salary Request**"* — only place that mentions both ASR and SR. The other three surfaces (`Receipt.tsx`, `RequestSummaryCard.tsx`, `useSubmitDialogContent.ts`) say only "Additional Salary Request". Pick one canonical wording.

**Suggested patch:**
```ts
it('returns overlapping-requests messaging when reason is OverlappingRequests', () => { /* … */ });
it('returns spouse-specific messaging when reason is OverSpouseCap', () => { /* … */ });
```

And consider adding an explicit `OverSpouseCap` branch with spouse-appropriate copy, or document that the generic copy is intentional for both `OverUserCap` and `OverSpouseCap`.

src/components/HrTools/AdditionalSalaryRequest/Shared/AdditionalSalaryRequestContext.tsx
```diff
@@ -255,7 +256,9 @@ export const AdditionalSalaryRequestProvider: React.FC<Props> = ({
       : spousePerson
     : undefined;
   const hasBoardCapException =
```
canac: [Important] **Semantic drift on `hasBoardCapException` — same name, new meaning.**

The boolean keeps its name but the source-of-truth changed from a *user-level privilege* (`user?.exceptionSalaryCap?.boardCapException`) to a *per-request server outcome* (`reason === BoardCapException`). The two are not equivalent:
- A user can have the board-cap-exception privilege without the current request triggering board-cap-exception routing.
- A request can come back with `reason === BoardCapException` for reasons that don't depend on the user's static privilege.
- During the request lifecycle, there are moments (no `latestAdditionalSalaryRequest` yet, mutation in flight, network error) where `reason` is undefined for a user who *does* have the privilege.

Downstream consumer `useAdditionalSalaryRequestForm.ts:164` uses this to skip the `additionalInfo` Yup rule. With the new derivation, that validation gate now toggles based on a server-derived field that may transiently be wrong or undefined — silently re-imposing (or skipping) a validation that the user previously did not see toggle.

**Recommend:**
- Rename to `requestReasonIsBoardCap` (or similar) to reflect the new semantics, and audit each consumer for desired behavior; or
- Keep `hasBoardCapException` derived from the user privilege and introduce a separate `progressiveApprovalReason` field for display-routing concerns.

Verify with product whether the form-validation skip should track the user privilege or the request outcome — both are defensible, but they're different rules.

src/components/HrTools/SalaryCalculator/StepNavigation/useSubmitDialogContent.ts
```diff
@@ -12,16 +15,15 @@ interface DialogContent {
...
 
   const progressiveApprovalTier = calculation?.progressiveApprovalTier;
+  const reason = calculation?.progressiveApprovalTierReason;
   const approvalRequired =
     !!progressiveApprovalTier &&
     progressiveApprovalTier.tier !== ProgressiveApprovalTierEnum.DivisionHead;
-  const hasBoardCapException =
-    hcmUser?.exceptionSalaryCap.boardCapException ?? false;
 
   if (!approvalRequired) {
```
canac, line 28: [Medium] **Gating still uses legacy `approvalRequired = tier !== DivisionHead`.**

The early-return on line 28 gates the entire branch chain on the legacy boolean. After this PR, `RequestSummaryCard.tsx` may render a non-trivial approval message for `tier === DivisionHead` paired with `reason === OverUserCap`/`OverSpouseCap`/`OverlappingRequests`, but the submit dialog will fall through to the generic "Your request will be sent to HR Services" copy. The two surfaces can disagree about whether approval is needed.

Consider gating on `reason` (or on `reason != null && tier`) to keep the submit dialog consistent with the status card the user just read.

src/components/HrTools/AdditionalSalaryRequest/RequestPage/Helper/CapSubContent.test.tsx
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import { ThemeProvider } from '@mui/material/styles';
 import { render } from '@testing-library/react';
 import { FormikProvider, useFormik } from 'formik';
+import { ProgressiveApprovalTierReasonEnum } from 'pages/api/graphql-rest.page.generated';
```
canac: [Suggestion] **Inconsistent enum import source.**

This test file imports `ProgressiveApprovalTierReasonEnum` from `pages/api/graphql-rest.page.generated`. The corresponding component (`CapSubContent.tsx`) and every other call site in this PR import from `src/graphql/types.generated`.

Both `.generated.ts` files declare identical enum values today, so runtime equality works. But TypeScript treats them as nominally distinct, and `rootFields.generated.ts` indicates the salary-request fields route to the primary GraphQL API server — so `src/graphql/types.generated` is the canonical source for client code.

**Fix:**
```ts
-import { ProgressiveApprovalTierReasonEnum } from 'pages/api/graphql-rest.page.generated';
+import { ProgressiveApprovalTierReasonEnum } from 'src/graphql/types.generated';
```

src/components/HrTools/SalaryCalculator/SalaryCalculation/ApprovalProcessCard/ApprovalProcessCard.tsx
```diff
@@ -53,13 +53,14 @@ export const ApprovalProcessCard: React.FC = () => {
       />
       <CardContent>
         <Typography data-testid="ApprovalProcessCard-status">
-          {tier === ProgressiveApprovalTierEnum.DivisionHead ? (
+          {tier === ProgressiveApprovalTierEnum.DivisionHead &&
```
canac, line 56: [Suggestion] **Tightened guard creates an untested branch with potentially contradictory copy.**

The guard changed from `tier === DivisionHead ?` to `tier === DivisionHead && overCapPerson ?` — a sound defensive improvement, but the case `tier === DivisionHead && !overCapPerson` now falls through to the `spouseName ? … : …` branches. Combined with the second Typography below ("Since your combined request is still within your combined Max Allowable Salary, no additional approvals are required"), a user could see one paragraph saying approval is required and another saying it isn't.

This combination only occurs if the server can return `tier === DivisionHead` with a reason other than `OverUserCap`/`OverSpouseCap` (e.g. `OverCombinedCap`). If the backend invariant is "DivisionHead ⇒ reason ∈ {OverUserCap, OverSpouseCap}", this is defensive-only and worth a code comment. If not, align the second Typography to also key off `reason`.

Either way, add a test for `tier: DivisionHead` with `reason: OverCombinedCap` (or `reason: null`).

src/components/HrTools/SalaryCalculator/Receipt/Receipt.tsx
```diff
@@ -49,13 +50,23 @@ export const ReceiptStep: React.FC = () => {
               It will be processed by HR Services within the next 2-3 business
               days. Please print a copy for your records.
             </Trans>
-          ) : boardCapException ? (
+          ) : reason === ProgressiveApprovalTierReasonEnum.BoardCapException ? (
             <Trans t={t}>
               You have a Board approved Maximum Allowable Salary (CAP) and your
               salary request exceeds that amount. As a result we need to get
               their approval for this request. We&apos;ll forward your request
               to them and get back to you with their decision.
```
canac, line 58: [Suggestion] **Copy drift — consider consolidating the OverlappingRequests message.**

Four surfaces now have subtly different wordings of the same "pending Additional Salary Request" message:
- `Receipt.tsx`: "Because you or your spouse has a pending Additional Salary Request, this request requires additional approval. **This will take** {{timeframe}}…"
- `RequestSummaryCard.tsx`: "You or your spouse has a pending Additional Salary Request, so this request needs additional approval. **This may take** {{timeframe}}…"
- `useSubmitDialogContent.ts`: "You or your spouse has a pending Additional Salary Request, so this request needs additional approval. **This will take** {{timeframe}}…"
- `getCapOverrides.ts`: "**Your spouse** has a pending Additional Salary Request **or Salary Request**, so this request needs additional approval." (no timeframe; mentions both ASR and SR)

Each distinct English string is its own i18n key (4× translation cost) and a future-drift risk. Recommend picking one canonical wording ("You or your spouse has…", "This will take…") and reusing it via a shared helper or `useApprovalReasonCopy()` hook.

Also: clarify whether the "or Salary Request" clause in `getCapOverrides` is intentional (the server condition fires for both) or a copy oversight (every other surface mentions only ASR).

src/components/HrTools/AdditionalSalaryRequest/RequestPage/Helper/getCapOverrides.ts
```diff
@@ -1,13 +1,13 @@
...
 interface CapOverridesOptions {
   splitAsr: boolean;
-  additionalApproval: boolean;
-  hasBoardCapException: boolean;
+  reason: ProgressiveApprovalTierReasonEnum | null | undefined;
 }
 
 export const getCapOverrides = (
-  { splitAsr, additionalApproval, hasBoardCapException }: CapOverridesOptions,
+  { splitAsr, reason }: CapOverridesOptions,
```
wjames111, line 10: Question: Is there anyway we could update the back-end so it already returns the reason title and reason content directly?

src/components/HrTools/AdditionalSalaryRequest/RequestPage/RequestPage.test.tsx
```diff
@@ -452,24 +452,11 @@ describe('RequestPage', () => {
...
-          calculations: {
-            currentSalaryCap: 500,
-            combinedCap: 100000,
-          },
-          spouseCalculations: {
-            currentSalaryCap: 50000,
-            pendingAsrAmount: 100,
-          },
-        },
-      },
```
wjames111, line 472: Praise: Switching over to using the approval reason from the server simplifies so many things!

wjames111 (approved):
Approved! Approved! Approved! Looks great though, simplifies things a lot.

src/components/HrTools/AdditionalSalaryRequest/RequestPage/Helper/getCapOverrides.ts
canac: I was thinking that for localization it's best to return a reason code enum and let the UI convert it to a human-readable message.

wjames111: Ahh yeah that's a fair point.

PR #1774, opened 2026-05-13
[MPDX-9536, MPDX-9570, MPDX-9572, MPDX-9573] - PDS fix calculations

wjames111:
## Description
This work aims to fix some of the calculations we came across in our review Monday. I've listed out the following tickets and their descriptions:
[MPDX-9536](https://jira.cru.org/browse/MPDX-9536) - Fix Goal Amount in the Goal Card. This number should match what's inside the form
[MPDX-9570](https://jira.cru.org/browse/MPDX-9570) - Bug showing 600% on Credit Card Fees. This appears to fixed although I don't believe I made changes to address it. Either someone already fix, or there was a weird merge conflict that caused the 600 bug.
[MPDX-9572](https://jira.cru.org/browse/MPDX-9572) - switch order of attrition, credit card fees, and assessment so it matches the order they show in the table. This has been corrected, but the formula itself has been updated as well so the label now reflects the updated formula.
[MPDX-9573](https://jira.cru.org/browse/MPDX-9573) - Assessment is wrong calculation. I triple checked the math, it should be correct based on the formula provided in the spreadsheet. However the number does not match the spreadsheet even though it should, I believe something is not working correctly in the spreadsheet.

## Testing

- Go to PDS GC
- Check each ticket is fixed

## Checklist:


src/components/HrTools/PdsGoalCalculator/calculations/salaryCalculation.ts
```diff
@@ -31,7 +31,7 @@ export const calculateSalaryTotals = (
     calculation.salaryOrHourly === DesignationSupportSalaryType.Salaried;
 
   const monthlyBase = isSalaried ? payRate / 12 : (payRate * hours * 52) / 12;
-  const grossMonthlyPay = monthlyBase * (1 + geographicMultiplier);
+  const grossMonthlyPay = monthlyBase * geographicMultiplier;
```
wjames111: **[Critical] F1: Geographic-multiplier convention split** — sev 9.5/10

**Flagged by:** Financial Reporting (9.5), Architecture (7.5, consolidated), Testing (supports). Independently verified.

This PR changes the formula to `monthlyBase * geographicMultiplier`, treating the map value as a full multiplicative factor (e.g. `1.06` for a 6% high-cost location, default `1`).

But the sibling calculator at [calculateNewStaffTotals.ts:144](src/components/HrTools/GoalCalculator/Shared/calculateNewStaffTotals.ts#L144) reads the SAME `goalGeographicConstantMap` as an additive delta:

```ts
return (geographicMultiplier + 1) * base + tenureIncrease + annualDebtSalary;
```

with default `0`. Both maps are populated from the same backend field `mpdGoalGeographicConstants.percentageMultiplier` via [useGoalCalculatorConstants.ts:50-54](src/hooks/useGoalCalculatorConstants.ts#L50-L54). Only one interpretation matches production data — the other silently produces grossly wrong salaries.

**Impact:**
- If the API returns `0.06` (additive): PDS multiplies salary by `0.06` → **~94% understatement** of monthly salary for Orlando staff.
- If the API returns `1.06` (multiplicative): the New Staff calculator computes `(1.06 + 1) * base` → **doubles compensation**.

Test fixtures themselves prove the divergence:
- `PdsGoalCalculatorTestWrapper.tsx:194-205` uses `1`, `1.06`, `1.12` (new)
- `GoalCalculatorTestWrapper.tsx:191` uses `0.06` (old)
- `useGoalCalculatorConstants.test.tsx:42-48` uses `0`, `0.12` (old)

**Required action — pick ONE before merge:**
1. **Align New Staff to multiplicative** (this PR's chosen convention): update `calculateNewStaffTotals.ts:144` to `geographicMultiplier * base + tenureIncrease + annualDebtSalary`, default `1`. Update `GoalCalculatorTestWrapper.tsx` and `useGoalCalculatorConstants.test.tsx` fixtures.
2. **Revert PDS to additive**: restore `monthlyBase * (1 + geographicMultiplier)` here and `?? 0` default in `pdsGoalConstants.ts:50`.
3. **Normalize at the hook** (architecturally preferred): translate at `useGoalCalculatorConstants.ts:50-54` so both downstream calculators consume one canonical form.

Document the choice and the backend value in the PR description for future maintainers.

src/components/HrTools/PdsGoalCalculator/calculations/pdsGoalConstants.ts
```diff
@@ -60,8 +44,10 @@ export const buildPdsGoalConstants = (
     return null;
   }
 
+  // Multiplier is the *full* factor applied to monthlyBase (e.g. 1.06 for a
+  // 6% high-cost location), so the no-adjustment default must be 1, not 0.
   const geographicMultiplier =
-    goalGeographicConstantMap.get(geographicLocation ?? '') ?? 0;
+    goalGeographicConstantMap.get(geographicLocation ?? '') ?? 1;
```
wjames111: **[Critical] F1: Default changed from 0 → 1; convention split with sister calculator** — sev 9.5/10

The `?? 1` default (and the comment at lines 47-48 explaining "full factor") is the source side of the F1 break. See the full finding on [salaryCalculation.ts:34](src/components/HrTools/PdsGoalCalculator/calculations/salaryCalculation.ts#L34) — same blocker, two source locations.

This line and `salaryCalculation.ts:34` must move together with whichever resolution wins, and the test fixture in `buildPdsGoalConstants.test.ts` (still asserting `0.06`/`0.12`) is now semantically contradictory with the comment above this line.

src/components/HrTools/PdsGoalCalculator/SupportItem/salaryBreakdown.tsx
```diff
@@ -77,12 +77,12 @@ export const buildSalaryBreakdownRows = (
       id: 'geographic-multiplier',
       category: t('Geographic Multiplier'),
       amount: geographicMultiplier,
-      format: 'percentage',
+      format: 'number',
```
wjames111: **[Important] U1+T5+F5: Geographic Multiplier renders as bare decimal "1.06" instead of "6%"** — sev 7.5/10

**Flagged by:** UX (7.5), Testing (6.0), Financial (5.0). Architecture supports — calls this "the missing test that would have caught F1."

**Issue:**
- Was `format: 'percentage'` (rendered "6.00%") → now `format: 'number'` (renders `numberFormat(1.06, locale)` = `"1.06"`).
- The "Amount" column elsewhere in this same DataGrid contains currencies (Pay Rate, Monthly Base, Gross Monthly Pay, Employer ½ FICA, Subtotal). A bare `"1.06"` sitting in a currency column is a perceptual trap — staff can easily read it as `$1.06`.
- The sibling `employer-fica` row at lines 93-95 demonstrates the correct pattern: rate goes in the `formula`, formatted with `percentageFormat`, NOT in the `amount` column.
- Codebase convention for rate-like values: `SummaryReport/PdsSummaryTable.tsx:69` and `SummaryReport/PdsSummaryHeaderCards.tsx:41` use `percentageFormat`.
- **No test asserts the rendered cell content** — a `getByTestId` assertion on the gridcell text would have caught this AND F1 simultaneously.

**Recommended fix (preferred):** move the multiplier into the `gross-monthly-pay` row's `formula` text using `percentageFormat`, and remove the standalone Geographic Multiplier row. Matches the employer-fica pattern.

**Alternative:** keep the row but display `(multiplier - 1)` as a percentage:
```ts
{
  id: 'geographic-multiplier',
  category: t('Geographic Multiplier'),
  amount: geographicMultiplier - 1,
  format: 'percentage',
},
```

Add a rendered test in `salaryBreakdown.test.tsx` asserting the cell text — e.g. `"+6%"` or `"1.06"` depending on which fix you ship.

src/components/HrTools/PdsGoalCalculator/calculations/usePdsSummaryData.test.ts
```diff
@@ -314,12 +310,13 @@ describe('usePdsSummaryData', () => {
...
-      // assessment = (7800 + 468 + 496.08) * 0.12 = 1051.69
-      // overallTotal = 7800 + 468 + 496.08 + 1051.69 = 9815.77
+      // adminBase = 7800 + 468 + 496.08 = 8764.08
+      // assessment = adminBase / 0.88 - adminBase ≈ 1195.10
+      // overallTotal = 7800 + 468 + 496.08 + 1195.10 ≈ 9959.18
       const { result } = renderHook(() =>
         usePdsSummaryData(defaultCalculation, defaultHcmUser),
       );
-      expect(result.current?.overallTotal).toBeCloseTo(9815.77, 0);
+      expect(result.current?.overallTotal).toBeCloseTo(9959.18, 0);
```
wjames111: **[Important] T1/T2/T3: Three e2e tests deleted with `calculatePdsGoalTotal.test.ts`, not migrated** — sev 7.0/10

**Flagged by:** Testing (7.5/7.0/6.5), Financial (supports — these are the root cause that let F1 land).

The deleted `calculatePdsGoalTotal.test.ts` had 6 cases; only 3 were migrated to this file (full-time salaried, Simple vs Detailed formType, legacy null formType). NOT migrated:

1. **T1 — Part-time hourly e2e** (sev 7.5). Part-time hourly is the second-most-common case and the most error-prone salary path because of `(payRate * hours * 52) / 12` annualization.

2. **T2 — `payRate: null` propagation** (sev 7.0). Currently no e2e regression test prevents `NaN` from propagating through `overallTotal`.

3. **T3 — Geographic-multiplier invariant `withGeo > withoutGeo`** (sev 6.5). This was the canonical guard that would have caught F1 directly. The existing `geographic multiplier` block at lines 203-232 only verifies the multiplier value flows through; it never asserts the multiplier affects `overallTotal`.

**Recommended test skeleton (add after the existing `overallTotal` test):**

```ts
it('computes correct overallTotal for a part-time hourly employee', () => {
  const calc = {
    ...defaultCalculation,
    salaryOrHourly: DesignationSupportSalaryType.Hourly,
    status: DesignationSupportStatus.PartTime,
    payRate: 25,
    hoursWorkedPerWeek: 20,
    benefits: null,
  };
  const { result } = renderHook(() => usePdsSummaryData(calc, defaultHcmUser));
  expect(result.current?.overallTotal).toBeCloseTo(/* compute */, 0);
});

it('returns a finite overallTotal when payRate is null', () => {
  const calc = { ...defaultCalculation, payRate: null };
  const { result } = renderHook(() => usePdsSummaryData(calc, defaultHcmUser));
  expect(Number.isFinite(result.current?.overallTotal)).toBe(true);
  expect(result.current?.overallTotal).toBeGreaterThanOrEqual(0);
});

it('produces a greater overallTotal when geographicLocation has uplift', () => {
  const withoutGeo = renderHook(() =>
    usePdsSummaryData({ ...defaultCalculation, geographicLocation: null }, defaultHcmUser),
  );
  const withGeo = renderHook(() =>
    usePdsSummaryData({ ...defaultCalculation, geographicLocation: 'Orlando, FL' }, defaultHcmUser),
  );
  expect(withGeo.result.current!.overallTotal).toBeGreaterThan(
    withoutGeo.result.current!.overallTotal,
  );
});
```

src/components/HrTools/PdsGoalCalculator/SupportItem/otherBreakdown.tsx
```diff
@@ -126,9 +126,15 @@ export const buildOtherBreakdownRows = (
     {
       id: 'assessment',
       category: t('Assessment'),
-      formula: t('(Subtotal + Credit Card Fees + Attrition) × {{rate}}', {
-        rate: percentageFormat(constants.adminRate, locale),
-      }),
+      formula: t(
+        '(Subtotal + Attrition + Credit Card Fees) ÷ {{divisor}} − (Subtotal + Attrition + Credit Card Fees)',
```
wjames111: **[Medium] T4: New assessment formula text has no rendered test + U2 readability** — sev 6.5/10

**Flagged by:** Testing (6.5), UX (6.5), Financial (4.0).

The formula string `'(Subtotal + Attrition + Credit Card Fees) ÷ {{divisor}} − (Subtotal + Attrition + Credit Card Fees)'` is ~88 chars after interpolation, duplicates the parenthesized subtotal, and will wrap/truncate in the DataGrid `category` column. Screen readers will recite the duplicated subtotal verbatim. None of the tests in `otherBreakdown.test.tsx` render this row and assert on the formula text.

**Recommended fix:** reframe the formula around `adminRate` itself (so the project's `percentageFormat` helper applies — closes the inline-`Intl.NumberFormat` smell at line 132 too):

```ts
formula: t(
  '(Subtotal + Attrition + Credit Card Fees) ÷ (1 − {{rate}})',
  { rate: percentageFormat(constants.adminRate, locale) },
),
```

This is shorter, matches the `attritionRate` (line 110) and `creditCardFeeRate` (line 120) patterns in the same file, and reads `"12%"` instead of `"0.88"` — staff already understand the admin rate as a percentage.

Add a render test in `otherBreakdown.test.tsx` asserting the formula renders as expected for `adminRate: 0.12`.

src/components/HrTools/PdsGoalCalculator/GoalCard/PdsGoalCard.tsx
```diff
@@ -24,23 +21,12 @@ export const PdsGoalCard: React.FC<PdsGoalCardProps> = ({ goal }) => {
...
-  const goalTotal = useMemo(() => {
-    const constants = buildPdsGoalConstants(
-      goalMiscConstants,
-      goalGeographicConstantMap,
-      goal.geographicLocation,
-      hcmUser?.fourOThreeB,
-    );
-    return constants ? calculatePdsGoalTotal(goal, constants) : 0;
-  }, [goal, goalMiscConstants, goalGeographicConstantMap, hcmUser]);
+  const summaryData = usePdsSummaryData(goal, hcmUser);
```
wjames111: **[Medium] C2: Redundant `useGoalCalculatorConstants` subscription** — sev 6.5/10

**Flagged by:** Architecture (6.5).

Line 24 calls `useGoalCalculatorConstants()` only to read `loading`, and line 28's `usePdsSummaryData(goal, hcmUser)` calls the same hook again internally (`usePdsSummaryData.ts:39`). Apollo's `cache-first` policy means no extra network hit, but you still have two subscriptions, two re-renders per cache update, and an architectural coupling smell — the card knows about an implementation detail of `usePdsSummaryData`.

More importantly: the parent's `loading` flag and the child's `summaryData === null` state can diverge. If the cache returns partial data such that `buildPdsGoalConstants` returns `null` (see [pdsGoalConstants.ts:37-45](src/components/HrTools/PdsGoalCalculator/calculations/pdsGoalConstants.ts#L37-L45)) but the query has resolved, the card renders `goalTotal = 0` without `loading=true`.

**Recommended fix:** have `usePdsSummaryData` return `{ data, loading }`:

```ts
// usePdsSummaryData.ts
export const usePdsSummaryData = (...) => {
  const { goalMiscConstants, goalGeographicConstantMap, loading } = useGoalCalculatorConstants();
  const data = useMemo(() => /* ... */, [...]);
  return { data, loading };
};

// PdsGoalCard.tsx
const { data: summaryData, loading: constantsLoading } = usePdsSummaryData(goal, hcmUser);
const goalTotal = summaryData?.overallTotal ?? 0;
```

Drop the direct `useGoalCalculatorConstants()` call on line 24.

src/components/HrTools/PdsGoalCalculator/calculations/buildPdsGoalConstants.test.ts
```diff
@@ -108,37 +108,37 @@ describe('buildPdsGoalConstants', () => {
...
+  it('defaults geographicMultiplier to 1 (no adjustment) for unknown location', () => {
     const result = buildPdsGoalConstants(
       buildMiscConstants(),
       defaultGeoMap,
       'US-XX',
       null,
     );
 
-    expect(result?.geographicMultiplier).toBe(0);
+    expect(result?.geographicMultiplier).toBe(1);
```
wjames111: **[Medium] F3: Test fixture `defaultGeoMap` still uses old additive values** — sev 6.0/10

**Flagged by:** Financial Reporting (6.0), Architecture (3.5), Testing (verified).

The new comment at [pdsGoalConstants.ts:47-48](src/components/HrTools/PdsGoalCalculator/calculations/pdsGoalConstants.ts#L47-L48) explicitly states "1.06 for a 6% high-cost location" (multiplicative). But the test fixture in this file (lines 46-49) still uses:

```ts
const defaultGeoMap: GoalGeographicConstantMap = new Map([
  ['US-CO', 0.06],
  ['US-NY', 0.12],
]);
```

and the assertion at line 67 still checks `geographicMultiplier: 0.06`, while line 152 asserts `0.12`. The tests pass only because `buildPdsGoalConstants` is a pure pass-through of map values.

With the new multiplicative semantic, fixtures `'US-CO': 0.06` describe a state where Colorado staff get 6% of their base salary — a 94% pay cut. The fixture documents the WRONG contract; a future developer reading these tests will assume additive values are correct.

**Recommended fix:** update fixture values to match the new convention:

```ts
const defaultGeoMap: GoalGeographicConstantMap = new Map([
  ['US-CO', 1.06],
  ['US-NY', 1.12],
]);
```

Then update the assertions at lines 67 and 152 to `1.06` and `1.12` respectively. (This same fix applies whether F1 is resolved by aligning New Staff to multiplicative or by reverting PDS to additive — in either case the fixtures should match the chosen convention.)

wjames111 (commented):
## Multi-Agent Code Review — PR #1774

**Mode:** standard · **Agents:** 5 specialized + 1 gap-review (Opus)

**Verdict: BLOCKERS_FOUND** — 1 critical issue must be resolved before merge.

### Top blockers

- **F1 (sev 9.5):** [salaryCalculation.ts:34](src/components/HrTools/PdsGoalCalculator/calculations/salaryCalculation.ts#L34) and [pdsGoalConstants.ts:50](src/components/HrTools/PdsGoalCalculator/calculations/pdsGoalConstants.ts#L50) reinterpret `goalGeographicConstantMap` as multiplicative (`1.06`), but the sibling [calculateNewStaffTotals.ts:144](src/components/HrTools/GoalCalculator/Shared/calculateNewStaffTotals.ts#L144) still reads the same field as additive (`(geographicMultiplier + 1) * base`, default `0`). Same backend field — two interpretations. Only one matches production data; the other is silently wrong by ~94% of salary.

### Risk Assessment

- Risk Score: **7/10** · Risk Level: **HIGH** · Required Reviewer: **SENIOR**
- Financial calculation semantics change for HR goal totals. A misunderstanding silently mis-states staff support targets.

### Findings summary

| Tier | Count |
|------|-------|
| Critical Blocker (9.0-10.0) | 1 |
| High Priority Blocker (8.0-8.9) | 0 |
| Important (7.0-7.9) | 3 |
| Medium (5.0-6.9) | 5 |
| Suggestions (<5.0) | 5 |

See inline comments for line-anchored details and recommended fixes.

---

These findings were surfaced during F1 investigation. They are on files outside the PR diff and cannot be posted as line comments. They are informational — they do not count toward the blocker count.

**[Critical 9.5/10] src/components/HrTools/GoalCalculator/Shared/calculateNewStaffTotals.ts:127-144** — Still uses additive convention `(geographicMultiplier + 1) * base` with `?? 0` default. Must be reconciled with whichever convention wins F1. If the multiplicative semantic ships (this PR's choice), this file silently doubles compensation for any New Staff goal with a known geographic location.

**[Medium 6.0/10] src/components/HrTools/GoalCalculator/GoalCalculatorTestWrapper.tsx:191** — Test fixture still uses additive value `percentageMultiplier: 0.06`. Must be updated alongside any convention decision.

**[Medium 6.0/10] src/hooks/useGoalCalculatorConstants.test.tsx:42,48** — Test fixture uses additive values (`0` and `0.12`). Same.

**[Suggestion 3.0/10] src/components/HrTools/PdsGoalCalculator/SupportItem/salaryBreakdown.test.tsx:144** — Stale formula string `'Pay Rate × (1 + Geographic Multiplier)'` no longer matches any production formula. (Line is unchanged context, not in the diff, so not flagged inline.) Update to `'Monthly Base × Geographic Multiplier'`.

**[Suggestion 3.5/10] Test file rename inconsistency** — `buildPdsGoalConstants.test.ts` was not renamed to `pdsGoalConstants.test.ts` to match the source rename. Pre-existing.

**[Suggestion 3.0/10] src/components/HrTools/PdsGoalCalculator/calculations/pdsGoalConstants.ts:10** — `PdsGoalTotalConstants` interface name is stale after `calculatePdsGoalTotal` was removed; consider renaming to `PdsGoalConstants`. (Line 10 is unchanged context, not in the diff.)

---

### Cross-cutting consistency check

**Operation:** "compute geographic salary adjustment"

| Code path | Default | Formula | Convention |
|-----------|---------|---------|------------|
| `PdsGoalCalculator/.../salaryCalculation.ts:34` | `?? 1` | `monthlyBase * geographicMultiplier` | Multiplicative |
| `GoalCalculator/Shared/calculateNewStaffTotals.ts:144` | `?? 0` | `(geographicMultiplier + 1) * base` | Additive |

Both read from the **same** `goalGeographicConstantMap` populated by [useGoalCalculatorConstants.ts:50-54](src/hooks/useGoalCalculatorConstants.ts#L50-L54). Resolution requires backend confirmation of what `mpdGoalGeographicConstants.percentageMultiplier` actually returns, then aligning whichever calculator is wrong (and the three test fixtures noted above).

**Architecturally preferred resolution:** Normalize at the hook boundary in `useGoalCalculatorConstants.ts:50-54` and use a branded type (`type GeographicFactor = number & { __brand: 'GeographicFactor' }`) so downstream calculators cannot reinterpret it silently.

---

*Generated by `/quality:agent-review` (standard mode, Opus). Reply `/dismiss: <reason>` on any non-blocking comment (severity < 7) to acknowledge.*

src/components/HrTools/PdsGoalCalculator/calculations/OtherExpenses.ts
```diff
@@ -55,8 +55,10 @@ export const calculateOtherExpenses = (
 
   const attrition = subtotal * constants.attritionRate;
   const creditCardFees = (subtotal + attrition) * constants.creditCardFeeRate;
-  const adminRate = constants.adminRate;
-  const assessment = (subtotal + creditCardFees + attrition) * adminRate;
+  const adminBase = subtotal + creditCardFees + attrition;
+  // Admin assessment is `adminRate` of the post-admin total, not a markup on
+  // `adminBase`, so gross up: assessment / (adminBase + assessment) = adminRate.
+  const assessment = adminBase / (1 - constants.adminRate) - adminBase;
```
wjames111, line 63: **[Suggestion] F2/T6: Gross-up formula has no guard against `adminRate >= 1`** — sev 4.0/10

**Flagged by:** Financial Reporting (revised 7.5 → 3.5 after debate), Testing (revised 4.5 → 3.0), Architecture (5.0), gap-review (3.0). Mostly downgraded to suggestion tier.

**Math verified correct:** `assessment / (adminBase + assessment) = adminRate` for any `0 ≤ adminRate < 1`. For `adminRate=0.12, adminBase=8314.64` → `assessment ≈ 1133.81` ✓. For `adminRate=0` → strict `0` ✓.

**Why downgraded:** the same unguarded gross-up pattern (`/ (1 - adminRate)`) is already established at [src/components/HrTools/GoalCalculator/Shared/calculateTotals.ts:92](src/components/HrTools/GoalCalculator/Shared/calculateTotals.ts#L92) and ships in production today. This PR brings PDS into alignment with an existing pattern — not a new risk.

**Remaining minor concern:** `adminRate` is a server-controlled constant from `mpdGoalMiscConstants.RATES.ADMIN_RATE.fee`. If anyone ever enters `12` instead of `0.12`, the formula produces a large negative `assessment`. If anyone enters `1.0`, the formula produces `Infinity` which then renders as the goal total.

**Optional hardening** in `buildPdsGoalConstants` (`pdsGoalConstants.ts:42`):

```ts
// Before
if (adminRate === undefined) { return null; }
// After
if (adminRate === undefined || adminRate < 0 || adminRate >= 1) { return null; }
```

Or add an invariant test asserting the formula is correct for a non-trivial rate:

```ts
it('produces an assessment that is `adminRate` of the post-admin total', () => {
  const result = calculateOtherExpenses(fullTime(), {
    ...defaultConstants,
    adminRate: 0.25,
  });
  const adminBase = result.subtotal + result.attrition + result.creditCardFees;
  expect(result.assessment / (adminBase + result.assessment)).toBeCloseTo(0.25, 10);
});
```

wjames111 (commented):
## Multi-Agent Code Review — PR #1774 (re-review)

**Mode:** Standard · **Agents:** 6 specialized reviewers + 1 consolidated debate round
**Risk:** 6/10 (MEDIUM) · **Verdict:** BLOCKERS FOUND

### Verdict summary

2 critical blockers must be resolved before merge. Both were independently flagged by 4–5 of the 6 agents with high confidence and concrete code/math citations.

1. **Geographic-multiplier convention split (sev 9.5)** — `PdsGoalCalculatorTestWrapper.tsx:197,201,205` mocks `percentageMultiplier: 1, 1.06, 1.12` (multiplicative) but `salaryCalculation.ts:34` (unchanged) still computes `monthlyBase * (1 + geographicMultiplier)` (additive). Every other test fixture and consumer in the repo (6 separate sites) uses the additive convention. Latent bug: any test exercising `geographicLocation: 'None'` against this wrapper silently produces doubled salary.
2. **Credit Card Fees display formula mismatch (sev 9.0)** — `otherBreakdown.tsx:119-121` still shows `(Subtotal + Attrition) × {{rate}}` but `OtherExpenses.ts:57-59` was changed to a gross-up. The displayed formula no longer matches the rendered number (~$31 discrepancy on the canonical $7800 subtotal example). The Assessment row's display formula was updated in the same hunk; Credit Card Fees was missed.

### Risk factors

- 14 files under `src/components/HrTools/PdsGoalCalculator/` (Medium-Risk per repo rules)
- Financial calculation domain — formula churned 4× across branch history
- Previous /agent-review on this PR flagged a severity 9.5 convention-split bug; this re-review finds it has been re-introduced in a different location (test wrapper) rather than resolved

### Math verification

- **Gross-up identities verified algebraically** — `creditCardFees / (subtotal + attrition + creditCardFees) === r` and `assessment / (adminBase + assessment) === adminRate`. Implementation correctness ✓
- **`PdsGoalCard.test.tsx:20` expected `$8,073.02` reproduces exactly**: subtotal $6300 × 1.06 / (0.94 × 0.88) = $8,073.02. Not a magic number ✓
- **`usePdsSummaryData.test.ts:319` expected `9995.16` reproduces exactly** with the new gross-up formulas ✓

### Findings overview

| Severity tier | Count | Issues |
|---|---|---|
| Critical (9.0-10) | 2 | F1 convention split, F2 CC fees display drift |
| High (8.0-8.9) | 0 | — |
| Important (7.0-7.9) | 0 | — |
| Medium (5.0-6.9) | 2 | F3 lost test coverage, F6 opaque gross-up display |
| Suggestion (<5.0) | 5 | F4 redundant subscription, F5 missing rate guards, F7/F8/F9 display polish |

### Agent summary

| Agent | Critical | Important | Medium | Suggestions |
|---|---|---|---|---|
| Architecture | 1 | 3 | — | 2 |
| Data Integrity | 2 | 1 | — | 2 |
| Testing | 0 | 3 | — | 4 |
| UX | 1 | 3 | — | 3 |
| Financial Reporting | 1 | 2 | — | 1 |
| Standards | 0 | 0 | — | 0 (all PASS) |

### Cross-cutting findings

- **Fix one, fix all violation**: developer updated Assessment label gross-up but missed Credit Card Fees label (identical math structure, adjacent row). F2 above.
- **Display layer needs a coordinated pass**: F2 (blocker) + F6 + F7 + F8 + F9 all touch `otherBreakdown.tsx` / `salaryBreakdown.tsx` formula rendering. Treat F2 as blocking; consolidate the rest into a polish ticket.

### Recommendations

1. **Revert the test wrapper** to additive convention (`0, 0.06, 0.12`) — minimum fix for F1.
2. **Update the Credit Card Fees display formula** to mirror the Assessment row's `÷ {{divisor}} − ...` pattern — fix for F2.
3. **Add a part-time-hourly overallTotal test** to `usePdsSummaryData.test.ts` to restore coverage lost when `calculatePdsGoalTotal.test.ts` was deleted.
4. **Optional follow-up**: document `MpdGoalGeographicConstant.percentageMultiplier` semantics in the GraphQL schema (the field name implies multiplicative but the consumer treats it as additive — naming is misleading).

### Open questions for human review

- The PR description for MPDX-9573 notes the assessment number "does not match the spreadsheet even though it should" — a finance reviewer should validate the gross-up math against the source-of-truth spreadsheet before merge.
- The geographic-multiplier convention has been flipped 4× in the branch history. Consider a one-line schema comment to permanently document the contract.

src/components/HrTools/PdsGoalCalculator/PdsGoalCalculatorTestWrapper.tsx
```diff
@@ -194,15 +194,15 @@ export const PdsGoalCalculatorTestWrapper: React.FC<
                     mpdGoalGeographicConstants: [
                       {
                         location: 'None',
-                        percentageMultiplier: 0,
+                        percentageMultiplier: 1,
```
wjames111: **[Critical] F1: Geographic-multiplier convention split (BLOCKER, re-introduced)** — sev 9.5/10

**Flagged by:** Architecture (9.5), Data Integrity (9.0), Financial Reporting (9.0), Testing (7.5). Debate confirmed and raised.

This PR changes the test mock `percentageMultiplier` values from `0 / 0.06 / 0.12` (additive convention) to `1 / 1.06 / 1.12` (multiplicative convention). But `salaryCalculation.ts:34` (NOT modified in this PR) still computes `monthlyBase * (1 + geographicMultiplier)`, which only makes sense with the additive convention.

**Six independent code paths confirm the additive convention is the established contract:**
1. `src/components/HrTools/PdsGoalCalculator/calculations/salaryCalculation.ts:34` — `monthlyBase * (1 + geographicMultiplier)`
2. `src/components/HrTools/GoalCalculator/Shared/calculateNewStaffTotals.ts:144` — `(geographicMultiplier + 1) * base`
3. `src/components/HrTools/GoalCalculator/GoalCalculatorTestWrapper.tsx:191` — Orlando = `0.06`
4. `src/hooks/useGoalCalculatorConstants.test.tsx:42,48` — `0` and `0.12`
5. `src/components/HrTools/PdsGoalCalculator/calculations/usePdsSummaryData.test.ts:38,73-74` — `GEO_MULTIPLIER = 0.06`, `None = 0`
6. `src/components/HrTools/PdsGoalCalculator/calculations/buildPdsGoalConstants.test.ts:67,152` — `0.06`, `0.12`

**Git history smoking gun** (verified via `git log -p`):
- Original: `monthlyBase * (1 + geographicMultiplier)` (additive)
- `fa11fb311` (May 11, zach): "Fix calculation of gross monthly pay by removing unnecessary multiplier" → `* geographicMultiplier` (multiplicative)
- `6ab40c702` (this PR, May 13): "Fix calculations" → REVERTED back to `* (1 + geographicMultiplier)` (additive)
- `2057ac41a` (this PR, May 13): "Fix tests" → test wrapper changed to `1 / 1.06 / 1.12` (multiplicative)

The wrapper change is left-over from the brief multiplicative period and was NOT updated when the calculation was reverted back to additive 7 hours later.

**Latent doubling proven:** any test using `geographicLocation: 'None'` would compute `monthlyBase * (1 + 1) = 2 * monthlyBase`. No current test exercises this path (only `null` or `'Orlando, FL'`), so CI is green. The bug is dormant only because the test for Orlando isn't asserting a specific salary number.

**Fix:**
```tsx
// Before:
mpdGoalGeographicConstants: [
  { location: 'None',         percentageMultiplier: 1 },
  { location: 'Orlando, FL',  percentageMultiplier: 1.06 },
  { location: 'New York, NY', percentageMultiplier: 1.12 },
],

// After (additive convention, consistent with rest of repo):
mpdGoalGeographicConstants: [
  { location: 'None',         percentageMultiplier: 0 },
  { location: 'Orlando, FL',  percentageMultiplier: 0.06 },
  { location: 'New York, NY', percentageMultiplier: 0.12 },
],
```

**Strongly recommended follow-up:** add a JSDoc comment to `MpdGoalGeographicConstant.percentageMultiplier` in the GraphQL schema documenting the additive convention (e.g., "stored as additive adjustment, 0.06 means +6%; consumers apply `(1 + value)`"). The field name implies multiplicative, which is what's causing the repeated developer confusion (4 flips in branch history).

src/components/HrTools/PdsGoalCalculator/SupportItem/otherBreakdown.tsx
wjames111: **[Critical] F2: Credit Card Fees display formula not updated (BLOCKER, fix-one-fix-all violation)** — sev 9.0/10

**Flagged by:** Data Integrity (9.0), UX (9.0), Architecture (7.5), Financial Reporting (7.5), Testing (7.0). 5/6 agents.

This hunk correctly updates the **Assessment** row's display formula to the gross-up form (`(Subtotal + Attrition + Credit Card Fees) ÷ {{divisor}} − ...`), matching the gross-up math in `OtherExpenses.ts:60-63`.

But the **Credit Card Fees row immediately above (lines 117-125, not in this hunk)** was NOT updated. It still reads:
```tsx
formula: t('(Subtotal + Attrition) × {{rate}}', {
  rate: percentageFormat(constants.creditCardFeeRate, locale),
}),
```
yet `OtherExpenses.ts:57-59` computes `creditCardFees = (subtotal + attrition) / (1 - creditCardFeeRate) - (subtotal + attrition)` — identical gross-up structure to Assessment.

**User-visible discrepancy** (verified numerically): with `subtotal=$7800`, `attrition=$468`, `creditCardFeeRate=0.06`, the displayed formula says `(8268) × 6% = $496.08` but the cell renders `$527.74`. A $31.66 gap on every full-time goal — staff reading the breakdown will not be able to reconcile the formula caption with the amount, defeating the purpose of showing the formula.

**Fix** — mirror the Assessment row pattern:
```tsx
// Before (otherBreakdown.tsx:117-125):
{
  id: 'credit-card-fees',
  category: t('Credit Card Fees'),
  formula: t('(Subtotal + Attrition) × {{rate}}', {
    rate: percentageFormat(constants.creditCardFeeRate, locale),
  }),
  amount: totals.creditCardFees,
  testId: 'other-credit-card-fees',
  bold: true,
},

// After:
{
  id: 'credit-card-fees',
  category: t('Credit Card Fees'),
  formula: t(
    '(Subtotal + Attrition) ÷ {{divisor}} − (Subtotal + Attrition)',
    {
      divisor: new Intl.NumberFormat(locale, {
        minimumFractionDigits: 2,
        maximumFractionDigits: 4,
      }).format(1 - constants.creditCardFeeRate),
    },
  ),
  amount: totals.creditCardFees,
  testId: 'other-credit-card-fees',
  bold: true,
},
```

The PR description for MPDX-9572 explicitly says "*the formula itself has been updated as well so the label now reflects the updated formula*" — this is one row away from satisfying that stated intent.

src/components/HrTools/PdsGoalCalculator/calculations/usePdsSummaryData.test.ts
```diff
@@ -300,8 +296,8 @@ describe('usePdsSummaryData', () => {
     });
 
     it('computes correct overallTotal for a full-time salaried employee', () => {
```
wjames111, line 298: **[Medium] F3: Part-time hourly `overallTotal` coverage lost from deleted test file** — sev 6.5/10

**Flagged by:** Testing (6.5), Architecture (6.0).

The deleted `calculatePdsGoalTotal.test.ts` (134 lines) contained an end-to-end test for a part-time hourly employee asserting an overall goal `≈ $434.83`. Migrating it to `usePdsSummaryData.test.ts` was nearly complete — but the part-time-hourly overall-total case was NOT ported.

Part-time hourly hits a different branch in `salaryCalculation.ts:33`:
```ts
const monthlyBase = isSalaried ? payRate / 12 : (payRate * hours * 52) / 12;
```
The hourly branch (`(payRate * hours * 52) / 12`) has no end-to-end test through `usePdsSummaryData.overallTotal`. The salary unit tests cover hourly math at the salary layer, and OtherExpenses tests cover part-time other-expense math, but nothing exercises both together via the hook.

**Coverage migration audit** (`git show` on the deleted file):
| Deleted case | Migrated? |
|---|---|
| Full-time salaried ≈$1038.21 (now $9,995.16) | ✅ (line 298) |
| Part-time hourly ≈$434.83 | ❌ **NOT migrated** |
| Null `payRate` returns positive | ⚠️ partial (only at salary unit level) |
| Geographic multiplier increases total | ✅ split across multiple tests |
| `formType: Simple` excludes reimbursables + 403b | ✅ (lines 339-354) |
| `formType: Detailed` includes them | ✅ (lines 356-370) |
| Legacy `formType: null` ≡ Detailed | ✅ (lines 372-386) |

**Suggested addition** (after the full-time test at line 320):
```ts
it('computes correct overallTotal for a part-time hourly employee', () => {
  const calc = {
    ...defaultCalculation,
    salaryOrHourly: DesignationSupportSalaryType.Hourly,
    status: DesignationSupportStatus.PartTime,
    payRate: 25,
    hoursWorkedPerWeek: 20,
    benefits: null,
  };
  const { result } = renderHook(() => usePdsSummaryData(calc, defaultHcmUser));
  expect(result.current?.overallTotal).toBeGreaterThan(0);
  expect(result.current?.otherTotals.workComp).toBeGreaterThan(0); // part-time only
  expect(result.current?.otherTotals.benefits).toBe(0);             // part-time => 0
});
```

src/components/HrTools/PdsGoalCalculator/SupportItem/otherBreakdown.tsx
```diff
@@ -126,9 +126,15 @@ export const buildOtherBreakdownRows = (
...
+        {
+          divisor: new Intl.NumberFormat(locale, {
+            minimumFractionDigits: 2,
+            maximumFractionDigits: 4,
+          }).format(1 - constants.adminRate),
+        },
+      ),
```
wjames111, line 137: **[Medium] F6: Gross-up formula is opaque to non-finance users** — sev 6.0/10

**Flagged by:** UX (7.5, revised to 6.0 by debate).

The new Assessment formula displays as: `(Subtotal + Attrition + Credit Card Fees) ÷ 0.88 − (Subtotal + Attrition + Credit Card Fees)`

Three readability problems for the target HR-staff audience:
1. **Bare `0.88` decimal** — users can't recognize this as `(1 − 12% admin rate)` without working out the algebra. Peer rows (Attrition, Credit Card Fees today) show percentages.
2. **Repeated subexpression** — `(Subtotal + Attrition + Credit Card Fees)` appears twice on a single category-formula caption line, making the row visually heavy compared to peer rows.
3. **Mathematically opaque** — the gross-up identity `x/(1−r) − x = x·r/(1−r)` is not intuitive without a finance background.

Alternatives, in increasing order of clarity:
- `(Subtotal + Attrition + Credit Card Fees) × {{rate}}` where `rate` is the pre-computed effective multiplier `adminRate / (1 − adminRate)` formatted as a percentage (e.g., `13.64%`). Single multiplication, parallel to peer rows.
- Add a tooltip via the existing `OtherBreakdownRow.tooltip` field that explains "Admin is `adminRate` of the post-admin total; this grosses up the pre-admin base." Keep the displayed formula short.
- Name the subexpression: `Base × {{rate}}` where `Base = Subtotal + Attrition + Credit Card Fees` (rendered as a separate sub-row or annotation).

**Note:** This finding consolidates with F2 — both touch `otherBreakdown.tsx` display rendering and should be addressed together. F2 (CC fees formula) blocks merge; F6 is a non-blocking polish task that can land in a follow-up display-layer cleanup.

src/components/HrTools/PdsGoalCalculator/calculations/OtherExpenses.ts
```diff
@@ -54,9 +54,13 @@ export const calculateOtherExpenses = (
...
-  const creditCardFees = (subtotal + attrition) * constants.creditCardFeeRate;
-  const adminRate = constants.adminRate;
-  const assessment = (subtotal + creditCardFees + attrition) * adminRate;
+  const creditCardFees =
+    (subtotal + attrition) / (1 - constants.creditCardFeeRate) -
+    (subtotal + attrition);
+  const adminBase = subtotal + creditCardFees + attrition;
+  // Admin assessment is `adminRate` of the post-admin total, not a markup on
+  // `adminBase`, so gross up: assessment / (adminBase + assessment) = adminRate.
+  const assessment = adminBase / (1 - constants.adminRate) - adminBase;
```
wjames111, line 63: **[Suggestion] F5: No guards on rate ≥ 1 for gross-up divisions** — sev 3.0/10 (revised down from 6.0 by debate)

**Flagged by:** Data Integrity (6.0). Debate lowered severity after finding the same pattern is established codebase-wide.

Lines 58 and 63 both divide by `(1 - rate)`. If `creditCardFeeRate >= 1` or `adminRate >= 1`, the result is `Infinity` or a negative number — silently propagating into `assessment`, `overallTotal`, and the rendered cell.

**Practical exposure is low**:
- Server constants are admin-managed and currently `0.06` / `0.12`.
- `pdsGoalConstants.ts:37-44` already validates that all 5 rates are *defined*, just not that they're in `[0,1)`.
- The same gross-up pattern (without guards) exists in `src/components/HrTools/GoalCalculator/Shared/calculateTotals.ts:92`, so this is a codebase-wide convention, not a PR-specific regression.

**Suggested follow-up** (not in this PR): add bounds validation to `buildPdsGoalConstants` and the sister `calculateTotals.ts` in one shot — return `null` if any rate is `>= 1` or `< 0`. Belongs in its own ticket; flagging here for visibility.

src/components/HrTools/PdsGoalCalculator/GoalCard/PdsGoalCard.tsx
```diff
@@ -24,23 +21,12 @@ export const PdsGoalCard: React.FC<PdsGoalCardProps> = ({ goal }) => {
   const accountListId = useAccountListId() ?? '';
   const [deletePdsGoalCalculation] = useDeletePdsGoalCalculationMutation();
 
-  const {
-    goalMiscConstants,
-    goalGeographicConstantMap,
-    loading: constantsLoading,
-  } = useGoalCalculatorConstants();
+  const { loading: constantsLoading } = useGoalCalculatorConstants();
```
wjames111: **[Suggestion] F4: Redundant `useGoalCalculatorConstants` subscription** — sev 3.0/10 (revised down from 6.5 by debate)

**Flagged by:** Architecture (6.5). Debate lowered severity — Apollo dedupes the request, no incorrect behavior.

This component calls `useGoalCalculatorConstants()` solely for `loading`, and then on line 28 calls `usePdsSummaryData(...)` which also calls `useGoalCalculatorConstants()` internally (see `usePdsSummaryData.ts:39-40`). Apollo's `cache-first` policy plus default request deduplication means **no extra network call** — this is purely stylistic.

A cleaner pattern would be to have `usePdsSummaryData` return `{ summaryData, loading, error }` so this component has a single source for loading state. But that's a small refactor and out of scope for this fix-focused PR — leaving as a suggestion for a follow-up.

src/components/HrTools/PdsGoalCalculator/SupportItem/salaryBreakdown.tsx
```diff
@@ -73,16 +73,12 @@ export const buildSalaryBreakdownRows = (
...
-      category: t('Geographic Multiplier'),
-      amount: geographicMultiplier,
-      format: 'percentage',
-    },
     {
       id: 'gross-monthly-pay',
       category: t('Gross Monthly Pay'),
-      formula: t('Monthly Base × (1 + Geographic Multiplier)'),
+      formula: t('Monthly Base × {{rate}}', {
+        rate: percentageFormat(1 + geographicMultiplier, locale),
```
wjames111, line 80: **[Suggestion] F7/F8/F9: Salary breakdown display polish** — sev 3.0/10 (revised down from 4.5–6.5 by debate)

**Flagged by:** UX (6.5), Architecture (4.5), Financial Reporting (3.5–6.0). Debate clarified the developer's intent was deliberate and these are subjective UX calls, not bugs.

Three related observations on the new Gross Monthly Pay formula display:

**F7 — `Monthly Base × 100%` for null-location**: with `geographicMultiplier = 0` (the default for null/unknown locations), the formula renders as `Monthly Base × 100%`. Technically correct, but visually a no-op multiplication. Git history (commit `ead7b6824` "Fix The Gross Monthly Pay display formula") shows this was a deliberate fix from displaying `0%`. Consider hiding the formula caption when `geographicMultiplier === 0`, or showing `Monthly Base + 0%` to make the additive intent explicit.

**F8 — Removed standalone Geographic Multiplier row**: the staff member can no longer see a labeled row indicating their geographic adjustment value. The multiplier is now embedded inline as `× 106%` in the Gross Monthly Pay formula. Git history shows this was intentional consolidation (commit `4136ea802`). Consider re-adding as a tooltip on the Gross Monthly Pay row for transparency, especially when the location is non-null.

**F9 — `1 + geographicMultiplier` computed in two places**: `salaryCalculation.ts:34` does the math, this line displays the same `1 +` adjustment. Minor drift risk; only 2 occurrences within the same feature folder. Not worth a helper today, but worth a `// keep in sync with salaryCalculation.ts:34` comment to flag the coupling.

Not blocking. Consolidate into a display-layer polish ticket with F6 (gross-up opaque formula).

src/components/HrTools/PdsGoalCalculator/SupportItem/otherBreakdown.tsx
wjames111: /dismiss might revisit this later. Right now getting the formula correct and keeping parity between calculations and the formula displays is the priority. That being said, 1 should be addressed and updated to use a percentage. I will update that in this PR.

src/components/HrTools/PdsGoalCalculator/calculations/OtherExpenses.ts
wjames111: /dismiss this is the general direction the codebase has taken. Might be worth making ensuring admin can only enter percentages in a follow up PR.

src/components/HrTools/PdsGoalCalculator/SupportItem/otherBreakdown.tsx
```diff
@@ -116,19 +116,25 @@ export const buildOtherBreakdownRows = (
     {
       id: 'credit-card-fees',
       category: t('Credit Card Fees'),
-      formula: t('(Subtotal + Attrition) × {{rate}}', {
-        rate: percentageFormat(constants.creditCardFeeRate, locale),
-      }),
+      formula: t(
+        '(Subtotal + Attrition) ÷ (1 − {{rate}}) − (Subtotal + Attrition)',
```
wjames111: **[High] Long gross-up formula labels will clip in the data grid.** — sev 8.0/10

**Flagged by:** UX (8.0), Architecture (rendering-contract confirmation).

The new formula here grows from ~30 chars to ~58 chars, and the Assessment formula at line 133 reaches ~100 chars. The category cell at `styledGrid.tsx:27-36` stacks the category label over the formula in a flex column. MUI X DataGrid defaults to `rowHeight={52}` and neither `SalarySection.tsx:47` nor `OtherSection.tsx:48` (both non-PR files) override it with `getRowHeight={() => 'auto'}`. The longest formula will wrap to 2-3 lines on viewports < ~1100px and on tablet/mobile, then clip at the third line.

**Cheapest fix (one line per non-PR container file):**

```tsx
// in SalarySection.tsx and OtherSection.tsx
<StyledGrid
  // ...existing props
  getRowHeight={() => 'auto'}
  // ...
/>
```

Alternative: shorten the in-PR formula labels — see Dismissed Findings F6 in the PR body (author has already pushed back on rewording for this PR scope; the container-level fix is the lower-friction path).

src/components/HrTools/PdsGoalCalculator/SupportItem/otherBreakdown.tsx
```diff
@@ -116,19 +116,25 @@ export const buildOtherBreakdownRows = (
...
       bold: true,
     },
     {
       id: 'assessment',
       category: t('Assessment'),
-      formula: t('(Subtotal + Credit Card Fees + Attrition) × {{rate}}', {
-        rate: percentageFormat(constants.adminRate, locale),
-      }),
+      formula: t(
+        '(Subtotal + Attrition + Credit Card Fees) ÷ (1 − {{rate}}) − (Subtotal + Attrition + Credit Card Fees)',
```
wjames111, line 133: **[Suggestion] Display summand order doesn't match code summand order.** — sev 4.5/10

**Flagged by:** Financial (3.5 → 4.5 in debate, parity hygiene).

- Code at `OtherExpenses.ts:60`: `adminBase = subtotal + creditCardFees + attrition`
- Display here: `(Subtotal + Attrition + Credit Card Fees) ÷ (1 − {{rate}}) − (Subtotal + Attrition + Credit Card Fees)`

Addition is commutative so the math is identical, but this is the kind of drift that erodes auditability when reviewers diff code against label. If a future engineer "fixes" the code to match the label (or vice-versa) and accidentally changes the formula semantics, this is how silent bugs land. Align order:

```diff
-      '(Subtotal + Attrition + Credit Card Fees) ÷ (1 − {{rate}}) − (Subtotal + Attrition + Credit Card Fees)',
+      '(Subtotal + Credit Card Fees + Attrition) ÷ (1 − {{rate}}) − (Subtotal + Credit Card Fees + Attrition)',
```

Also worth adding a one-line comment near these two gross-up display strings:

```tsx
// Mirrors the gross-up calculation in `calculateOtherExpenses` — do not simplify
// to `(Subtotal + Attrition) × {{rate}}` (the old wrong markup form).
```

src/components/HrTools/PdsGoalCalculator/GoalCard/PdsGoalCard.tsx
```diff
@@ -24,23 +21,12 @@ export const PdsGoalCard: React.FC<PdsGoalCardProps> = ({ goal }) => {
...
-    const constants = buildPdsGoalConstants(
-      goalMiscConstants,
-      goalGeographicConstantMap,
-      goal.geographicLocation,
-      hcmUser?.fourOThreeB,
-    );
-    return constants ? calculatePdsGoalTotal(goal, constants) : 0;
-  }, [goal, goalMiscConstants, goalGeographicConstantMap, hcmUser]);
+  const summaryData = usePdsSummaryData(goal, hcmUser);
+  const goalTotal = summaryData?.overallTotal ?? 0;
```
wjames111, line 30: **[Important] Goal Amount display jumps ~9.5× on existing data.** — sev 7.5/10

**Flagged by:** Financial (7.5), UX-3 (7.0, merged).

This line correctly replaces the old `calculatePdsGoalTotal(goal, constants)` return (which was `otherExpenses.assessment` — only the admin line item) with the full `overallTotal = subtotal + attrition + creditCardFees + assessment`. The math is correct (verified algebraically: with `payRate=60000` and the test fixture, the new value matches the test's `toBeCloseTo(9995.16, 0)` at `usePdsSummaryData.test.ts:319`). The previously-displayed value matched only the assessment component, so the old card was buggy in two ways (wrong field shown, wrong formula).

**Non-code actions before merge:**

1. Confirm with the business owner that the "Goal Amount" label was always intended to show the full funding goal (not just the assessment line).
2. Search for downstream consumers of `goalTotal` / `overallTotal` (other dashboards, exports, MCC/SOSA aggregations) calibrated against the previous (buggy) value.
3. Add release notes / in-product callout — staff will see their goal jump ~9.5× and may reasonably suspect the calculator is broken.

The code itself is fine; this is a comms blocker, not a code blocker.

src/components/HrTools/PdsGoalCalculator/GoalCard/PdsGoalCard.tsx
wjames111: **[Suggestion] Redundant `useGoalCalculatorConstants` call (keystone refactor).** — sev 4.0/10

**Flagged by:** Architecture (3.5, collapsed C1+C2+S1 in debate), UX-S2 (3.5).

`usePdsSummaryData` invokes `useGoalCalculatorConstants` internally at `usePdsSummaryData.ts:39-40`. This component invokes it again purely for the `loading` flag. Apollo's `fetchPolicy: 'cache-first'` (`useGoalCalculatorConstants.ts:71`) keeps the cost cheap, but:

1. The hook is now called twice in the render tree.
2. The `loading` boolean from this call may not match the `data` availability inside the other call, creating a small race where `constantsLoading=false` but `summaryData=null`.

**Keystone fix:** expose `loading` from `usePdsSummaryData`, drop this hook from the component:

```ts
// usePdsSummaryData.ts — extend return shape
export interface UsePdsSummaryDataResult {
  data: PdsSummaryData | null;
  loading: boolean;
  error?: ApolloError;
}
return { data, loading: constantsLoading || hcmLoading, ...etc };
```

```tsx
// PdsGoalCard.tsx — collapsed
const { data: summaryData, loading: summaryLoading } = usePdsSummaryData(goal, hcmUser);
const goalTotal = summaryData?.overallTotal ?? 0;
// ...
loading={summaryLoading || hcmLoading}
```

Also resolves the `$0 vs unknown` empty-state ambiguity (Architecture C2 / UX-S3): once `loading` flows from a single hook, the component can render `t('Unable to calculate')` when `!summaryLoading && summaryData === null`.

src/components/HrTools/PdsGoalCalculator/SupportItem/salaryBreakdown.tsx
```diff
@@ -73,16 +73,12 @@ export const buildSalaryBreakdownRows = (
...
-      id: 'geographic-multiplier',
-      category: t('Geographic Multiplier'),
-      amount: geographicMultiplier,
-      format: 'percentage',
-    },
     {
       id: 'gross-monthly-pay',
       category: t('Gross Monthly Pay'),
-      formula: t('Monthly Base × (1 + Geographic Multiplier)'),
+      formula: t('Monthly Base × {{rate}}', {
```
wjames111, line 79: **[Medium] `Monthly Base × 100%` reads as noise when geographicMultiplier=0.** — sev 5.5/10

**Flagged by:** UX (6.5 → 5.5 in debate).

When `geographicMultiplier = 0.06`, this renders as `Monthly Base × 106%` — clean. But the default (unknown location, null location, or location not in the geo map — see `buildPdsGoalConstants.test.ts:111-141`) is `geographicMultiplier = 0`, which renders as `Monthly Base × 100%`. That reads as noise — a multiplication by 100% communicates nothing.

Additionally, by removing the standalone `geographic-multiplier` row above, staff lose the ability to validate "is the multiplier actually being applied?" without working backward from the formula.

**Recommended:** conditionally render the formula:

```diff
-      formula: t('Monthly Base × {{rate}}', {
-        rate: percentageFormat(1 + geographicMultiplier, locale),
-      }),
+      formula:
+        geographicMultiplier === 0
+          ? undefined
+          : t('Monthly Base × {{rate}}', {
+              rate: percentageFormat(1 + geographicMultiplier, locale),
+            }),
```

Or, better, restore the geographic-multiplier row as a separate line in the breakdown so the input is auditable independently of its product. The row pattern is the established convention in the sibling `otherBreakdown.tsx` (intermediate values like Attrition, 403b, Work Comp all get their own rows even though they could be inlined into subtotal formulas).

src/components/HrTools/PdsGoalCalculator/calculations/OtherExpenses.test.ts
```diff
@@ -137,19 +137,18 @@ describe('calculateOtherExpenses', () => {
   });
 
   describe('credit card fees', () => {
-    it('is 6% of (subtotal + attrition)', () => {
+    it('grosses up (subtotal + attrition) so that fees are `creditCardFeeRate` of the post-fees total', () => {
       const result = calculateOtherExpenses(fullTime(), defaultConstants);
-      // (7400 + 444) * 0.06
-      expect(result.creditCardFees).toBeCloseTo(470.64);
+      // (7400 + 444) / (1 - 0.06) - (7400 + 444) ≈ 500.68
```
wjames111, line 142: **[Medium] Asymmetric edge-case coverage: missing `creditCardFeeRate === 0` test.** — sev 6.0/10

**Flagged by:** Financial (6.0), Testing supports.

The sibling `assessment` describe block at line 153 has `it('returns 0 when adminRate is 0', ...)`. The credit card fees describe block does not have the equivalent for `creditCardFeeRate`. New gross-up formula `(x + attr) / (1 − 0) − (x + attr) === 0` evaluates correctly at the degenerate case mathematically, but the assertion is missing. Asymmetric coverage between parallel formulas masks future refactor regressions.

**Patch:**

```ts
describe('credit card fees', () => {
  it('grosses up (subtotal + attrition) so that fees are `creditCardFeeRate` of the post-fees total', () => {
    const result = calculateOtherExpenses(fullTime(), defaultConstants);
    expect(result.creditCardFees).toBeCloseTo(500.68);
  });

  it('returns 0 when creditCardFeeRate is 0', () => {
    const result = calculateOtherExpenses(fullTime(), {
      ...defaultConstants,
      creditCardFeeRate: 0,
    });
    expect(result.creditCardFees).toBe(0);
  });
});
```

src/components/HrTools/PdsGoalCalculator/GoalCard/PdsGoalCard.test.tsx
```diff
@@ -17,7 +17,7 @@ describe('PdsGoalCard', () => {
       </PdsGoalCalculatorTestWrapper>,
     );
 
-    expect(await findByText('$849.44')).toBeInTheDocument();
+    expect(await findByText('$8,073.02')).toBeInTheDocument();
```
wjames111, line 34: **[Suggestion] `$8,073.02` is not algebraically derivable from any visible test input.** — sev 4.5/10

**Flagged by:** Testing (6.0 → 4.5 in debate; Financial supports — confirmed by reading `PdsGoalCalculatorTestWrapper.tsx`).

The wrapper at `PdsGoalCalculatorTestWrapper.tsx:35-60` only pins 9 fields (name, formType, status, salaryOrHourly, payRate=50000, hoursWorkedPerWeek=null, benefits=1500, geographicLocation=null). The 9 reimbursable fields (`ministryCellPhone`, `ministryInternet`, `mpdNewsletter`, `mpdMiscellaneous`, `accountTransfers`, `otherMonthlyReimbursements`, `conferenceRetreatCosts`, `ministryTravelMeals`, `otherAnnualReimbursements`) and `averageHoursPerWeek` are populated by `graphql-ergonomock` from seed `'seed'` (set in `__tests__/util/graphqlMocking.tsx:26`).

`$8,073.02` is the result of (a) the new formula AND (b) ergonomock's seeded RNG for those fields. The assertion will silently change if ergonomock is upgraded, the seed is changed, or a new field is added to the fragment.

**Cheapest fix:** pin every reimbursable field to a known value in `PdsGoalCalculatorTestWrapper.tsx`'s `calculationsDefault`, then add a comment here showing the hand-derivation of `$8,073.02`. The math is then reproducible for the next maintainer who has to update this assertion.

Less critical because `usePdsSummaryData.test.ts:319` independently pins `9995.16` for `payRate=60000` — drift in the underlying formula will be caught upstream, and this test failure becomes a useful canary rather than a duplicated trap. But it should still be made deterministic.

src/components/HrTools/PdsGoalCalculator/calculations/usePdsSummaryData.test.ts
```diff
@@ -313,13 +309,31 @@ describe('usePdsSummaryData', () => {
...
+        salaryOrHourly: DesignationSupportSalaryType.Hourly,
+        status: DesignationSupportStatus.PartTime,
+        payRate: 25,
+        hoursWorkedPerWeek: 20,
+        benefits: null,
+      };
+      const { result } = renderHook(() =>
+        usePdsSummaryData(calc, defaultHcmUser),
+      );
+      expect(result.current?.overallTotal).toBeGreaterThan(0);
```
wjames111: **[Medium] Part-time test assertions are weak compared to the deleted equivalent.** — sev 5.0/10

**Flagged by:** Testing #2 (5.5), expanded to cover other null paths.

The deleted `calculatePdsGoalTotal.test.ts` asserted `toBeCloseTo(434.83, 0)` on a specific computed part-time total. The replacement here only checks `overallTotal > 0`, `workComp > 0`, and `benefits === 0`. A formula regression on the part-time branch (e.g., flipping the workComp sign, missing the gross-up, wrong constants) would still satisfy these assertions.

**Add the numeric pin:**

```ts
// Hand-derivation (verify against your wrapper constants):
// grossMonthlyPay = (25 * 20 * 52) / 12 = 2166.67
// employerFica ≈ 173.33; salarySubtotal ≈ 2340
// reimbursable monthly + annual / 12 ≥ floor 300 (depends on wrapper defaults)
// 403b = grossMonthlyPay * 0.08 ≈ 173.33
// workComp = grossMonthlyPay * 0.17 ≈ 368.33
// subtotal ≈ ... attrition ≈ ... ccFees = (.)/(.94) - .; assessment = .
expect(result.current?.overallTotal).toBeCloseTo(<computed>, 0);
expect(result.current?.otherTotals.workComp).toBeCloseTo(<computed>, 1);
expect(result.current?.otherTotals.benefits).toBe(0);
```

**Also add (this describe block is the right home):**

1. **null `payRate` test** — the gross-up `(x)/(1-rate)` would NaN-cascade if `payRate=null` isn't guarded all the way through.
2. **geographic multiplier flowthrough** — `withGeo.overallTotal > baseline.overallTotal` (the deleted file had this; the new file lost it).

wjames111 (commented):
## Multi-Agent Review — PR #1774

**Verdict:** BLOCKERS_FOUND — 1 high-priority blocker plus 1 important issue must be resolved (or human-acknowledged) before merge.

**Risk:** HIGH (7/10) — financial-math change with display-side impact. The PR has two PR-level wins (correct gross-up math; consolidating goal-card display onto the same `usePdsSummaryData` source as the rest of the calculator). Net architectural debt is significantly reduced.

**Agents:** 5 (Architecture, Testing, Standards, UX, Financial Reporting). Debate round resolved 4 challenges and merged 2 duplicate findings.

---

### Top-of-Pile (one-liners)

1. **[High 8.0] Long gross-up formula labels will clip in the salary/other breakdown grids.** Comment posted at otherBreakdown.tsx:120.
2. **[Important 7.5] Goal card total jumped ~9.5× ($849.44 → $8,073.02).** This is a genuine bug fix (old code showed only the `assessment` line item; new code shows full `overallTotal`), but staff and downstream consumers need to know. Comment posted at PdsGoalCard.tsx:29.

---

### High Priority Blocker (sev 8.0)

**Severity: 8.0/10 — Flagged by: UX (8.0), Architecture (confirmed via codebase read)**

**[High 8.0] otherBreakdown.tsx:120 / 133 + salaryBreakdown.tsx:79 — Long formula labels will clip in the data grid.**

The new gross-up formula strings are ~58-100 characters (vs. ~30 for the old strings). The category cell stacks `category` over `formula` in a flex column inside MUI X DataGrid's default fixed `rowHeight={52}`. Neither `SalarySection.tsx:47` nor `OtherSection.tsx:48` (both non-PR files) sets `getRowHeight={() => 'auto'}`. The longest formula — `(Subtotal + Attrition + Credit Card Fees) ÷ (1 − 12%) − (Subtotal + Attrition + Credit Card Fees)` — will wrap to 2-3 lines on desktops < ~1100px wide and on mobile/tablet, and the third line will clip.

**Recommended fix (one line in each non-PR container file):**

```tsx
// SalarySection.tsx and OtherSection.tsx (BOTH are outside this PR's diff)
<StyledGrid
  // ...existing props
  getRowHeight={() => 'auto'}
  // ...
/>
```

Alternatively, shorten the in-PR formula labels (use a friendlier label in the cell + the verbose math in a tooltip — see Dismissed Findings F6 below; the author has already pushed back on that direction for this PR).

---

### Important Issue (sev 7.5)

**Severity: 7.5/10 — Flagged by: Financial (7.5), UX-3 (7.0, merged)**

**[Important 7.5] PdsGoalCard.tsx:29 — Goal Amount display jumps ~9.5× on existing data.**

The old `calculatePdsGoalTotal` function returned only `otherExpenses.assessment` (one line item). The new code reads `summaryData.overallTotal` = `subtotal + attrition + creditCardFees + assessment` (the full funding goal). This is the correct fix — the field was always labelled "Goal Amount" but was previously displaying only the admin-assessment line. Math verified algebraically: with `payRate=60000` and the test fixture, the new value is $9,995.16 vs. the previously-displayed assessment-only $1,199.42.

**Action required (non-code):**

- Confirm with the business owner that the goal card was always intended to show the full funding goal.
- Search for downstream consumers of `goalTotal` / `overallTotal` (other dashboards, exports, MCC/SOSA aggregations) that may be calibrated against the previous (buggy) value.
- Add release notes / in-product callout: staff will see their goal jump ~9.5× and reasonably suspect the calculator is broken.

**Comment posted at the line.**

---

### Medium Priority (sev 5.0-6.9)

Line-level comments posted on each. Summary:

| Sev | Finding | File:Line |
|-----|---------|-----------|
| 6.0 | Asymmetric test coverage: no `creditCardFees === 0` test when `rate = 0` (the sibling `assessment` test has one) | OtherExpenses.test.ts:142 |
| 6.0 | `SalaryBreakdownRow` lacks the `tooltip` field that `OtherBreakdownRow` has. UX-2 (dismissed) and UX-4 (below) both want this — the affordance is now load-bearing. | salaryBreakdown.tsx (interface) |
| 5.5 | Stale formula text in test fixture (`salaryBreakdown.test.tsx:140, 147`) — still uses `'Pay Rate × (1 + Geographic Multiplier)'`. Plus: new production formula strings have no test coverage. *Lines 140/147 are outside this PR's diff — see body, not a line comment.* | salaryBreakdown.test.tsx:140,147 (not in diff) |
| 5.5 | `Monthly Base × {{rate}}` renders as `Monthly Base × 100%` when geographicMultiplier=0 (default/unknown location) — noise, not signal. Conditionally render the formula. | salaryBreakdown.tsx:79 |
| 5.0 | Missing tests for null `payRate`, null `hoursWorkedPerWeek`, and geo-key-with-null-multiplier paths in `usePdsSummaryData`. Gross-up `(x)/(1−r)` will NaN-cascade. | usePdsSummaryData.test.ts |

---

### Suggestions (sev < 5.0) — informational, do not block verdict

- **[Suggestion 4.5]** PdsGoalCard.test.tsx:20 — `$8,073.02` is not algebraically derivable; depends on `graphql-ergonomock` seeded RNG for the 9 reimbursable fields the wrapper doesn't pin. Cheap fix: pin the reimbursable fields in `PdsGoalCalculatorTestWrapper.tsx`. Comment posted.
- **[Suggestion 4.5]** usePdsSummaryData.test.ts:334 — Part-time test only asserts `> 0`. The deleted test had a `toBeCloseTo(434.83, 0)` numeric pin. Add an exact value. Comment posted.
- **[Suggestion 4.5]** otherBreakdown.tsx:133 — Display formula reads `(Subtotal + Attrition + Credit Card Fees)`; code computes `subtotal + creditCardFees + attrition`. Mathematically equivalent but a parity-hygiene issue: code-vs-display drift is how silent bugs land. Align order. Comment posted.
- **[Suggestion 4.5]** otherBreakdown.tsx:120 — Add an inline comment near the gross-up display strings explaining that the algebraic identity `x ÷ (1−r) − x` is intentional and mirrors `OtherExpenses.ts`. Prevents a future maintainer from "simplifying" back to `x × r` (reintroducing the bug this PR fixes).
- **[Suggestion 4.5]** usePdsSummaryData.test.ts — No test asserts that a non-zero `geographicMultiplier` actually flows into `overallTotal`. Add a `withGeo > withoutGeo` test. The deleted file had this; the new file lost it.
- **[Suggestion 4.0]** PdsGoalCard.tsx:24 — Redundant `useGoalCalculatorConstants()` call: `usePdsSummaryData` invokes it internally. Cheap (cache-first) but architecturally noisy and creates a small race where `constantsLoading=false` but `summaryData=null`. Keystone fix: expose `loading` from `usePdsSummaryData` and drop this hook from the component. Comment posted.
- **[Suggestion 4.0]** Add an end-to-end hand-verifiable `usePdsSummaryData` test with a fully zeroed reimbursable payload so the math is reproducible from comments (no `ergonomock` dependency).
- **[Suggestion 4.0]** A11Y-1: Removing the `geographic-multiplier` row changes the salary table structure; confirm no other view/aria-relationship depends on this row.
- **[Suggestion 4.0]** Three new translation keys are not yet in `public/locales/en/translation.json` and one old key (`'Monthly Base × (1 + Geographic Multiplier)'`) is now orphaned. Run `yarn extract` before merge so translators (Crowdin) see the new strings. *Translation file is not in this PR's diff.*
- **[Suggestion 3.5]** A11Y-2: Unicode `−` (U+2212, true minus) is new to this codebase; screen-reader pronunciation varies. The tooltip-affordance approach (UX-2, dismissed) would side-step this. Low risk; keep the character for visual parity with `÷` and `×`.
- **[Suggestion 3.0]** PdsGoalCard does not communicate "summary unavailable" — when `buildPdsGoalConstants` returns `null` post-load, the card renders `$0.00` indistinguishable from a real zero goal. *(Pre-existing — flagged for visibility; the consolidation in this PR makes it slightly more invisible.)*
- **[Suggestion 3.0]** Removal of `consistency with calculatePdsGoalTotal` test block — appropriate (the second source-of-truth function is gone), but the structural cross-check safety net is permanently retired. Informational.

---

- **[Pre-existing, sev 4.0]** PdsGoalCard.tsx — `summaryData == null && !loading` collapses silently to `$0.00`. Not introduced by this PR; documented above as a Suggestion.
- **[Pre-existing, sev 2.5]** A11Y-3 — `InfoIcon` tooltip in `otherBreakdown.tsx:163-167` isn't wrapped in an `IconButton`, so keyboard users can't focus it. Pre-dates the PR.

---

### Dismissed Findings (Acknowledged by Developer)

These were flagged again by the relevant agents in this review pass but match prior `/dismiss` replies. They do not count toward the verdict.

- **[Dismissed, sev 5.5]** otherBreakdown.tsx:137 — Gross-up formulas opaque to non-finance users (UX-2, this run). Dismissed by @wjames111: *"might revisit this later. Right now getting the formula correct and keeping parity between calculations and the formula displays is the priority."* Financial agent supports this prioritization (literal-math display is more auditable than a friendlier rewording).
- **[Dismissed, sev 2.0]** OtherExpenses.ts:63 — No guards on `rate ≥ 1` for gross-up divisions (Financial #4, this run, conceded in debate down from 4.5). Dismissed by @wjames111: *"this is the general direction the codebase has taken."* Confirmed by codebase scan — the sister `GoalCalculator/Shared/calculateTotals.ts:92` follows the same pattern.

---

### Cross-Cutting Checks

- **"Fix one, fix all":** Architecture and Financial both verified `src/components/HrTools/GoalCalculator/Shared/calculateTotals.ts:92` already uses the gross-up pattern. The PDS calculator is being brought into consistency with the (correct) sister — no parallel-fix gap.
- **Safeguard parity:** N/A — no auth/validation/data-mutation operations changed.
- **Dependency impact:** Architecture verified no callers of the deleted `calculatePdsGoalTotal` remain. Safe deletion.

---

### Review Summary Table

| Agent | Blocker | Important | Medium | Suggestions | Confidence |
|---|---|---|---|---|---|
| Architecture | 0 | 0 | 0 | 6 | High |
| Testing | 0 | 0 | 2 | 7 | Medium-High |
| Standards | 0 | 0 | 1 | 1 | High |
| UX | 1 | 0 | 2 (after dismissals) | 7 | High |
| Financial Reporting | 0 | 1 | 1 (after dismissals) | 3 | High |
| **Active total** | **1** | **1** | **5** | **11** | |

---

*Generated by `/quality:agent-review` (multi-agent cross-examination review).*

src/components/HrTools/PdsGoalCalculator/SupportItem/salaryBreakdown.tsx
wjames111: /dismiss

src/components/HrTools/PdsGoalCalculator/SupportItem/otherBreakdown.tsx
dr-bizz: Is looks like you're using a weird character here for the minus. 
```suggestion
        '(Subtotal + Attrition) ÷ (1 - {{rate}}) - (Subtotal + Attrition)',
```

src/components/HrTools/PdsGoalCalculator/SupportItem/otherBreakdown.tsx
dr-bizz:
```suggestion
        '(Subtotal + Attrition + Credit Card Fees) ÷ (1 - {{rate}}) - (Subtotal + Attrition + Credit Card Fees)',
```

src/components/HrTools/PdsGoalCalculator/SupportItem/salaryBreakdown.tsx
dr-bizz:
```suggestion
      formula: t('Monthly Base x {{rate}}', {
```

dr-bizz: I guess there is no multiplier that is `0`?

dr-bizz (approved):
This is great. It works well.

I just spotted that the salary pay rate should be the monthly rate, not the yearly rate. We can keep it as the yearly rate, but we will need to make it so obvious to staff that this is the case.

Image


Also, we should say that the percentage is here:

Image

PR #1778, opened 2026-05-15
[MPDX-9574] - Show required fields on load in PDS Goal Calculator

wjames111:
## Description

- Required fields in the PDS Goal Calculator Setup step now display their validation error (red border + helper text, e.g. "Goal Name is a required field") immediately on load, instead of staying silent until the user touches them.
- Removed the `touched`/empty-value gate in the PDS-only `AutosaveTextField` so the underlying schema error surfaces on first render.
- Made `Benefits` optional in the Setup schema (was conditionally required for non-Part-time staff). The positive-number check still applies if a value is entered.
- Updated `AutosaveTextField` tests to assert the new on-load behavior.

Related ticket: [MPDX-9574](https://jira.cru.org/browse/MPDX-9574)

## Testing

- Go to the PDS Goal Calculator and create a new goal (or open one with an empty Goal Name).
- On the Setup step, confirm that empty required fields (Goal Name, Employment Status, Pay Type, Pay Rate, Hours Worked when Hourly) show their red error state and required-field helper text on load — no need to click into them first.
- Confirm that `Benefits` does NOT show a required-field error when empty, regardless of Full-time / Part-time status.
- Enter a negative value into `Benefits` and confirm the "Benefits must be a positive number" error still appears.
- Fill in the required fields and confirm errors clear and the Continue button enables as expected.

## Checklist:

### DIFF
--- a/src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.tsx
@@ -65,14 +65,10 @@ export const SetupStep: React.FC = () => {
               .positive(t('Hours Worked must be a positive number')),
           otherwise: (s) => s.optional().nullable(),
         }),
-        benefits: yup.number().when('status', {
-          is: (val: string) => val !== DesignationSupportStatus.PartTime,
-          then: (s) =>
-            s
-              .required(t('Benefits is a required field'))
-              .positive(t('Benefits must be a positive number')),
-          otherwise: (s) => s.optional().nullable(),
-        }),
+        benefits: yup
+          .number()
+          .nullable()
+          .positive(t('Benefits must be a positive number')),
       }),
     [t],
   );
--- a/src/components/HrTools/PdsGoalCalculator/Shared/Autosave/AutosaveTextField.test.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/Shared/Autosave/AutosaveTextField.test.tsx
@@ -232,6 +232,50 @@ describe('AutosaveTextField', () => {
       );
       expect(input).toHaveAttribute('aria-invalid', 'true');
     });
+
+    it('clears validation error after user fills required field', async () => {
+      const { findByRole } = renderRequired();
+
+      const input = await findByRole('textbox', { name: 'Goal Name' });
+      await waitFor(() =>
+        expect(input).toHaveAttribute('aria-invalid', 'true'),
+      );
+
+      userEvent.type(input, 'Filled in');
+
+      await waitFor(() =>
+        expect(input).not.toHaveAttribute('aria-invalid', 'true'),
+      );
+      expect(input).toHaveAccessibleDescription('Enter the goal name');
+    });
+  });
+
+  describe('optional field', () => {
+    const optionalSchema = yup.object({
+      name: yup.string().nullable(),
+    });
+
+    it('does not show validation error for an empty optional field on load', async () => {
+      const { findByRole } = render(
+        <PdsGoalCalculatorTestWrapper
+          calculationMock={{ ...calculationMock, name: '' }}
+          onCall={mutationSpy}
+        >
+          <AutosaveTextField
+            label="Goal Name"
+            fieldName="name"
+            schema={optionalSchema}
+            helperText="Enter the goal name"
+          />
+        </PdsGoalCalculatorTestWrapper>,
+      );
+
+      const input = await findByRole('textbox', { name: 'Goal Name' });
+      await waitFor(() => expect(input).toHaveValue(''));
+
+      expect(input).not.toHaveAttribute('aria-invalid', 'true');
+      expect(input).toHaveAccessibleDescription('Enter the goal name');
+    });
   });
 
   describe('select input', () => {
--- a/src/components/HrTools/PdsGoalCalculator/Shared/Autosave/AutosaveTextField.test.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/Shared/Autosave/AutosaveTextField.test.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { MenuItem } from '@mui/material';
-import { fireEvent, render, waitFor } from '@testing-library/react';
+import { render, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import * as yup from 'yup';
 import {
@@ -221,28 +221,16 @@ describe('AutosaveTextField', () => {
         </PdsGoalCalculatorTestWrapper>,
       );
 
-    it('hides validation error for an empty untouched field', async () => {
+    it('shows validation error for an empty required field on load', async () => {
       const { findByRole } = renderRequired();
 
       const input = await findByRole('textbox', { name: 'Goal Name' });
       await waitFor(() => expect(input).toHaveValue(''));
-
-      expect(input).toHaveAccessibleDescription('Enter the goal name');
-      expect(input).not.toHaveAttribute('aria-invalid', 'true');
-    });
-
-    it('shows validation error after the field is touched', async () => {
-      const { findByRole } = renderRequired();
-
-      const input = await findByRole('textbox', { name: 'Goal Name' });
-      await waitFor(() => expect(input).toHaveValue(''));
-
-      fireEvent.focus(input);
-      fireEvent.blur(input);
 
       await waitFor(() =>
         expect(input).toHaveAccessibleDescription('Goal Name is required'),
       );
+      expect(input).toHaveAttribute('aria-invalid', 'true');
     });
   });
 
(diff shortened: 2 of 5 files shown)


src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.tsx
```diff
@@ -65,14 +65,10 @@ export const SetupStep: React.FC = () => {
...
         }),
-        benefits: yup.number().when('status', {
-          is: (val: string) => val !== DesignationSupportStatus.PartTime,
-          then: (s) =>
-            s
-              .required(t('Benefits is a required field'))
-              .positive(t('Benefits must be a positive number')),
-          otherwise: (s) => s.optional().nullable(),
-        }),
+        benefits: yup
```
wjames111, line 68: [Medium] **Schema / completion contradiction for `benefits`**

Flagged by UX (6.5), Architecture (5.0), Financial (3.5), Standards (server-parity warn). After this PR, `benefits` is the only required-for-completion field with no field-level error feedback when blank.

- This schema accepts `null` and shows no error.
- [`pdsCompletion.ts:25`](src/components/HrTools/PdsGoalCalculator/Shared/pdsCompletion.ts#L25) still requires `(benefits ?? 0) > 0` for non-PartTime to mark Setup complete.
- Every other required field (name, status, salaryOrHourly, payRate, hoursWorkedPerWeek) now renders its required-error on load — benefits is the orphan.

**User impact:** A Full-time user leaves Benefits blank, sees no in-field error, but the Setup section stays flagged "incomplete" in the side nav with no actionable cue at the field.

**Fix options** (pick one; needs product call):
1. (Cleanest, matches PR title "Make Benefits Optional") — Remove the benefits check from `pdsCompletion.isSetupComplete` so the section can mark complete with `benefits = null`.
2. (Conservative) — Restore the conditional `.required()` on benefits to match the existing completion rule.
3. (Bandaid) — Keep both as-is and add a non-error helper hint on Benefits like "Optional, but required to mark Setup complete."

**Also confirm server parity:** the upstream API should not still reject `null` benefits for non-PartTime, or saves will fail silently after the client validation passed. (`DesignationSupportCalculationUpdateInput.benefits` is `InputMaybe<Float>`, suggesting the server accepts it — confirm with a quick smoke test.)

src/components/HrTools/PdsGoalCalculator/Shared/Autosave/AutosaveTextField.tsx
```diff
@@ -23,11 +23,8 @@ export const AutosaveTextField: React.FC<AutosaveTextFieldProps> = ({
     schema,
     saveOnChange: props.select,
   });
-  const [touched, setTouched] = useState(false);
 
-  const showValidationError = Boolean(
-    fieldProps.error && (touched || fieldProps.value !== ''),
-  );
+  const showValidationError = Boolean(fieldProps.error);
```
wjames111, line 27: [Suggestion] **Behavior change applies to all consumers of `AutosaveTextField`, not just `benefits`**

Removing the `touched` gate flips error-display semantics for every consumer (`formType`, `status`, `salaryOrHourly`, `payRate`, `hoursWorkedPerWeek`, `benefits`). On first paint of a new PDS goal, the user will see four required-field errors simultaneously across the Setup card. The PR title says this is intentional — flagging only to confirm with design and worth a screenshot in the PR description.

**Adjacent note (low priority):** Three `AutosaveTextField` files now exist in the repo (`PdsGoalCalculator/Shared/`, `SalaryCalculator/Autosave/`, `GoalCalculator/CalculatorSettings/Categories/Autosave/`) and have diverged on the touched-state question. This PR closes the gap between PDS and the other two; a follow-up to consolidate them into one shared component would prevent future drift.

src/components/HrTools/PdsGoalCalculator/Shared/Autosave/AutosaveTextField.test.tsx
```diff
@@ -221,28 +221,16 @@ describe('AutosaveTextField', () => {
         </PdsGoalCalculatorTestWrapper>,
       );
 
-    it('hides validation error for an empty untouched field', async () => {
+    it('shows validation error for an empty required field on load', async () => {
```
wjames111, line 224: [Suggestion] **Missing inverse / recovery test coverage**

The new test asserts `aria-invalid='true'` on load for an empty required field — good. Two adjacent gaps would round out the coverage:

1. **No-error inverse:** assert that an empty *optional* field does NOT have `aria-invalid='true'` on load. This is exactly the scenario the PR's other change (making `benefits` optional) creates. Without this assertion, a future regression that re-tightens the schema would not fail any test.

2. **Recovery path:** assert that a field with a validation error becomes valid (and `aria-invalid` clears) after the user types a valid value. The negative-number tests at lines 97 and 178 cover valid→invalid; nothing currently covers invalid→valid.

Example skeleton:
```tsx
it('clears validation error after user fills required field', async () => {
  const { findByRole } = renderRequired();
  const input = await findByRole('textbox', { name: 'Goal Name' });
  await waitFor(() => expect(input).toHaveAttribute('aria-invalid', 'true'));
  userEvent.type(input, 'Filled in');
  await waitFor(() =>
    expect(input).not.toHaveAttribute('aria-invalid', 'true'),
  );
  expect(input).toHaveAccessibleDescription('Enter the goal name');
});
```

wjames111 (commented):
## Multi-Agent Code Review — APPROVED WITH SUGGESTIONS

**Risk:** LOW (2/10) — small validation behavior change scoped to PDS Goal Calculator.
**Verdict:** No blockers. 1 medium finding + several lower-priority suggestions.
**Agents:** Architecture, Testing, Standards, UX, Financial Reporting (5 agents, full debate round).

### Headline finding

With this PR, `benefits` is the only required-for-completion field that no longer renders a field-level error when blank, while `pdsCompletion.ts:25` still gates Setup completion on `benefits > 0` for non-PartTime. Every other required field (name, status, salaryOrHourly, payRate, hoursWorkedPerWeek) now surfaces its error on load — benefits is the orphan. See the inline comment on `SetupStep.tsx:68` for the suggested fix paths.

### What's working

- All downstream `benefits` consumers are null-safe (`OtherExpenses.ts:47`, `pdsCompletion.ts:25`, `PdsSummaryTable.tsx:139`, `otherBreakdown.tsx:89`). No financial-correctness regression.
- Standards checklist: clean pass — named exports, i18n via `t()`, no `any`, no `console.log`, no `new Date()`, no unused imports.
- Removing the `touched` state brings the PDS `AutosaveTextField` into alignment with sibling `SalaryCalculator/Autosave/AutosaveTextField` and `GoalCalculator/.../Autosave/AutosaveTextField`, neither of which had a touched gate. Net debt: better.
- `DesignationSupportCalculationUpdateInput.benefits` is already `InputMaybe<Float>`, so the server accepts `null` — schema relaxation is server-compatible.

---

These findings reference files outside the PR diff and cannot be posted as line comments.

**[Medium 5.5] `src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.test.tsx`** — Missing positive test for the schema change
- The Yup schema for `benefits` was changed from conditional-required to always-optional. The existing test at `SetupStep.test.tsx:355` uses `benefits: null` but only asserts the field is visible, never asserts it does NOT render `aria-invalid='true'` on load. A future regression that re-introduces `.required()` on benefits would pass all existing tests.
- Suggested test:
  ```tsx
  it('does not show validation error for empty Benefits when Full-time', async () => {
    const { findByRole } = renderSetup({
      calculationMock: { ...fullTimeSalariedMock, benefits: null },
    });
    const benefits = await findByRole('spinbutton', { name: 'Benefits' });
    await waitFor(() => expect(benefits).toHaveValue(null));
    expect(benefits).not.toHaveAttribute('aria-invalid', 'true');
  });
  ```

**[Suggestion 3.5] `SetupStep.test.tsx:360`** — Stale comment
- The comment `// Benefits is visible and required when Full-time` is now inaccurate; benefits is no longer required. Update or extend the test to assert benefits is not in an error state when empty.

**[Suggestion 4.0] `src/components/HrTools/PdsGoalCalculator/SummaryReport/PdsSummaryTable.tsx:139`** — Display honesty for `benefits = null`
- With benefits now optional, a Full-time user can save the goal with `benefits = null`. `calculateOtherExpenses` coerces null → 0, and the Summary table renders `$0.00` for the Benefits line — indistinguishable from a real entered $0. Per the rule "A report with zero donations should render 'no data' — not `$0.00` that looks like real data," consider rendering "—" or "Not entered" when `calculation.benefits == null && isFullTime`. Mitigation already in place: `SetupSectionList` flags the section as incomplete via `pdsCompletion.ts:25`, so the user has another signal — this is a polish-tier nice-to-have, not a blocker.

---

### Severity distribution

| Severity | Count |
|----------|-------|
| Critical (9.0-10.0) | 0 |
| High (8.0-8.9) | 0 |
| Important (7.0-7.9) | 0 |
| Medium (5.0-6.9) | 3 |
| Suggestion (< 5.0) | 9 |

### Auto-approval

This PR is LOW risk + APPROVED_WITH_SUGGESTIONS — qualifies for auto-approval. The repo's `ai-review-auto-approve.yml` workflow will be triggered after this review posts.

src/components/HrTools/PdsGoalCalculator/Shared/Autosave/AutosaveTextField.tsx
wjames111: /dismiss this should be done in a follow up PR. Not priority right now

PR #1780, opened 2026-05-15
MPDX-9568 PDS Goal Calculator: section descriptions, field labels, and form fixes

wjames111:
## Description

Bundles UI/text refinements across the PDS Goal Calculator from six related Jira tickets, plus a small Pay Type bug fix.

- **MPDX-9569** — Show the geographic location next to the multiplier amount in the Salary breakdown (e.g. `4.0% (Atlanta)`), update the Gross Monthly Pay formula to include the location, and append each location's percentage to the options in the Location dropdown.
- **MPDX-9568** — Replace the "Total Reimbursable Expenses" title info tooltip with a description below the heading explaining the `$300/month` minimum. On the Monthly Reimbursable grid, show the per-month maximums inline with the Ministry Cell Phone and Ministry Internet field names and add a description above the grid clarifying that amounts above the max will not be saved.
- **MPDX-9567** — Remove the title-level info tooltips from the Annual / Monthly / Total Reimbursable sections (replaced with descriptive text) and add row-level info icons on Ministry Cell Phone / Ministry Internet explaining why those fields are prepopulated.
- **MPDX-9564** — Split the Form Type helper text so the `Default` description and the `Simple` description render on separate lines.
- **MPDX-9563** — Round the Average Hours Worked Per Week to two decimal places in the grid footer and when applying the value to Hours Worked.
- **MPDX-9474** — Add descriptive headings at the top of the Reimbursable Expenses, Support Items (with Salary / Other subsections), and Summary Report steps.

Additional minor fix:

- Switching Pay Type between Salaried and Hourly now clears the Pay Rate, preventing stale hourly/salary values from carrying over to the new pay type.

Related Jira tickets: MPDX-9569, MPDX-9568, MPDX-9567, MPDX-9564, MPDX-9563, MPDX-9474.

## Testing

Navigate to the PDS Goal Calculator (HR Tools → PDS Goal Calculator) and walk through each step:

**Setup step**
- Form Type helper text shows the `Default` sentence and `Simple` sentence on two separate lines.
- Open the Location dropdown — each option is suffixed with its percentage, e.g. `Orlando, FL (6%)`.
- Switch Pay Type from Hourly → Salaried (and back) and confirm the Pay Rate field clears each time.
- In the Hours Per Week grid, enter values that don't divide evenly (e.g. Regular Week 40 hrs × 48 wks plus Travel 7 hrs × 4 wks) and confirm the footer average and the value applied to Hours Worked both show two decimal places (`37.46`).

**Reimbursable Expenses step**
- Confirm the step now has a heading and description at the top.
- Monthly section: shows a description below the heading; Ministry Cell Phone and Ministry Internet rows show `(max $35/mo)` / `(max $30/mo)` inline with the field name and an info icon whose tooltip reads "Pre-filled with the maximum allowed amount. Edit to a lower value if needed."
- Annual section: shows a description below the heading and no title info icon.
- Total Reimbursable Expenses: shows the `$300` minimum description in place of the previous info tooltip; the floor still applies (totals below `$300` snap to `$300`).

**Support Items step**
- Step has a heading and description, and both the Salary and Other subsections have their own descriptions.
- With a geographic location set (e.g. Atlanta), the Geographic Multiplier row shows `(Atlanta)` next to the percentage and the Gross Monthly Pay formula reads `Monthly Base × (1 + Geographic Multiplier (Atlanta))`.
- With no location set, the suffix and the location in the formula are both omitted.

**Summary step**
- Shows a heading and description above the summary table.

## Checklist:

### DIFF
--- a/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/MonthlyReimbursableSection.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/MonthlyReimbursableSection.tsx
@@ -23,13 +23,7 @@ export const MonthlyReimbursableSection: React.FC = () => {
     ? calculateReimbursableTotals(calculation).monthlySubtotal
     : 0;
 
-  const buildCappedLabel = (label: string, max: number | undefined) =>
-    max !== undefined
-      ? t('{{label}} (max {{max}}/mo)', {
-          label,
-          max: currencyFormat(max, 'USD', locale),
-        })
-      : label;
+  const formatMaxPerMonth = (max: number) => currencyFormat(max, 'USD', locale);
 
   const prepopulatedTooltip = t(
     'Pre-filled with the maximum allowed amount. Edit to a lower value if needed.',
@@ -38,13 +32,23 @@ export const MonthlyReimbursableSection: React.FC = () => {
   const fields: ReimbursableField[] = [
     {
       fieldName: 'ministryCellPhone',
-      label: buildCappedLabel(t('Ministry Cell Phone'), phoneMax),
+      label:
+        phoneMax !== undefined
+          ? t('Ministry Cell Phone (max {{max}}/mo)', {
+              max: formatMaxPerMonth(phoneMax),
+            })
+          : t('Ministry Cell Phone'),
       max: phoneMax,
       tooltip: prepopulatedTooltip,
     },
     {
       fieldName: 'ministryInternet',
-      label: buildCappedLabel(t('Ministry Internet'), internetMax),
+      label:
+        internetMax !== undefined
+          ? t('Ministry Internet (max {{max}}/mo)', {
+              max: formatMaxPerMonth(internetMax),
+            })
+          : t('Ministry Internet'),
       max: internetMax,
       tooltip: prepopulatedTooltip,
     },
--- a/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/MonthlyReimbursableSection.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/MonthlyReimbursableSection.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useGoalCalculatorConstants } from 'src/hooks/useGoalCalculatorConstants';
+import { useLocale } from 'src/hooks/useLocale';
+import { currencyFormat } from 'src/lib/intlFormat';
 import { usePdsGoalCalculator } from '../Shared/PdsGoalCalculatorContext';
 import { calculateReimbursableTotals } from '../calculations/reimbursableExpenses';
 import {
@@ -10,6 +12,7 @@ import {
 
 export const MonthlyReimbursableSection: React.FC = () => {
   const { t } = useTranslation();
+  const locale = useLocale();
   const { calculation } = usePdsGoalCalculator();
   const { goalMiscConstants } = useGoalCalculatorConstants();
   const phoneMax = goalMiscConstants.REIMBURSEMENTS_WITH_MAXIMUM?.PHONE?.fee;
@@ -20,16 +23,30 @@ export const MonthlyReimbursableSection: React.FC = () => {
     ? calculateReimbursableTotals(calculation).monthlySubtotal
     : 0;
 
+  const buildCappedLabel = (label: string, max: number | undefined) =>
+    max !== undefined
+      ? t('{{label}} (max {{max}}/mo)', {
+          label,
+          max: currencyFormat(max, 'USD', locale),
+        })
+      : label;
+
+  const prepopulatedTooltip = t(
+    'Pre-filled with the maximum allowed amount. Edit to a lower value if needed.',
+  );
+
   const fields: ReimbursableField[] = [
     {
       fieldName: 'ministryCellPhone',
-      label: t('Ministry Cell Phone'),
+      label: buildCappedLabel(t('Ministry Cell Phone'), phoneMax),
       max: phoneMax,
+      tooltip: prepopulatedTooltip,
     },
     {
       fieldName: 'ministryInternet',
-      label: t('Ministry Internet'),
+      label: buildCappedLabel(t('Ministry Internet'), internetMax),
       max: internetMax,
+      tooltip: prepopulatedTooltip,
     },
     { fieldName: 'mpdNewsletter', label: t('MPD Newsletter') },
     { fieldName: 'mpdMiscellaneous', label: t('MPD Miscellaneous') },
@@ -43,6 +60,9 @@ export const MonthlyReimbursableSection: React.FC = () => {
   return (
     <ReimbursableExpensesGrid
       title={t('Monthly Reimbursable Expenses')}
+      description={t(
+        'Ministry Cell Phone and Ministry Internet reimbursements are capped at the per-month maximums shown next to each field name. Amounts above the maximum will not be saved.',
+      )}
       fields={fields}
       subtotalLabel={t('Subtotal Monthly')}
       subtotalValue={monthlySubtotal}
--- a/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/MonthlyReimbursableSection.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/MonthlyReimbursableSection.tsx
@@ -61,7 +61,7 @@ export const MonthlyReimbursableSection: React.FC = () => {
     <ReimbursableExpensesGrid
       title={t('Monthly Reimbursable Expenses')}
       description={t(
-        'Ministry Cell Phone and Ministry Internet reimbursements are capped at the per-month maximums shown next to each field name. Amounts above the maximum will not be saved.',
+        'Ministry Cell Phone and Ministry Internet reimbursements are capped at the per-month maximums shown next to each field name. Amounts entered above the maximum will be saved as the maximum.',
       )}
       fields={fields}
       subtotalLabel={t('Subtotal Monthly')}
--- a/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesGrid.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesGrid.tsx
@@ -134,11 +134,7 @@ export const ReimbursableExpensesGrid: React.FC<
       <Stack direction="row" alignItems="center" gap={0.5}>
         <span>{row.label}</span>
         <Tooltip title={row.tooltip}>
-          <InfoIcon
-            color="action"
-            fontSize="small"
-            aria-label={row.tooltip}
-          />
+          <InfoIcon color="action" fontSize="small" aria-label={row.tooltip} />
         </Tooltip>
       </Stack>
     );
--- a/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesGrid.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesGrid.tsx
@@ -4,12 +4,14 @@ import {
   Box,
   Card,
   FormHelperText,
+  IconButton,
   Stack,
   Tooltip,
   Typography,
   styled,
 } from '@mui/material';
 import { GridColDef, GridRenderCellParams } from '@mui/x-data-grid';
+import { useSnackbar } from 'notistack';
 import { useTranslation } from 'react-i18next';
 import { useLocale } from 'src/hooks/useLocale';
 import { currencyFormat } from 'src/lib/intlFormat';
@@ -68,6 +70,7 @@ export const ReimbursableExpensesGrid: React.FC<
 }) => {
   const { t } = useTranslation();
   const locale = useLocale();
+  const { enqueueSnackbar } = useSnackbar();
   const { calculation } = usePdsGoalCalculator();
   const saveField = useSaveField();
   const [cellErrors, setCellErrors] = useState<Record<string, string>>({});
@@ -102,10 +105,17 @@ export const ReimbursableExpensesGrid: React.FC<
       return newRow;
     }
 
-    const amount =
-      field.max !== undefined && newRow.amount > field.max
-        ? field.max
-        : newRow.amount;
+    let amount = newRow.amount;
+    if (field.max !== undefined && newRow.amount > field.max) {
+      amount = field.max;
+      enqueueSnackbar(
+        t('{{label}} reduced to its maximum of {{max}}.', {
+          label: field.label,
+          max: currencyFormat(field.max, 'USD', locale),
+        }),
+        { variant: 'info' },
+      );
+    }
 
     setCellErrors((prev) => {
       if (!(cellKey in prev)) {
@@ -130,7 +140,14 @@ export const ReimbursableExpensesGrid: React.FC<
       <Stack direction="row" alignItems="center" gap={0.5}>
         <span>{row.label}</span>
         <Tooltip title={row.tooltip}>
-          <InfoIcon color="action" fontSize="small" aria-label={row.tooltip} />
+          <IconButton
+            size="small"
+            aria-label={row.tooltip}
+            disableRipple
+            sx={{ p: 0 }}
+          >
+            <InfoIcon color="action" fontSize="small" />
+          </IconButton>
         </Tooltip>
       </Stack>
     );
--- a/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesGrid.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesGrid.tsx
@@ -76,18 +76,6 @@ export const ReimbursableExpensesGrid: React.FC<
     return null;
   }
 
-  const validateAmount = (field: ReimbursableField, amount: number) => {
-    if (amount < 0) {
-      return t('Amount must be positive');
-    }
-    if (field.max !== undefined && amount > field.max) {
-      return t('Amount cannot exceed {{max}}', {
-        max: currencyFormat(field.max, 'USD', locale),
-      });
-    }
-    return null;
-  };
-
   const rows: ReimbursableRow[] = [
     ...fields.map((field) => ({
       id: field.fieldName,
@@ -104,25 +92,33 @@ export const ReimbursableExpensesGrid: React.FC<
       return newRow;
     }
 
-    const { amount } = newRow;
     const cellKey = `${field.fieldName}-amount`;
-    const error = validateAmount(field, amount);
+
+    if (newRow.amount < 0) {
+      setCellErrors((prev) => ({
+        ...prev,
+        [cellKey]: t('Amount must be positive'),
+      }));
+      return newRow;
+    }
+
+    const amount =
+      field.max !== undefined && newRow.amount > field.max
+        ? field.max
+        : newRow.amount;
 
     setCellErrors((prev) => {
-      const next = { ...prev };
-      if (error) {
-        next[cellKey] = error;
-      } else {
-        delete next[cellKey];
+      if (!(cellKey in prev)) {
+        return prev;
       }
+      const next = { ...prev };
+      delete next[cellKey];
       return next;
     });
 
-    if (!error) {
-      await saveField({ [field.fieldName]: amount });
-    }
+    await saveField({ [field.fieldName]: amount });
 
-    return newRow;
+    return { ...newRow, amount };
   };
 
   const renderLabelCell = (params: GridRenderCellParams<ReimbursableRow>) => {
--- a/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesGrid.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesGrid.tsx
@@ -24,17 +24,19 @@ export interface ReimbursableField {
   fieldName: ReimbursableFieldName;
   label: string;
   max?: number;
+  tooltip?: string;
 }
 
 interface ReimbursableRow {
   id: ReimbursableFieldName | 'total';
   label: string;
   amount: number;
+  tooltip?: string;
 }
 
 interface ReimbursableExpensesGridProps {
   title: string;
-  titleTooltip?: string;
+  description?: string;
   fields: ReimbursableField[];
   subtotalLabel: string;
   subtotalValue: number;
@@ -58,7 +60,7 @@ export const ReimbursableExpensesGrid: React.FC<
   ReimbursableExpensesGridProps
 > = ({
   title,
-  titleTooltip,
+  description,
   fields,
   subtotalLabel,
   subtotalValue,
@@ -91,6 +93,7 @@ export const ReimbursableExpensesGrid: React.FC<
       id: field.fieldName,
       label: field.label,
       amount: calculation[field.fieldName] ?? 0,
+      tooltip: field.tooltip,
     })),
     { id: 'total', label: subtotalLabel, amount: subtotalValue },
   ];
@@ -122,6 +125,25 @@ export const ReimbursableExpensesGrid: React.FC<
     return newRow;
   };
 
+  const renderLabelCell = (params: GridRenderCellParams<ReimbursableRow>) => {
+    const { row } = params;
+    if (!row.tooltip) {
+      return row.label;
+    }
+    return (
+      <Stack direction="row" alignItems="center" gap={0.5}>
+        <span>{row.label}</span>
+        <Tooltip title={row.tooltip}>
+          <InfoIcon
+            color="action"
+            fontSize="small"
+            aria-label={row.tooltip}
+          />
+        </Tooltip>
+      </Stack>
+    );
+  };
+
   const renderAmountCell = (params: GridRenderCellParams<ReimbursableRow>) => {
     const cellKey = `${params.id}-amount`;
     const error = cellErrors[cellKey];
@@ -145,6 +167,7 @@ export const ReimbursableExpensesGrid: React.FC<
       flex: 1,
       minWidth: 200,
       sortable: false,
+      renderCell: renderLabelCell,
     },
     {
       field: 'amount',
@@ -162,14 +185,14 @@ export const ReimbursableExpensesGrid: React.FC<
 
   return (
     <>
-      <Stack direction="row" alignItems="center" gap={0.5} pb={2}>
+      <Box pb={2}>
         <Typography variant="h6">{title}</Typography>
-        {titleTooltip && (
-          <Tooltip title={titleTooltip}>
-            <InfoIcon color="action" aria-label={titleTooltip} />
-          </Tooltip>
+        {description && (
+          <Typography variant="body2" color="text.secondary" pt={0.5}>
+            {description}
+          </Typography>
         )}
-      </Stack>
+      </Box>
       <StyledCard>
         <BaseGrid
           rows={rows}
--- a/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesStep.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesStep.tsx
@@ -11,7 +11,9 @@ export const ReimbursableExpensesStep: React.FC = () => {
   return (
     <>
       <Box pb={4}>
-        <Typography variant="h6">{t('Reimbursable Expenses')}</Typography>
+        <Typography variant="h6" component="h2">
+          {t('Reimbursable Expenses')}
+        </Typography>
         <Typography pt={1}>
           {t(
             'Enter the ministry expenses you are reimbursed for each year. Monthly entries are used as-is and annual entries are divided by 12; the combined total is included in your support goal.',
--- a/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesStep.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesStep.tsx
@@ -1,12 +1,23 @@
 import React from 'react';
-import { Divider } from '@mui/material';
+import { Box, Divider, Typography } from '@mui/material';
+import { useTranslation } from 'react-i18next';
 import { AnnualReimbursableSection } from './AnnualReimbursableSection';
 import { MonthlyReimbursableSection } from './MonthlyReimbursableSection';
 import { TotalReimbursableSection } from './TotalReimbursableSection';
 
 export const ReimbursableExpensesStep: React.FC = () => {
+  const { t } = useTranslation();
+
   return (
     <>
+      <Box pb={4}>
+        <Typography variant="h6">{t('Reimbursable Expenses')}</Typography>
+        <Typography pt={1}>
+          {t(
+            'Enter the ministry expenses you are reimbursed for each year. Monthly entries are used as-is and annual entries are divided by 12; the combined total is included in your support goal.',
+          )}
+        </Typography>
+      </Box>
       <MonthlyReimbursableSection />
       <Divider sx={{ mx: -4, my: 4 }} />
       <AnnualReimbursableSection />
--- a/src/components/HrTools/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
@@ -24,6 +24,8 @@ import {
 import { useSnackbar } from 'notistack';
 import { useTranslation } from 'react-i18next';
 import { BaseGrid } from 'src/components/HrTools/GoalCalculator/SharedComponents/GoalCalculatorGrid/BaseGrid';
+import { useLocale } from 'src/hooks/useLocale';
+import { numberFormat } from 'src/lib/intlFormat';
 import {
   useCreateDesignationSupportHoursItemMutation,
   useDeleteDesignationSupportHoursItemMutation,
@@ -64,6 +66,7 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
   onApply,
 }) => {
   const { t } = useTranslation();
+  const locale = useLocale();
   const { enqueueSnackbar } = useSnackbar();
   const { calculation, trackMutation } = usePdsGoalCalculator();
   const [createHoursItem] = useCreateDesignationSupportHoursItemMutation();
@@ -114,6 +117,9 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
     () => (totalWeeks > 0 ? totalHours / totalWeeks : 0),
     [totalHours, totalWeeks],
   );
+
+  const roundedAverageHoursPerWeek =
+    Math.round(averageHoursPerWeek * 100) / 100;
 
   const weeksRemaining = MAX_TOTAL_WEEKS - totalWeeks;
 
@@ -490,7 +496,10 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
             {t('Average Hours Worked Per Week')}
           </Typography>
           <Typography variant="body2" fontWeight="bold">
-            {averageHoursPerWeek.toFixed(2)}
+            {numberFormat(roundedAverageHoursPerWeek, locale, {
+              minimumFractionDigits: 2,
+              maximumFractionDigits: 2,
+            })}
           </Typography>
         </FooterRow>
       </StyledCard>
@@ -509,7 +518,7 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
           <Button
             variant="contained"
             disabled={weeksRemaining > 0}
-            onClick={() => onApply(Math.round(averageHoursPerWeek * 100) / 100)}
+            onClick={() => onApply(roundedAverageHoursPerWeek)}
           >
             {t('Apply to Hours Worked')}
           </Button>
--- a/src/components/HrTools/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
@@ -490,7 +490,7 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
             {t('Average Hours Worked Per Week')}
           </Typography>
           <Typography variant="body2" fontWeight="bold">
-            {averageHoursPerWeek.toFixed(1)}
+            {averageHoursPerWeek.toFixed(2)}
           </Typography>
         </FooterRow>
       </StyledCard>
@@ -509,7 +509,7 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
           <Button
             variant="contained"
             disabled={weeksRemaining > 0}
-            onClick={() => onApply(averageHoursPerWeek)}
+            onClick={() => onApply(Math.round(averageHoursPerWeek * 100) / 100)}
           >
             {t('Apply to Hours Worked')}
           </Button>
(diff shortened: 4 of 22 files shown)


src/components/HrTools/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
```diff
@@ -509,7 +509,7 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
           <Button
             variant="contained"
             disabled={weeksRemaining > 0}
-            onClick={() => onApply(averageHoursPerWeek)}
+            onClick={() => onApply(Math.round(averageHoursPerWeek * 100) / 100)}
```
wjames111: [High] **Display vs. applied rounding divergence — BLOCKER.**

This line uses `Math.round(x*100)/100` while [line 493](src/components/HrTools/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx#L493) uses `.toFixed(2)`. These rounding strategies diverge on `.xx5` floating-point boundaries because the underlying IEEE-754 representation of e.g. `0.175` is `0.17499...`, which `.toFixed(2)` rounds to `"0.17"` while `Math.round(0.175 * 100) / 100` lifts to `0.18` (the multiplication scales the error past the half-boundary).

Verified divergent cases (Financial agent, via Node):

| totalHours | x = h/52 | Displayed `.toFixed(2)` | Applied `Math.round(*100)/100` |
|---|---|---|---|
| 9.1 | 0.175 | **0.17** | **0.18** |
| 139.1 | 2.675 | **2.67** | **2.68** |
| 152.1 | 2.925 | **2.92** | **2.93** |

Reachable from valid user input (the grid accepts decimal hours per week, weeksRemaining=0 gates the Apply button at totalWeeks=52). The drifted value flows through `saveField({ hoursWorkedPerWeek })` → `calculateSalaryTotals` (`monthlyBase = payRate * hours * 52 / 12`) → geographic multiplier → FICA → goal total. At $25/hr, ~$1.08/mo base drift; ~$13–15/yr per staff member.

Breaks WYSIWYS ("what you see is what you save"), the cardinal UX rule for goal/donation UIs. Also a Standards violation per `.CLAUDE/rules/code-review.md` Financial focus area: *"Rounding should happen at the display boundary via intlFormat / Intl.NumberFormat, not sprinkled through calculation code. Flag any `.toFixed(n)` used inside aggregation logic."*

**Fix — round once and reuse:**

```tsx
const roundedAverage = Math.round(averageHoursPerWeek * 100) / 100;

// line 493:
<Typography variant="body2" fontWeight="bold">
  {numberFormat(roundedAverage, locale, { minimumFractionDigits: 2, maximumFractionDigits: 2 })}
</Typography>

// line 512:
onClick={() => onApply(roundedAverage)}
```

Add a regression test pairing displayed text with the `onApply` argument for an `.xx5` boundary (e.g., totalHours=9.1).

**Bigger picture:** the project already has `Math.round(x*100)/100` at `useSaveField.ts:40` and `AddAppealForm.tsx:81`. Consider extracting a shared `roundToCents(value)` helper in `src/lib/intlFormat.ts` so future authors can't reinvent a `.toFixed` divergence.

*Cross-agent consensus: Financial Reporting 8.0 (conceded from 9.0), Architecture, Testing (withdrew T3 in favor of this), UX, Standards (documented violation).*

src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.tsx
```diff
@@ -148,9 +160,20 @@ export const SetupStep: React.FC = () => {
               schema={schema}
               select
               label={t('Form Type')}
-              helperText={t(
-                'Default includes reimbursable expenses and 403b contributions in the goal total. Simple excludes them; existing entries are preserved and will count again if you switch back.',
-              )}
+              helperText={
```
wjames111, line 163: [Medium] Form Type `helperText` now renders two `<Box display="block">` spans, but no test asserts both sentences appear or that they render on separate lines. A future refactor could collapse them silently.

Add to [SetupStep.test.tsx](src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.test.tsx):

```tsx
it('renders both Form Type helper text sentences', async () => {
  const { findByText } = renderSetup({
    calculationMock: { ...fullTimeSalariedMock, formType: DesignationSupportFormType.Detailed },
  });
  expect(await findByText(/Default includes reimbursable expenses/)).toBeInTheDocument();
  expect(await findByText(/Simple excludes them/)).toBeInTheDocument();
});
```

Minor (sev 3.5, Suggestion): splitting one sentence across two `t()` keys means translators can't reflow word order for languages with different syntax. Acceptable trade-off for the two-clause structure but worth noting. (UX flagged at 7.0 then withdrew to 3.5 during debate — the standards rules don't prohibit multi-key splits.)

src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.tsx
```diff
@@ -74,12 +77,21 @@ export const SetupStep: React.FC = () => {
...
   const { goalGeographicConstantMap } = useGoalCalculatorConstants();
   const saveField = useSaveField();
+  const locale = useLocale();
 
   const locations = useMemo(
     () => Array.from(goalGeographicConstantMap.keys()),
     [goalGeographicConstantMap],
   );
 
+  const getLocationLabel = (location: string) => {
```
wjames111, line 87: [Medium] The 0%/undefined fallback branch (returns bare `location` with no parens) is untested. Only the `Orlando, FL (6%)` happy path is exercised in [SetupStep.test.tsx#L293](src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.test.tsx#L293).

Add a test using a constants mock with multiplier=0 (e.g., a `None` entry) asserting the option label has no parens. Negative multipliers (discount locations) are also unhandled by the guard — currently they would render as `(-5%)`; clarify whether that's intentional or whether the guard should be `multiplier > 0`.

Minor (sev 3.5): wrap in `useCallback([goalGeographicConstantMap, locale])` to match the surrounding `useMemo` style — `Autocomplete` will otherwise re-render its option list on every parent render.

wjames111 (commented):
## Multi-Agent Code Review — `/quality:agent-review` (standard mode)

**Verdict: BLOCKERS_FOUND** — 1 high-priority blocker, plus 7 medium-priority items and ~13 informational suggestions.

**Agents:** Architecture, Testing, Standards, UX, Financial Reporting (+ dependency-impact). Security and Data Integrity skipped (no auth/API/Apollo-cache paths touched).

**Risk:** 10/10 by formula (driven by file count — 10 production components touched), but the actual changes are narrow per-file (UI text + a few calc tweaks). Blast radius is fully contained in `src/components/HrTools/PdsGoalCalculator/`; no external API or `index.ts` re-exports affected. The `titleTooltip → description` interface rename has zero orphaned callers.

---

### Highlights

#### 🔴 High Priority Blocker (sev 8.0)

**HoursPerWeekGrid: displayed average can differ from applied average.** [HoursPerWeekGrid.tsx:493](https://github.com/CruGlobal/mpdx-react/blob/MPDX-9568/src/components/HrTools/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx#L493) uses `.toFixed(2)`, but [line 512](https://github.com/CruGlobal/mpdx-react/blob/MPDX-9568/src/components/HrTools/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx#L512) applies `Math.round(x*100)/100`. These diverge on `.xx5` floating-point boundaries — verified with concrete inputs (totalHours=9.1 → user sees `0.17`, but `0.18` is saved). The drifted value propagates through `saveField → calculateSalaryTotals → goal totals`. Also a documented Standards violation per `.CLAUDE/rules/code-review.md` (`.toFixed` in calculation paths). See inline comment for the fix. Cross-agent consensus (Financial, Architecture, Testing, UX, Standards).

#### 🟡 Medium Priority (sev 5.0-6.9)

1. **Missing test for multi-line Form Type helperText** — `SetupStep.tsx`
2. **Untested `getLocationLabel` 0%/undefined fallback branch** — `SetupStep.tsx`
3. **Pay Type TextField departs from `AutosaveTextField` pattern undocumented** — `SetupStep.tsx`
4. **Over-max reimbursable value retained in cell with red border despite "will not be saved" description** — `MonthlyReimbursableSection.tsx` + `ReimbursableExpensesGrid.tsx`
5. **Non-descriptive translation key `t('({{location}})')`** — `salaryBreakdown.tsx`
6. **Heading hierarchy: step + subsection both use `variant="h6"` (a11y)** — multiple step components

All details + before/after code blocks are in the inline review comments below.

#### 🟢 Suggestions (sev < 5)

- Pay Type switch silently clears `payRate` — consider visible inline feedback.
- Move SummaryReportStep heading above the loading guard.
- Extract a shared `roundToCents` helper to `src/lib/intlFormat.ts` (addresses root cause of the blocker — 3 sites already use this pattern).
- `MonthlyReimbursableSection.fields` rebuilt every render; `getLocationLabel` not memoized.
- 5 new heading+description blocks lack render tests (low risk — presentational JSX behind translation keys).
- InfoIcon `aria-label` may double-announce with the visible `(max $35/mo)` label.
- Various small testing edge cases (undefined max, negative multiplier, empty-string location).

- `HoursPerWeekGrid.test.tsx` uses several `await waitFor(() => getBy*)` patterns where the project prefers `findBy*`. Not introduced by this PR.

### Unresolved Debate

UX flagged Pay Type TextField's disabled-state visual treatment may differ from surrounding `AutosaveTextField` peers (sev 6.0). Architecture countered factually that `AutosaveTextField` has no saving adornment / spinner, so the visuals match (sev 3.5). The two arguments address different aspects. Recommendation: a visual check with `isMutating=true` will resolve it conclusively.

---

_Generated by `/quality:agent-review` — multi-agent review with cross-examination debate. 5 agents + dependency impact. Mode: standard, model: opus._

src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.tsx
```diff
@@ -178,19 +201,29 @@ export const SetupStep: React.FC = () => {
           </Grid>
 
           <Grid item xs={12}>
-            <AutosaveTextField
-              fieldName="salaryOrHourly"
-              schema={schema}
+            <TextField
```
wjames111, line 207: [Medium] Pay Type field departs from the `AutosaveTextField` pattern used by every other select in this form.

**The departure is justified** — `saveField({ salaryOrHourly, payRate: null })` writes two fields atomically and `AutosaveTextField` only writes its bound `fieldName`. But it's undocumented; a future contributor may "normalize" it back to `AutosaveTextField` and reintroduce the stale-`payRate` bug. Add a one-line comment:

```tsx
{/* Manual TextField (not AutosaveTextField) because changing Pay Type must
    atomically clear payRate as well; AutosaveTextField only writes the
    single bound fieldName. */}
<TextField
  fullWidth
  ...
```

**Suggestion (sev 4.5):** clearing `payRate: null` is silent — a user with a typed salary/hourly value sees it disappear without explanation. Consider visible inline feedback after the change, e.g. *"Pay rate reset — enter new value for [Hourly/Salary]"*. The two-option toggle doesn't warrant a confirm dialog, but a brief helper text would prevent surprise.

**Unresolved (sev 3.5–6.0, see review body "Unresolved Debate"):** UX flagged that the disabled-state visual treatment of this raw `TextField` may differ from surrounding `AutosaveTextField` cells when the row is locked. Architecture countered that `AutosaveTextField` has no saving adornment, so visuals match. A visual check with `isMutating=true` will resolve this conclusively.

*Architecture (6.0) + Standards (Suggestion-tier, departure is Formik-adjacent so not a standards violation) + Financial (the silent `payRate` reset).*

src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/MonthlyReimbursableSection.tsx
```diff
@@ -43,6 +60,9 @@ export const MonthlyReimbursableSection: React.FC = () => {
   return (
     <ReimbursableExpensesGrid
       title={t('Monthly Reimbursable Expenses')}
+      description={t(
+        'Ministry Cell Phone and Ministry Internet reimbursements are capped at the per-month maximums shown next to each field name. Amounts above the maximum will not be saved.',
```
wjames111: [Medium] Description states *"Amounts above the maximum will not be saved"*, but the actual UX behavior is that the over-max value is **retained** in the cell with a red border + helper text — the cell is NOT visually reverted to the previously saved value. Users may believe they entered $50 when $35 is what's persisted.

Verified in [ReimbursableExpensesGrid.tsx](src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesGrid.tsx) — `processRowUpdate` returns `newRow` even when `validateAmount` fails, so the DataGrid keeps the over-max value in its state and `renderAmountCell` renders it inside `ErrorCell`.

Resolve the description/behavior mismatch by either:
- **Clipping** the typed value to `max` automatically and saving the clipped value (description becomes accurate; user gets immediate persisted feedback), OR
- **Reverting** the cell on validation failure to the previously saved value (matches the description literally).

The red border with no inline `helperText` also nudges against the project's UX rule that errors should appear next to the field, not via color-only state.

*Co-owned Financial 5.5 (conceded from 7.0) + UX (independent flag) + Architecture supports.*

src/components/HrTools/PdsGoalCalculator/SupportItem/salaryBreakdown.tsx
```diff
@@ -39,14 +44,23 @@ export const buildSalaryBreakdownRows = (
...
+  const { salaryOrHourly, geographicLocation } = calculation;
   const payRate = calculation.payRate ?? 0;
   const hoursPerWeek = calculation.hoursWorkedPerWeek ?? 0;
   const isSalaried = salaryOrHourly === DesignationSupportSalaryType.Salaried;
 
   const { monthlyBase, grossMonthlyPay, employerFica, subtotal } =
     calculateSalaryTotals(calculation, constants);
 
+  const geographicMultiplierSuffix = geographicLocation
+    ? t('({{location}})', { location: geographicLocation })
```
wjames111: [Medium] Non-descriptive translation key `t('({{location}})')` — the key is literally two parentheses around a placeholder. Gives translators zero context, fragile to extraction collisions (any other `t('({{x}})')` collides on the same key), and the wrapper characters may not be appropriate for every locale (some use angle brackets, different spacing).

Location names like `Atlanta`, `Orlando, FL` are proper nouns that typically don't translate. Recommended — drop the `t()` wrapper:

```tsx
const geographicMultiplierSuffix = geographicLocation
  ? `(${geographicLocation})`
  : undefined;
```

If you do want localized wrapping, use a descriptive key, e.g. `t('Location suffix: ({{location}})', { location: geographicLocation })`.

Note: nested `t()` interpolation in [MonthlyReimbursableSection.tsx](src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/MonthlyReimbursableSection.tsx#L33) (`t('{{label}} (max {{max}}/mo)', { label, max })` where `label` is itself a `t()` result) works fine at runtime — i18next does final-string substitution — but smoke-test with `yarn extract` to confirm both outer and inner keys appear in `public/locales/en/*.json`.

src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesStep.tsx
```diff
@@ -1,12 +1,23 @@
...
 import { MonthlyReimbursableSection } from './MonthlyReimbursableSection';
 import { TotalReimbursableSection } from './TotalReimbursableSection';
 
 export const ReimbursableExpensesStep: React.FC = () => {
+  const { t } = useTranslation();
+
   return (
     <>
+      <Box pb={4}>
+        <Typography variant="h6">{t('Reimbursable Expenses')}</Typography>
```
wjames111: [Medium] Heading hierarchy issue (a11y). The step-level heading uses `variant="h6"`, and each subsection inside it (Monthly / Annual / Total Reimbursable, Salary / Other) also uses `variant="h6"`. Screen readers using heading navigation (NVDA/JAWS H key) cannot distinguish "new step" from "subsection of the current step" — they're announced as peers.

Set the `component` prop explicitly to decouple semantic level from visual size:

```tsx
// Step-level heading:
<Typography variant="h6" component="h2">{t('Reimbursable Expenses')}</Typography>

// Subsection headings (inside ReimbursableExpensesGrid, OtherSection, SalarySection, etc.):
<Typography variant="h6" component="h3">{title}</Typography>
```

Applies symmetrically to:
- [SupportItemStep.tsx:11](src/components/HrTools/PdsGoalCalculator/SupportItem/SupportItemStep.tsx#L11)
- [SummaryReportStep.tsx:25](src/components/HrTools/PdsGoalCalculator/SummaryReport/SummaryReportStep.tsx#L25)
- [OtherSection.tsx:45](src/components/HrTools/PdsGoalCalculator/SupportItem/OtherSection.tsx#L45)
- [SalarySection.tsx:43](src/components/HrTools/PdsGoalCalculator/SupportItem/SalarySection.tsx#L43)
- And [ReimbursableExpensesGrid.tsx:186](src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesGrid.tsx#L186), [TotalReimbursableSection.tsx:30](src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/TotalReimbursableSection.tsx#L30) (subsections)

**Bonus suggestion (sev 3.0):** the same `<Box pb={4}><Typography variant="h6">...</Typography><Typography pt={1}>...</Typography></Box>` block now appears in three step components. A small `StepHeader` shared component in `PdsGoalCalculator/Shared/` would normalize i18n and accessibility (e.g., the `component="h2"` fix above) in one place.

src/components/HrTools/PdsGoalCalculator/SummaryReport/SummaryReportStep.tsx
```diff
@@ -17,5 +20,17 @@ export const SummaryReportStep: React.FC = () => {
     return <Loading loading />;
   }
```
wjames111, line 22: [Suggestion] The new heading + description block is only rendered **after** the loading guard returns. Users see no heading or description while `calculationLoading` is true. Consider showing the heading and description always and gating only the table:

```tsx
return (
  <>
    <Box pb={4}>
      <Typography variant="h6">{t('Summary')}</Typography>
      <Typography pt={1}>
        {t('Review your complete PDS goal and current support progress. ...')}
      </Typography>
    </Box>
    {calculationLoading ? (
      <Loading loading />
    ) : (
      <PdsSummaryTable supportRaised={supportRaised} />
    )}
  </>
);
```

Low-priority — depends on intended loading experience (is the bare `<Loading />` spinner with no context preferable, or the full layout with a loading placeholder for the table?).

src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.tsx
```diff
@@ -178,19 +201,32 @@ export const SetupStep: React.FC = () => {
...
+                Type must atomically clear payRate as well; AutosaveTextField
+                only writes the single bound fieldName. */}
+            <TextField
+              fullWidth
+              size="small"
+              variant="outlined"
               select
               label={t('Pay Type')}
+              value={calculation?.salaryOrHourly ?? ''}
+              disabled={!calculation || isMutating}
```
wjames111, line 215: [Important] **Pay Type now bypasses `AutosaveTextField`; `isMutating`-disabled state and no-op guard are untested.**

This PR replaces `AutosaveTextField` for Pay Type with a manual `<TextField>` so it can atomically clear `payRate` (line 219). That's a reasonable trade-off, but it makes `disabled={!calculation || isMutating}` the **only** guard left against concurrent mutations on this field — and there is no test exercising:

1. `isMutating === true` disabling the select (so users can't fire a concurrent save during pay-type switching, racing the `payRate: null` clear with another pending mutation)
2. The no-op guard at line 218 (`if (newValue !== calculation?.salaryOrHourly)`) — re-selecting the same option should NOT send `payRate: null` and wipe a valid `payRate`
3. `disabled` state when `calculation` is null

The two new tests at `SetupStep.test.tsx:479-526` cover only the happy-path `payRate` clearing.

**Suggested test (covers both gaps at once):**

```tsx
it('disables Pay Type while a mutation is in flight', async () => {
  const { findByRole } = renderSetup({
    calculationMock: fullTimeSalariedMock,
    mocks: { UpdatePdsGoalCalculation: () => new Promise(() => {}) as never },
    onCall: mutationSpy,
  });
  const payType = await findByRole('combobox', { name: /Pay Type/ });
  userEvent.click(payType);
  userEvent.click(await findByRole('option', { name: 'Hourly' }));
  await waitFor(() => expect(payType).toHaveAttribute('aria-disabled', 'true'));
});

it('does not save when Pay Type is set to its current value', async () => {
  const { findByRole, getByRole } = renderSetup({
    calculationMock: fullTimeHourlyMock,
    onCall: mutationSpy,
  });
  const payType = await findByRole('combobox', { name: /Pay Type/ });
  await waitFor(() => expect(payType).toHaveTextContent('Hourly'));
  userEvent.click(payType);
  userEvent.click(getByRole('option', { name: 'Hourly' }));
  await new Promise((r) => setTimeout(r, 0));
  expect(mutationSpy).not.toHaveGraphqlOperation('UpdatePdsGoalCalculation', {
    attributes: { payRate: null },
  });
});
```

Note: this severity is **above the 7.0 dismiss threshold** — please address in this PR or open a follow-up ticket.

— *Testing (7.5, ↑ from 6.5 after debate)*

src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.tsx
```diff
@@ -178,19 +201,32 @@ export const SetupStep: React.FC = () => {
...
+              onChange={(event) => {
+                const newValue = event.target
+                  .value as DesignationSupportSalaryType;
+                if (newValue !== calculation?.salaryOrHourly) {
+                  saveField({ salaryOrHourly: newValue, payRate: null });
```
wjames111, line 220: [Medium] **Pay Type switch silently clears Pay Rate.**

Selecting Hourly/Salaried atomically nulls `payRate`. If the user already entered a value (e.g., `50000` salary), switching to Hourly silently wipes it — no confirmation, no toast, and no helper text near Pay Type warns that switching clears Pay Rate. The schema then marks Pay Rate as a required error, which the user has to rediscover.

The new test even acknowledges the behavior: the test name says *"the user must re-enter it before Continue becomes enabled"*.

**Mitigation options (increasing strength):**

1. Helper text under Pay Type explaining the clear:
   ```tsx
   <TextField
     ...
     label={t('Pay Type')}
     helperText={t('Changing this clears Pay Rate.')}
   />
   ```
2. Confirmation dialog when switching with an existing `payRate`.
3. Don't clear — let the user re-enter when they notice the helper text changed (`Enter yearly salary` vs `Enter hourly rate`).

— *UX (6.0), endorsed by Architecture and Financial*

src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/MonthlyReimbursableSection.tsx
```diff
@@ -20,16 +23,30 @@ export const MonthlyReimbursableSection: React.FC = () => {
     ? calculateReimbursableTotals(calculation).monthlySubtotal
     : 0;
 
+  const buildCappedLabel = (label: string, max: number | undefined) =>
+    max !== undefined
+      ? t('{{label}} (max {{max}}/mo)', {
```
wjames111: [Medium] **Nested translation: `buildCappedLabel` composes pre-translated strings inside another `t()` call.**

```ts
t('{{label}} (max {{max}}/mo)', { label: t('Ministry Cell Phone'), max: ... })
```

This isn't a hard `.CLAUDE/rules/code-review.md` violation (the key is static and uses `{{}}` placeholders, not template-literal interpolation), but it damages translator UX:

- The outer key contains *English word order* — translators of `'{{label}} (max {{max}}/mo)'` can't reorder the noun, can't inflect for gender, can't pluralize.
- For locales that need different parenthetical conventions (e.g., Japanese full-width `（最大$35/月）`, or German compound-noun ordering), the structure is unfixable from translations alone.
- Crowdin extraction may pick up the inner labels independently, so translators see them without context.

**Suggested fix — write each field's capped form as its own translatable string:**

```tsx
const formatMaxPerMonth = (max: number) =>
  currencyFormat(max, 'USD', locale);

{
  fieldName: 'ministryCellPhone',
  label: phoneMax !== undefined
    ? t('Ministry Cell Phone (max {{max}}/mo)', { max: formatMaxPerMonth(phoneMax) })
    : t('Ministry Cell Phone'),
  ...
},
{
  fieldName: 'ministryInternet',
  label: internetMax !== undefined
    ? t('Ministry Internet (max {{max}}/mo)', { max: formatMaxPerMonth(internetMax) })
    : t('Ministry Internet'),
  ...
},
```

— *UX (6.5), Architecture (6.0 post-debate), Standards (3.5 post-debate). Consensus: 6.0.*

src/lib/intlFormat.ts
```diff
@@ -1,8 +1,19 @@
...
+
+export const numberFormat = (
+  value: number,
+  locale: string,
+  options?: NumberFormatOptions,
+): string =>
   new Intl.NumberFormat(locale, {
     style: 'decimal',
+    minimumFractionDigits: options?.minimumFractionDigits,
+    maximumFractionDigits: options?.maximumFractionDigits,
```
wjames111, line 16: [Medium] **New `numberFormat` options bag is not directly unit-tested.**

`numberFormat` is a shared lib with ~144 importers app-wide; this PR extends its signature with optional `minimumFractionDigits` / `maximumFractionDigits`. The new options are exercised end-to-end in `HoursPerWeekGrid.test.tsx`, but `src/lib/intlFormat.test.ts > describe('numberFormat')` has no test for the new options — including the NaN-fallback path (`Number.isFinite(value) ? value : 0`) combined with `minimumFractionDigits`.

**Suggested unit tests:**

```ts
describe('numberFormat with options', () => {
  it('respects minimumFractionDigits', () => {
    expect(numberFormat(40, 'en-US', { minimumFractionDigits: 2 })).toEqual('40.00');
  });
  it('respects maximumFractionDigits', () => {
    expect(numberFormat(0.1234, 'en-US', { maximumFractionDigits: 2 })).toEqual('0.12');
  });
  it('applies fraction-digit options to the NaN fallback', () => {
    expect(
      numberFormat(NaN, 'en-US', { minimumFractionDigits: 2, maximumFractionDigits: 2 }),
    ).toEqual('0.00');
  });
  it('formats per locale', () => {
    expect(numberFormat(0.18, 'de-DE', { minimumFractionDigits: 2 })).toEqual('0,18');
  });
});
```

— *Testing (6.0)*

src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesGrid.tsx
```diff
@@ -101,25 +92,48 @@ export const ReimbursableExpensesGrid: React.FC<
...
+
+    if (newRow.amount < 0) {
+      setCellErrors((prev) => ({
+        ...prev,
+        [cellKey]: t('Amount must be positive'),
+      }));
+      return newRow;
+    }
+
+    const amount =
```
wjames111: [Medium] **Silent over-max clipping removes the only signal that a typed amount was rewritten.**

The previous `validateAmount` set a per-cell error ("cannot exceed {{max}}") and skipped the save. This PR replaces that with implicit clipping: if `newRow.amount > field.max`, the code silently substitutes `field.max` and saves. A user types `999` in a $35-max cell, sees their entry rewritten to `35` with zero acknowledgement.

The `MonthlyReimbursableSection` description (*"Amounts entered above the maximum will be saved as the maximum"*) explains the policy at the section header level, but visually distant from the cell once the user is editing — and screen-reader users editing the cell won't re-encounter the description without re-navigating.

**Disagreement on severity:** Financial agent rates this 3.0 (description text covers it, no financial bug). UX rates 6.0 (WCAG 3.3.1 Error Identification — value transformations diverging from user input should be communicated). Architecture revised down to 4.0 after debate (documented contract via description + per-row tooltip + new test). **Consensus: 5.0 — UX clarity concern, not a financial-correctness bug.**

**Suggested mitigation — non-blocking inline feedback or aria-live announcement:**

```tsx
// inside processRowUpdate, when clipping occurs
if (field.max !== undefined && newRow.amount > field.max) {
  enqueueSnackbar(
    t('{{label}} reduced to its maximum of {{max}}.', {
      label: field.label,
      max: currencyFormat(field.max, 'USD', locale),
    }),
    { variant: 'info' },
  );
}
```

Alternative: surface the clip via a brief `FormHelperText` under the cell with `role="status"` (keeps feedback proximate to the field, works for keyboard/AT users).

— *UX (6.0), Architecture (4.0), Financial (3.0). Consensus: 5.0.*

src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesGrid.tsx
```diff
@@ -101,25 +92,48 @@ export const ReimbursableExpensesGrid: React.FC<
...
+  const renderLabelCell = (params: GridRenderCellParams<ReimbursableRow>) => {
+    const { row } = params;
+    if (!row.tooltip) {
+      return row.label;
+    }
+    return (
+      <Stack direction="row" alignItems="center" gap={0.5}>
+        <span>{row.label}</span>
+        <Tooltip title={row.tooltip}>
+          <InfoIcon color="action" fontSize="small" aria-label={row.tooltip} />
```
wjames111: [Suggestion] **Info icon inside `Tooltip` may not be keyboard-reachable for sighted users.**

`<Tooltip><InfoIcon ... aria-label={row.tooltip} /></Tooltip>` — `InfoIcon` is an `<svg>` and is not focusable by default. MUI's `Tooltip` only opens on hover/focus, so keyboard-only sighted users cannot trigger the visual tooltip.

Disagreement note: the `aria-label` on the icon DOES expose the text to screen readers (so AT users are covered, satisfying WCAG 2.1.1 Keyboard), but it doesn't help sighted keyboard users (WCAG 1.4.13 Content on Hover or Focus).

**Suggested fix — wrap in `IconButton` to make the trigger focusable:**

```tsx
<Stack direction="row" alignItems="center" gap={0.5}>
  <span>{row.label}</span>
  <Tooltip title={row.tooltip}>
    <IconButton size="small" aria-label={row.tooltip} disableRipple sx={{ p: 0 }}>
      <InfoIcon color="action" fontSize="small" />
    </IconButton>
  </Tooltip>
</Stack>
```

— *UX (6.0), Architecture (3.5 — argues `aria-label` covers AT users). Consensus: 4.5.*

src/components/HrTools/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
```diff
@@ -115,6 +118,9 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
     [totalHours, totalWeeks],
   );
 
+  const roundedAverageHoursPerWeek =
+    Math.round(averageHoursPerWeek * 100) / 100;
```
wjames111: [Suggestion] **`roundedAverageHoursPerWeek` derivation is not memoized; autosave path at lines 287/340 still uses the un-rounded `newAverage`.**

Two small follow-ups for the new rounding logic introduced here:

**1. Memoization consistency.** `averageHoursPerWeek` on line 116-119 is `useMemo`'d, but `roundedAverageHoursPerWeek` is computed inline on every render. Style nit only — collapse the rounding into the same memo so consumers can't accidentally reach for the un-rounded version:

```tsx
const averageHoursPerWeek = useMemo(() => {
  if (totalWeeks <= 0) return 0;
  return Math.round((totalHours / totalWeeks) * 100) / 100;
}, [totalHours, totalWeeks]);
// Then drop `roundedAverageHoursPerWeek` entirely.
```

This also eliminates the new WYSIWYS test as a tautology — display, Apply, and any future consumer all read the same value.

**2. Autosave-divergence at lines 287 and 340.** The autosave calls still pass un-rounded `newAverage`:

```tsx
saveField({ averageHoursPerWeek: newAverage });  // line 287, 340
```

*This was originally flagged severity 6.5 but downgraded after Financial Agent verified that `averageHoursPerWeek` is a scratch column never consumed by salary math (downstream pay reads `hoursWorkedPerWeek`, which gets the rounded value via Apply).* No user-visible bug today. But: a one-line fix is essentially free, defends against any future server-side consumer (PDFs, exports, Salary Calculator handoff), and keeps the WYSIWYS contract truthful end-to-end.

— *Architecture (3.5 — memoization), Financial (3.0 post-debate — autosave divergence). Consensus: 3.5.*

src/components/HrTools/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
```diff
@@ -509,7 +518,7 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
           <Button
             variant="contained"
             disabled={weeksRemaining > 0}
-            onClick={() => onApply(averageHoursPerWeek)}
+            onClick={() => onApply(roundedAverageHoursPerWeek)}
```
wjames111: [Suggestion] **Defensive comment: this is the WYSIWYS boundary.**

`Math.round((x) * 100) / 100` uses half-away-from-zero rounding; `Intl.NumberFormat` (used for display at line 499) defaults to half-to-even. They agree here only because the display formats the *pre-rounded* `roundedAverageHoursPerWeek` (line 122). If a future developer changes the display layer to format the un-rounded `averageHoursPerWeek` (e.g., to add more precision), the WYSIWYS invariant silently breaks.

A short code comment near line 122 documenting *why* the format input must remain `roundedAverageHoursPerWeek` would prevent that drift.

— *Financial (3.5)*

wjames111 (commented):
# Multi-Agent Code Review — PR #1780

**Mode:** Standard · **Agents:** Architecture, Testing, Standards, UX, Financial Reporting (+ Coverage-Gap Review) · **Risk:** 10/10 (CRITICAL, breadth-driven across 13 non-test files in a single feature)

## Verdict: APPROVED_WITH_SUGGESTIONS

No CRITICAL (≥9.0) or HIGH-priority (≥8.0) blockers were found. The PR is well-scoped, the WYSIWYS rounding in `HoursPerWeekGrid` is correctly implemented (display = applied = persisted to `hoursWorkedPerWeek`), and i18n discipline is generally strong (static keys, no template-literal interpolation inside `t()`).

**Findings summary (post-debate consensus):**

| Tier | Count | Notes |
|---|---|---|
| Critical (≥9.0) | 0 | — |
| High (8.0–8.9) | 0 | — |
| Important (7.0–7.9) | 1 | Cannot be `/dismiss`'d — strongly recommended |
| Medium (5.0–6.9) | 5 | Block CLEAN verdict but not merge |
| Suggestions (<5.0) | ~20 | Informational only |

> ⚠️ **1 Important finding (severity 7.0–7.9)** — strongly recommended but doesn't block merge and can't be `/dismiss`'d. Consider addressing in this PR or creating a follow-up ticket.

## Risk Assessment

Risk score is capped at 10/10 but the *nature* of the changes is low-blast-radius: 12 medium-risk component files in a single feature (PdsGoalCalculator) plus 1 shared lib (`intlFormat.ts`) with a backward-compatible signature extension. No auth, no API, no schema, no migrations.

## Dependency Impact

- **`src/lib/intlFormat.ts`** — 144 importers across the app. `numberFormat` signature is extended with an optional `options` param (backward compatible). 9 files import `numberFormat` directly; signature change is non-breaking.
- All other changes are leaf or internally-encapsulated within `src/components/HrTools/PdsGoalCalculator/**`.

## Cross-Examination Highlights

The debate round produced significant re-calibration:

- **Silent over-max clipping** (`ReimbursableExpensesGrid.tsx`): Financial (3.0) vs UX (6.0) vs Architecture (4.0 post-debate). UX argues description text ≠ in-context feedback per WCAG 3.3.1; Architecture/Financial argue the documented contract (section description + per-row tooltip + new test) is adequate. **Consensus: medium-priority UX clarity concern, not a financial bug.**
- **Autosave divergence** (`HoursPerWeekGrid.tsx:287, 340`): Financial originally 6.5; revised DOWN to 3.0 after verifying `averageHoursPerWeek` is a scratch column never read by salary math (downstream pay uses `hoursWorkedPerWeek` which receives the rounded value via Apply).
- **Pay Type bypass of AutosaveTextField** (`SetupStep.tsx:214`): Architecture's 6.5 revised DOWN to 4.5 (Yup rule is only `.required()` and `MenuItem` prevents empty value, so the validity gate is functionally inert here). BUT Testing's `isMutating` test gap went UP to 7.5 because it's now the only validity gate left.
- **Translation nesting** (`MonthlyReimbursableSection.tsx:26-32`): Standards revised DOWN to 3.5 (not a literal Checklist violation — uses `{{}}` placeholders, not template-literal interpolation). UX held at 6.5 (translator-grammar concern). **Consensus: 6.0 (translator-context smell, not a hard violation).**

## Important Issues (Severity 7.0–7.9)

See inline comment on `SetupStep.tsx:214`.

## Medium-Priority Issues (Severity 5.0–6.9)

See inline comments on:
- `SetupStep.tsx:219` — Pay Type clears Pay Rate without warning
- `MonthlyReimbursableSection.tsx:28` — `buildCappedLabel` translation nesting
- `intlFormat.ts:16` — `numberFormat` options not unit-tested
- `ReimbursableExpensesGrid.tsx:105` — Silent over-max clipping

## Suggestions (Severity <5.0)

Lower-priority findings inline on `ReimbursableExpensesGrid.tsx:133`, `HoursPerWeekGrid.tsx:122`, `HoursPerWeekGrid.tsx:521`.

### Other suggestions (not inlined — informational only)

- **Step-intro duplication** across `SupportItemStep`, `SummaryReportStep`, and `ReimbursableExpensesStep` — same `<Box pb={4}><Typography variant="h6" component="h2">…</Typography><Typography pt={1}>…</Typography></Box>` pattern in 3 files. Extract `<StepIntro title description />` to `Shared/`. (3.5/10)
- **`OtherSection.tsx` missing description test** — sibling sections (Annual/Total/Monthly Reimbursable) added description-render assertions in this PR; `OtherSection` did not. (2.5/10)
- **Heading hierarchy nit** — `ReimbursableExpensesGrid.tsx:181` and `TotalReimbursableSection.tsx` render `variant="h6"` without `component="h3"`, producing a muddled outline under the step's `h2`. (4.0/10)
- **AmountSuffix screen reader gap** — `salaryBreakdown.tsx:160-164` renders `0%(Atlanta)` without whitespace separation in screen-reader output. (4.0/10)
- **`getLocationLabel` template literal** — `SetupStep.tsx:87-93` builds Autocomplete label with `` `${location} (${percent})` ``; not a violation (wraps data only, no English copy) but inconsistent with neighboring `t()` usage. (2.5/10)
- **`geographicMultiplierSuffix` template literal** — `salaryBreakdown.tsx:1194-1196` same pattern. (2.5/10)
- **Description tests** — `SupportItemStep`, `SalarySection`, `SummaryReportStep`, `ReimbursableExpensesStep` add new description copy without colocated assertions. Static i18n strings without branching are low-value to test, but `SalarySection` has no test file at all. (2.5/10)
- **HoursPerWeekGrid locale-formatted average not tested** — new `numberFormat` call with `minimumFractionDigits: 2` lacks non-US-locale assertion. (3.5/10)

### Pre-debate findings that were withdrawn

- ~~HoursPerWeekGrid autosave divergence at lines 287/340 (severity 6.5)~~ — Financial Agent verified `averageHoursPerWeek` is never consumed by salary math; downgraded to 3.0. A one-line fix would still tighten data hygiene for any future server-side consumer (PDFs, exports).

## Standards Checklist

All mandatory items in `.CLAUDE/rules/code-review.md` pass:
- Named exports only · PascalCase file naming · Colocated tests ✓
- i18n via `useTranslation`/`t()`, static keys, `{{}}` interpolation ✓
- No `any`, no `@ts-ignore`, no `new Date()`, no debug output, no empty `catch` ✓
- Apollo `optimisticResponse` includes `__typename` and `id` (pre-existing pattern) ✓
- No `.graphql`/migration changes (N/A items) ✓

## How to respond

- For **severity <7 findings** you disagree with, reply `/dismiss: <reason>` on the comment.
- For the **Important finding (severity 7.5)**, address in this PR or open a follow-up ticket — it can't be `/dismiss`'d.

PR #1791, opened 2026-05-20
MPDX-PDS-cleanup: error handling and rounding fixes in PDS calculator

wjames111:
## Description

Cleanup pass on the PDS Goal Calculator to tighten error handling, surface failures to users, and fix display rounding so the summary table line items sum exactly to the submitted monthly goal.

- Surface mutation failures via snackbar / Alert instead of swallowing them silently (`PdsGoalCalculator`, `SummaryReportStep`)
- Re-throw errors from `useSaveField` and revert the local input back to the cached value in `useAutoSave` when a save fails, so the UI no longer shows what the user typed after Apollo rolls back the optimistic response
- Refetch `PdsGoalCalculations` after each autosave to keep dependent queries in sync
- Round the four gross-up components (`subtotal`, `attrition`, `creditCardFees`, `assessment`) at the display boundary in `usePdsSummaryData` so the table line items sum exactly to `overallTotal` and the submitted monthly goal — and drop the now-redundant `Math.round` at submit time
- Replace the deprecated `Loading` component with a `CircularProgress` in `SummaryReportStep` and add an `<Alert>` error state for the support-raised query
- Replace the icon-only title-attribute error indicator in `ReimbursableExpensesGrid` with a visible `ErrorOutlineIcon` for better a11y
- Extract a `roundHundredths` helper in `HoursPerWeekGrid` and add explicit `.catch(() => {})` on fire-and-forget autosave calls so unhandled promise rejections don't surface
- Trim stale `setRightPanelContent` / `setIsDrawerOpen` from the `PdsGoalCalculatorContext` memo dependency array

No Jira ticket — pure cleanup pass.

## Testing

- Open the PDS Goal Calculator on an account list
- In the Setup step, change Hours Per Week, Salary/Hourly, Pay Rate, and Geographic Location — confirm each change saves and the right-hand panel reflects the new value
- Open Chrome DevTools → Network → Offline, then edit any Setup field — confirm the input reverts to the previous value and an error snackbar appears
- In the Reimbursable Expenses step, enter a negative amount in a row — confirm the red error icon + value render in the cell and no mutation fires
- Navigate to the Summary Report step — confirm the line items in the summary table sum exactly to the displayed "Overall Total"
- Click "Save monthly goal" on the summary — confirm a success snackbar; with the network offline, confirm an error snackbar with the failure message
- Force the `AccountListSupportRaised` query to error (e.g. via DevTools network throttling/block) — confirm the Alert error state renders in place of the summary table

## Checklist:

### DIFF
--- a/src/components/HrTools/PdsGoalCalculator/PdsGoalCalculator.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/PdsGoalCalculator.tsx
@@ -1,4 +1,5 @@
 import React, { useState } from 'react';
+import { ApolloError } from '@apollo/client';
 import { useSnackbar } from 'notistack';
 import { useTranslation } from 'react-i18next';
 import { DirectionButtons } from 'src/components/HrTools/Shared/CalculationReports/DirectionButtons/DirectionButtons';
@@ -77,7 +78,7 @@ const MainContent: React.FC = () => {
     if (!accountListId || !summaryData?.overallTotal) {
       return;
     }
-    const monthlyGoal = Math.round(summaryData.overallTotal);
+    const monthlyGoal = summaryData.overallTotal;
     await updateAccountPreferences({
       variables: {
         input: {
@@ -96,6 +97,14 @@ const MainContent: React.FC = () => {
             formattedTotal: currencyFormat(monthlyGoal, 'USD', locale),
           }),
           { variant: 'success' },
+        );
+      },
+      onError: (err: ApolloError) => {
+        enqueueSnackbar(
+          t('Error while updating your monthly goal - {{error}}', {
+            error: err.message,
+          }),
+          { variant: 'error' },
         );
       },
     });
--- a/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesGrid.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesGrid.tsx
@@ -1,4 +1,5 @@
 import React, { useState } from 'react';
+import ErrorOutlineIcon from '@mui/icons-material/ErrorOutline';
 import InfoIcon from '@mui/icons-material/Info';
 import {
   Box,
@@ -56,6 +57,8 @@ const ErrorCell = styled(Box)(({ theme }) => ({
   display: 'flex',
   alignItems: 'center',
   justifyContent: 'center',
+  gap: theme.spacing(0.5),
+  color: theme.palette.error.main,
 }));
 
 export const ReimbursableExpensesGrid: React.FC<
@@ -159,7 +162,12 @@ export const ReimbursableExpensesGrid: React.FC<
     const formatted = currencyFormat(params.value ?? 0, 'USD', locale);
 
     if (error) {
-      return <ErrorCell title={error}>{formatted}</ErrorCell>;
+      return (
+        <ErrorCell>
+          <ErrorOutlineIcon fontSize="small" aria-hidden="true" />
+          <span>{formatted}</span>
+        </ErrorCell>
+      );
     }
 
     if (params.id === 'total' && subtotalTestId) {
--- a/src/components/HrTools/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
@@ -62,6 +62,8 @@ interface HoursPerWeekGridProps {
 
 const MAX_TOTAL_WEEKS = 52;
 
+const roundHundredths = (n: number) => Math.round(n * 100) / 100;
+
 export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
   onApply,
 }) => {
@@ -121,7 +123,7 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
     // matches the value submitted via saveField/onApply. Math.round uses
     // half-away-from-zero; Intl.NumberFormat uses half-to-even — they only
     // agree because the formatter receives this already-rounded value.
-    return Math.round((totalHours / totalWeeks) * 100) / 100;
+    return roundHundredths(totalHours / totalWeeks);
   }, [totalHours, totalWeeks]);
 
   const weeksRemaining = MAX_TOTAL_WEEKS - totalWeeks;
@@ -276,9 +278,7 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
         0,
       );
       const newAverage =
-        newTotalWeeks > 0
-          ? Math.round((newTotalHours / newTotalWeeks) * 100) / 100
-          : 0;
+        newTotalWeeks > 0 ? roundHundredths(newTotalHours / newTotalWeeks) : 0;
 
       try {
         if (!entryId.startsWith('temp-') && !entryId.startsWith('default-')) {
@@ -291,7 +291,7 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
         }
         saveField({
           averageHoursPerWeek: newAverage,
-        });
+        }).catch(() => {});
       } catch {
         setEntries(previousEntries);
         enqueueSnackbar(t('Failed to delete entry. Please try again.'), {
@@ -342,12 +342,10 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
         0,
       );
       const newAverage =
-        newTotalWeeks > 0
-          ? Math.round((newTotalHours / newTotalWeeks) * 100) / 100
-          : 0;
+        newTotalWeeks > 0 ? roundHundredths(newTotalHours / newTotalWeeks) : 0;
       saveField({
         averageHoursPerWeek: newAverage,
-      });
+      }).catch(() => {});
 
       return { ...newRow, weeks: Math.max(0, clampedWeeks) };
     },
--- a/src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.tsx
@@ -114,7 +114,7 @@ export const SetupStep: React.FC = () => {
     setRightPanelContent(
       <HoursPerWeekGrid
         onApply={(average) => {
-          saveField({ hoursWorkedPerWeek: average });
+          saveField({ hoursWorkedPerWeek: average }).catch(() => {});
         }}
       />,
     );
@@ -219,7 +219,10 @@ export const SetupStep: React.FC = () => {
                 const newValue = event.target
                   .value as DesignationSupportSalaryType;
                 if (newValue !== calculation?.salaryOrHourly) {
-                  saveField({ salaryOrHourly: newValue, payRate: null });
+                  saveField({
+                    salaryOrHourly: newValue,
+                    payRate: null,
+                  }).catch(() => {});
                 }
               }}
             >
@@ -317,9 +320,9 @@ export const SetupStep: React.FC = () => {
               options={locations}
               getOptionLabel={getLocationLabel}
               value={calculation?.geographicLocation ?? 'None'}
-              onChange={(_, newValue: string | null) =>
-                saveField({ geographicLocation: newValue })
-              }
+              onChange={(_, newValue: string | null) => {
+                saveField({ geographicLocation: newValue }).catch(() => {});
+              }}
               disabled={!calculation}
               size="small"
               renderInput={(params: AutocompleteRenderInputParams) => (
--- a/src/components/HrTools/PdsGoalCalculator/Shared/Autosave/useSaveField.test.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/Shared/Autosave/useSaveField.test.tsx
@@ -1,7 +1,16 @@
 import React from 'react';
-import { renderHook, waitFor } from '@testing-library/react';
+import { act, renderHook, waitFor } from '@testing-library/react';
 import { PdsGoalCalculatorTestWrapper } from '../../PdsGoalCalculatorTestWrapper';
+import { usePdsGoalCalculator } from '../PdsGoalCalculatorContext';
 import { useSaveField } from './useSaveField';
+
+const mockEnqueue = jest.fn();
+jest.mock('notistack', () => ({
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  ...jest.requireActual('notistack'),
+  useSnackbar: () => ({ enqueueSnackbar: mockEnqueue }),
+}));
 
 const mutationSpy = jest.fn();
 
@@ -18,7 +27,28 @@ const Wrapper: React.FC<{ children: React.ReactNode }> = ({ children }) => (
   </PdsGoalCalculatorTestWrapper>
 );
 
+const ErrorWrapper: React.FC<{ children: React.ReactNode }> = ({
+  children,
+}) => (
+  <PdsGoalCalculatorTestWrapper
+    calculationMock={{
+      id: 'goal-1',
+      name: 'Test Goal',
+      payRate: 50000,
+    }}
+    onCall={mutationSpy}
+    updatePdsGoalCalculationError
+  >
+    {children}
+  </PdsGoalCalculatorTestWrapper>
+);
+
 describe('useSaveField', () => {
+  beforeEach(() => {
+    mockEnqueue.mockClear();
+    mutationSpy.mockClear();
+  });
+
   it('should update the calculation when a value changes', async () => {
     const { result } = renderHook(useSaveField, { wrapper: Wrapper });
 
@@ -35,6 +65,46 @@ describe('useSaveField', () => {
           name: 'New Name',
         },
       }),
+    );
+  });
+
+  it('shows an error snackbar and rolls back the optimistic update when the mutation fails', async () => {
+    const { result } = renderHook(
+      () => ({
+        saveField: useSaveField(),
+        calculation: usePdsGoalCalculator().calculation,
+      }),
+      { wrapper: ErrorWrapper },
+    );
+
+    await waitFor(() =>
+      expect(result.current.calculation?.name).toBe('Test Goal'),
+    );
+
+    let savePromise!: Promise<unknown>;
+    act(() => {
+      savePromise = result.current.saveField({ name: 'New Name' });
+    });
+
+    // The optimistic response is written to the cache immediately, so the
+    // calculation in context briefly reflects the in-flight value.
+    await waitFor(() =>
+      expect(result.current.calculation?.name).toBe('New Name'),
+    );
+
+    await act(async () => {
+      await expect(savePromise).rejects.toThrow();
+    });
+
+    // After the mutation fails, Apollo rolls the optimistic update back and
+    // the cache returns to the original value.
+    await waitFor(() =>
+      expect(result.current.calculation?.name).toBe('Test Goal'),
+    );
+
+    expect(mockEnqueue).toHaveBeenCalledWith(
+      'Failed to save changes. Please try again.',
+      { variant: 'error' },
     );
   });
 });
--- a/src/components/HrTools/PdsGoalCalculator/Shared/Autosave/useSaveField.ts
+++ b/src/components/HrTools/PdsGoalCalculator/Shared/Autosave/useSaveField.ts
@@ -44,12 +44,18 @@ export const useSaveField = () => {
                 },
               },
             },
+            refetchQueries: ['PdsGoalCalculations'],
+            awaitRefetchQueries: false,
           }),
         );
-      } catch {
+      } catch (error) {
         enqueueSnackbar(t('Failed to save changes. Please try again.'), {
           variant: 'error',
         });
+        // Re-throw so callers (e.g. useAutoSave) can revert local field
+        // state back to the cache value after Apollo rolls back the
+        // optimistic update.
+        throw error;
       }
     },
     [calculation, trackMutation, updatePdsGoalCalculation, enqueueSnackbar, t],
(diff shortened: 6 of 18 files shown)


src/components/HrTools/PdsGoalCalculator/PdsGoalCalculator.tsx
```diff
@@ -98,6 +99,14 @@ const MainContent: React.FC = () => {
           { variant: 'success' },
         );
       },
+      onError: (err: ApolloError) => {
+        enqueueSnackbar(
+          t('Error while updating your monthly goal - {{error}}', {
```
wjames111, line 104: [Medium] **Raw `err.message` exposed in user-facing snackbar.** (UX + Data Integrity agree.)

`err.message` from Apollo can include backend stack fragments or internal validation phrasing not designed for end users. For non-English locales, the result is half-translated: the `t(...)` wrapper is localized, but `err.message` is always English from the backend.

This matches an existing precedent (`MinisterHousingAllowance.tsx:89`), so it's a codebase convention rather than a novel anti-pattern — flagged as Medium, not blocker.

**Suggested fix:**
```tsx
onError: (err: ApolloError) => {
  enqueueSnackbar(
    t('Error while updating your monthly goal. Please try again.'),
    { variant: 'error' },
  );
  // Log err for diagnostics, but don't surface raw backend text to the user.
},
```

src/components/HrTools/PdsGoalCalculator/PdsGoalCalculator.tsx
```diff
@@ -77,7 +78,7 @@ const MainContent: React.FC = () => {
     if (!accountListId || !summaryData?.overallTotal) {
       return;
     }
-    const monthlyGoal = Math.round(summaryData.overallTotal);
+    const monthlyGoal = summaryData.overallTotal;
```
wjames111, line 81: [Suggestion] **Defense-in-depth: keep `Math.round` as a tail guard.**

The removal of `Math.round` here is safe today because `usePdsSummaryData` now rounds the four gross-up components, making `overallTotal` invariant-integer. But this couples this file to an implementation detail of the hook. If a future refactor introduces an unrounded `overallTotal` (e.g., a split into `rawOverallTotal` + `displayOverallTotal`), this mutation variable will silently start carrying sub-cent precision into `UpdateAccountPreferences`.

**Suggested fix:**
```tsx
// Defense in depth: the goal sent to the server must be a whole-dollar value
// regardless of how usePdsSummaryData computes overallTotal.
const monthlyGoal = Math.round(summaryData.overallTotal);
```

src/components/HrTools/PdsGoalCalculator/Shared/Autosave/useSaveField.ts
```diff
@@ -44,12 +44,18 @@ export const useSaveField = () => {
                 },
               },
             },
+            refetchQueries: ['PdsGoalCalculations'],
```
wjames111, line 47: [Medium] **`refetchQueries: ['PdsGoalCalculations']` is redundant and adds latency cost per keystroke.** (Architecture + Data Integrity agree.)

The `UpdatePdsGoalCalculation` mutation already returns the full `PdsGoalCalculationFields` fragment including `id`, so Apollo cache normalization updates the cached `DesignationSupportCalculation:<id>` entity automatically. The list query reads the same normalized entity — no refetch needed.

Worse, `PdsGoalCalculations` is driven by `useFetchAllPages`. A `refetchQueries` call only refetches the **first page** (with `after: undefined`), discarding the accumulated pagination state for that page until the `useEffect` re-paginates. So the refetch can't actually refresh later pages, but it does cost one network round-trip per saved field.

**Suggested fix:** Drop both lines.
```ts
}),
// Apollo cache normalization on the returned `id` already keeps both
// the detail view and the list query in sync.
```

If a list-level field (e.g. `updatedAt`) appears stale, the right fix is to include it in the mutation's selection set, not to refetch.

src/components/HrTools/PdsGoalCalculator/Shared/Autosave/useSaveField.test.tsx
```diff
@@ -37,4 +67,44 @@ describe('useSaveField', () => {
...
+      await expect(savePromise).rejects.toThrow();
+    });
+
+    // After the mutation fails, Apollo rolls the optimistic update back and
+    // the cache returns to the original value.
+    await waitFor(() =>
+      expect(result.current.calculation?.name).toBe('Test Goal'),
+    );
+
+    expect(mockEnqueue).toHaveBeenCalledWith(
```
wjames111, line 105: [Important] **Race-prone snackbar assertion — missing `waitFor`.**

`expect(mockEnqueue).toHaveBeenCalledWith(...)` is reached after `await act(async () => { await expect(savePromise).rejects.toThrow(); })`. The `enqueueSnackbar` call happens **inside the `catch` block** of `saveField` after the awaited `trackMutation` re-throws. The current ordering works today, but it depends on micro-task scheduling — any future refactor that yields between the rejection and the catch could fail this assertion intermittently.

Also: this test never verifies the mutation was actually invoked. If the optimistic-response → rollback path short-circuited (e.g. a future change that fails the variable check before sending), this test would still pass.

**Suggested fix:**
```tsx
await waitFor(() =>
  expect(mockEnqueue).toHaveBeenCalledWith(
    'Failed to save changes. Please try again.',
    { variant: 'error' },
  ),
);
expect(mutationSpy).toHaveGraphqlOperation('UpdatePdsGoalCalculation', {
  attributes: { id: 'goal-1', name: 'New Name' },
});
```

wjames111 (commented):
## Multi-Agent Code Review — Standard Mode

6 specialized review agents (Architecture, Data Integrity, Financial Reporting, Testing, UX, Standards) + dependency-impact + coverage-gap analysis.

### Verdict: APPROVED_WITH_SUGGESTIONS

No HIGH or CRITICAL blockers (severity ≥ 8) in modified files. **2 Important findings (severity 7.0–7.9)** should be addressed before merge — these can't be dismissed via `/dismiss`. Several MEDIUM findings posted as line comments.

The core financial change (rounding refactor in `usePdsSummaryData`) is mathematically sound — Financial Reporting agent confirmed no precision issues. `overallTotal` is now invariant-integer (sum of four `Math.round` values), making the mutation-site `Math.round` removal safe.

### Risk Assessment

- **Risk Score:** 10/10 (capped) — **CRITICAL**
- **Drivers:** 1 high-risk file (`src/components/Shared/Autosave/useAutosave.ts`) + 8 medium-risk source files + financial-reporting domain + shared-hook blast radius into GoalCalculator / SalaryCalculator.
- **Reviewer level required:** Senior.
- **Auto-approve:** NOT eligible due to CRITICAL risk.

### Important Issues in this PR (severity 7.0–7.9)

1. **`ReimbursableExpensesGrid.tsx:132` — missing `.catch` on `saveField` call** — Severity 7.5
   - Flagged by Architecture + Dependency Impact analysis.
   - This call site is **not in the diff hunks** so cannot be anchored here, but the issue is caused by this PR: `useSaveField` now re-throws, and every other call site in this PR added `.catch(() => {})` — except this one. `processRowUpdate` will receive the re-thrown rejection; MUI's DataGrid has no `onProcessRowUpdateError` wired, causing a console warning and uncoordinated double rollback (Apollo optimistic + MUI row reset). Fix: `await saveField({ [field.fieldName]: amount }).catch(() => {});`

2. **`useSaveField.test.tsx:105` — race-prone snackbar assertion** — see line comment.

### Medium / Suggestion findings

See line comments for the rest. Key themes:
- `refetchQueries: ['PdsGoalCalculations']` is redundant given cache normalization (Arch + Data agree).
- Raw `err.message` interpolation in user-facing snackbar (UX + Data agree).
- Five `.catch(() => {})` sites are functionally correct but fragile — coupled to `useSaveField`'s snackbar; consider a non-throwing API.
- Missing failure-path tests for the 5 `.catch(() => {})` sites and the new `ErrorOutlineIcon`.
- `CircularProgress` missing accessible name (pre-existing repo pattern).

### Findings on Related Files (Not in This PR — informational)

**`src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/Autosave/useSaveField.ts`** — Severity 9.0 (Architecture)
- The PR changes the shared `useAutoSave` to silently revert `internalValue` on save failure. PDS adds a snackbar before re-throwing to make this visible. **GoalCalculator's parallel `useSaveField` has NO snackbar** — after this PR merges, a failed save in GoalCalculator will silently revert the user's input with zero feedback.
- Recommendation: Mirror the PDS pattern (try/catch + snackbar + rethrow) in GoalCalculator's `useSaveField`, OR move the error feedback into `useAutoSave` itself. Open a follow-up ticket or expand this PR's scope.

**`src/components/HrTools/PdsGoalCalculator/SupportItem/otherBreakdown.tsx`** — Severity 7.0 (Data Integrity)
- `OtherSection.tsx` → `buildOtherBreakdownRows` calls `calculateOtherExpenses(...)` directly with UNROUNDED values. The new `PdsSummaryTable` consumes the ROUNDED values from `usePdsSummaryData`. The "Other" breakdown and the SummaryReport will show different amounts for the same line items (e.g., $426.26 vs $426; $968.76 vs $969). Recommendation: have `OtherSection` consume `summaryData.otherTotals` instead.

**`src/components/HrTools/PdsGoalCalculator/SummaryReport/PdsSummaryTable.tsx`** — Severity 4.0 (Financial)
- Salary/reimbursable/403b/workComp/benefits rows remain unrounded; subtotal/attrition/creditCardFees/assessment rows are now integer. A user auditing "row 2A + 2B + 2C = row 2" by eye won't get exact equality. Decide whether to round the component lines for visual consistency.

### Standards Checklist (notable items)

- ✅ Named exports, file naming, i18n with `t()`, generated operation types, no `console.log`/`new Date()`/unused imports.
- ⚠️ `@ts-ignore` on `jest.requireActual('notistack')` matches 10+ repo precedents but the inline reason is just an eslint-disable, not a "why".
- ⚠️ Empty `.catch(() => {})` blocks are flagged by literal-rule reading but are intentional (snackbar fires upstream in `useSaveField`). Consider an inline comment at each site.
- ⚠️ `PdsGoalCalculatorContext.tsx` removes 2 setState setters from useMemo deps — correct (setters are stable) but verify `react-hooks/exhaustive-deps` lints clean.

### Dependency Impact

- `useAutoSave` — 11 consumers (PDS, GoalCalculator, SalaryCalculator) — silent behavior change affects all.
- `useSaveField` (PDS) — 5 consumers — all safe except `ReimbursableExpensesGrid.tsx:132`.
- `usePdsSummaryData` — 2 consumers, both inside PDS.

### Agent Summary

|---|---|---|---|---|---|
| Architecture | 1 (related-file) | 0 | 1 | 7 | High |
| Data Integrity | 0 | 0 | 1 (related-file) | 5 | High |
| Financial Reporting | 0 | 0 | 0 | 3 | High |
| Testing | 0 | 0 | 1 | 7 | High |
| UX | 0 | 0 | 0 | 5 | High |
| **PR-file total** | **0** | **0** | **2** | **~22** | **High** |

*Security agent skipped (no auth/api/env/CSP/CI triggers).*

---

🤖 Generated by `/quality:agent-review` (standard mode)

src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.tsx
```diff
@@ -114,7 +114,7 @@ export const SetupStep: React.FC = () => {
     setRightPanelContent(
       <HoursPerWeekGrid
         onApply={(average) => {
-          saveField({ hoursWorkedPerWeek: average });
+          saveField({ hoursWorkedPerWeek: average }).catch(() => {});
```
wjames111, line 117: [Medium] **Empty `.catch(() => {})` pattern is fragile across 5 sites** (Architecture, Testing, Standards agree).

This (and 4 other sites in this PR: `SetupStep.tsx:225,324`, `HoursPerWeekGrid.tsx:294,348`) silently swallows the re-thrown rejection from `useSaveField`. It works today because `useSaveField` enqueues a snackbar BEFORE re-throwing. If a future contributor moves the snackbar out of `useSaveField`, every one of these 5 sites becomes a silent failure with no warning.

The codebase has zero other `.catch(() => {})` occurrences outside this PR — this is a new pattern.

CLAUDE.md / `.claude/rules/code-review.md` explicitly forbids "empty `catch {}` blocks that swallow errors silently." Strict reading: violation. Spirit reading: acceptable because the error IS surfaced upstream.

**Recommended fix (one of):**
1. **Document the intent** at each site with an inline comment: `.catch(() => {}); // snackbar shown by useSaveField`.
2. **Refactor `useSaveField`** to not re-throw — instead, accept an optional `onError` callback, so non-autosave callers don't need any boundary handling:
```ts
const saveField = useCallback(
  async (attributes, { onError }: { onError?: () => void } = {}) => {
    try { return await trackMutation(...); }
    catch (error) {
      enqueueSnackbar(t('Failed to save changes. Please try again.'), { variant: 'error' });
      onError?.();
    }
  },
  [...],
);
```
Option 2 also lets us delete every `.catch(() => {})` AND fixes the cross-feature regression in GoalCalculator (see review body).

src/components/HrTools/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
```diff
@@ -291,7 +291,7 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
         }
         saveField({
           averageHoursPerWeek: newAverage,
-        });
+        }).catch(() => {});
```
wjames111, line 294: [Medium] **No failure-path test for this `.catch(() => {})` site.**

This and `HoursPerWeekGrid.tsx:348` (the `processRowUpdate` `.catch`) have no test that exercises the rejection path. If the upstream snackbar in `useSaveField` is ever moved or refactored, these empty catches become silent regressions with zero coverage.

**Suggested test (`HoursPerWeekGrid.test.tsx`):**
```tsx
it('shows snackbar but does not crash when delete fails', async () => {
  const { getByText, getByLabelText, findByText } = render(
    <PdsGoalCalculatorTestWrapper
      calculationMock={defaultCalculationMock}
      onCall={mutationSpy}
      updatePdsGoalCalculationError
    >
      <HoursPerWeekGrid />
  );
  await findByText('Regular Week');
  userEvent.click(getByText('Add Entry'));
  userEvent.click(getByLabelText('Delete'));
  expect(await findByText(/Failed to save changes/)).toBeInTheDocument();
});
```

src/components/HrTools/PdsGoalCalculator/SummaryReport/SummaryReportStep.tsx
```diff
@@ -11,13 +10,17 @@ export const SummaryReportStep: React.FC = () => {
...
+  const { data, error } = useAccountListSupportRaisedQuery({
     variables: { accountListId },
   });
   const supportRaised = data?.accountList.receivedPledges ?? 0;
 
   if (calculationLoading) {
-    return <Loading loading />;
+    return (
+      <Box sx={{ display: 'flex', justifyContent: 'center', py: 2 }}>
+        <CircularProgress size={24} />
```
wjames111, line 21: [Medium] **`CircularProgress` missing accessible name / `aria-live`.**

Screen-reader users will hear nothing while the data is loading. The `<CircularProgress size={24} />` has no `aria-label`, and the wrapping `<Box>` has no `role="status"` or `aria-live="polite"`.

This matches existing repo precedent — `PdsGoalsList.tsx`, `CreateGoalDialog.tsx`, and the shared modals all use bare `<CircularProgress>` without labels. So this is a missed improvement, not a regression. Flagged for visibility, not blocker.

**Suggested fix:**
```tsx
<Box
  role="status"
  aria-live="polite"
  sx={{ display: 'flex', justifyContent: 'center', py: 2 }}
>
  <CircularProgress size={24} aria-label={t('Loading support progress')} />
</Box>
```

src/components/HrTools/PdsGoalCalculator/SummaryReport/SummaryReportStep.tsx
```diff
@@ -11,13 +10,17 @@ export const SummaryReportStep: React.FC = () => {
   const { t } = useTranslation();
   const accountListId = useAccountListId() ?? '';
   const { calculationLoading } = usePdsGoalCalculator();
-  const { data } = useAccountListSupportRaisedQuery({
+  const { data, error } = useAccountListSupportRaisedQuery({
```
wjames111, line 13: [Suggestion] **Missing loading state for `supportLoading`.**

This branch only handles `calculationLoading`. If `calculationLoading` is false but the support-raised query is still in flight, `supportRaised` defaults to `0` and `<PdsSummaryTable>` renders with 0% progress until the query resolves. Pre-existing behavior, lightly worsened by adding the explicit `error` branch — adding `supportLoading` would close the gap.

**Suggested fix:**
```tsx
const { data, loading: supportLoading, error } = useAccountListSupportRaisedQuery({
  variables: { accountListId },
});

if (calculationLoading || supportLoading) {
  return (
    <Box role="status" aria-live="polite" sx={{ display: 'flex', justifyContent: 'center', py: 2 }}>
      <CircularProgress size={24} aria-label={t('Loading')} />
    </Box>
  );
}
```

src/components/Shared/Autosave/useAutosave.ts
```diff
@@ -66,6 +66,15 @@ export const useAutoSave = <Value extends string | number>({
     };
   }, [fieldName, errorMessage, markValid, markInvalid]);
 
+  // If a save fails (e.g. Apollo rolls back the optimistic response),
+  // revert the local input back to the canonical value from the cache so
+  // it doesn't keep showing what the user typed.
+  const revertOnFailure = (savePromise: Promise<unknown>) => {
```
wjames111, line 72: [Suggestion] **`revertOnFailure` captures `value` in closure at call-time** (Architecture + Data Integrity agree).

`revertOnFailure` is defined inline on each render. The `.catch` callback closes over `value` at the moment the function ran (when the user typed/blurred). If a concurrent save lands and updates the cache between save start and rejection, the revert restores the captured `value` rather than the latest cache value. `useSyncedState` self-corrects on the next render, but the user may see a momentary stale value.

**Suggested fix (use a ref to read the latest value at revert time):**
```ts
const valueRef = useRef(value);
useEffect(() => { valueRef.current = value; }, [value]);

const revertOnFailure = (savePromise: Promise<unknown>) => {
  savePromise.catch(() => {
    setInternalValue(valueRef.current?.toString() ?? '');
  });
};
```

**Important cross-feature note** (see review body): the broader concern with this hook is that it silently reverts in *any* consumer whose `saveValue` rejects. PDS adds a snackbar in its own `useSaveField` before re-throwing, but **GoalCalculator's `useSaveField` has no snackbar** — after this PR merges, a failed save in GoalCalculator will silently roll back with zero feedback. Either: (a) push error feedback into `useAutoSave` (default snackbar, or accept an `onSaveError` callback), or (b) update GoalCalculator's `useSaveField` to mirror the PDS pattern.

src/components/HrTools/PdsGoalCalculator/calculations/usePdsSummaryData.ts
```diff
@@ -74,7 +74,18 @@ export const usePdsSummaryData = (
       salaryTotals,
       reimbursableTotals.total,
     );
-    const otherTotals = calculateOtherExpenses(calculation, otherConstants);
+    const rawOtherTotals = calculateOtherExpenses(calculation, otherConstants);
+
+    // Round the four gross-up components once at the display boundary so the
+    // line items rendered in the summary table sum exactly to overallTotal
+    // (and match what gets submitted as the monthly goal).
+    const otherTotals = {
```
wjames111, line 82: [Suggestion] **Hook now serves as a display boundary** (Architecture + Financial Reporting agree).

The repo rule says "rounding should happen at the display boundary via `intlFormat` / `Intl.NumberFormat`, not sprinkled through calculation code." This PR moves `Math.round` INTO the calculation hook. The justification (sum-consistency: displayed line items must add to the displayed total, and the submitted `monthlyGoal` must equal the displayed total) is legitimate — but it also means any future consumer of `usePdsSummaryData` (e.g., a CSV export, an analytics call) gets pre-rounded values.

Also: the salary subtotal / reimbursable / 403b / workComp / benefits rows in `PdsSummaryTable` remain unrounded. Mixed precision in the table — see review body.

**Consider exposing both raw and display variants:**
```ts
return {
  ...,
  otherTotals: rawOtherTotals,        // for calculations / exports
  displayOtherTotals: otherTotals,    // for the summary table & monthlyGoal
  overallTotal,                       // the rounded display total
};
```

The Financial Reporting agent verified the rounding math itself is sound: `overallTotal` is invariant-integer (sum of 4 `Math.round` integers), well within `Number.MAX_SAFE_INTEGER`.

src/components/HrTools/PdsGoalCalculator/ReimbursableExpenses/ReimbursableExpensesGrid.tsx
```diff
@@ -159,7 +162,12 @@ export const ReimbursableExpensesGrid: React.FC<
     const formatted = currencyFormat(params.value ?? 0, 'USD', locale);
 
     if (error) {
-      return <ErrorCell title={error}>{formatted}</ErrorCell>;
+      return (
+        <ErrorCell>
+          <ErrorOutlineIcon fontSize="small" aria-hidden="true" />
```
wjames111, line 167: [Medium] **New `ErrorOutlineIcon` has no test coverage.**

The new error-cell rendering (`ErrorOutlineIcon` with `aria-hidden="true"`, color from `theme.palette.error.main`) is not asserted anywhere in `ReimbursableExpensesGrid.test.tsx`. The only test change in this file is `expect(mutationSpy).not.toHaveGraphqlOperation('UpdatePdsGoalCalculation')`. Without an icon assertion, removing the icon would not fail any test.

**Accessibility note** (verified safe): the error is announced to screen-readers via the existing `<FormHelperText role="alert">` (around line 220-224), so `aria-hidden="true"` on the icon is correct. The amount text alone would be color-only signaling, but the icon + colored amount + announced alert combine for dual-channel feedback.

**Suggested test:**
```tsx
it('renders the error icon when a cell has a validation error', async () => {
  // ... trigger negative-amount validation ...
  const errorCell = await findByRole('cell', { name: /\-5/ });
  expect(errorCell.querySelector('[data-testid="ErrorOutlineIcon"]')).toBeInTheDocument();
});
```

PR #1792, opened 2026-05-21
[no-Jira] Migrate to TypeScript 6

canac:
## Description

Upgrade to TypeScript 6.0

## Testing

- [x] Test that CSV upload and profile image upload still works

## Checklist:

### DIFF
--- a/app.d.ts
+++ b/app.d.ts
@@ -5,4 +5,5 @@ declare module '*.graphql' {
   export = value;
 }
 
+declare module '*.css';
 declare module '*.svg';
--- a/package.json
+++ b/package.json
@@ -141,7 +141,7 @@
     "prettier": "^3.6.2",
     "puppeteer": "24.4.0",
     "ts-essentials": "^9.3.0",
-    "typescript": "~5.6.3",
+    "typescript": "~6.0.0",
     "url-loader": "^4.1.1",
     "webpack": "^5.96.1"
   },
--- a/package.json
+++ b/package.json
@@ -68,7 +68,6 @@
     "next-auth": "^4.24.11",
     "next-optimized-images": "^2.6.2",
     "next-pwa": "^5.6.0",
-    "node-fetch": "^3.3.2",
     "notistack": "^2.0.5",
     "prop-types": "^15.8.1",
     "react": "^18.2.0",
--- a/pages/api/uploads/csv-upload.page.ts
+++ b/pages/api/uploads/csv-upload.page.ts
@@ -54,7 +54,7 @@ const uploadFile = async (
     }
 
     const csvFile = new File(
-      [await readFile(file.filepath)],
+      [new Uint8Array(await readFile(file.filepath))],
       file.originalFilename ?? 'csvFile',
     );
 
--- a/pages/api/uploads/csv-upload.page.ts
+++ b/pages/api/uploads/csv-upload.page.ts
@@ -1,8 +1,7 @@
-import { readFile } from 'fs/promises';
+import { openAsBlob } from 'node:fs';
 import { NextApiRequest, NextApiResponse } from 'next';
 import formidable, { IncomingForm } from 'formidable';
 import { getToken } from 'next-auth/jwt';
-import fetch, { File, FormData } from 'node-fetch';
 
 export const config = {
   api: {
@@ -54,7 +53,7 @@ const uploadFile = async (
     }
 
     const csvFile = new File(
-      [new Uint8Array(await readFile(file.filepath))],
+      [await openAsBlob(file.filepath)],
       file.originalFilename ?? 'csvFile',
     );
 
--- a/.pnp.cjs
+++ b/.pnp.cjs
@@ -122,7 +122,6 @@ const RAW_RUNTIME_STATE =
           ["next-compose-plugins", "npm:2.2.1"],\
           ["next-optimized-images", "npm:2.6.2"],\
           ["next-pwa", "virtual:9909ff5388c6b6a3a46f12eb37c0afb449fcd1eedb9f02d871bde711a076c929583f48ecc4b85fa6d71478b076104a25f83dee45bc69687a22f551c576d7595d#npm:5.6.0"],\
-          ["node-fetch", "npm:3.3.2"],\
           ["node-mocks-http", "virtual:9909ff5388c6b6a3a46f12eb37c0afb449fcd1eedb9f02d871bde711a076c929583f48ecc4b85fa6d71478b076104a25f83dee45bc69687a22f551c576d7595d#npm:1.16.2"],\
           ["notistack", "virtual:9909ff5388c6b6a3a46f12eb37c0afb449fcd1eedb9f02d871bde711a076c929583f48ecc4b85fa6d71478b076104a25f83dee45bc69687a22f551c576d7595d#npm:2.0.5"],\
           ["prettier", "npm:3.6.2"],\
@@ -12612,13 +12611,6 @@ const RAW_RUNTIME_STATE =
       }]\
     ]],\
     ["data-uri-to-buffer", [\
-      ["npm:4.0.1", {\
-        "packageLocation": "./.yarn/cache/data-uri-to-buffer-npm-4.0.1-5c66a78beb-0d0790b67f.zip/node_modules/data-uri-to-buffer/",\
-        "packageDependencies": [\
-          ["data-uri-to-buffer", "npm:4.0.1"]\
-        ],\
-        "linkType": "HARD"\
-      }],\
       ["npm:6.0.2", {\
         "packageLocation": "./.yarn/cache/data-uri-to-buffer-npm-6.0.2-1725fff558-8b6927c33f.zip/node_modules/data-uri-to-buffer/",\
         "packageDependencies": [\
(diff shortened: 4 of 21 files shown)


package.json
```diff
@@ -141,7 +141,7 @@
     "prettier": "^3.6.2",
     "puppeteer": "24.4.0",
     "ts-essentials": "^9.3.0",
-    "typescript": "~5.6.3",
+    "typescript": "~6.0.0",
```
canac, line 143: [Important] The `String(t(...))` churn this PR adds in 7 component files is a workaround for stale i18next typings — `i18next@21.9.2` and `react-i18next@11.18.6` (3 majors behind) declare `TFunction` returning `string | object | undefined | null`. Newer versions return `string` by default, or this can be fixed today by augmenting `react-i18next` with `interface CustomTypeOptions { returnNull: false; returnObjects: false; returnEmptyString: false }`. Either approach eliminates the wrap entirely **and** prevents the ~73 other unwrapped `aria-label={t(...)}` sites in the codebase (which compile today only because looser MUI prop typings tolerate the wider type) from becoming landmines when MUI next tightens a prop signature. Recommend opening a follow-up ticket — not blocking this PR.

app.d.ts
```diff
@@ -5,4 +5,5 @@ declare module '*.graphql' {
   export = value;
 }
 
+declare module '*.css';
```
canac, line 8: [Medium] With the new `tsconfig.json` `types: ["node", "jest", "@testing-library/jest-dom"]` narrowing, ambient `google.maps` types resolve only because `@googlemaps/js-api-loader` carries a transitive `/// <reference types="google.maps" />`. Any future upgrade of that package that drops the reference will silently lose those ambient types. Recommend adding an explicit reference here so the chain isn't load-bearing on a transitive dep's internals:

```ts
/// <reference types="google.maps" />
declare module '*.graphql' { ... }
declare module '*.css';
declare module '*.svg';
```

src/components/Reports/ExpectedMonthlyTotalReport/Table/ExpectedMonthlyTotalReportTable.tsx
```diff
@@ -86,10 +87,7 @@ export const ExpectedMonthlyTotalReportTable: React.FC<Props> = ({
         </AccordionSummary>
         <AccordionDetails>
           <TableContainer component={Paper}>
-            <Table
-              style={{ minWidth: 700 }}
-              aria-label={t('Expected monthly total report table')}
-            >
+            <Table style={{ minWidth: 700 }} aria-label={label}>
```
canac: [Medium] Inconsistent fix style — this file hoists `const label = t('Expected monthly total report table')` at line 59 and uses `aria-label={label}` here, while the other 7 components touched in this PR all use the inline `aria-label={String(t('...'))}` pattern. Both work, but mixing two idioms across a single PR makes the rationale opaque for future contributors. Pick one. Suggested for consistency with the rest of the PR:

```tsx
<Table
  style={{ minWidth: 700 }}
  aria-label={String(t('Expected monthly total report table'))}
>
```

(Flagged independently by Architecture, Testing, UX, and Standards agents.)

pages/api/uploads/csv-upload.page.ts
```diff
@@ -54,7 +54,7 @@ const uploadFile = async (
     }
 
     const csvFile = new File(
-      [await readFile(file.filepath)],
+      [new Uint8Array(await readFile(file.filepath))],
```
canac: [Suggestion] `new Uint8Array(Buffer)` is byte-identical to passing the `Buffer` directly (`Buffer extends Uint8Array`, the constructor copies elements via `byteOffset`/`byteLength` — no risk of reading past the buffer). The wrap is the correct TS 6 / updated-DOM-lib fix for `BlobPart` typing. However, this endpoint and `upload-person-avatar.page.ts` have no automated test coverage (pre-existing, excluded by `jest.config.js` coverage policy). Smoke-test both upload paths (CSV import + avatar upload) end-to-end before merge to confirm no upstream multipart `Content-Length` / mime-detection regression in the formidable → upstream-REST flow.

canac (commented):
## Multi-Agent Code Review

**Verdict: APPROVED WITH SUGGESTIONS** — no blockers, mechanical TS 5.6 → 6.0 bump. 1 Important + 4 Medium/Suggestion findings posted as inline comments.

### Risk Assessment
- **Score:** 10/10 (capped) — **CRITICAL**
- Driven by typescript compiler-version blast radius across the whole repo, **not** by semantic risk in the diffs themselves. The actual changes are minimal and mechanical.
- Gate merge on a green whole-repo `yarn lint:ts` + `yarn test` CI run, plus a manual smoke test of the two upload paths (CSV import + avatar upload) given the `new Uint8Array(...)` wrap.

### Agents Run
Security, Architecture, Testing, UX, Standards, Financial Reporting (Data Integrity skipped — no schema/migration/Apollo cache/mutation/pagination changes).

### Key Findings

1. **[Important 7.0]** Root cause of every `String(t(...))` wrap is stale `i18next@21.9.2` / `react-i18next@11.18.6`. Newer versions return `string` by default, or augment via `interface CustomTypeOptions { returnNull: false }`. Eliminates the churn and de-risks the ~73 other unwrapped `aria-label={t(...)}` sites that compile today only because looser MUI prop typings tolerate the wider return type. Follow-up ticket recommended — not blocking.

2. **[Medium 6.5]** Inconsistent fix style: `ExpectedMonthlyTotalReportTable.tsx` hoists `const label = t(...)` while the other 7 files inline `String(t(...))`. Pick one. Flagged by 4 agents (Architecture, Testing, UX, Standards).

3. **[Medium 6.0]** New `tsconfig.json` `types: ["node", "jest", "@testing-library/jest-dom"]` narrowing is load-bearing on a transitive `/// <reference types="google.maps" />` from `@googlemaps/js-api-loader`. Add an explicit reference in `app.d.ts` so a future dep upgrade can't silently drop ambient types.

4. **[Medium 5.0]** Selective application — only ~7 of ~80 `aria-label={t(...)}` sites were wrapped (rest compile via looser MUI prop typings). Confirm `yarn lint:ts` passes whole-repo, not just changed files, before merging.

5. **[Suggestion 4.0]** `@types/testing-library__jest-dom` is now redundant — the runtime `@testing-library/jest-dom` package ships its own types (and is referenced explicitly in the new `types` array). Remove from devDependencies.

6. **[Suggestion 4.0]** No tests cover the two upload endpoints (pre-existing, excluded by jest coverage policy). `new Uint8Array(Buffer)` wrap is byte-identical at runtime, but smoke-test both upload paths before merge.

### What's Clean
- **Security:** No findings. `NextAuthOptions['providers']` swap is type-equivalent to `Provider[]`. `new Uint8Array(buffer)` copies bytes 1:1 (no risk of reading past buffer). `typescript` correctly placed in `devDependencies`. No NextAuth callback / CSP / env-var / redirect changes.
- **Financial Reporting:** No calculation, aggregation, currency, or date logic touched.
- **Standards Checklist:** All applicable items PASS. No new `any`, no `@ts-ignore`, no debug output, no commented-out code, no `new Date()`.
- **Architecture positives:** `moduleResolution: "bundler"` is the correct choice for Next 15 + TS 6. Removing `baseUrl` while adding explicit `paths` is the right move. `declare module '*.css'` confirmed needed (2 real CSS imports in `_app.page.tsx`).

### Dependency Impact
Global: `tsconfig.json`, `app.d.ts`, `package.json`, `.yarn/sdks/typescript/package.json`. No breaking runtime changes.

---

### Findings on Related Files (Not in Diff Hunk)

The following finding is on a file in this PR but on a line outside the diff hunk, so it cannot be posted as a line-anchored comment:

**[Suggestion 4.0] `package.json:116`** — `@types/testing-library__jest-dom` is now redundant. The runtime `@testing-library/jest-dom` package ships its own types and is referenced explicitly in the new `tsconfig.json` `types` array. Remove this DefinitelyTyped package from devDependencies: `yarn remove @types/testing-library__jest-dom`. Informational — not blocking.

canac:
@wjames111 This is not urgent at all. But I remembered I had this working in a branch a few weeks ago, so I might as well push it up. This should take care of all the breaking changes so that when TypeScript 7.0 comes out (which is supposed to be ~10x faster 🚀 ) it will be an easy upgrade.

pages/api/uploads/csv-upload.page.ts
wjames111: From Claude:  new Uint8Array(typedArray) allocates a brand-new ArrayBuffer and copies the bytes (per ECMA-262 24.1.1.4). readFile already returns a Buffer (which extends Uint8Array), so for every uploaded file we now pay 2× peak memory. Zero-copy alternative that still satisfies the TS6 BlobPart typing:
```Suggestion
const buf = await readFile(file.filepath);
const csvFile = new File( [new Uint8Array(buf.buffer, buf.byteOffset, buf.byteLength)],
file.originalFilename ?? 'csvFile',
 );
```

src/components/Coaching/CoachingDetail/LevelOfEffort/LevelOfEffort.tsx
wjames111, line 1: Question why are we only updating this in these 8 files and not the rest of the codebase?

app.d.ts
wjames111, line 8: Optional: Would you mind adding a comment on why this is necessary? I'm a little confused by the PR comment above.

wjames111 (approved):
Looks great, looking forward to the upgrade! Claude mentioned this as well, "the String(t(...)) workaround papered over a typing problem that's better solved with one module-augmentation block in app.d.ts" if this is possible to do a in a module wouldn't it be better just to do it there?

app.d.ts
canac: It's so that we can do `import './helpjuice.css'`

wjames111: Ahh okay, then disregard, I didn't realize we were importing any css files.

src/components/Coaching/CoachingDetail/LevelOfEffort/LevelOfEffort.tsx
canac: It's only an issue when passing `t(...)` to `aria-label`.

PR #1797, opened 2026-05-21
MPDX-9611 Scope PDS autosave disable to per-field saves

wjames111:
## Description

- The Form Type, Employment Status, and Pay Type selects in the PDS Goal Calculator Setup step briefly rendered disabled whenever an unrelated field (e.g. Benefits) was being saved, because they were gated on the global `isMutating` flag from the calculator context.
- Added per-field saving state to `PdsGoalCalculatorContext`: a new `isFieldSaving(fieldName)` selector and a `trackFieldMutation(promise, fields)` wrapper that tags the fields being written for the duration of the mutation.
- `useSaveField` now uses `trackFieldMutation` and passes `Object.keys(attributes)`, so each in-flight save tags only the fields it actually writes.
- `usePdsGoalAutoSave` and the manual Pay Type select in `SetupStep` now check `isFieldSaving(fieldName)` instead of `isMutating`. Same-field race protection on `formType` (and friends) is preserved — only the cross-field flicker is gone.
- Added a regression test that saving an unrelated field does not disable a `saveOnChange` select.
- Jira: [MPDX-9611](https://jira.cru.org/browse/MPDX-9611)

## Testing

- Open the PDS Goal Calculator Setup step for a Full-time Salaried goal so all of Form Type, Employment Status, Pay Type, Pay Rate, and Benefits are visible.
- Edit the Benefits field, then tab/click away to trigger the autosave.
- Watch Form Type, Employment Status, and Pay Type while the Benefits save is in flight — they should remain interactive (no flicker to a disabled state).
- Toggle Form Type (Default ↔ Simple) rapidly and confirm the Form Type select itself is briefly disabled while its own save is in flight — same-field race protection still applies.
- Toggle Pay Type (Salaried ↔ Hourly) and confirm the Pay Type select disables only during its own save, and the atomic `salaryOrHourly + payRate: null` write still happens.

## Checklist:

### DIFF
--- a/src/components/HrTools/PdsGoalCalculator/PdsGoalCalculator.test.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/PdsGoalCalculator.test.tsx
@@ -79,11 +79,8 @@ describe('PdsGoalCalculator', () => {
           name: 'Test Goal',
         }}
         extraMocks={{
-          // Pin the mutation response so the post-save cache matches the
-          // expected SALARIED + payRate: null state. Without this, the
-          // auto-generated mock fills salaryOrHourly with the first enum
-          // value (HOURLY), which re-introduces the Hours Worked
-          // requirement and keeps Continue disabled after typing.
+          // Pin the mutation response so the auto-generated mock doesn't fill
+          // salaryOrHourly with HOURLY (the first enum value).
           UpdatePdsGoalCalculation: {
             updateDesignationSupportCalculation: {
               designationSupportCalculation: {
@@ -122,9 +119,7 @@ describe('PdsGoalCalculator', () => {
       ).not.toBeInTheDocument();
     });
     // Switching Pay Type clears payRate, so the user must re-enter it before
-    // Continue becomes enabled. The Pay Rate input is disabled while the
-    // atomic { salaryOrHourly, payRate: null } save is in flight, so wait
-    // for the field to be enabled before typing.
+    // Continue becomes enabled.
     const payRateInput = await findByRole('spinbutton', {
       name: 'Annual Pay Rate',
     });
--- a/src/components/HrTools/PdsGoalCalculator/PdsGoalCalculator.test.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/PdsGoalCalculator.test.tsx
@@ -6,6 +6,7 @@ import {
   DesignationSupportSalaryType,
   DesignationSupportStatus,
 } from 'src/graphql/types.generated';
+import { UpdatePdsGoalCalculationMutation } from './GoalsList/PdsGoalCalculations.generated';
 import { PdsGoalCalculator } from './PdsGoalCalculator';
 import { PdsGoalCalculatorTestWrapper } from './PdsGoalCalculatorTestWrapper';
 
@@ -69,7 +70,9 @@ describe('PdsGoalCalculator', () => {
 
   it('re-enables Continue after switching from Hourly to Salaried and entering a new Pay Rate', async () => {
     const { findByRole, getByRole, queryByRole } = render(
-      <PdsGoalCalculatorTestWrapper
+      <PdsGoalCalculatorTestWrapper<{
+        UpdatePdsGoalCalculation: UpdatePdsGoalCalculationMutation;
+      }>
         calculationMock={{
           salaryOrHourly: DesignationSupportSalaryType.Hourly,
           hoursWorkedPerWeek: null,
@@ -123,6 +126,8 @@ describe('PdsGoalCalculator', () => {
     const payRateInput = await findByRole('spinbutton', {
       name: 'Annual Pay Rate',
     });
+    // Pay Type's atomic save (salaryOrHourly + payRate: null) locks payRate
+    // until the mutation resolves. Wait for it before typing.
     await waitFor(() => expect(payRateInput).not.toBeDisabled());
     userEvent.type(payRateInput, '50000');
 
--- a/src/components/HrTools/PdsGoalCalculator/PdsGoalCalculator.test.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/PdsGoalCalculator.test.tsx
@@ -78,6 +78,26 @@ describe('PdsGoalCalculator', () => {
           benefits: 1500,
           name: 'Test Goal',
         }}
+        extraMocks={{
+          // Pin the mutation response so the post-save cache matches the
+          // expected SALARIED + payRate: null state. Without this, the
+          // auto-generated mock fills salaryOrHourly with the first enum
+          // value (HOURLY), which re-introduces the Hours Worked
+          // requirement and keeps Continue disabled after typing.
+          UpdatePdsGoalCalculation: {
+            updateDesignationSupportCalculation: {
+              designationSupportCalculation: {
+                id: 'goal-1',
+                name: 'Test Goal',
+                status: DesignationSupportStatus.FullTime,
+                salaryOrHourly: DesignationSupportSalaryType.Salaried,
+                payRate: null,
+                hoursWorkedPerWeek: null,
+                benefits: 1500,
+              },
+            },
+          },
+        }}
       >
         <PdsGoalCalculator />
       </PdsGoalCalculatorTestWrapper>,
@@ -102,10 +122,13 @@ describe('PdsGoalCalculator', () => {
       ).not.toBeInTheDocument();
     });
     // Switching Pay Type clears payRate, so the user must re-enter it before
-    // Continue becomes enabled.
+    // Continue becomes enabled. The Pay Rate input is disabled while the
+    // atomic { salaryOrHourly, payRate: null } save is in flight, so wait
+    // for the field to be enabled before typing.
     const payRateInput = await findByRole('spinbutton', {
       name: 'Annual Pay Rate',
     });
+    await waitFor(() => expect(payRateInput).not.toBeDisabled());
     userEvent.type(payRateInput, '50000');
 
     await waitFor(() => expect(continueButton).not.toBeDisabled());
--- a/src/components/HrTools/PdsGoalCalculator/PdsGoalCalculatorTestWrapper.tsx
+++ b/src/components/HrTools/PdsGoalCalculator/PdsGoalCalculatorTestWrapper.tsx
@@ -1,4 +1,4 @@
-import React, { ComponentProps } from 'react';
+import React from 'react';
 import { ThemeProvider } from '@mui/material/styles';
 import { MockLinkCallHandler } from 'graphql-ergonomock/dist/apollo/MockLink';
 import { merge, mergeWith } from 'lodash';
@@ -132,7 +132,9 @@ export type GoalCalculatorConstantsMock = DeepPartial<
   GoalCalculatorConstantsQuery['constant']
 >;
 
-export interface PdsGoalCalculatorTestWrapperProps {
+export interface PdsGoalCalculatorTestWrapperProps<
+  TExtraMocks = Record<string, never>,
+> {
   children?: React.ReactNode;
   withProvider?: boolean;
   calculationsMock?: PdsGoalCalculationsMock;
@@ -141,15 +143,25 @@ export interface PdsGoalCalculatorTestWrapperProps {
   userMock?: GetUserMock;
   constantsMock?: GoalCalculatorConstantsMock;
   supportRaisedMock?: number;
-  /** Extra GqlMockedProvider mocks merged on top of the defaults. */
-  extraMocks?: ComponentProps<typeof GqlMockedProvider>['mocks'];
+  /**
+   * Extra GqlMockedProvider mocks merged on top of the defaults. Pass the
+   * operation map as a generic argument so mock shapes are type-checked:
+   * `<PdsGoalCalculatorTestWrapper<{ UpdatePdsGoalCalculation: UpdatePdsGoalCalculationMutation }> extraMocks={...} />`.
+   * Each value may be a partial mock object or a resolver function (e.g. to
+   * throw and exercise error paths).
+   */
+  extraMocks?: {
+    [K in keyof TExtraMocks]?:
+      | DeepPartial<TExtraMocks[K]>
+      | ((...args: never[]) => unknown);
+  };
   onCall?: MockLinkCallHandler;
   router?: React.ComponentProps<typeof TestRouter>['router'];
 }
 
-export const PdsGoalCalculatorTestWrapper: React.FC<
-  PdsGoalCalculatorTestWrapperProps
-> = ({
+export const PdsGoalCalculatorTestWrapper = <
+  TExtraMocks = Record<string, never>,
+>({
   children,
   withProvider = true,
   calculationsMock,
@@ -161,7 +173,7 @@ export const PdsGoalCalculatorTestWrapper: React.FC<
   extraMocks,
   onCall,
   router,
-}) => {
+}: PdsGoalCalculatorTestWrapperProps<TExtraMocks>): React.ReactElement => {
   return (
     <ThemeProvider theme={theme}>
       <TestRouter
@@ -179,96 +191,98 @@ export const PdsGoalCalculatorTestWrapper: React.FC<
             GetUser: GetUserQuery;
             AccountListSupportRaised: AccountListSupportRaisedQuery;
           }>
-            mocks={{
-              PdsGoalCalculations: {
-                designationSupportCalculations: merge(
-                  {},
-                  calculationsDefault,
-                  calculationsMock,
-                ),
-              },
-              PdsGoalCalculation: {
-                designationSupportCalculation: merge(
-                  {},
-                  calculationDefault,
-                  calculationMock,
-                ),
-              },
-              HcmUser: {
-                hcm:
-                  hcmUserMock === null
-                    ? []
-                    : [merge({}, hcmUserDefault.hcm[0], hcmUserMock)],
-              },
-              ...(userMock ? { GetUser: userMock } : {}),
-              ...(supportRaisedMock !== undefined
-                ? {
-                    AccountListSupportRaised: {
-                      accountList: {
-                        id: 'account-list-1',
-                        receivedPledges: supportRaisedMock,
+            mocks={merge(
+              {
+                PdsGoalCalculations: {
+                  designationSupportCalculations: merge(
+                    {},
+                    calculationsDefault,
+                    calculationsMock,
+                  ),
+                },
+                PdsGoalCalculation: {
+                  designationSupportCalculation: merge(
+                    {},
+                    calculationDefault,
+                    calculationMock,
+                  ),
+                },
+                HcmUser: {
+                  hcm:
+                    hcmUserMock === null
+                      ? []
+                      : [merge({}, hcmUserDefault.hcm[0], hcmUserMock)],
+                },
+                ...(userMock ? { GetUser: userMock } : {}),
+                ...(supportRaisedMock !== undefined
+                  ? {
+                      AccountListSupportRaised: {
+                        accountList: {
+                          id: 'account-list-1',
+                          receivedPledges: supportRaisedMock,
+                        },
                       },
+                    }
+                  : {}),
+                GoalCalculatorConstants: {
+                  constant: mergeWith(
+                    {},
+                    {
+                      mpdGoalBenefitsConstants: [],
+                      mpdGoalGeographicConstants: [
+                        {
+                          location: 'None',
+                          percentageMultiplier: 0,
+                        },
+                        {
+                          location: 'Orlando, FL',
+                          percentageMultiplier: 0.06,
+                        },
+                        {
+                          location: 'New York, NY',
+                          percentageMultiplier: 0.12,
+                        },
+                      ],
+                      mpdGoalMiscConstants: [
+                        {
+                          category:
+                            MpdGoalMiscConstantCategoryEnum.AdditionalRates,
+                          label: MpdGoalMiscConstantLabelEnum.EmployerFicaRate,
+                          fee: 0.08,
+                        },
+                        {
+                          category:
+                            MpdGoalMiscConstantCategoryEnum.AdditionalRates,
+                          label:
+                            MpdGoalMiscConstantLabelEnum.PartTimeWorkCompensation,
+                          fee: 0.17,
+                        },
+                        {
+                          category: MpdGoalMiscConstantCategoryEnum.Rates,
+                          label: MpdGoalMiscConstantLabelEnum.AttritionRate,
+                          fee: 0.06,
+                        },
+                        {
+                          category:
+                            MpdGoalMiscConstantCategoryEnum.AdditionalRates,
+                          label: MpdGoalMiscConstantLabelEnum.CreditCardFeeRate,
+                          fee: 0.06,
+                        },
+                        {
+                          category: MpdGoalMiscConstantCategoryEnum.Rates,
+                          label: MpdGoalMiscConstantLabelEnum.AdminRate,
+                          fee: 0.12,
+                        },
+                      ],
                     },
-                  }
-                : {}),
-              GoalCalculatorConstants: {
-                constant: mergeWith(
-                  {},
-                  {
-                    mpdGoalBenefitsConstants: [],
-                    mpdGoalGeographicConstants: [
-                      {
-                        location: 'None',
-                        percentageMultiplier: 0,
-                      },
-                      {
-                        location: 'Orlando, FL',
-                        percentageMultiplier: 0.06,
-                      },
-                      {
-                        location: 'New York, NY',
-                        percentageMultiplier: 0.12,
-                      },
-                    ],
-                    mpdGoalMiscConstants: [
-                      {
-                        category:
-                          MpdGoalMiscConstantCategoryEnum.AdditionalRates,
-                        label: MpdGoalMiscConstantLabelEnum.EmployerFicaRate,
-                        fee: 0.08,
-                      },
-                      {
-                        category:
-                          MpdGoalMiscConstantCategoryEnum.AdditionalRates,
-                        label:
-                          MpdGoalMiscConstantLabelEnum.PartTimeWorkCompensation,
-                        fee: 0.17,
-                      },
-                      {
-                        category: MpdGoalMiscConstantCategoryEnum.Rates,
-                        label: MpdGoalMiscConstantLabelEnum.AttritionRate,
-                        fee: 0.06,
-                      },
-                      {
-                        category:
-                          MpdGoalMiscConstantCategoryEnum.AdditionalRates,
-                        label: MpdGoalMiscConstantLabelEnum.CreditCardFeeRate,
-                        fee: 0.06,
-                      },
-                      {
-                        category: MpdGoalMiscConstantCategoryEnum.Rates,
-                        label: MpdGoalMiscConstantLabelEnum.AdminRate,
-                        fee: 0.12,
-                      },
-                    ],
-                  },
-                  constantsMock,
-                  (_objValue, srcValue) =>
-                    Array.isArray(srcValue) ? srcValue : undefined,
-                ),
+                    constantsMock,
+                    (_objValue, srcValue) =>
+                      Array.isArray(srcValue) ? srcValue : undefined,
+                  ),
+                },
               },
-              ...extraMocks,
-            }}
+              extraMocks,
+            )}
             onCall={onCall}
           >
             {withProvider ? (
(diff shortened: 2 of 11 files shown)


src/components/HrTools/PdsGoalCalculator/Shared/Autosave/usePdsGoalAutoSave.ts
```diff
@@ -15,17 +15,20 @@ export const usePdsGoalAutoSave = ({
...
-    // calculation, so a stale final value silently understates the total.
-    disabled: !calculation || (options.saveOnChange === true && isMutating),
+    // Block change-driven (select) autosaves while a save for this same field
+    // is in flight: rapid back-and-forth toggles can otherwise land out of
+    // order in the Apollo cache. formType is the load-bearing case — its
+    // value reshapes the goal calculation, so a stale final value silently
+    // understates the total. Other fields' saves do not affect this one.
+    disabled:
+      !calculation ||
+      (options.saveOnChange === true && isFieldSaving(fieldName)),
```
wjames111: [High] **Cross-field stale-write race** (consensus from Data Integrity 8.0, UX 6.5, Financial Reporting 6.5).

The `UpdatePdsGoalCalculation` response carries the **full** `PdsGoalCalculationFields` fragment. If save A (e.g., on `formType`) is in flight and the user fires save B (e.g., on `status`, or the Pay Type select which atomically writes `{ salaryOrHourly, payRate: null }`), B's server response can echo the *pre-A* `formType` value. Apollo's last-write-wins normalization then stamps the stale `formType` into the cache — exactly the silent total-miscalculation the comment above warns about.

A second related case: Pay Type writes `{ salaryOrHourly, payRate: null }`. Both are correctly marked in `savingFieldCounts`, but the Pay Rate input is `saveOnChange=false`, so this `disabled` check ignores `isFieldSaving('payRate')` and a user-typed value in Pay Rate can race the atomic clear.

**Suggested fix** (the smallest change that closes both races):
```ts
disabled: !calculation || isFieldSaving(fieldName),
```
This makes blur-driven fields also disable while their own save is in flight. Combined with reverting the Pay Type select to disable on any in-flight mutation (see the comment on `SetupStep.tsx:217`), the load-bearing `formType` is fully protected.

After the fix, the test at lines 147-169 ("does not disable blur-driven fields while a save is in flight") will need to be inverted, and the new test at lines 171-203 will need to verify the cross-field RACE is prevented rather than that selects remain enabled.

src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.tsx
```diff
@@ -214,7 +214,7 @@ export const SetupStep: React.FC = () => {
               label={t('Pay Type')}
               helperText={t('Changing this clears Pay Rate.')}
               value={calculation?.salaryOrHourly ?? ''}
-              disabled={!calculation || isMutating}
+              disabled={!calculation || isFieldSaving('salaryOrHourly')}
```
wjames111: [High] Pay Type's manual `TextField` writes a multi-field payload `{ salaryOrHourly, payRate: null }` (line 222) and reshapes the form. Narrowing its in-flight guard to `isFieldSaving('salaryOrHourly')` only protects one of the two fields the mutation touches, and leaves the cross-field race described on `usePdsGoalAutoSave.ts:32`.

For this specific load-bearing select, consider keeping the broad guard:
```ts
const { calculation, hcmUser, isMutating, isFieldSaving, setRightPanelContent } =
  usePdsGoalCalculator();
...
disabled={!calculation || isMutating}
```
`isMutating` is still exposed and still has a consumer (`pages/.../pdsGoalCalculator/[pdsGoalId].page.tsx` for `SavingStatus`), so this requires no extra plumbing.

src/components/HrTools/PdsGoalCalculator/Shared/PdsGoalCalculatorContext.tsx
```diff
@@ -39,8 +39,19 @@ export type PdsGoalCalculatorType = {
 
   /** Whether any mutations are currently in progress */
   isMutating: boolean;
+  /** Whether a save mutation tagged with the given field name is in flight */
+  isFieldSaving: (fieldName: string) => boolean;
```
wjames111: [Important] **Loose `string` typing erases the schema constraint that exists upstream** (Architecture 7.0 + Standards pattern deviation).

The new API accepts arbitrary strings:
- `isFieldSaving: (fieldName: string) => boolean` (this line)
- `trackFieldMutation: <T>(mutation: Promise<T>, fields: string[]) => Promise<T>` (line 51-54)

But callers actually have `keyof DesignationSupportCalculationUpdateInput` available (`usePdsGoalAutoSave.ts` already enforces it). `SetupStep.tsx:217` passes the bare literal `'salaryOrHourly'` — a typo or rename would compile cleanly and silently leave the Pay Type select un-disabled.

**Suggested patch:**
```ts
import { DesignationSupportCalculationUpdateInput } from 'src/graphql/types.generated';
...
isFieldSaving: (
  fieldName: keyof DesignationSupportCalculationUpdateInput,
) => boolean;
...
trackFieldMutation: <T>(
  mutation: Promise<T>,
  fields: Array<keyof DesignationSupportCalculationUpdateInput>,
) => Promise<T>;
```
And in `useSaveField.ts:48`, isolate the `Object.keys` widening with one cast:
```ts
Object.keys(attributes) as Array<keyof DesignationSupportCalculationUpdateInput>,
```

src/components/HrTools/PdsGoalCalculator/Shared/Autosave/usePdsGoalAutoSave.test.tsx
```diff
@@ -167,4 +167,38 @@ describe('usePdsGoalAutoSave', () => {
...
+    act(() => {
+      result.current.name.onBlur();
+    });
+
+    // The unrelated formType select must NOT flicker disabled while `name`
+    // is saving — only same-field saves should disable a select.
+    await waitFor(() =>
+      expect(mutationSpy).toHaveGraphqlOperation('UpdatePdsGoalCalculation'),
+    );
+    expect(result.current.formType.disabled).toBe(false);
```
wjames111, line 242: [Important] **This test does not actually catch the regression it claims to.**

The final assertion runs *after* `await waitFor(() => mutationSpy...)`. By that point, the mock mutation has resolved and `savingFieldCounts.name` is back to 0. The assertion passes both pre-fix (broad `isMutating`) and post-fix.

Compare with the working pattern at lines 124-145, which polls for `true` *then* `false` to catch the transition.

**Suggested patch** — sample synchronously mid-save:
```diff
     act(() => {
       result.current.name.onBlur();
     });

+    // Sample synchronously mid-save: savingFieldCounts.name === 1 here, but
+    // the mock mutation has not yet resolved. Pre-fix (gating on isMutating),
+    // formType.disabled would have been true at this moment.
+    expect(result.current.formType.disabled).toBe(false);
+
     await waitFor(() =>
       expect(mutationSpy).toHaveGraphqlOperation('UpdatePdsGoalCalculation'),
     );
     expect(result.current.formType.disabled).toBe(false);
```

src/components/HrTools/PdsGoalCalculator/Shared/PdsGoalCalculatorContext.tsx
```diff
@@ -108,6 +119,44 @@ export const PdsGoalCalculatorProvider: React.FC<Props> = ({ children }) => {
...
+                next[field] = remaining;
+              }
+            }
+            return next;
+          });
+        }),
+      );
+    },
+    [trackMutation],
+  );
```
wjames111, line 164: [Medium] **No test for multi-field atomic save.** The Pay Type path passes `['salaryOrHourly', 'payRate']` into `trackFieldMutation` (`useSaveField.ts:48`). This is the entire reason `fields` is an array, but no test verifies that BOTH fields are marked saving during a multi-field write. Consider a direct context test that calls `trackFieldMutation` with two fields and asserts `isFieldSaving` returns `true` for both.

src/components/HrTools/PdsGoalCalculator/Shared/PdsGoalCalculatorContext.tsx
```diff
@@ -108,6 +119,44 @@ export const PdsGoalCalculatorProvider: React.FC<Props> = ({ children }) => {
...
+    <T,>(mutation: Promise<T>, fields: string[]): Promise<T> => {
+      setSavingFieldCounts((prev) => {
+        const next = { ...prev };
+        for (const field of fields) {
+          next[field] = (next[field] ?? 0) + 1;
+        }
+        return next;
+      });
+      return trackMutation(
+        mutation.finally(() => {
```
wjames111, line 147: [Medium] **No test for the failure path.** `mutation.finally(...)` correctly decrements on rejection in principle, but no test exercises that branch. A future refactor moving the decrement into `.then(...)` would leak counts (the field stays disabled forever) and no test would catch it. A test using a rejecting Apollo mock and asserting `isFieldSaving(field) === false` after rejection would close the gap.

src/components/HrTools/PdsGoalCalculator/Shared/Autosave/usePdsGoalAutoSave.ts
```diff
@@ -15,17 +15,20 @@ export const usePdsGoalAutoSave = ({
...
-    // Block change-driven (select) autosaves while a save is in flight: rapid
-    // back-and-forth toggles can otherwise land out of order in the Apollo
-    // cache. formType is the load-bearing case — its value reshapes the goal
-    // calculation, so a stale final value silently understates the total.
-    disabled: !calculation || (options.saveOnChange === true && isMutating),
+    // Block change-driven (select) autosaves while a save for this same field
+    // is in flight: rapid back-and-forth toggles can otherwise land out of
+    // order in the Apollo cache. formType is the load-bearing case — its
+    // value reshapes the goal calculation, so a stale final value silently
+    // understates the total. Other fields' saves do not affect this one.
```
wjames111: [Suggestion] The comment says "Other fields' saves do not affect this one," but as noted on line 32, that's true for the UI disable behavior but **not** for cache state — a concurrent save's full-fragment response can briefly stamp the cache and momentarily un-do this field's optimistic write. If the team accepts that trade-off, please reword the comment to acknowledge it. If the team adopts the fix on line 32, the whole comment block can be tightened.

wjames111 (commented):
## Multi-Agent Code Review — VERDICT: BLOCKERS_FOUND

**6 agents** · Architecture · Data Integrity · Testing · UX · Standards · Financial Reporting

### Top blocker
**Cross-field stale-write race** (consensus from Data Integrity, UX, Financial Reporting). The PR narrows the in-flight save guard from "any mutation in flight" to "same field in flight," but `UpdatePdsGoalCalculation` responses carry the **full** `PdsGoalCalculationFields` fragment. A second save's response can land after a first save's, stamping stale field values (notably `formType`) into the Apollo cache — the exact silent goal-total miscalculation the in-line comment warns about. See inline comment on `usePdsGoalAutoSave.ts`.

### Risk assessment
- Risk: **5/10 — MEDIUM**
- Reviewer level recommended: **mid-level / senior**

### Findings summary
| Tier | Count |
|---|---|
| Critical (9-10) | 0 |
| High Priority Blocker (8-8.9) | 1 |
| Important (7-7.9) | 2 |
| Medium (5-6.9) | 2 |
| Suggestions (<5) | ~10 |

### Verdict reasoning
The per-field disable guard is correctly motivated for rapid same-field toggling, but is insufficient because (a) the server response carries the full fragment, so cross-field responses can interleave and stomp the cache, and (b) the Pay Type → `payRate: null` multi-field atomic save isn't protected on the Pay Rate text field (which has `saveOnChange=false`). Either widen the guard back for `formType`/`salaryOrHourly`, or drop the `saveOnChange === true` qualifier so blur-driven fields also disable during their own save.

---

### Findings on Related / Pre-existing Code (Not in This PR's Diff)

**[Suggestion] `SetupStep.tsx:316-332` (Geographic Multiplier `Autocomplete`)**
- Severity: 4.0/10
- Flagged by: UX, Data Integrity
- Note: `disabled={!calculation}` only — no in-flight protection of any kind. Same race class as the blocker above. **Pre-existing** (not introduced by this PR). Worth a follow-up ticket, not a blocker here.

---

### Standards & Quality summary
- Named exports: PASS
- TypeScript hygiene: PASS (no `any`, no `@ts-ignore`, no `!`)
- i18n: PASS (no new strings)
- Lint/TS clean for touched files: PASS
- Test colocation + `GqlMockedProvider` typed generics: PASS
- `isMutating` is NOT dead code — page-level `SavingStatus` still consumes it via `pages/.../pdsGoalCalculator/[pdsGoalId].page.tsx`.

---

_To resolve_: address the blocker comment on `usePdsGoalAutoSave.ts`, then either fix or reply `/dismiss: <reason>` to severity < 7 findings you disagree with. Severity 7.0-7.9 findings cannot be dismissed but don't trigger BLOCKERS_FOUND on their own.

wjames111 (commented):
## Multi-Agent Code Review — Standard Mode

**Verdict:** APPROVED WITH SUGGESTIONS
**Risk:** 5/10 (MEDIUM)
**Agents:** 5 specialized reviewers (Architecture, Data Integrity, Testing, UX, Standards)
**Diff:** 7 files, +214/-9

### Headline

The per-field saving state machine is correct under concurrency, handles resolve and reject paths, and is well-tested. No blocking issues in PR files. **The most consequential finding is a related-file finding (not in this PR):** the PR description names Pay Type as one of three selects being fixed, but the Pay Type select still binds to the global `isMutating` flag — see the "Findings on Related Files" section below.

### Summary of in-PR Findings

| Tier | Count | Notes |
|---|---|---|
| Critical (9-10) | 0 | — |
| High (8-9) | 0 | — |
| Important (7-8) | 0 in-PR (1 related-file, informational) | Cannot be dismissed if any |
| Medium (5-7) | 7 | Posted as line comments below |
| Suggestions (<5) | ~10 | Informational |

All Standards-checklist items PASS. `yarn lint` + `yarn lint:ts` confirmed clean. Apollo cache normalization preserved (`id` + `__typename` in optimistic response).

### Cross-cutting consistency check

The new `trackFieldMutation` is correctly invoked by the single consumer (`useSaveField`) with `Object.keys(attributes)`, so multi-field atomic writes (e.g. `salaryOrHourly + payRate: null`) correctly tag both fields as saving for the lifetime of the single mutation. `HoursPerWeekGrid` still uses `trackMutation` (correctly — it's a multi-step hours workflow, not a per-field save).

---

These were identified during review on files related to the changes but not directly modified in this PR. They cannot be posted as line comments.

**[Important] `src/components/HrTools/PdsGoalCalculator/Setup/SetupStep.tsx:217`** — Pay Type select still uses global `isMutating`
- Severity: 7.5/10
- Flagged by: Architecture + UX (consensus)
- Problem: The PR description states the fix removes the disabled flicker for *"Form Type, Employment Status, and Pay Type selects"* in the Setup step. Form Type and Employment Status route through `AutosaveTextField → usePdsGoalAutoSave`, which the diff updates correctly. But Pay Type is a manual `<TextField>` (it must atomically clear `payRate`), and at `SetupStep.tsx:217` it still reads `disabled={!calculation || isMutating}`. After this PR merges, Pay Type will still flicker disabled while Benefits / Goal Name / Pay Rate is being saved — exactly the behavior the PR claims to remove.
- Recommended fix (single-line change you can land on this branch as a follow-up commit):
```tsx
// SetupStep.tsx:41 — pull isFieldSaving instead of isMutating
const { calculation, hcmUser, isFieldSaving, setRightPanelContent } =
  usePdsGoalCalculator();

// SetupStep.tsx:217
disabled={
  !calculation ||
  isFieldSaving('salaryOrHourly') ||
  isFieldSaving('payRate')
}
```
`useSaveField` already passes `Object.keys(attributes)`, so the multi-field atomic write tags both fields and only Pay Type + Pay Rate will disable during their own save.

---

### Pre-existing notes (informational)

- `useSaveField.ts:52` `catch { ... enqueueSnackbar ... }` drops the error object entirely (no `console.error`, no Sentry). Server validation messages are invisible. Pre-existing pattern; the PR rewrites lines around it.
- Form-wide: `disabled` is used as a "saving in progress" signal where `aria-busy` would be more semantically correct. Pattern is repo-wide; the codebase already uses `aria-busy` elsewhere (`CsvHeaders.tsx:337`, `CsvPreview.tsx:346`).

src/components/HrTools/PdsGoalCalculator/Shared/Autosave/usePdsGoalAutoSave.ts
```diff
@@ -26,6 +26,6 @@ export const usePdsGoalAutoSave = ({
     // back-and-forth toggles can otherwise land out of order in the Apollo
     // cache. formType is the load-bearing case — its value reshapes the goal
     // calculation, so a stale final value silently understates the total.
-    disabled: !calculation || (options.saveOnChange === true && isMutating),
+    disabled: !calculation || isFieldSaving(fieldName),
```
wjames111: [Medium] **Comment above is now stale.** It says "Block change-driven (select) autosaves while a save is in flight" — that described the pre-PR condition `options.saveOnChange === true && isMutating`. The new gate `isFieldSaving(fieldName)` disables **both** change-driven and blur-driven fields whenever the field's own save is in flight (the new test at `usePdsGoalAutoSave.test.tsx:148` explicitly verifies this for the blur-driven case). Future readers will be misled.

Suggested rewrite:
```ts
// Disable a field while ITS OWN save is in flight (including multi-field
// atomic saves like salaryOrHourly + payRate). Prevents rapid back-and-forth
// writes from landing out of order in the Apollo cache while leaving
// unrelated fields interactive. formType is the load-bearing case — its value
// reshapes the goal calculation, so a stale final value silently understates
// the total.
disabled: !calculation || isFieldSaving(fieldName),
```

wjames111: [Medium] **Behavioral change beyond the PR description.** The previous test "does not disable blur-driven fields while a save is in flight" was an explicit guarantee that the new test replaces. The PR description only mentions select flicker — it doesn't flag that blur-driven text inputs (Goal Name, Benefits, Pay Rate when typed) will now also briefly disable for ~RTT after the user tabs out. There's no spinner and no `aria-busy`, so the disabled state semantically reads as "unavailable" rather than "saving."

Options:
- **(a) Restore previous semantic for blur-driven fields:** `disabled: !calculation || (options.saveOnChange === true && isFieldSaving(fieldName))` — keeps the select bug-fix while preserving the old blur UX.
- **(b) Keep the new gate but signal correctly:** add `aria-busy={fieldProps.disabled && !!calculation}` in `AutosaveTextField`. The repo already uses this pattern at `CsvHeaders.tsx:337`, `CsvPreview.tsx:346`, `CsvValues.tsx:263`.

If (b) is the choice, please expand the PR description to call out the new blur-disable behavior — otherwise reviewers will assume the previous semantic is preserved.

src/components/HrTools/PdsGoalCalculator/Shared/PdsGoalCalculatorContext.tsx
```diff
@@ -108,6 +120,48 @@ export const PdsGoalCalculatorProvider: React.FC<Props> = ({ children }) => {
   const [isDrawerOpen, setIsDrawerOpen] = useState<boolean>(true);
   const { trackMutation, isMutating } = useTrackMutation();
 
+  const [savingFieldCounts, setSavingFieldCounts] = useState<
+    Record<string, number>
+  >({});
```
wjames111, line 125: [Medium] **State type widens `keyof` to plain `string`** (3-agent consensus: Architecture 4.0, Standards 5.5, Data Integrity 3.0). The public API (`isFieldSaving`, `trackFieldMutation`) is typed `keyof DesignationSupportCalculationUpdateInput`, but the internal state accepts any string. A typo inside the provider would silently store a non-field key.

```ts
// before
const [savingFieldCounts, setSavingFieldCounts] = useState<
  Record<string, number>
>({});

// after
const [savingFieldCounts, setSavingFieldCounts] = useState<
  Partial<Record<keyof DesignationSupportCalculationUpdateInput, number>>
>({});
```

The reads at lines 129, 141, 150 already work with the tighter shape (all use indexed access + nullish coalescing).

src/components/HrTools/PdsGoalCalculator/Shared/PdsGoalCalculatorContext.test.tsx
```diff
@@ -153,6 +153,68 @@ describe('PdsGoalCalculatorContext', () => {
     expect(result.current.percentComplete).toBe(100);
   });
 
+  describe('trackFieldMutation', () => {
```
wjames111, line 156: [Medium] **Missing concurrent-same-field test.** The ref-counter `(savingFieldCounts[field] ?? 0) + 1` only matters when two mutations are in flight on the same field simultaneously. The two new tests cover ONE in-flight mutation at a time, so a future "simplification" to a boolean Set (or a typo like `< 0` instead of `<= 0`) would only break under concurrent same-field saves and these tests would still pass.

Suggested addition:
```ts
it('keeps a field saving while a second concurrent save on the same field is in flight', async () => {
  const { result } = renderUsePdsGoalCalculator();
  let resolve1!: (v: unknown) => void;
  let resolve2!: (v: unknown) => void;
  const p1 = new Promise((r) => { resolve1 = r; });
  const p2 = new Promise((r) => { resolve2 = r; });
  let t1!: Promise<unknown>;
  let t2!: Promise<unknown>;
  act(() => {
    t1 = result.current.trackFieldMutation(p1, ['payRate']);
    t2 = result.current.trackFieldMutation(p2, ['payRate']);
  });
  expect(result.current.isFieldSaving('payRate')).toBe(true);
  await act(async () => { resolve1(undefined); await t1; });
  expect(result.current.isFieldSaving('payRate')).toBe(true); // p2 still pending
  await act(async () => { resolve2(undefined); await t2; });
  expect(result.current.isFieldSaving('payRate')).toBe(false);
});
```

src/components/HrTools/PdsGoalCalculator/Shared/Autosave/useSaveField.ts
```diff
@@ -45,14 +45,23 @@ export const useSaveField = () => {
               },
             },
           }),
+          Object.keys(attributes) as Array<
+            keyof DesignationSupportCalculationUpdateInput
+          >,
         );
       } catch {
```
wjames111, line 52: [Medium] **Error snackbar path is untested** (pre-existing gap, but flagged because the PR rewrites lines 27-56 around it). No test (in this PR or in `useSaveField.test.tsx`) exercises the failure path. If the snackbar message changes or the `try/catch` is accidentally dropped during a future refactor, nothing fires.

Consider adding a `useSaveField.test.tsx` case that mocks `UpdatePdsGoalCalculation` to throw and asserts the snackbar appears.

Separately (pre-existing, do not address here): the `catch {}` drops the error object entirely. No `console.error`, no Sentry, no rethrow — server validation messages are invisible to dev tooling. Worth a follow-up ticket.

src/components/HrTools/PdsGoalCalculator/PdsGoalCalculatorTestWrapper.tsx
```diff
@@ -141,6 +141,8 @@ export interface PdsGoalCalculatorTestWrapperProps {
   userMock?: GetUserMock;
   constantsMock?: GoalCalculatorConstantsMock;
   supportRaisedMock?: number;
+  /** Extra GqlMockedProvider mocks merged on top of the defaults. */
+  extraMocks?: ComponentProps<typeof GqlMockedProvider>['mocks'];
```
wjames111: [Medium] **`extraMocks` is effectively typed `unknown`** (Testing + Standards consensus). `ComponentProps<typeof GqlMockedProvider>['mocks']` resolves to `DeepPartial<unknown>` because the default generic on `GqlMockedProvider` is `unknown`. The new test at `PdsGoalCalculator.test.tsx:81-97` supplies an `UpdatePdsGoalCalculation` shape with no compile-time check that `salaryOrHourly` is a valid field or that `DesignationSupportSalaryType.Salaried` is the right enum.

Repo convention (`.claude/rules/code-review.md` Testing Focus Areas): *"Component tests use `GqlMockedProvider<{ OperationName: OperationNameQuery }>` with typed generics so mock shapes are type-checked at compile time."*

Either narrow the type to a per-wrapper operation map:
```ts
extraMocks?: Partial<{
  UpdatePdsGoalCalculation: UpdatePdsGoalCalculationMutation;
  // ...other operations as needed
}>;
```
or thread a generic through the wrapper signature so callers can pass their own operation map at the call site.

src/components/HrTools/PdsGoalCalculator/PdsGoalCalculatorTestWrapper.tsx
```diff
@@ -264,6 +267,7 @@ export const PdsGoalCalculatorTestWrapper: React.FC<
                     Array.isArray(srcValue) ? srcValue : undefined,
                 ),
               },
+              ...extraMocks,
```
wjames111: [Medium] **Shallow spread, but JSDoc says "merged on top of the defaults"** (line 144). The `...extraMocks` spread is operation-level: if a caller passes `extraMocks={{ PdsGoalCalculation: { ... } }}`, it REPLACES the carefully-merged default at lines 244-263 entirely. Today no caller does this (the only use is `UpdatePdsGoalCalculation`, which isn't in defaults), but the first developer who tries to override a nested field of `PdsGoalCalculation` will get a confusing empty-default test failure.

Either deep-merge:
```tsx
mocks={merge({/* default operations */}, extraMocks)}
```
or fix the JSDoc to describe actual behavior:
```ts
/**
 * Extra GqlMockedProvider mocks. Entries shallow-replace the matching default
 * operation entirely; to override a nested field of an existing default, use
 * the typed prop (calculationMock, hcmUserMock, etc.) instead.
 */
```

src/components/HrTools/PdsGoalCalculator/PdsGoalCalculator.test.tsx
```diff
@@ -106,6 +123,7 @@ describe('PdsGoalCalculator', () => {
     const payRateInput = await findByRole('spinbutton', {
       name: 'Annual Pay Rate',
     });
+    await waitFor(() => expect(payRateInput).not.toBeDisabled());
```
wjames111, line 131: [Suggestion] Worth a one-line comment explaining why this wait is now required. Pay Type's atomic multi-field save (`salaryOrHourly + payRate: null`) now locks `payRate` until the mutation lands — without the new wait, `userEvent.type` would fire while `payRate` is disabled. A future cleanup might "simplify away" the wait without understanding it's load-bearing.

```ts
// Pay Type's atomic save (salaryOrHourly + payRate: null) locks payRate
// until the mutation resolves. Wait for it before typing.
await waitFor(() => expect(payRateInput).not.toBeDisabled());
```

src/components/HrTools/PdsGoalCalculator/Shared/Autosave/usePdsGoalAutoSave.test.tsx
```diff
@@ -161,10 +162,63 @@ describe('usePdsGoalAutoSave', () => {
       result.current.onBlur();
     });
 
-    expect(result.current.disabled).toBe(false);
+    await waitFor(() => expect(result.current.disabled).toBe(true));
```
wjames111: [Suggestion] **Timing-sensitive assertion.** `await waitFor(...toBe(true))` followed immediately by `await waitFor(...toBe(false))` depends on a microtask boundary existing between the save-start and the mock-resolve. If the mock ever resolves synchronously, the `true` window could collapse to zero and `waitFor` would only ever see `false`. Other tests in this file sample synchronously right after `act` (e.g. line 194 elsewhere in the suite), which is more robust:

```ts
act(() => result.current.onBlur());
expect(result.current.disabled).toBe(true); // sampled mid-save, sync
await waitFor(() => expect(result.current.disabled).toBe(false));
```

src/components/HrTools/PdsGoalCalculator/Shared/PdsGoalCalculatorContext.tsx
```diff
@@ -108,6 +120,48 @@ export const PdsGoalCalculatorProvider: React.FC<Props> = ({ children }) => {
...
+      mutation: Promise<T>,
+      fields: Array<keyof DesignationSupportCalculationUpdateInput>,
+    ): Promise<T> => {
+      setSavingFieldCounts((prev) => {
+        const next = { ...prev };
+        for (const field of fields) {
+          next[field] = (next[field] ?? 0) + 1;
+        }
+        return next;
+      });
```
wjames111, line 145: [Suggestion] The increment loop here and the decrement loop at lines 149-156 are symmetric, so duplicate field names in `fields` are handled correctly (n duplicates → +n then -n). Today the only caller passes `Object.keys(attributes)` which is unique by construction, so duplicates are unreachable. A one-line `const uniqueFields = Array.from(new Set(fields));` at the top would make the invariant bulletproof and self-documenting. Optional — math works either way.

PR #1803, opened 2026-05-26
MPDX-9632 Hide ASR balance and CAP when no request exists yet

wjames111:
## Description

- Hide the **Primary Account Balance** and **Your Maximum Allowable Salary (CAP)** amounts on the ASR About form when no `latestAdditionalSalaryRequest` exists yet. These values are sourced from the latest request's `calculations`, so before any ASR has been submitted they would render as a meaningless `$0.00`.
- The name + person-number header still renders; only the balance/CAP row is hidden via the existing `showContent` prop on `NameDisplay`.
- This change is scoped to `AboutForm` — `NewForm`, `EditForm`, and `ViewForm` always render with a populated ASR because `RequestPage.handleContinue()` creates one before transitioning to those screens.
- Jira: [MPDX-9632](https://jira.cru.org/browse/MPDX-9632)

## Testing

- Go to **HR Tools → Additional Salary Request** as a user who has never submitted an ASR (or whose only ASR is in a status outside `IN_PROGRESS / ACTION_REQUIRED / PENDING / APPROVED_NOT_PAID / APPROVED_AND_PAID`).
- On the **About this Form** step, confirm the name card still shows your preferred name and person number, and that the **Primary Account Balance** and **Your Maximum Allowable Salary (CAP)** row is no longer shown.
- As a user who *does* have a current ASR (any of the statuses above), confirm the same About step still shows the balance and CAP amounts as before.
- Continue past the About step to verify NewForm/EditForm/ViewForm are unaffected — balance and CAP continue to display there.

## Checklist:

### DIFF
--- a/src/components/HrTools/AdditionalSalaryRequest/AboutForm/AboutForm.test.tsx
+++ b/src/components/HrTools/AdditionalSalaryRequest/AboutForm/AboutForm.test.tsx
@@ -70,9 +70,7 @@ const TestWrapper: React.FC<TestWrapperProps> = ({ asrMocks }) => (
   <ThemeProvider theme={theme}>
     <TestRouter router={router}>
       <SnackbarProvider>
-        <GqlMockedProvider<AboutFormMocks>
-          mocks={{ ...hcmMocks, ...asrMocks }}
-        >
+        <GqlMockedProvider<AboutFormMocks> mocks={{ ...hcmMocks, ...asrMocks }}>
           <AdditionalSalaryRequestProvider>
             <AboutForm />
           </AdditionalSalaryRequestProvider>
--- a/src/components/HrTools/AdditionalSalaryRequest/AboutForm/AboutForm.test.tsx
+++ b/src/components/HrTools/AdditionalSalaryRequest/AboutForm/AboutForm.test.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { ThemeProvider } from '@mui/material/styles';
-import { render } from '@testing-library/react';
+import { render, waitFor } from '@testing-library/react';
 import { SnackbarProvider } from 'notistack';
 import TestRouter from '__tests__/util/TestRouter';
 import { GqlMockedProvider } from '__tests__/util/graphqlMocking';
@@ -64,13 +64,17 @@ type AboutFormMocks = {
 
 interface TestWrapperProps {
   asrMocks?: Partial<AboutFormMocks>;
+  onCall?: jest.Mock;
 }
 
-const TestWrapper: React.FC<TestWrapperProps> = ({ asrMocks }) => (
+const TestWrapper: React.FC<TestWrapperProps> = ({ asrMocks, onCall }) => (
   <ThemeProvider theme={theme}>
     <TestRouter router={router}>
       <SnackbarProvider>
-        <GqlMockedProvider<AboutFormMocks> mocks={{ ...hcmMocks, ...asrMocks }}>
+        <GqlMockedProvider<AboutFormMocks>
+          mocks={{ ...hcmMocks, ...asrMocks }}
+          onCall={onCall}
+        >
           <AdditionalSalaryRequestProvider>
             <AboutForm />
           </AdditionalSalaryRequestProvider>
@@ -113,16 +117,22 @@ describe('AboutForm', () => {
   });
 
   it('should hide financial data when no request has been created yet', async () => {
+    const mutationSpy = jest.fn();
     const { findByText, queryByText } = render(
       <TestWrapper
         asrMocks={{
           AdditionalSalaryRequest: { latestAdditionalSalaryRequest: null },
         }}
+        onCall={mutationSpy}
       />,
     );
 
-    // Wait for the query to resolve before asserting absence.
+    // The CardHeader name proves the form rendered; absence assertions below
+    // document that only the financial portion is hidden when no ASR exists.
     expect(await findByText('Doc, John')).toBeInTheDocument();
+    await waitFor(() =>
+      expect(mutationSpy).toHaveGraphqlOperation('AdditionalSalaryRequest'),
+    );
     expect(queryByText(/Primary Account Balance/i)).not.toBeInTheDocument();
     expect(
       queryByText(/Your Maximum Allowable Salary \(CAP\)/i),
--- a/src/components/HrTools/AdditionalSalaryRequest/AboutForm/AboutForm.test.tsx
+++ b/src/components/HrTools/AdditionalSalaryRequest/AboutForm/AboutForm.test.tsx
@@ -6,6 +6,7 @@ import TestRouter from '__tests__/util/TestRouter';
 import { GqlMockedProvider } from '__tests__/util/graphqlMocking';
 import theme from 'src/theme';
 import { HcmQuery } from '../../Shared/HcmData/Hcm.generated';
+import { AdditionalSalaryRequestQuery } from '../AdditionalSalaryRequest.generated';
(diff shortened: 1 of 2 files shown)


src/components/HrTools/AdditionalSalaryRequest/AboutForm/AboutForm.tsx
```diff
@@ -15,9 +15,8 @@ export const AboutForm: React.FC = () => {
   const theme = useTheme();
 
   const { name, accountNumber, primaryAccountBalance } = useFormUserInfo();
-  const individualCap =
-    requestData?.latestAdditionalSalaryRequest?.calculations.currentSalaryCap ??
-    0;
+  const latestRequest = requestData?.latestAdditionalSalaryRequest;
```
wjames111, line 18: [Suggestion] **Style / minor coupling.** Two optional tweaks at this site:

1. Consider extracting the visibility predicate as a named local for readability: `const hasRequest = !!latestRequest;` and then `showContent={hasRequest}` below.
2. Note that `primaryAccountBalance` from `useFormUserInfo()` is derived from the same `latestAdditionalSalaryRequest?.calculations?.staffAccountBalance` chain — when `latestRequest` is null, both `amountOne` and `amountTwo` props passed to `NameDisplay` are already `0`. They're effectively dead values whenever `showContent` is false. Harmless but a minor coupling smell; current code is clear enough to keep as-is.

Flagged by: Architecture Agent (Sev 3).

src/components/HrTools/AdditionalSalaryRequest/AboutForm/AboutForm.tsx
```diff
@@ -96,7 +95,7 @@ export const AboutForm: React.FC = () => {
         titleTwo={t('Your Maximum Allowable Salary (CAP)')}
         amountTwo={individualCap}
         spouseComponent={<SpouseComponent />}
-        showContent
+        showContent={!!latestRequest}
```
wjames111, line 98: [Suggestion] **Three observations on `showContent={!!latestRequest}`** — all optional, none blocking:

1. **UX — empty-state messaging:** when `!!latestRequest` is false, the `CardHeader` (name + person #) still renders but `CardContent` is silently empty. A first-time user may wonder where their balance is. Consider a follow-up ticket for a short helper line (e.g., *"Your Primary Account Balance and CAP will appear here after you create a request."*). The surrounding instructional copy probably covers this in practice. *(UX Agent, Sev ~3-4)*

2. **Defense-in-depth:** `calculations` is non-null in the schema (`schema.graphql:884`), so `latestRequest?.calculations.currentSalaryCap` is safe. You *could* tighten the gate to `!!latestRequest?.calculations` in case the REST proxy ever drifts and returns a partial object, but the schema contract is the right place to trust. Recommend leaving as-is. *(Financial Reporting Agent, Sev 3)*

3. **`showContent` semantics:** this is the first dynamic `showContent` consumer in the codebase — all other callers (`NewForm`, `EditForm`, `ViewForm`, `NewSalaryCalculatorLanding`) pass it as a static literal. The de facto contract is now "do we have data to render". Forward-looking note only, not actionable here. *(Architecture Agent, Sev 2)*

src/components/HrTools/AdditionalSalaryRequest/AboutForm/AboutForm.test.tsx
```diff
@@ -86,17 +98,39 @@ describe('AboutForm', () => {
...
+    const { findByText, queryByText } = render(
+      <TestWrapper
+        asrMocks={{
+          AdditionalSalaryRequest: { latestAdditionalSalaryRequest: null },
+        }}
+      />,
+    );
+
+    // Wait for the query to resolve before asserting absence.
+    expect(await findByText('Doc, John')).toBeInTheDocument();
```
wjames111, line 132: [Suggestion] **Robustness of the absence assertion.** This `findByText('Doc, John')` comes from the `Hcm` query, but the values being asserted absent (Primary Account Balance, CAP) are derived from the independent `AdditionalSalaryRequest` query. The test works in practice because `GqlMockedProvider` resolves both queries on the same tick — but the wait doesn't strictly prove the ASR query resolved to null vs. "not yet loaded." Optional improvements:

- Use `mutationSpy` + `toHaveGraphqlOperation('AdditionalSalaryRequest', ...)` to confirm the ASR query was issued and settled.
- Add an explicit `findByText` assertion for the `CardHeader` (name) being visible — documents that only the financial portion is intentionally hidden.

Flagged by: Architecture + Testing Agents (cross-agreement, Sev 3-4).

wjames111 (commented):
## AI Multi-Agent Code Review — PR #1803

**Verdict:** CLEAN — No findings at severity ≥ 5.
**Risk:** LOW (1/10)
**Agents:** 5 (Architecture, Testing, Standards, UX, Financial Reporting)

This change correctly fixes the empty-state display bug: previously, when no ASR existed, the user saw `$0.00` for Primary Account Balance and CAP — exactly the "misleading zero" anti-pattern the project's Financial Reporting rules flag. The fix gates the `NameDisplay` CardContent on `!!latestRequest`, hiding both amounts when no request has been created yet.

### Cross-Agent Cross-Checks
- **Flash-of-empty-state concern** was investigated and refuted — `AdditionalSalaryRequest.tsx:64-82` already gates the parent on `requestLoading`/`requestError`, so `AboutForm` only mounts after the query resolves.
- **Schema check:** `calculations` is non-null on `AdditionalSalaryRequest` (schema.graphql:884), so `latestRequest?.calculations.currentSalaryCap` is safe.

### Summary

| Agent | Critical | High | Important | Suggestions |
|---|---|---|---|---|
| Architecture | 0 | 0 | 0 | 4 |
| Testing | 0 | 0 | 0 | 2 |
| Standards | 0 | 0 | 0 | 0 |
| UX | 0 | 0 | 0 | 3 |
| Financial Reporting | 0 | 0 | 0 | 2 |
| **Total (deduped)** | **0** | **0** | **0** | **7** |

### Standards Checklist

All applicable items PASS. Named exports, no `any`, typed `GqlMockedProvider<AboutFormMocks>`, `findBy*` for async, no `new Date()`, no debug output, GraphQL `id` fields preserved.

### Financial Checklist

- Rounding at display boundary only: Yes (via `currencyFormat` in `NameDisplay`)
- Null/undefined amounts handled: Yes — this PR is the fix
- Server-provided aggregations preferred: Yes

All suggestions are informational and do not block merge. See inline comments.

src/components/HrTools/AdditionalSalaryRequest/AboutForm/AboutForm.test.tsx
```diff
@@ -86,17 +98,39 @@ describe('AboutForm', () => {
...
+  it('should hide financial data when no request has been created yet', async () => {
+    const { findByText, queryByText } = render(
+      <TestWrapper
+        asrMocks={{
+          AdditionalSalaryRequest: { latestAdditionalSalaryRequest: null },
+        }}
+      />,
+    );
+
+    // Wait for the query to resolve before asserting absence.
```
wjames111: [Suggestion] **Nit on the comment.** The comment says *"Wait for the query to resolve"* — the awaited text actually comes from the `Hcm` query, not the `AdditionalSalaryRequest` query under test. Either drop the comment or clarify it (e.g., *"Wait for initial data so the component has rendered before checking absence."*). Minor — keep if you prefer the guard-rail intent, since the assertion's purpose is non-obvious without it. *(Standards / Testing Agent, Sev 2)*

src/components/HrTools/AdditionalSalaryRequest/AboutForm/AboutForm.tsx
wjames111: /dismiss

PR #1804, opened 2026-05-26
MPDX-9613 Gate MPD and PDS Goal Calculators by peopleGroupSupportType

wjames111:
## Description

- Restrict the MPD Goal Calculator to US Staff whose `peopleGroupSupportType` is `SupportedRmo` (senior staff).
- Restrict the PDS Goal Calculator to US Staff whose `peopleGroupSupportType` is `Designation`.
- Add `MpdGoalCalc` and `PdsGoalCalc` values to `RequiredUserGroupEnum` and extend `useUsStaffGroups` with matching `inMpdGoalCalcIneligibleGroup` / `inPdsGoalCalcIneligibleGroup` flags.
- Wrap both Goal Calculator pages in `UserTypeAccess` so ineligible users see the "Access to this feature is limited." page.
- Hide each Goal Calculator entry from the HR Tools nav for ineligible users.
- Eligibility for Goal Calculators is evaluated against the logged-in user (not spouse), consistent with the Salary Calculator check.

Jira: [MPDX-9613](https://jira.cru.org/browse/MPDX-9613)

## Testing

- Go to **HR Tools** as a US Staff user whose `peopleGroupSupportType` is `SupportedRmo`
- Check that the **MPD Goal Calculator** nav item is visible and the page loads normally
- Check that the **Paid with Designation Support Goal Calculator** nav item is hidden, and visiting its URL directly shows the limited-access page
- Sign in as a US Staff user whose `peopleGroupSupportType` is `Designation`
- Check that the **Paid with Designation Support Goal Calculator** nav item is visible and the page loads normally
- Check that the **MPD Goal Calculator** nav item is hidden, and visiting its URL directly shows the limited-access page
- Sign in as a Non-Cru user
- Check that both Goal Calculator pages show the limited-access page

## Checklist:

### DIFF
--- a/pages/accountLists/[accountListId]/hrTools/goalCalculator/[goalCalculationId].page.tsx
+++ b/pages/accountLists/[accountListId]/hrTools/goalCalculator/[goalCalculationId].page.tsx
@@ -23,7 +23,10 @@ import {
   MultiPageMenu,
   NavTypeEnum,
 } from 'src/components/Shared/MultiPageLayout/MultiPageMenu/MultiPageMenu';
-import { UserTypeAccess } from 'src/components/Shared/UserTypeAccess/UserTypeAccess';
+import {
+  RequiredUserGroupEnum,
+  UserTypeAccess,
+} from 'src/components/Shared/UserTypeAccess/UserTypeAccess';
 import { ReportPageWrapper } from 'src/components/Shared/styledComponents/ReportPageWrapper';
 import { useAccountListId } from 'src/hooks/useAccountListId';
 import { getAppName } from 'src/lib/getAppName';
@@ -139,7 +142,10 @@ export const GoalCalculatorPage: React.FC = () => {
         <title>{`${appName} | ${t('Reports - Goal Calculation')}`}</title>
       </Head>
       {accountListId ? (
-        <UserTypeAccess>
+        <UserTypeAccess
+          requireStaffAccount
+          requireUserGroups={RequiredUserGroupEnum.MpdGoalCalc}
+        >
           <ReportPageWrapper>
             <GoalCalculatorProvider>
               <GoalCalculatorContent
--- a/pages/accountLists/[accountListId]/hrTools/goalCalculator/[goalCalculationId].page.tsx
+++ b/pages/accountLists/[accountListId]/hrTools/goalCalculator/[goalCalculationId].page.tsx
@@ -142,10 +142,7 @@ export const GoalCalculatorPage: React.FC = () => {
         <title>{`${appName} | ${t('Reports - Goal Calculation')}`}</title>
       </Head>
       {accountListId ? (
-        <UserTypeAccess
-          requireStaffAccount
-          requireUserGroups={RequiredUserGroupEnum.MpdGoalCalc}
-        >
+        <UserTypeAccess requireUserGroups={RequiredUserGroupEnum.MpdGoalCalc}>
           <ReportPageWrapper>
             <GoalCalculatorProvider>
               <GoalCalculatorContent
--- a/pages/accountLists/[accountListId]/hrTools/goalCalculator/index.page.tsx
+++ b/pages/accountLists/[accountListId]/hrTools/goalCalculator/index.page.tsx
@@ -14,7 +14,10 @@ import {
   MultiPageMenu,
   NavTypeEnum,
 } from 'src/components/Shared/MultiPageLayout/MultiPageMenu/MultiPageMenu';
-import { UserTypeAccess } from 'src/components/Shared/UserTypeAccess/UserTypeAccess';
+import {
+  RequiredUserGroupEnum,
+  UserTypeAccess,
+} from 'src/components/Shared/UserTypeAccess/UserTypeAccess';
 import { ReportPageWrapper } from 'src/components/Shared/styledComponents/ReportPageWrapper';
 import { useAccountListId } from 'src/hooks/useAccountListId';
 import { getAppName } from 'src/lib/getAppName';
@@ -36,7 +39,7 @@ export const GoalCalculatorPage: React.FC = () => {
         <title>{`${appName} | ${t('HR Tools | MPD Goal Calculator')}`}</title>
       </Head>
       {accountListId ? (
-        <UserTypeAccess>
+        <UserTypeAccess requireUserGroups={RequiredUserGroupEnum.MpdGoalCalc}>
           <ReportPageWrapper>
             <SidePanelsLayout
               isScrollBox={false}
--- a/pages/accountLists/[accountListId]/hrTools/goalCalculator/index.page.tsx
+++ b/pages/accountLists/[accountListId]/hrTools/goalCalculator/index.page.tsx
@@ -39,7 +39,10 @@ export const GoalCalculatorPage: React.FC = () => {
         <title>{`${appName} | ${t('HR Tools | MPD Goal Calculator')}`}</title>
       </Head>
       {accountListId ? (
-        <UserTypeAccess requireUserGroups={RequiredUserGroupEnum.MpdGoalCalc}>
+        <UserTypeAccess
+          requireStaffAccount
+          requireUserGroups={RequiredUserGroupEnum.MpdGoalCalc}
+        >
           <ReportPageWrapper>
             <SidePanelsLayout
               isScrollBox={false}
--- a/pages/accountLists/[accountListId]/hrTools/goalCalculator/index.page.tsx
+++ b/pages/accountLists/[accountListId]/hrTools/goalCalculator/index.page.tsx
@@ -39,10 +39,7 @@ export const GoalCalculatorPage: React.FC = () => {
         <title>{`${appName} | ${t('HR Tools | MPD Goal Calculator')}`}</title>
       </Head>
       {accountListId ? (
-        <UserTypeAccess
-          requireStaffAccount
-          requireUserGroups={RequiredUserGroupEnum.MpdGoalCalc}
-        >
+        <UserTypeAccess requireUserGroups={RequiredUserGroupEnum.MpdGoalCalc}>
           <ReportPageWrapper>
             <SidePanelsLayout
               isScrollBox={false}
--- a/pages/accountLists/[accountListId]/hrTools/pdsGoalCalculator/index.page.tsx
+++ b/pages/accountLists/[accountListId]/hrTools/pdsGoalCalculator/index.page.tsx
@@ -38,10 +38,7 @@ const PdsGoalCalculatorPage: React.FC = () => {
         <title>{`${appName} | ${t('HR Tools | Paid with Designation Support Goal Calculator')}`}</title>
       </Head>
       {accountListId ? (
-        <UserTypeAccess
-          requireStaffAccount
-          requireUserGroups={RequiredUserGroupEnum.PdsGoalCalc}
-        >
+        <UserTypeAccess requireUserGroups={RequiredUserGroupEnum.PdsGoalCalc}>
           <ReportPageWrapper>
             <SidePanelsLayout
               isScrollBox={false}
--- a/pages/accountLists/[accountListId]/hrTools/pdsGoalCalculator/index.page.tsx
+++ b/pages/accountLists/[accountListId]/hrTools/pdsGoalCalculator/index.page.tsx
@@ -14,6 +14,10 @@ import {
   MultiPageMenu,
   NavTypeEnum,
 } from 'src/components/Shared/MultiPageLayout/MultiPageMenu/MultiPageMenu';
+import {
+  RequiredUserGroupEnum,
+  UserTypeAccess,
+} from 'src/components/Shared/UserTypeAccess/UserTypeAccess';
 import { ReportPageWrapper } from 'src/components/Shared/styledComponents/ReportPageWrapper';
 import { useAccountListId } from 'src/hooks/useAccountListId';
 import { getAppName } from 'src/lib/getAppName';
@@ -34,33 +38,35 @@ const PdsGoalCalculatorPage: React.FC = () => {
         <title>{`${appName} | ${t('HR Tools | Paid with Designation Support Goal Calculator')}`}</title>
       </Head>
       {accountListId ? (
-        <ReportPageWrapper>
-          <SidePanelsLayout
-            isScrollBox={false}
-            leftPanel={
-              <MultiPageMenu
-                isOpen={isNavListOpen}
-                selectedId="pdsGoalCalculator"
-                onClose={handleNavListToggle}
-                navType={NavTypeEnum.HrTools}
-              />
-            }
-            leftOpen={isNavListOpen}
-            leftWidth="290px"
-            headerHeight={multiPageHeaderHeight}
-            mainContent={
-              <>
-                <MultiPageHeader
-                  isNavListOpen={isNavListOpen}
-                  onNavListToggle={handleNavListToggle}
-                  title={t('Paid with Designation Support Goal Calculator')}
-                  headerType={HeaderTypeEnum.HrTools}
+        <UserTypeAccess requireUserGroups={RequiredUserGroupEnum.PdsGoalCalc}>
+          <ReportPageWrapper>
+            <SidePanelsLayout
+              isScrollBox={false}
+              leftPanel={
+                <MultiPageMenu
+                  isOpen={isNavListOpen}
+                  selectedId="pdsGoalCalculator"
+                  onClose={handleNavListToggle}
+                  navType={NavTypeEnum.HrTools}
                 />
-                <PdsGoalsList />
-              </>
-            }
-          />
-        </ReportPageWrapper>
+              }
+              leftOpen={isNavListOpen}
+              leftWidth="290px"
+              headerHeight={multiPageHeaderHeight}
+              mainContent={
+                <>
+                  <MultiPageHeader
+                    isNavListOpen={isNavListOpen}
+                    onNavListToggle={handleNavListToggle}
+                    title={t('Paid with Designation Support Goal Calculator')}
+                    headerType={HeaderTypeEnum.HrTools}
+                  />
+                  <PdsGoalsList />
+                </>
+              }
+            />
+          </ReportPageWrapper>
+        </UserTypeAccess>
       ) : (
         <Loading loading />
       )}
--- a/pages/accountLists/[accountListId]/hrTools/pdsGoalCalculator/index.page.tsx
+++ b/pages/accountLists/[accountListId]/hrTools/pdsGoalCalculator/index.page.tsx
@@ -38,7 +38,10 @@ const PdsGoalCalculatorPage: React.FC = () => {
         <title>{`${appName} | ${t('HR Tools | Paid with Designation Support Goal Calculator')}`}</title>
       </Head>
       {accountListId ? (
-        <UserTypeAccess requireUserGroups={RequiredUserGroupEnum.PdsGoalCalc}>
+        <UserTypeAccess
+          requireStaffAccount
+          requireUserGroups={RequiredUserGroupEnum.PdsGoalCalc}
+        >
           <ReportPageWrapper>
             <SidePanelsLayout
               isScrollBox={false}
--- a/src/components/Shared/MultiPageLayout/MultiPageMenu/MultiPageMenu.test.tsx
+++ b/src/components/Shared/MultiPageLayout/MultiPageMenu/MultiPageMenu.test.tsx
@@ -517,6 +517,7 @@ describe('MultiPageMenu', () => {
                     },
                     asrEit: { asrEligibility: true },
                     salaryRequestEligible: true,
+                    designationSupportCalculatorEligible: true,
                   },
                 ],
               },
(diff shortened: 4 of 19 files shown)


src/hooks/useUsStaffGroups.ts
```diff
@@ -54,23 +54,38 @@ export function useUsStaffGroups(skip?: boolean) {
       (graphQLError) => graphQLError.extensions?.code === 'NO_STAFF_ACCOUNT',
     ) ?? false;
 
+  const userSupportType = user?.staffInfo?.peopleGroupSupportType;
+  const userSupportTypeKnown = userSupportType !== undefined;
```
wjames111: [Important] **Null vs undefined asymmetry creates fail-open behavior.**

The `peopleGroupSupportType` field is `Maybe<PeopleGroupSupportTypeEnum>` (nullable). Two failure modes diverge:

- `userSupportType === undefined` (HCM not loaded yet, or `staffInfo` field absent) → `userSupportTypeKnown = false` → both `inMpdGoalCalcIneligibleGroup` and `inPdsGoalCalcIneligibleGroup` short-circuit to `hasNoStaffAccount`. If no `NO_STAFF_ACCOUNT` error, the user sees **BOTH** calculators — a fail-open + contradictory state (a person can't be both `SupportedRmo` AND `Designation`).
- `userSupportType === null` (server explicitly returned null) → `userSupportTypeKnown = true`, predicate evaluates `null !== SupportedRmo` → user is gated out of both calculators.

This asymmetry is also the root cause of the brief UI gate-bypass on direct URL navigation: during the HCM-loading window, the page mounts and renders content before the gate evaluates.

Flagged by: Security, Architecture, Testing, UX.

**Recommended fix** (verify intent with product before applying):
```ts
const userSupportTypeKnown = userSupportType != null;
```
Or drop `userSupportTypeKnown` entirely and treat unknown as ineligible (fail-closed):
```ts
const inMpdGoalCalcIneligibleGroup =
  hasNoStaffAccount ||
  userSupportType !== PeopleGroupSupportTypeEnum.SupportedRmo;
```
Note: changing this is a deliberate behavior shift — the existing `defaults all to false when array is empty` test (line ~250-260) will need updating.

src/hooks/useUsStaffGroups.ts
```diff
@@ -54,23 +54,38 @@ export function useUsStaffGroups(skip?: boolean) {
...
     () => ({
       inAsrIneligibleGroup,
       inSalaryCalcIneligibleGroup,
       inMhaIneligibleGroup,
+      inMpdGoalCalcIneligibleGroup,
+      inPdsGoalCalcIneligibleGroup,
       hasNoStaffAccount,
       loading: loading && !data,
     }),
     [
```
wjames111, line 88: [Suggestion] `data` in the `useMemo` deps is redundant — all derived booleans already reflect `data` changes via their own computations, and `loading: loading && !data` covers the data transition. Harmless but noisy. Same pattern existed before this PR, so this is convention compliance, not a regression.

src/hooks/useHrToolsNavItems.ts
```diff
@@ -27,6 +29,7 @@ export function useHrToolsNavItems(): NavItems[] {
     {
       id: 'goalCalculator',
       title: t('MPD Goal Calculator'),
+      hideItem: inMpdGoalCalcIneligibleGroup,
```
wjames111: [Medium] **Nav-item flash for MPD Goal Calculator during HCM load.**

`useUsStaffGroups()` already exposes `loading`, but this hook discards it. While HCM is in flight, `inMpdGoalCalcIneligibleGroup` is `false` (per the fail-open at useUsStaffGroups.ts:58), so the item renders. When HCM resolves with `peopleGroupSupportType: Designation`, the item disappears — a visible content shift on every page mount that consumes the HR Tools nav.

Same issue exists for ASR/Salary/MHA, so this PR follows the established pattern — but it doubles the number of items affected by the flash.

Flagged by: UX, Architecture.

**Fix:**
```ts
const {
  ...,
  loading: hcmLoading,
} = useUsStaffGroups();
// ...
hideItem: hcmLoading || inMpdGoalCalcIneligibleGroup,
```
Add `hcmLoading` to the `useMemo` deps. Also addresses the brief security exposure window (UI fail-open during loading).

src/hooks/useHrToolsNavItems.ts
```diff
@@ -41,6 +44,7 @@ export function useHrToolsNavItems(): NavItems[] {
     {
       id: 'pdsGoalCalculator',
       title: t('Paid with Designation Support Goal Calculator'),
+      hideItem: inPdsGoalCalcIneligibleGroup,
```
wjames111: [Medium] **Same nav-item flash for PDS Goal Calculator.** See companion comment on line 32. Apply the same `hcmLoading || ...` treatment here:
```ts
hideItem: hcmLoading || inPdsGoalCalcIneligibleGroup,
```

src/hooks/useUsStaffGroups.test.tsx
```diff
@@ -271,9 +283,79 @@ describe('useUsStaffGroups', () => {
...
         inMhaIneligibleGroup: true,
+        inMpdGoalCalcIneligibleGroup: true,
+        inPdsGoalCalcIneligibleGroup: true,
         hasNoStaffAccount: true,
         loading: false,
       });
     });
   });
+
+  describe('Goal Calculator eligibility', () => {
```
wjames111: [Important] **Missing tests for `peopleGroupSupportType: null` and `undefined`.**

The new `Goal Calculator eligibility` describe block covers `SupportedRmo`, `Designation`, `SupportedNonRmo`, `None`, and a spouse case — but not the two nullish cases that drive the asymmetric fail-open behavior (see comment on useUsStaffGroups.ts:58):

- `peopleGroupSupportType: undefined` (field omitted) — currently treated as **eligible** for both calcs (fail-open)
- `peopleGroupSupportType: null` (explicit null from server) — currently treated as **ineligible** for both calcs (fail-closed)

Whichever contract the team intends, add tests to lock it in. UX recommends the test also assert the **rendered body copy** of `LimitedAccess` so the misleading "contact support@mpdx.org" message doesn't silently survive a refactor.

**Suggested additions:**
```tsx
it('marks user as ineligible for both goal calcs when peopleGroupSupportType is null', async () => {
  const { result } = renderUseUsStaffGroups(
    buildHcmMock({
      peopleGroupSupportType: null as unknown as PeopleGroupSupportTypeEnum,
    }),
  );
  await waitFor(() => {
    expect(result.current.inMpdGoalCalcIneligibleGroup).toBe(true);
    expect(result.current.inPdsGoalCalcIneligibleGroup).toBe(true);
  });
});

it('treats peopleGroupSupportType undefined as <chosen behavior>', async () => {
  const { result } = renderUseUsStaffGroups({
    hcm: [{ asrEit: { asrEligibility: true }, salaryRequestEligible: true, mhaEit: { mhaEligibility: true }, staffInfo: { /* peopleGroupSupportType intentionally omitted */ } }],
  } as HcmQuery);
  await waitFor(() => {
    // Pin to the chosen behavior
  });
});
```

Note: Do not write tests against the current fail-open behavior if you plan to also adopt the architectural fix on useUsStaffGroups.ts:58 — write tests against the **fixed** behavior.

src/components/Shared/UserTypeAccess/UserTypeAccess.test.tsx
```diff
@@ -128,6 +134,54 @@ describe('UserTypeAccess', () => {
     ).toBeInTheDocument();
   });
 
+  it('should render child component when user is in the MPD goal calculator group (SupportedRmo)', async () => {
+    const { findByText } = render(
+      <TestComponent
+        requireUserGroups={RequiredUserGroupEnum.MpdGoalCalc}
```
wjames111, line 146: [Medium] **The four new MpdGoalCalc/PdsGoalCalc LimitedAccess tests assert only the heading, not the body copy.**

Compare to the existing `requireStaffAccount` test at line ~185-198 which asserts both heading (`'Access to this feature is limited.'`) and body (`/our records show that you do not have a staff account/i`). Without the body assertion, a regression that swaps the underlying `getLimitedText.tsx` branch for these enums would silently pass.

**Add a `getByText` (or `findByText`) check** for the expected body copy on lines 140, 150, 164, and 174. This is the load-bearing assertion when the gate predicate evolves.

src/components/Shared/MultiPageLayout/MultiPageMenu/MultiPageMenu.test.tsx
```diff
@@ -520,4 +524,82 @@ describe('MultiPageMenu', () => {
       expect(getByText('Ministry Partner Reminders')).toBeInTheDocument();
     });
   });
+
+  it('hides the PDS Goal Calculator nav item for SupportedRmo (senior) staff', async () => {
```
wjames111: [Suggestion] **Tests assert one item hidden + one item visible, but not that the other unaffected HR Tools nav items (Salary Calculation Form, MHA Calculation Tool, Additional Salary Request, Savings Fund Transfer, Ministry Partner Reminders) remain present.**

A regression that nukes the entire HR Tools menu when HCM data is present would still pass these tests as long as the one expected item is rendered. Add `getByText('Salary Calculation Form')` (or one of the others) to both new test cases to lock in the menu shape.

pages/accountLists/[accountListId]/hrTools/pdsGoalCalculator/index.page.tsx
```diff
@@ -34,33 +38,35 @@ const PdsGoalCalculatorPage: React.FC = () => {
...
-            leftWidth="290px"
-            headerHeight={multiPageHeaderHeight}
-            mainContent={
-              <>
-                <MultiPageHeader
-                  isNavListOpen={isNavListOpen}
-                  onNavListToggle={handleNavListToggle}
-                  title={t('Paid with Designation Support Goal Calculator')}
-                  headerType={HeaderTypeEnum.HrTools}
+        <UserTypeAccess requireUserGroups={RequiredUserGroupEnum.PdsGoalCalc}>
```
Comment (wjames111) [inline pages/accountLists/[accountListId]/hrTools/pdsGoalCalculator/index.page.tsx:41]: [Medium] **This page newly routes through `UserTypeAccess`'s HCM-loading path — first paint will be a tiny 30px floating spinner instead of the calculator chrome.**

Before this PR: the page rendered `ReportPageWrapper`/`SidePanelsLayout`/`MultiPageHeader` immediately and only the inner `PdsGoalsList` data queries showed loading. After: while `useUsStaffGroups()` is fetching HCM, `UserTypeAccess` returns `<Loading loading />` (a 30px `CircularProgress` inside a `Fab` fixed at top:50%/left:50% per `src/components/Loading/Loading.tsx:64-73`), blanking the whole page. PDS direct-URL/bookmark users will see a blank screen with a small spinning dot until HCM resolves.

MPD page already had `<UserTypeAccess>` but without `requireUserGroups`, so HCM wasn't being queried and this loading state was effectively bypassed. This PR adds the HCM query for both pages.

**Recommendation** (out of scope but worth noting): render the layout shell (`ReportPageWrapper` + `SidePanelsLayout` + `MultiPageHeader`) and use an MUI `Skeleton` for the main content, instead of replacing the whole page with the floating spinner. This is a wider pattern issue (Salary/MHA/ASR all use the same loading path), so consider tackling it in `UserTypeAccess` itself.

wjames111 (commented):
## Multi-Agent Code Review

**Verdict:** APPROVED WITH SUGGESTIONS — no merge blockers. Several Important and Medium findings recommended for follow-up.

**Risk:** HIGH (7/10) — modifies a shared gating component (`UserTypeAccess`, 15 dependents) and a hook consumed by multiple HR Tools pages.

**Agents:** Security · Architecture · Testing · UX · Standards (debated, with rebuttals)

---

### Top 3 Findings

1. **PDS detail page is not gated at all** (severity 7.5, files NOT in this PR diff — see "Related Files" section)
2. **Null vs undefined `peopleGroupSupportType` creates fail-open / asymmetric behavior** (severity 7.0, useUsStaffGroups.ts:58)
3. **MPD/PDS goal-calc nav items flash visible during HCM load** (severity 6.5, useHrToolsNavItems.ts)

These are not blockers — the PR's stated objective (gating list pages by support type) is correctly implemented and tested. Findings are improvements to make the gating consistent and robust.

---

### Cross-Cutting Consistency Check — Detail-Page Gating Gap

The PR title commits to gating MPD and PDS Goal Calculators by `peopleGroupSupportType`. Verified the gate is applied at the **list** page level only:

| Page | Auth | Support-type gate |
|------|------|------------------|
| `goalCalculator/index.page.tsx` (list) | UsStaff | ✅ MpdGoalCalc |
| `pdsGoalCalculator/index.page.tsx` (list) | UsStaff | ✅ PdsGoalCalc |
| `goalCalculator/[goalCalculationId].page.tsx` (detail) | UsStaff (no `requireUserGroups`) | ❌ MISSING |
| `pdsGoalCalculator/[pdsGoalId].page.tsx` (detail) | ❌ NO `UserTypeAccess` wrapper | ❌ MISSING |

For comparison, `salaryCalculator/[calculationId]/index.page.tsx:62` and `mhaCalculator/[requestId].page.tsx:69` both pass `requireUserGroups={...}` to their detail page wrappers — that pattern is established in this repo. A user with an ineligible `peopleGroupSupportType` who has a valid `goalCalculationId` / `pdsGoalId` (bookmark, shared link, browser history) can deep-link directly to the detail page and bypass the new gate. Server-side authorization on the upstream MPDX API should be the ultimate defense, but UI parity with the existing pattern is the highest-leverage UI fix.

---

These files were identified as having gating-gap issues directly caused by this PR's design choices, but the files themselves are not modified by this PR. They cannot be posted as line comments.

**[Important] `pages/accountLists/[accountListId]/hrTools/pdsGoalCalculator/[pdsGoalId].page.tsx`** — No `UserTypeAccess` wrapper at all
- Severity: 7.5/10
- Flagged by: Architecture, Security
- Problem: The PDS detail page has no `UserTypeAccess` wrapper. After this PR ships, an ineligible US Staff user (or even a non-US-Staff user) can deep-link a `pdsGoalId` and reach the detail page directly.
- Recommended Action: Wrap with `<UserTypeAccess requireUserGroups={RequiredUserGroupEnum.PdsGoalCalc}>` to match the new list-page gating.

**[Medium] `pages/accountLists/[accountListId]/hrTools/goalCalculator/[goalCalculationId].page.tsx:142`** — `UserTypeAccess` is missing `requireUserGroups`
- Severity: 6.5/10
- Flagged by: Architecture, Security
- Problem: The MPD detail page wraps in `<UserTypeAccess>` with no `requireUserGroups` — so it only checks `userType === UsStaff`, not the support-type gate. Pre-existing inconsistency, but in scope to fix given this PR's stated goal.
- Recommended Action: Change to `<UserTypeAccess requireUserGroups={RequiredUserGroupEnum.MpdGoalCalc}>`.

---

### Other Notes (Not Inline)

**[Suggestion] Server-side enforcement of `peopleGroupSupportType` is unverified** (severity 6.5, Security)
- `UserTypeAccess` is a UI-only gate. Per `.claude/rules/code-review.md` Security Focus Areas: "every Yup rule, every disabled button, every required field must have a server-side equivalent." The upstream MPDX GraphQL API for `goalCalculations`/`designationSupportCalculations` and their mutations should enforce `peopleGroupSupportType` independently. Not visible in this repo — verify with the API team. Same posture already exists for ASR/Salary/MHA, so this is not a regression.

**[Suggestion] LimitedAccess copy is misleading when `peopleGroupSupportType` is null/None** (severity 4.5, UX, pre-existing)
- A US Staff user whose `peopleGroupSupportType` is genuinely `None` or `null` sees "Our records show that you are not part of the user group… contact support@mpdx.org to change your user group" (`src/components/Shared/LimitedAccess/getLimitedText.tsx:42-51`). The actionable fix is with HR/HCM data ownership, not MPDX support. Pre-existing copy.

**[Suggestion] Senior staff landing on PDS page has no signposting to MPD calc** (severity 5.5, UX)
- A `SupportedRmo` user navigating to PDS (or `Designation` user to MPD) hits a generic "limited access" wall with only a "Back to Dashboard" CTA. The nav menu correctly hides the wrong calculator, so they have a way forward — but the page itself doesn't signpost. Out of scope for this PR.

**[Pre-existing] `PdsGoalCalculatorPage` uses `export default`** (severity 2.0, Standards)
- Next.js Pages Router requires default export. This is a project-wide carve-out from CLAUDE.md's "named exports only." Not introduced by this PR.

---

### Standards Compliance — All Pass

- Named exports for new code, file naming, i18n via `t()`, generated GraphQL types used, typed `GqlMockedProvider`, `findBy*` for async assertions — all PASS.
- `yarn lint:ci` and `yarn lint:ts` clean on changed files.
- No `any`, no debug output, no `new Date()`, no commented code.

---

*To dismiss a non-blocking suggestion (severity < 7), reply `/dismiss: <reason>` on the comment.*

src/hooks/useUsStaffGroups.test.tsx
wjames111: Not sure what this should default to.

pages/accountLists/[accountListId]/hrTools/goalCalculator/index.page.tsx
```diff
@@ -36,7 +39,7 @@ export const GoalCalculatorPage: React.FC = () => {
         <title>{`${appName} | ${t('HR Tools | MPD Goal Calculator')}`}</title>
       </Head>
       {accountListId ? (
-        <UserTypeAccess>
+        <UserTypeAccess requireUserGroups={RequiredUserGroupEnum.MpdGoalCalc}>
```
Comment (kegrimes) [inline pages/accountLists/[accountListId]/hrTools/goalCalculator/index.page.tsx:42]: Do you know if MPD Goal calculator requires a staff account? If so, we would need to add the `requireStaffAccount` flag!

pages/accountLists/[accountListId]/hrTools/pdsGoalCalculator/index.page.tsx
Comment (kegrimes) [inline pages/accountLists/[accountListId]/hrTools/pdsGoalCalculator/index.page.tsx:41]: Same question! Do we need a staff account?

pages/accountLists/[accountListId]/hrTools/goalCalculator/index.page.tsx
```diff
@@ -36,7 +39,7 @@ export const GoalCalculatorPage: React.FC = () => {
...
           <ReportPageWrapper>
             <SidePanelsLayout
               isScrollBox={false}
```
Comment (kegrimes) [inline pages/accountLists/[accountListId]/hrTools/goalCalculator/index.page.tsx:45]: I think you would also need to add the `UserTypeAccess` component to the `[goalCalculationId].page.tsx`. It is very unlikely that a user could navigate to that page since they would need to know the id, but what if it was bookmarked before their user group changed? In that case, I think they would be able to navigate to the page.

pages/accountLists/[accountListId]/hrTools/pdsGoalCalculator/index.page.tsx
```diff
@@ -34,33 +38,35 @@ const PdsGoalCalculatorPage: React.FC = () => {
...
+                    onNavListToggle={handleNavListToggle}
+                    title={t('Paid with Designation Support Goal Calculator')}
+                    headerType={HeaderTypeEnum.HrTools}
+                  />
+                  <PdsGoalsList />
+                </>
+              }
+            />
+          </ReportPageWrapper>
+        </UserTypeAccess>
```
Comment (kegrimes) [inline pages/accountLists/[accountListId]/hrTools/pdsGoalCalculator/index.page.tsx:69]: Same thing with PDS! I think we may need to add `UserTypeAccess` to `[pdsGoalId].page.test.tsx`

src/components/Shared/UserTypeAccess/UserTypeAccess.tsx
```diff
@@ -35,26 +37,30 @@ export const UserTypeAccess: React.FC<UserTypeAccessProps> = ({
...
   const {
     inAsrIneligibleGroup,
     inSalaryCalcIneligibleGroup,
     inMhaIneligibleGroup,
+    inMpdGoalCalcIneligibleGroup,
+    inPdsGoalCalcIneligibleGroup,
     loading: hcmLoading,
   } = useUsStaffGroups(skip);
 
+  const ineligibleByGroup: Record<RequiredUserGroupEnum, boolean> = {
```
kegrimes, line 51: Great idea using the Record! Thanks for simplifying that

src/hooks/useHrToolsNavItems.ts
```diff
@@ -10,53 +10,62 @@ export function useHrToolsNavItems(): NavItems[] {
...
-    () => hrToolsNavItems.filter((item) => !item.hideItem),
-    [
-      t,
-      inAsrIneligibleGroup,
-      inSalaryCalcIneligibleGroup,
-      inMhaIneligibleGroup,
-      hasNoStaffAccount,
-    ],
-  );
+    return [
```
kegrimes, line 27: I think we need to hide the tab fully if each item is hidden. 

When I opened MPDX with my account, I could see the "HR Tools" dropdown but nothing inside because I do not have a staff account id and I am not eligible for any of the request forms. I'm not sure if this is a practical case, but I think it would be good to cover to prevent users from seeing what I see!

src/hooks/useUsStaffGroups.ts
```diff
@@ -54,23 +54,35 @@ export function useUsStaffGroups(skip?: boolean) {
...
 
+  const userSupportType = user?.staffInfo?.peopleGroupSupportType;
+
   const inAsrIneligibleGroup = hasNoStaffAccount || allAsrIneligible;
   const inSalaryCalcIneligibleGroup =
     hasNoStaffAccount || user?.salaryRequestEligible === false;
   const inMhaIneligibleGroup = hasNoStaffAccount || allMhaIneligible;
+  const inMpdGoalCalcIneligibleGroup =
+    hasNoStaffAccount ||
+    userSupportType !== PeopleGroupSupportTypeEnum.SupportedRmo;
```
kegrimes: If senior staff are the only ones who are eligible for MPD goal calculator, I don't think this check is correct. The status codes for senior staff are...

- SC - Staff Full time
Employee - Staff, Supported - RMO, Active - Payroll Eligible, Full-time regular
- SCO - Staff Full time only
Employee - Staff non-RMO Spouse, Supported - RMO, Active - Payroll Eligible, Full-time regular

...but an intern can also be `Supported - RMO` (if you impersonate nataley.guetherman@cru.org from this [excel sheet](https://docs.google.com/spreadsheets/d/1vX8MRuY96nztM_lIwxwHSF2JH77Rbc2AiPTY4bmflLM/edit?gid=0#gid=0), you will be able to see MPD)

For MHA, ASR, and Salary calc, I have the eligibility validators checking the fields instead of doing it like this on the frontend. Is there an eligibility field for MPD (and even PDS)? If there is, I would use that field instead and verify that it is for a senior staff only (you can check Salary calc in the backend for reference - senior staff only).

But that might need to be another PR for the backend if those eligibility checks are not correct for MPD and PDS. If that is the case, then I would keep these checks on the frontend but fix senior staff to only be those status codes listed above. Senior staff is the one group that needs all 4 fields to check. You can look at the status codes [here](https://docs.google.com/spreadsheets/d/1FajMaIa0ocsA2lzKgx22FSj71NOAHqIlWhXUSBwBrbc/edit?gid=338348075#gid=338348075) to verify)

kegrimes (changes requested):
Thank you for doing this Will, it looks amazing! I did have a few suggestions and found a couple bugs.

Also just to let you know, I have already updated the backend to cache all possible US staff groups (senior staff, new staff, PTFS, paid with designation, and interns) into the User record. This will eliminate the extra call we are making to HCM so, a lot of this work will be refactored! But, thank you for adding MPD and PDS calculators to the list! That is one less thing I need to do.

Would you mind adding MPD and PDS goal calculators to this document with the groups who are eligible for them? This is just so we can keep track of who is eligible for what!

https://docs.google.com/document/d/1OWkY1iDCE-DGlVuYMYpLcvtPucmjD-j1wQt9AsQxxIs/edit?tab=t.0

pages/accountLists/[accountListId]/hrTools/goalCalculator/index.page.tsx
Comment (wjames111) [reply pages/accountLists/[accountListId]/hrTools/goalCalculator/index.page.tsx:45]: Good catch, my bad!

kegrimes:
@wjames111 Now that I am thinking about it, I don't think the PDS calculator requires a staff account id because paid with designation staff members do not have one at all.
Screenshot 2026-05-27 at 3 47 07 PM

wjames111:
@kegrimes Okay so I made the changes you requested. Also `inMpdGoalCalcIneligibleGroup` now uses `salaryRequestEligible` and I hardcoded `inPdsGoalCalcIneligibleGroup` as true, until we can get a backend PR up that adds a PDS Goal Calculator eligibility validator. However in order to create that backend work we'll need to resolve them needing a staff account Id. Does that sounds right?

pages/accountLists/[accountListId]/hrTools/goalCalculator/[goalCalculationId].page.tsx
```diff
@@ -139,7 +142,10 @@ export const GoalCalculatorPage: React.FC = () => {
         <title>{`${appName} | ${t('Reports - Goal Calculation')}`}</title>
       </Head>
       {accountListId ? (
-        <UserTypeAccess>
+        <UserTypeAccess
+          requireStaffAccount
```
Comment (kegrimes) [inline pages/accountLists/[accountListId]/hrTools/goalCalculator/[goalCalculationId].page.tsx]: This is my fault, but PDS and Goal calculator do NOT require a staff account. PDS is because of my previous comment (when I figured out paid with designation users do not have a staff account) and Caleb told me goal calculator does not.

So, I would remove the prop from all 4 pages.

src/hooks/useUsStaffGroups.ts
```diff
@@ -58,19 +58,27 @@ export function useUsStaffGroups(skip?: boolean) {
   const inSalaryCalcIneligibleGroup =
     hasNoStaffAccount || user?.salaryRequestEligible === false;
   const inMhaIneligibleGroup = hasNoStaffAccount || allMhaIneligible;
+  const inMpdGoalCalcIneligibleGroup =
+    hasNoStaffAccount || user?.salaryRequestEligible === false;
```
kegrimes: I think we need to remove the `hasNoStaffAccount` check because the MPD goal calculator does not require a staff account id. 

Evidence: When I impersonate timothy.ethington@cru.org (New staff - should not be able to see salary calc or mpd goal calculator), it was not showing Salary calculator in HR Tools, however, MPD goal calculator was showing, which is incorrect. 

```suggestion
  const inMpdGoalCalcIneligibleGroup = user?.salaryRequestEligible === false;
```

src/hooks/useUsStaffGroups.ts
kegrimes, line 59: Can you update this after the new Alithya change? I think we want to check for both "NO_STAFF_ACCOUNT" and "HCM_PERSON_NOT_FOUND", but you may want to double check from [Daniel's PR](https://github.com/CruGlobal/mpdx_api/pull/3335). We are now getting a different error with the new parameter. My personal account should not be seeing any request forms, but now they are appearing because of this.

Maybe we need to rename the variable to `hasNoHcmData`?

```
{
    "data": null,
    "errors": [
        {
            "message": "HCM person not found for staff_account_id=, person_number=001317644",
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ],
            "path": [
                "hcm"
            ],
            "extensions": {
                "code": "HCM_PERSON_NOT_FOUND"
            }
        }
    ]
}
```

kegrimes (changes requested):
We are almost there! Just a couple things. I want to be able to test the HR Tools tab not showing up when the user does not have access to any of those reports!  The new Alithya change affected that, so we need to change that.

kegrimes:
I also reached out to Daniel to see if he could merge his PR to unblock us for the backend work!

[Update]: It has been merged to production!

src/hooks/useUsStaffGroups.ts
```diff
@@ -47,31 +47,45 @@ export function useUsStaffGroups(skip?: boolean) {
...
 
-  const inAsrIneligibleGroup = hasNoStaffAccount || allAsrIneligible;
+  const inAsrIneligibleGroup = hasNoHcmData || allAsrIneligible;
   const inSalaryCalcIneligibleGroup =
-    hasNoStaffAccount || user?.salaryRequestEligible === false;
-  const inMhaIneligibleGroup = hasNoStaffAccount || allMhaIneligible;
+    hasNoHcmData || user?.salaryRequestEligible === false;
+  const inMhaIneligibleGroup = hasNoHcmData || allMhaIneligible;
+  const inMpdGoalCalcIneligibleGroup = user?.salaryRequestEligible === false;
+  const inPdsGoalCalcIneligibleGroup = hasNoHcmData || allPdsGoalCalcIneligible;
```
kegrimes: I actually think we need to separate these errors. I was impersonating myself and it was still allowing me to see the MPD goal calculator even though I did not have any HCM data. I think we need to do something like this:

```
  const hasNoStaffAccount =
    error?.graphQLErrors.some(
      (graphQLError) => graphQLError.extensions?.code === 'NO_STAFF_ACCOUNT',
    ) ?? false;

  const hcmPersonNotFound =
    error?.graphQLErrors.some(
      (graphQLError) =>
        graphQLError.extensions?.code === 'HCM_PERSON_NOT_FOUND',
    ) ?? false;

  const hasNoHcmData = hasNoStaffAccount || hcmPersonNotFound;

  const inAsrIneligibleGroup = hasNoHcmData || allAsrIneligible;
  const inSalaryCalcIneligibleGroup =
    hasNoHcmData || user?.salaryRequestEligible === false;
  const inMhaIneligibleGroup = hasNoHcmData || allMhaIneligible;
  const inMpdGoalCalcIneligibleGroup =
    hcmPersonNotFound || user?.salaryRequestEligible === false;
  const inPdsGoalCalcIneligibleGroup =
    hcmPersonNotFound || allPdsGoalCalcIneligible;
```

I already tested all 5 scenarios and they should all be working with this logic. MPD and PDS calculators don't require a staff account id, so even without one HCM still gets called (we have a person number) and can come back with `HCM_PERSON_NOT_FOUND`. We only want to check that error against these 2 calculators. `NO_STAFF_ACCOUNT` just means no account id exists, and since these reports don't require one, we don't want that error to hide them. ASR, Salary, and MHA do require a staff account id, so they hide on either error (`hasNoHcmData`).

So the distinction is:
- `NO_STAFF_ACCOUNT` = no id exists (raised before HCM hit)
- `HCM_PERSON_NOT_FOUND` = HCM data was reached, but there is no data

kegrimes: If you are wanting to test these for yourself, kathryn.hougham@cru.org is the only account that will work for paid with designation users. She has her person number set up correctly in staging!

kegrimes (approved):
Great work Will!

I have one more bug fix for you, but other than that, this PR is good to go so I am going to approve. Thank you for all your hard work! It will be easier for me to update the frontend customized work with MPD and PDS set up!

PR #1809, opened 2026-05-27
MPDX-9608 Remove extra HCM call

kegrimes:
## Description

Currently, we are calling HCM on every dashboard page load which is expensive. Instead, the backend has been updated to cache US staff groups into the User record which we have access to through a new graphql field, `usStaffGroup`. 

Using this new field will eliminate the extra call to HCM. Also, instead of hiding "HR Tools" until we find out if the user should be able to see it, I flipped that logic to only have the tab appear only if a user is allowed to see it.

Jira Ticket: [MPDX-9608](https://jira.cru.org/browse/MPDX-9608)

## Testing

Senior Staff test: ✅
- Go to dashboard
- Ensure "HR Tools" is present and open tab
- Check that user can see Salary Calculation Form, Savings Fund Transfer, MPD Goal Calculator, MHA Calculation Tool, Additional Salary Request, and Ministry Partner Reminders (total: 6)
- Go to `/hrTools/pdsGoalCalculator` and verify limited access message appears

New Staff test: ✅
- Impersonate: timothy.ethington@cru.org
- Go to dashboard
- Ensure "HR Tools" is present and open tab
- Check that user can see Savings Fund Transfer, Additional Salary Request, and Ministry Partner Reminders (total: 3)
- Go to `/hrTools/salaryCalculator`, `/hrTools/goalCalculator`, `/hrTools/mhaCalculator`, and `/hrTools/pdsGoalCalculator` and verify limited access message appears

Part-time Field Staff test: ✅
- Impersonate: christina.garber@athletesinaction.org 
- Go to dashboard
- Ensure "HR Tools" is present and open tab
- Check that user can see Savings Fund Transfer and Ministry Partner Reminders (total: 2)
- Go to `/hrTools/salaryCalculator`, `/hrTools/goalCalculator`, `/hrTools/mhaCalculator`, `/hrTools/additionalSalaryRequest`, and `/hrTools/pdsGoalCalculator` and verify limited access message appears

Paid with Designation Staff test: ✅
- Impersonate: kathryn.hougham@cru.org
- Go to dashboard
- Ensure "HR Tools" is present and open tab
- Check that user can see Paid with Designation Support Goal Calculator (total: 1)
- Go to `/hrTools/salaryCalculator`, `/hrTools/staffSavingFund`, `/hrTools/goalCalculator`, `/hrTools/mhaCalculator`, `/hrTools/additionalSalaryRequest`, and `/hrTools/partnerReminders` and verify limited access message appears

Intern test: ✅
- Impersonate: nataley.guetherman@cru.org 
- Go to dashboard
- Ensure "HR Tools" is present and open tab
- Check that user can see Savings Fund Transfer and Ministry Partner Reminders (total: 2)
- Go to `/hrTools/salaryCalculator`, `/hrTools/goalCalculator`, `/hrTools/mhaCalculator`, `/hrTools/additionalSalaryRequest`, and `/hrTools/pdsGoalCalculator` and verify limited access message appears

National Staff (Expatriate) test: Asking Randy for another test account ❗️
- Impersonate: ?
- Go to dashboard
- Ensure "HR Tools" is present and open tab
- Check that user can see Salary Calculation Form, Savings Fund Transfer, MHA Calculation Tool, Additional Salary Request, and Ministry Partner Reminders (total: 5)
- Go to `/hrTools/goalCalculator` and `/hrTools/pdsGoalCalculator` and verify limited access message appears

Locally tested National Staff ✅

## Checklist:


src/hooks/useIneligibleByGroup.ts
```diff
@@ -0,0 +1,49 @@
+import { useMemo } from 'react';
+import { useGetUserQuery } from 'src/components/User/GetUser.generated';
+import { UsStaffGroupEnum } from 'src/graphql/types.generated';
+
+export function useIneligibleByGroup() {
+  const { data, loading: userLoading, error: userError } = useGetUserQuery();
+
+  const usStaffGroup = data?.user.usStaffGroup;
```
kegrimes, line 15: 🟡 **[Important] Spouse-eligibility dimension dropped.** The deleted `useUsStaffGroups` hook evaluated ASR/MHA/PDS eligibility across the logged-in user *and spouse* (`people.every(...)`, with an explicit non-RMO-spouse branch). This new hook reads only `data.user.usStaffGroup` — a per-individual field (`schema.graphql:12739`) with no spouse dimension. A user previously eligible *via their spouse* will now be wrongly hidden the tab AND blocked by `UserTypeAccess`; for MHA/MPD/PDS there is no inner server re-check to let them back in (only ASR has one). Confirm against MPDX-9608 that dropping spouse eligibility is intentional, and document it. Flagged by Financial, Data Integrity, and Architecture agents + adjudication.

src/hooks/useIneligibleByGroup.ts
```diff
@@ -0,0 +1,49 @@
...
+  const isNationalExpat = usStaffGroup === UsStaffGroupEnum.NationalExpat;
+
+  const inAsrIneligibleGroup =
+    !isSeniorStaff &&
+    usStaffGroup !== UsStaffGroupEnum.NewStaff &&
+    !isNationalExpat;
+  const inSalaryCalcIneligibleGroup = !isSeniorStaff && !isNationalExpat;
+  const inMhaIneligibleGroup = !isSeniorStaff && !isNationalExpat;
+  const inMpdGoalCalcIneligibleGroup = !isSeniorStaff;
+  const inPdsGoalCalcIneligibleGroup =
```
kegrimes: 🟡 **[Important] PDS excludes Senior Staff; two enum values unhandled.** `inPdsGoalCalcIneligibleGroup = usStaffGroup !== PaidWithDesignation` means only `PaidWithDesignation` is eligible — Senior Staff are *denied* PDS Goal Calc, whereas the old logic gated on `designationSupportCalculatorEligible` independent of seniority. Also, `UsStaffGroupEnum` has 6 values but only 4 are referenced here: `Intern` and `PartTimeFieldStaff` fall through to deny-everything implicitly. Confirm the full eligibility matrix against the HR spec (esp. the Senior-Staff-denied-PDS row and the two unhandled groups). Consider per-tool eligible-group allowlists to make the rules self-documenting.

src/hooks/useIneligibleByGroup.ts
```diff
@@ -0,0 +1,49 @@
+import { useMemo } from 'react';
+import { useGetUserQuery } from 'src/components/User/GetUser.generated';
+import { UsStaffGroupEnum } from 'src/graphql/types.generated';
+
+export function useIneligibleByGroup() {
+  const { data, loading: userLoading, error: userError } = useGetUserQuery();
+
+  const usStaffGroup = data?.user.usStaffGroup;
+  const userType = data?.user.userType;
+  const hasNoStaffAccount = !data?.user.staffAccountId;
```
kegrimes, line 18: 🔵 **[Medium] `hasNoStaffAccount` semantics changed + a gate dropped.** Old code derived 'no staff account' from the GraphQL error code `NO_STAFF_ACCOUNT`; this uses `!data?.user.staffAccountId`. More importantly, the old `HCM_PERSON_NOT_FOUND` gate (which fed `inMpdGoalCalc`/`inPdsGoalCalc`) has no replacement, so a user with a staff account but no HCM person record now passes the client gate (worst case: a confusing screen — ASR still server-checks; other tools don't). Confirm `!staffAccountId` is the canonical signal and dropping the HCM-person-not-found distinction is intended. Adjudicated down from 6.0.

src/hooks/useIneligibleByGroup.ts
```diff
@@ -0,0 +1,49 @@
...
+    !isSeniorStaff &&
+    usStaffGroup !== UsStaffGroupEnum.NewStaff &&
+    !isNationalExpat;
+  const inSalaryCalcIneligibleGroup = !isSeniorStaff && !isNationalExpat;
+  const inMhaIneligibleGroup = !isSeniorStaff && !isNationalExpat;
+  const inMpdGoalCalcIneligibleGroup = !isSeniorStaff;
+  const inPdsGoalCalcIneligibleGroup =
+    usStaffGroup !== UsStaffGroupEnum.PaidWithDesignation;
+
+  return useMemo(
```
kegrimes, line 46: ⚪ **[Suggestion] `useMemo` adds no value here.** All derived booleans are primitives recomputed on every render *outside* the memo; the memo only stabilizes the wrapper object, and consumers immediately destructure primitives out of it. Harmless and pattern-consistent with the deleted hook, so not a regression — but you could return the object literal directly for clarity. Also consider adding a one-line comment noting that unknown/undefined `usStaffGroup` intentionally fails closed (deny), so a future maintainer adding a backend enum value knows to update this hook.

src/hooks/useNavPages.tsx
```diff
@@ -122,7 +122,8 @@ export function useNavPages(coachingAccountCount: boolean, isSearch = false) {
               showInNav: true,
               hideTab:
                 (!!data && !showTab) ||
```
kegrimes, line 124: 🔵 **[Medium] `hideTab` loading-inversion is intentional but untested.** This changes the HR Tools tab from shown-while-loading/hidden-once-empty to hidden-while-loading/shown-after-load. UX and Architecture agents agree the new direction is the better trade-off (avoids flashing a tab an ineligible user could click), but there is no `useNavPages` test asserting the tab is absent while pending and present after resolve — a future accidental sign-flip wouldn't be caught. Add a NavMenu/useNavPages test for the loading state.

src/components/Shared/UserTypeAccess/UserTypeAccess.test.tsx
```diff
@@ -57,12 +60,7 @@ const TestComponent: React.FC<TestComponentProps> = ({
               },
```
kegrimes: 🔵 **[Medium] Tests pass for the wrong reason (stale `Hcm` mock + dead props).** The component now derives eligibility solely from `useIneligibleByGroup` (GetUser), but this test still mocks the `Hcm` query and passes `asrEligible`/`salaryRequestEligible`/`designationSupportCalculatorEligible`/etc. props that no longer affect any assertion. The 'ineligible' cases pass only because the default `usStaffGroup = PartTimeFieldStaff` is ineligible for everything — so `asrEligible={false}` is a no-op and the assertions would still pass if set to `true`. Drive each case via `usStaffGroup` explicitly and delete the inert `Hcm`/`StaffAccount` mocks, generics, and props. Flagged by Testing + Standards agents.

src/hooks/useIneligibleByGroup.test.tsx
```diff
@@ -0,0 +1,153 @@
...
+            },
+          },
+        }}
+      >
+        {children}
+      </GqlMockedProvider>
+    ),
+  });
+
+describe('useIneligibleByGroup', () => {
```
kegrimes, line 35: ⚪ **[Suggestion] Add a test for the resolved `undefined`-`usStaffGroup` case.** The 6 enum values + loading + error are well covered, but not a *resolved* user with no `usStaffGroup` (a real new-user/missing-data state), which should yield all five `in*IneligibleGroup` flags `true` (default-deny). This is the most likely real-world edge to silently regress to default-allow. Note: `GqlMockedProvider` auto-fills the enum, so you may need to force it via the MultiPageMenu render path — worth a comment either way.

kegrimes (commented):
## 🔍 Multi-Agent Code Review — Verdict: APPROVED WITH SUGGESTIONS

7 specialized agents (Security, Architecture, Data Integrity, Testing, UX, Standards, Financial-Reporting) + dependency analysis + 1 adjudication round. **No merge blockers.** The refactor cleanly removes two network round-trips, has zero dangling references to the deleted `useUsStaffGroups` hook, and is well-tested at the unit level.

⚠️ **3 Important findings (severity 7.0–7.9)** — these don't block merge and can't be `/dismiss`ed. Two of them depend on the **MPDX-9608 eligibility spec**: this PR collapses multi-signal HCM eligibility (per-person flags, spouse, error codes) into a single `usStaffGroup` enum, and code alone cannot prove the mapping is faithful.

### Please confirm before merge
1. **Spouse dimension dropped** — old logic evaluated ASR/MHA/PDS against the user *and spouse*; new logic reads only the logged-in user's `usStaffGroup`. Adjudication confirmed only **ASR** has an inner server re-check, so for MHA/MPD/PDS this client hook is the authoritative gate — a spouse-eligible user is now wrongly *hidden* with no fallback. Confirm intended.
2. **Eligibility matrix** — PDS Goal Calc now excludes Senior Staff (only `PaidWithDesignation` eligible); `Intern` and `PartTimeFieldStaff` are never referenced and fall through to deny-everything. Confirm the full 6-value matrix vs the HR spec.

### Eligibility truth table (new logic, ✅ = eligible)
| usStaffGroup | ASR | Salary | MHA | MPD Goal | PDS Goal |
|---|:--:|:--:|:--:|:--:|:--:|
| SeniorStaff | ✅ | ✅ | ✅ | ✅ | ❌ |
| NationalExpat | ✅ | ✅ | ✅ | ❌ | ❌ |
| NewStaff | ✅ | ❌ | ❌ | ❌ | ❌ |
| PaidWithDesignation | ❌ | ❌ | ❌ | ❌ | ✅ |
| Intern | ❌ | ❌ | ❌ | ❌ | ❌ |
| PartTimeFieldStaff | ❌ | ❌ | ❌ | ❌ | ❌ |
| undefined/null | ❌ | ❌ | ❌ | ❌ | ❌ (default-deny, fail-closed ✓) |

### Security
No issues — this is client-side display gating, no server authorization boundary changed, fail-state is fail-closed (deny).

### Dependency impact
`UserTypeAccess` has high blast radius (13 render sites) but its public props are unchanged → all sites stay source-compatible. `useUsStaffGroups` is fully removed with 0 remaining references.

---

These were identified during review on files related to the change but not modified in this PR, so they can't be posted as line comments.

**[Important] `src/components/Layouts/Primary/TopBar/Items/NavMenu/NavMenu.test.tsx` (~lines 130–175)** — Severity 7.0 — Flagged by: Testing
This is a real consumer of `useNavPages → useHrToolsNavItems → useIneligibleByGroup`, but it was not migrated. It still mocks the now-inert `Hcm` query and provides no `usStaffGroup`/`staffAccountId`; the suite passes (11/11) only because `GqlMockedProvider` auto-fills the enum, so the eligibility assertions no longer verify the logic they claim to (false-positive coverage). Recommended: migrate to mock `GetUser.usStaffGroup` + `staffAccountId` explicitly and drop the dead `Hcm` mock + unused enum imports. Ideally done in this PR since it's the same refactor. *(Informational — does not count toward the verdict.)*

---
*Severity scale: 🔴 Critical 9–10 · 🟠 High 8–8.9 · 🟡 Important 7–7.9 · 🔵 Medium 5–6.9 · ⚪ Suggestion <5. Inline comments below are tagged with machine-readable severity.*

src/hooks/useIneligibleByGroup.ts
kegrimes: /dismiss This is correct. PDS calculator should only be available to users who are paid with designation.

src/hooks/useIneligibleByGroup.ts
kegrimes: /dismiss we are not calling HCM anymore on the first dashboard load. This means we will not see either of those errors and it doesn't matter because `usStaffGroup` field is based on HCM data. If that field doesn't get set, then users won't be able to access reports that use it anywy.

src/hooks/useNavPages.tsx
kegrimes: /dismiss code coverage is good

src/hooks/useIneligibleByGroup.ts
kegrimes: /dismiss

src/hooks/useIneligibleByGroup.ts
kegrimes: /dismiss I think the logic before was incorrect. User's part of a US staff group should only be able to see what their group allows them to. If a user is ASR ineligible and their spouse is eligible, then the spouse would be able to create an ASR for the user. We shouldn't let un-permitted users see forms they aren't able to fill out.

kegrimes:
@canac Before, we were showing tabs based on user or spouse eligibility since forms like ASR can switch to the spouse's view. However, I am struggling to see how we can do this since we would need an HCM call to get spouse eligibility which would be adding back the extra call this PR was removing in the first place.

I haven't thought of another possible solution yet, but it seems like we may need to take the tradeoff if we want the customized login to work correctly.

canac:
So the issue is that the backend only remembers that a user is senior staff, but if their spouse is senior staff, they should have access to ASRs? It seems like if we want to avoid an HCM GraphQL call on the frontend, then the backend needs to remember more information. Maybe a boolean field for whether either the user or spouse is eligible for ASRs. And possibly the same for MHA.

Although, I wonder if Scott would be OK with making so that if John isn't eligible for ASRs and Jane is, that John can't submit an ASR on Jane's behalf.

kegrimes:
@canac I don't have a test account for National Staff (Expatriate). Randy said he could look into getting me one next week, so it will be a while. The only difference between a senior staff member and NX staff members is that they should not be able to see MPD goal calculator.

src/components/Shared/UserTypeAccess/UserTypeAccess.tsx
```diff
@@ -80,13 +68,7 @@ export const UserTypeAccess: React.FC<UserTypeAccessProps> = ({
...
 
   if (requireStaffAccount) {
-    if (staffAccountError) {
-      return <LimitedAccess userGroupError />;
-    }
-    if (staffAccountLoading && !staffAccountData) {
-      return <Loading loading />;
-    }
-    if (!staffAccountData?.staffAccount?.id) {
+    if (hasNoStaffAccount) {
```
canac:
```suggestion
  if (requireStaffAccount && hasNoStaffAccount) {
```

src/hooks/useIneligibleByGroup.ts
```diff
@@ -0,0 +1,70 @@
+import { useMemo } from 'react';
+import { useGetUserQuery } from 'src/components/User/GetUser.generated';
+import { UsStaffGroupEnum } from 'src/graphql/types.generated';
+
+const isInEligibleGroup = (
+  group: UsStaffGroupEnum | null | undefined,
+  eligibleGroups: UsStaffGroupEnum[],
+) => {
+  return group !== null && eligibleGroups.includes(group as UsStaffGroupEnum);
```
canac: Should we check group !== undefined? Or `typeof group === 'string' && ...' to skip null and undefined with one check?

src/hooks/useIneligibleByGroup.ts
```diff
@@ -0,0 +1,70 @@
...
+};
+
+export function useIneligibleByGroup() {
```
canac, line 12: praise: The access checks are really easy to read and maintain now!

canac (approved):
This looks great! I spot-checked a few of the impersonations but not all of them. I'm trusting that you've been testing them. Feel free to merge before testing National Staff (Expatriate). You could also hardcode the status to NationalExpat locally to test that the frontend works correctly.

PR #1811, opened 2026-05-29
MPDX-9648, MPDX-9649 Savings fund transfer & MPGA UAT

kegrimes:
## Description

Jira ticket: [MPDX-9648](https://jira.cru.org/browse/MPDX-9648)
- Round CSV values to nearest hundredth place

Jira ticket: [MPDX-9649](https://jira.cru.org/browse/MPDX-9649)
- Change all "Stop Date" occurrences to "End Date" in Savings Fund Transfer

## Testing

MPGA test:
- Impersonate: brian.funkhouser@cru.org
- Download the MPGA income table
- Verify all values are rounded to the nearest hundredth place

Savings Fund Transfer test:
- Impersonate: veronika.bailey@cru.org
- Go to `/hrTools/staffSavingFund/transfers`
- Verify Transfer History table has "End Date" as column header
- Download transfer history csv file
- Check that column header says "End Date"

## Checklist:

### DIFF
--- a/src/components/Reports/MPGAIncomeExpensesReport/CustomExport/CustomExport.tsx
+++ b/src/components/Reports/MPGAIncomeExpensesReport/CustomExport/CustomExport.tsx
@@ -2,12 +2,30 @@ import { buildURI } from 'react-csv/lib/core';
 import { ReportTypeEnum } from '../Helper/MPGAReportEnum';
 import { DataFields } from '../mockData';
 
-export const createTable = (csvHeader: string[], data: DataFields[]) => {
+const round = (value: number, locale: string) =>
+  new Intl.NumberFormat(locale, {
+    style: 'decimal',
+    minimumFractionDigits: 2,
+    maximumFractionDigits: 2,
+    trailingZeroDisplay: 'stripIfInteger',
+    useGrouping: false,
+  }).format(value);
+
+export const createTable = (
+  csvHeader: string[],
+  data: DataFields[],
+  locale: string,
+) => {
   const csvData = data.map((item) => {
-    const monthlyData = item.monthly.map((month) =>
-      month === 0 ? '-' : month,
+    const monthlyData = item.monthly.map((monthlyAmount) =>
+      monthlyAmount === 0 ? '-' : round(monthlyAmount, locale),
     );
-    return [item.description, ...monthlyData, item.average, item.total];
+    return [
+      item.description,
+      ...monthlyData,
+      round(item.average, locale),
+      round(item.total, locale),
+    ];
   });
 
   return [csvHeader, ...csvData];
@@ -17,6 +35,7 @@ export const exportToCsv = (
   data: DataFields[],
   reportType: ReportTypeEnum,
   months: string[],
+  locale: string,
 ) => {
   const title =
     reportType === ReportTypeEnum.Income
@@ -46,7 +65,7 @@ export const exportToCsv = (
   ];
 
   const csvHeader = ['Description', ...last12Months, 'Average', 'Total'];
-  const csvData = createTable(csvHeader, dataWithTotal);
+  const csvData = createTable(csvHeader, dataWithTotal, locale);
 
   const csvBlob = buildURI(csvData, true);
 
--- a/src/components/HrTools/SavingsFundTransfer/DownloadTable/downloadTable.test.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/DownloadTable/downloadTable.test.tsx
@@ -18,7 +18,7 @@ const mockHeaders = [
   'Schedule',
   'Status',
   'Transfer Date',
-  'Stop Date',
+  'End Date',
   'Note',
 ];
 
--- a/src/components/HrTools/SavingsFundTransfer/DownloadTable/downloadTable.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/DownloadTable/downloadTable.tsx
@@ -73,7 +73,7 @@ export const downloadCSV = (
     t('Schedule'),
     t('Status'),
     t('Transfer Date'),
-    t('Stop Date'),
+    t('End Date'),
     t('Note'),
   ];
 
--- a/src/components/HrTools/SavingsFundTransfer/IntroPage/IntroPage.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/IntroPage/IntroPage.tsx
@@ -99,14 +99,14 @@ export const SavingsFundTransfer: React.FC<SavingsFundTransferProps> = ({
               <StyledListItem>
                 <ListItemText
                   primary={t(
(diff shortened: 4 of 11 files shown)


kegrimes:
## 🤖 Multi-Agent Code Review — Quick Mode ✅ Approve

**Agents:** 🧪 Testing · 👤 UX · 📋 Standards · **Risk:** 🟢 LOW · 10 files (+64 / −25)

Two logical changes: (1) label rename **"Stop Date" → "End Date"** across `SavingsFundTransfer/**`, and (2) a `round()`-to-2-decimals helper added to the MPGA CSV export (with a new test).

### Verdict: ✅ Approve / safe to merge — no real blockers.

---

### 🚫 Critical Blockers
**None.**

### ⚠️ Flagged by agents but cleared on verification

**"Missing translation keys will break the UI" → false positive.**
`src/lib/i18n.ts` is configured with `keySeparator: false`, `nsSeparator: false`, and no `saveMissing`/missing-key handler. i18next therefore returns the *key itself* when a translation is absent — and in MPDX the English string **is** the key. So `t('End date updated successfully')` renders correctly for English users; there is no runtime breakage. The only minor follow-ups (non-blocking): non-English locales fall back to English until CrowdIn refreshes (normal `yarn extract` pipeline), and 3 orphaned `Stop date…` entries remain in `public/locales/en/translation.json` (pruned by `yarn extract`).

**"CSV rounding may mismatch the on-screen report" → resolved by this PR.**
On-screen tables already format via `amountFormat`/`zeroAmountFormat` (Intl, 2 decimals). The CSV export was the outlier emitting raw floats — this PR brings the download into line with the display. This *is* the fix.

### 💡 Optional, non-blocking suggestions
- **Rounding edge cases** — `CustomExport.tsx:5`. The new test covers round-down/round-up/average/total but not negatives or an exact half-cent (`x.xx5`) boundary; worth one extra assertion if expense reports can go negative.
- **Aggregate-vs-rounded (verified correct)** — `CustomExport.tsx:34-42` sums **raw** values and rounds only at the `createTable` boundary. Rounded monthly columns may therefore not sum exactly to the rounded total (≤1¢ drift) — the standard, acceptable display-boundary tradeoff.

### ✅ What's solid
- Rename is **complete and consistent** across source + tests (column headers, IconButton titles, snackbar success/error messages, intro help text, CSV/print/download headers). No leftover user-facing "Stop Date" strings.
- New rounding logic is **tested**; existing tests updated in lockstep.
- All strings stay wrapped in `t()`; no hardcoded text, no interpolation-in-`t()`, no dynamic keys.
- Standards clean: named exports, no `any`/`@ts-ignore`/non-null assertions, no `console.log`/`new Date()`, accessible names preserved, financial rounding at the display boundary.

| Agent | Critical | Important | Suggestions | After verification |
|---|---|---|---|---|
| 🧪 Testing | 0 | 0 | 2 | clean |
| 👤 UX | 1 | 2 | — | 1 false-positive, 1 resolved-by-PR |
| 📋 Standards | 1 | 0 | — | 1 false-positive |

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) — multi-agent review (quick mode)

src/components/Reports/MPGAIncomeExpensesReport/CustomExport/CustomExport.tsx
```diff
@@ -2,12 +2,19 @@ import { buildURI } from 'react-csv/lib/core';
 import { ReportTypeEnum } from '../Helper/MPGAReportEnum';
 import { DataFields } from '../mockData';
 
+const round = (value: number) => Math.round(value * 100) / 100;
+
 export const createTable = (csvHeader: string[], data: DataFields[]) => {
   const csvData = data.map((item) => {
     const monthlyData = item.monthly.map((month) =>
-      month === 0 ? '-' : month,
+      month === 0 ? '-' : round(month),
```
zweatshirt: What do you think of changing `month` to something else here? `round(month)` is kind of confusing to read. Also wondering if we want the 0 to remain 0 instead of '-', to give spreadsheet software an easier time with calculations.

src/components/Reports/MPGAIncomeExpensesReport/CustomExport/CustomExport.tsx
```diff
@@ -2,12 +2,19 @@ import { buildURI } from 'react-csv/lib/core';
 import { ReportTypeEnum } from '../Helper/MPGAReportEnum';
 import { DataFields } from '../mockData';
 
+const round = (value: number) => Math.round(value * 100) / 100;
+
```
zweatshirt, line 13: I think we might want something like this instead of rounding this way. This would fix some weird edge cases that would occur with `Math.round` or `toFixed(2)`. I think `Math.round(value * 100) / 100` also gets rid of any trailing zeroes e.g. 10.200 becomes 10.2 instead of 10.20.

```
const round = (value: number, locale: string) =>
    style: 'decimal',
    minimumFractionDigits: 2,
    maximumFractionDigits: 2,
    useGrouping: false,  // no commas — keeps Excel happy
  }).format(value);
```

zweatshirt (changes requested):
This is great! Just a quick suggestion, feel free to dismiss and readd me for review if needed :)

src/components/Reports/MPGAIncomeExpensesReport/CustomExport/CustomExport.tsx
kegrimes: Good point about the name change!

I think I kept it like that because it matches the actual table, so I think I might leave that for now. If it becomes something stakeholders or users want changed, it would be an easy fix!

src/components/Reports/MPGAIncomeExpensesReport/CustomExport/CustomExport.tsx
kegrimes: I like this! Thanks

zweatshirt: We also have the `currencyFormat` in `intlFormat` but wasn't sure if that fit the needs here.

Or formatDecimal in `useFormatters` which also isn't quite what we need but could maybe be changed to accept arguments eventually:
```
const formatDecimal = (value: number | null | undefined) =>
      style: 'decimal',
      minimumFractionDigits: 2,
      maximumFractionDigits: 4,
    }).format(value ?? 0);
 ```

zweatshirt:
Not sure why, impersonating Brian in the preview environment, the report itself is for Mickey and Minnie. Super odd. Did you run into this issue?

kegrimes:
@zweatshirt Keely might be creating the helpducks article for these reports. I know she needed the test account to display Mickey and Minnie data so she isn't screenshotting real data!

zweatshirt (approved):
Looks great! Sorry for the delay. Would we be able to get the test coverage up before merging?

kegrimes:
@zweatshirt I think the project coverage is what matters most and it seems like the percentage did not drop! I think we should be good to merge as is.

PR #1812, opened 2026-05-29
MPDX-9653 - Simplify From Account -> To Account behavior in Savings Fund Transfer

zweatshirt:
## Description

- Ensures that the 'From' account is disabled, and that the 'Transfer to' button is removed to fulfill ticket requirements. 
    - This heavily implies the ability to swap From and To accounts in the selects should be removed as well.
- https://jira.cru.org/browse/MPDX-9653

## Testing

- Go to Savings Fund Transfer on any test account
- Test the behavior of selecting the From button, ensuring that a user cannot change the From account.
- Ensure that the To button is removed from every card.

## Checklist:

### DIFF
--- a/src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
@@ -1,6 +1,5 @@
 import { useState } from 'react';
-import East from '@mui/icons-material/East';
-import SwapHorizIcon from '@mui/icons-material/SwapHoriz';
+import { ArrowRightAlt } from '@mui/icons-material';
 import {
   Alert,
   Box,
@@ -11,12 +10,10 @@ import {
   FormHelperText,
   FormLabel,
   Grid,
-  IconButton,
   InputLabel,
   Radio,
   RadioGroup,
   TextField,
-  Tooltip,
   Typography,
 } from '@mui/material';
 import { Formik } from 'formik';
@@ -309,26 +306,17 @@ export const TransferModal: React.FC<TransferModalProps> = ({
                             {t('From Account')}
                           </InputLabel>
                           <TransferModalSelect
-                            notSelected={transferTo}
-                            funds={funds}
+                            funds={funds.filter(
+                              (fund) => fund.fundType === transferFrom,
+                            )}
                             label={t('From Account')}
                             labelId="transferFrom"
                             name="transferFrom"
                             value={transferFrom}
-                            disabled={!isNew}
-                            onChange={handleChange}
-                            onBlur={handleBlur}
-                            error={
-                              touched.transferFrom &&
-                              Boolean(errors.transferFrom)
-                            }
+                            readOnly
+                            IconComponent={() => null}
                             required
                           />
-                          {touched.transferFrom && errors.transferFrom && (
-                            <FormHelperText error={true}>
-                              {errors.transferFrom}
-                            </FormHelperText>
-                          )}
                         </FormControl>
                       </Grid>
 
@@ -342,18 +330,7 @@ export const TransferModal: React.FC<TransferModalProps> = ({
                           alignItems: 'center',
                         }}
                       >
-                        <IconButton
-                          onClick={() => {
-                            setFieldValue('transferFrom', transferTo);
-                            setFieldValue('transferTo', transferFrom);
-                          }}
-                          color="primary"
-                          disabled={!transferFrom || !transferTo}
-                        >
-                          <Tooltip title={t('Swap')}>
-                            <SwapHorizIcon />
-                          </Tooltip>
-                        </IconButton>
+                        <ArrowRightAlt />
                       </Grid>
 
                       <Grid item xs={12} sm={5.5}>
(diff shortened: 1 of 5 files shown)


zweatshirt (commented):
## Multi-Agent Code Review — PR #1812

**Mode:** standard · **Risk:** LOW (3/10) · **Verdict:** CLEAN · **Blockers:** 0

Locks the From account in the Savings Fund Transfer modal (read-only, full-contrast `Select` with the dropdown arrow hidden) and removes the "TRANSFER TO" entry point from the balance card. Net deletion; mostly simplification. The full `SavingsFundTransfer` suite passes and `lint:ts` is clean.

### Resolved since earlier iterations
The From account was briefly a `disabled` Select, which greyed out the must-read balance and flattened the negative-balance red. Switching to **`readOnly`** restores full contrast (and keeps the value in the tab order / a11y tree so screen-reader users can still read the source fund + balance), while the hidden arrow + read-only behavior keep it locked. Verified: clicking it opens no menu.

### Suggestions (informational — severity < 5, none block)

**[Suggestion] `src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx` — From `Select` reserved arrow gutter** (3.5/10, UX)
With the arrow hidden via `IconComponent={() => null}`, MUI still reserves right-padding on `.MuiSelect-select`, so the From value sits with dead space vs the To field. Optional: `sx={{ '& .MuiSelect-select': { pr: '14px !important' } }}`.

**[Suggestion] `src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx` — lock test only covers mouse** (4.0/10, Testing)
The "locks the From Account" test clicks and asserts no listbox; a MUI Select can also open via Enter/ArrowDown. `readOnly` blocks that too, but it isn't pinned by a test. Optional: add `userEvent.type(fromAccount, '{enter}')` then assert no listbox.

**[Suggestion] `src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx` — `findBy` convention** (3.5/10, Testing)
The synchronous `getByRole('option', { name: /staff savings/i })` after the `waitFor` would be marginally more idiomatic as `await findByRole(...)`. Not flaky as-is.

**[Suggestion] `src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx` — Note field `disabled={!isNew}` dropped** (4.5/10, Architecture)
The Note field is now editable in edit mode, but `updateRecurringTransfer` has no `description` field, so an edit-mode note change would be silently dropped. Proven **unreachable today** (editable rows are always `Monthly`; the Note renders only for `OneTime`) — latent debt, not a live bug. Consider restoring `disabled={!isNew}` to make intent explicit.

### Clean dimensions
- **Financial:** no money/balance/date logic changed; deficit-warning math still keys on the correct source fund.
- **Standards:** named exports, no orphaned imports, `lint:ci` 0 errors, `IconComponent={() => null}` trips no rule.
- **Architecture:** net technical-debt reduction (swap + dual-direction complexity removed).
- **Security / Data Integrity:** not triggered (no api/auth/env/workflow or graphql/cache/schema changes).

zweatshirt:
@kegrimes This was a suggested change by Scott. I kind of liked the original design more, and so made the suggestion to him that we keep the original design. I'll wait for his response before merging this. I was curious what you think of this visually and whether you think it satisfies the ticket description

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
```diff
@@ -309,19 +306,19 @@ export const TransferModal: React.FC<TransferModalProps> = ({
...
+                              (fund) => fund.fundType === transferFrom,
+                            )}
                             label={t('From Account')}
                             labelId="transferFrom"
                             name="transferFrom"
                             value={transferFrom}
-                            disabled={!isNew}
-                            onChange={handleChange}
-                            onBlur={handleBlur}
+                            readOnly
```
kegrimes, line 316: Could we add something like this to make it unclickable?
```suggestion
                            readOnly
                            tabIndex={-1}
                            sx={{ pointerEvents: 'none' }}
```

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
kegrimes: Do we need this anymore?

kegrimes (approved):
I think this looks very good Zach!

Honestly, I did like the original design a lot, but I understand why he would want this change. It was really repetitive to have 6 buttons for 3 different fund types. I think that in terms of money, most people would be thinking in the direction of "transfer from --> transfer to" and not really the other way around.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
zweatshirt: I am leaning toward making it clickable (hoverable) for accessibility if that's ok. Otherwise users with screen readers can't know what the from account is. At least that was the intention. With that being said, I should test what the screen reader actually produces on hover here.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
zweatshirt: Great point, probably not

PR #1814, opened 2026-06-01
MPDX-9650 - Improve validation in Savings Fund Transfer

zweatshirt:
## Description

- Ensures the start and end date can't be the same date, since SAA has validation that raises in error for this case.
- Adds information text for the recurring end date
https://jira.cru.org/browse/MPDX-9580

## Testing

- Go to SavingsFundTransfer
- Ensure that the user cannot set the start date and the end date to be the same date.

## Checklist:

### DIFF
--- a/src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
@@ -119,14 +119,14 @@ const transferSchema = (locale: string) =>
         is: (schedule: ScheduleEnum) => schedule !== ScheduleEnum.OneTime,
         then: (schema) =>
           schema.test(
-            'end>=start',
-            i18n.t('End date must be after transfer date'),
+            'end>start',
+            i18n.t('End date must be at least one day after the transfer date'),
             function (end) {
               const start = this.parent.transferDate as DateTime | null;
               if (!end || !start) {
                 return true;
               }
-              return end.toMillis() >= start.toMillis();
+              return end.startOf('day') > start.startOf('day');
             },
           ),
         otherwise: (schema) => schema.notRequired(),
@@ -491,7 +491,11 @@ export const TransferModal: React.FC<TransferModalProps> = ({
                             }}
                             error={touched.endDate && Boolean(errors.endDate)}
                             helperText={
-                              touched.endDate && (errors.endDate as string)
+                              touched.endDate && errors.endDate
+                                ? errors.endDate
+                                : t(
+                                    'The transfer will no longer recur after this date. If left blank, the transfer will recur indefinitely until manually stopped.',
+                                  )
                             }
                           />
                         </Grid>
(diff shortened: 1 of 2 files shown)


src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
```diff
@@ -491,7 +491,11 @@ export const TransferModal: React.FC<TransferModalProps> = ({
                             }}
                             error={touched.endDate && Boolean(errors.endDate)}
                             helperText={
-                              touched.endDate && (errors.endDate as string)
+                              touched.endDate && errors.endDate
```
zweatshirt, line 494: [Suggestion] The `touched.endDate && errors.endDate ? errors.endDate : t(...)` expression is correct but slightly over-conditioned — the leading `touched.endDate &&` only gates the error branch, while the default helper text always renders in the false branch. Behavior is verified correct across all touched/error states, so this is optional polish, not a fix.

zweatshirt (commented):

**Mode:** standard · **Risk:** 1/10 (LOW) · **Blockers:** 0

Agents run: Architecture, UX/a11y, Financial (domain). Testing + Standards were verified inline (their key concerns are addressed below). This is a small, focused change — end-date validation now requires the end date to be **strictly at least one calendar day after** the transfer date (day-granular Luxon comparison), plus a permanent descriptive helper text on the End Date field, with 4 colocated tests.

### Findings
Only two severity-3 suggestions (informational — do not block merge). See the inline comment for the first; the second is below.

### Verified — concerns from skipped agents
- **Type safety:** Dropping the `as string` cast on `errors.endDate` is correct and type-safe. Because `endDate` is `DateTime<boolean> | null`, Formik's `FormikErrors` mapped type resolves it to `string | undefined` (the `| null` defeats the `extends object` branch). The sibling `transferDate` (non-nullable) resolves to a `FormikErrors<DateTime>` object — which is exactly why line 478 still needs its cast. `tsc --noEmit` is clean.
- **Tests using past dates:** `12/01/2024` also fails the recurring `start-date` rule, but each new test asserts only the end-date error's presence/absence, which is independent and correct in isolation. Suite passes. Minor smell — future dates would read cleaner — but not a defect.

### Findings on Related Files (Not on Changed Lines)

**[Suggestion] TransferModal.tsx:478** — Cast asymmetry between sibling date fields
- Severity: 3.0/10 · Flagged by: Architecture
- The End Date helper (this PR) now passes `errors.endDate` directly, while the adjacent Transfer Date helper at line 478 still uses `(errors.transferDate as string)`. The new pattern is the better one; line 478 is the outlier and is out of scope here. Consider dropping that cast if the line is touched in a future PR.

### Domain note (non-blocking, pre-existing)
The submit path sends `recurringStart`/`recurringEnd` as full `toISO()` timestamps with a local offset, while validation compares at day granularity. Not reachable as a bug here (both fields are local-midnight date-only values), but worth a one-line confirmation with the API team that the backend treats these as **calendar dates**, not raw UTC instants — to avoid a boundary day-shift. This concerns the pre-existing `toISO()` submission, not the lines this PR adds.

PR #1781, opened 2026-05-18
MPDX-9452 - Maintenance: Remove SITE_URL

zweatshirt:
## Description

Remove `SITE_URL`

Follow up: Remove the SITE_URL variable from Amplify env vars once this is merged.

## Testing

- Go to the preview environment
- Test login and logout behavior
- Test any number of pages where API calls may be affected by these changes.

## Checklist:

- [x] I have given my PR a title with the format "MPDX-(JIRA#) (summary sentence max 80 chars)"
- [x] I have applied the appropriate labels (_Add the label "Preview" to automatically create a preview environment_)
- [x] I have run the Claude Code `/pr-review` command locally and fixed any relevant suggestions
- [x] I have requested a review from another person on the project
- [ ] I have tested my changes in preview or in staging
- [x] I have cleaned up my commit history

PR #1816, opened 2026-06-03
HS-1639514 - Fix Google Import, MailChimp, Organization, and PrayerLetters redirects

zweatshirt:
## Description

Caused by a minor regression from https://github.com/CruGlobal/mpdx-react/pull/1781
- https://secure.helpscout.net/conversation/3341814677/1639514?viewId=8745276

#### Issue:

useOauthUrl.ts built the OAuth button URLs by reading window.location.origin directly inside getRedirectUrl. The integration accordions call these getters during render to set each button's href, and render also runs on the server. 

On a full page load (the OAuth redirect landing, a direct URL, or a refresh), window is undefined on the server, so the read threw ReferenceError: window is not defined and crashed the page render. This was introduced by PR #1781 (Remove SITE_URL), which deleted the env var that had been the SSR-safe branch of `process.env.SITE_URL || window.location.origin`, leaving only the window read.

Fix: Resolve the origin in a client side only useEffect and hold it in state that starts
as an empty string. 

SSR and the first hydration render use the empty value, so no window access happens during render and the server/client markup matches. After mount, the effect sets the real window.location.origin and rerenders,
patching the button hrefs with the user's host. This keeps the redirect correct across all domains without depending on a build-time env var. Fixes all four OAuth flows (Google, MailChimp, DonorHub/Organization, prayerletters.com),
which share getRedirectUrl.

## Testing

- Go to MPDX Tools > Import from Google > Select 'Connect Google Account'
    - In production this will cause a server side error, in the preview environment this will resolve correctly. 
- You can test also that refreshes in the Settings work when the PrayerLetters and MailChimp accordions are open as well.

## Checklist:

### DIFF
--- a/src/components/Settings/integrations/useOauthUrl.test.tsx
+++ b/src/components/Settings/integrations/useOauthUrl.test.tsx
@@ -1,0 +1,79 @@
+import React, { PropsWithChildren } from 'react';
+import { renderHook } from '@testing-library/react';
+import { Session } from 'next-auth';
+import { useSession } from 'next-auth/react';
+import { renderToString } from 'react-dom/server';
+import TestRouter from '__tests__/util/TestRouter';
+import { IntegrationAccordion } from 'src/components/Shared/Forms/Accordions/AccordionEnum';
+import { useOauthUrl } from './useOauthUrl';
+
+jest.mock('next-auth/react');
+
+const accountListId = 'account-list-1';
+const apiToken = 'apiToken';
+
+const Wrapper = ({ children }: PropsWithChildren) => (
+  <TestRouter router={{ query: { accountListId } }}>{children}</TestRouter>
+);
+
+const renderUseOauthUrl = () =>
+  renderHook(() => useOauthUrl(), { wrapper: Wrapper });
+
+beforeEach(() => {
+  process.env.OAUTH_URL = 'https://auth.mpdx.org';
+  (useSession as jest.Mock).mockReturnValue({
+    data: { user: { apiToken } } as Session,
+  });
+});
+
+describe('useOauthUrl', () => {
+  it('builds the Google OAuth url with the redirect back to integrations', () => {
+    const { result } = renderUseOauthUrl();
+
+    expect(result.current.getGoogleOauthUrl()).toBe(
+      `https://auth.mpdx.org/auth/user/google?account_list_id=${accountListId}&redirect_to=${encodeURIComponent(
+        `http://localhost/accountLists/${accountListId}/settings/integrations?selectedTab=${IntegrationAccordion.Google}`,
+      )}&access_token=${apiToken}`,
+    );
+  });
+
+  it('includes the organization id in the DonorHub OAuth url', () => {
+    const { result } = renderUseOauthUrl();
+
+    expect(result.current.getOrganizationOauthUrl('org-1')).toBe(
+      `https://auth.mpdx.org/auth/user/donorhub?account_list_id=${accountListId}&redirect_to=${encodeURIComponent(
+        `http://localhost/accountLists/${accountListId}/settings/integrations?selectedTab=${IntegrationAccordion.Organization}`,
+      )}&access_token=${apiToken}&organization_id=org-1`,
+    );
+  });
+
+  it('renders on the server when window is undefined', () => {
+    const Probe = () => {
+      const { getGoogleOauthUrl } = useOauthUrl();
+      return <a href={getGoogleOauthUrl()}>{'link'}</a>;
+    };
+
+    const originalWindow = global.window;
+    // @ts-expect-error simulating a server environment with no window
+    delete global.window;
+
+    try {
+      let html = '';
+      expect(() => {
+        html = renderToString(
+          <Wrapper>
+            <Probe />
+          </Wrapper>,
+        );
+      }).not.toThrow();
+
+      expect(html).toContain(
+        `redirect_to=${encodeURIComponent(
+          `/accountLists/${accountListId}/settings/integrations?selectedTab=${IntegrationAccordion.Google}`,
+        )}`,
+      );
+    } finally {
+      global.window = originalWindow;
+    }
+  });
+});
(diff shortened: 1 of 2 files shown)


src/components/Settings/integrations/useOauthUrl.ts
```diff
@@ -6,12 +7,15 @@ export const useOauthUrl = () => {
...
-  const getRedirectUrl = (accordion: IntegrationAccordion) => {
-    const domain = window.location.origin;
-    return encodeURIComponent(
-      `${domain}/accountLists/${accountListId}/settings/integrations?selectedTab=${accordion}`,
+  const [origin, setOrigin] = useState('');
+  useEffect(() => {
+    setOrigin(window.location.origin);
+  }, []);
+
+  const getRedirectUrl = (accordion: IntegrationAccordion) =>
```
zweatshirt, line 15: [Suggestion] During SSR and the first client render `origin` is `''`, so `redirect_to` is host-relative until the post-mount effect fires. This is functionally safe — these getters are only navigated on user click (post-hydration), and the Organization flows build the URL at click time, so `origin` is always populated by then; no open-redirect risk (no user input in the target). Noted for awareness only. If you ever want server-rendered links to be absolute, you could fall back to `window.location.origin` at call time when `origin === ''` — but that adds branching for a near-impossible race and isn't recommended here.

src/components/Settings/integrations/useOauthUrl.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+    const { result } = renderUseOauthUrl();
+
+    expect(result.current.getGoogleOauthUrl()).toBe(
+      `https://auth.mpdx.org/auth/user/google?account_list_id=${accountListId}&redirect_to=${encodeURIComponent(
+        `http://localhost/accountLists/${accountListId}/settings/integrations?selectedTab=${IntegrationAccordion.Google}`,
+      )}&access_token=${apiToken}`,
+    );
+  });
+
+  it('includes the organization id in the DonorHub OAuth url', () => {
```
zweatshirt, line 40: [Suggestion] `getMailChimpOauthUrl` and `getPrayerlettersOauthUrl` aren't asserted here, so their distinct path segments (`/auth/user/mailchimp`, `/auth/user/prayer_letters` — note the underscore) aren't pinned by this unit test. They are covered with full-URL assertions at the component level (`MailchimpAccordion.test.tsx:143`, `PrayerlettersAccordion.test.tsx:112-117`), so this is minor. An `it.each` over all four getters would close the gap cheaply (`getOrganizationOauthUrl` stays separate since it takes an arg).

zweatshirt (commented):
## 🤖 Multi-Agent Code Review — ✅ CLEAN

**4 agents** (Security, Architecture, Testing, Standards) · **standard** mode · **Risk: LOW (1/10)**

No blockers, no important issues, no standards violations. This is a correct, idiomatic, net-debt-reducing SSR-safety fix. Six low-severity suggestions (all < 2.5/10), all informational.

### What the change does
Defers the `window.location.origin` read out of render and into a `useEffect` (`useState('')` default), fixing the SSR `ReferenceError: window is not defined` crash on the integrations page (introduced when `SITE_URL` was removed). Server + first hydration render use `''` (markup matches → clean hydration), then the effect populates the real origin and React patches the button hrefs.

### Consensus theme — transient host-relative `redirect_to` (all 4 agents)
During SSR / pre-hydration, `origin` is `''`, so `redirect_to` is host-relative until the effect runs. **All agents independently concluded this is functionally safe**: the `href` getters are only navigated on user click (post-hydration), and the Organization flows build the URL at click time, so `origin` is always populated by interaction time. No open-redirect risk (no user input flows into the target). Strictly better than the crash it replaces. Optional belt-and-suspenders (fall back to `window.location.origin` at call time when `origin === ''`) was considered and **not** recommended as required.

### Notes
- **Architecture:** the `useState('')`+`useEffect` choice is the correct hydration-safe pattern; explicitly *don't* copy the `typeof window` lazy-initializer from `useLocalStorage.ts` (would cause a hydration mismatch here). `useEffect([])` deps correct. Net technical debt reduced.
- **Standards:** all applicable checklist items PASS — named export, hook naming/colocation, no `any`, `@ts-expect-error` carries an explanatory comment, exact full-URL test assertions, `window` read confined to an effect.
- **"Fix one, fix all":** verified the only other `window.location.origin` reference is in `_document.page.tsx` inside a browser-executed `<script>` string (safe). No remaining instances of this bug — fix is complete.

### Suggestions (informational, no `/dismiss` needed)
- Add `getMailChimpOauthUrl` / `getPrayerlettersOauthUrl` to this hook's test (their path segments `mailchimp` / `prayer_letters` aren't pinned here — though both are covered at the component level). An `it.each` over all four getters closes it cheaply.
- Optional: assert the no-session error path (`useRequiredSession` throws).

| Security | 0 | 0 | 0 | 1 | High |
| Architecture | 0 | 0 | 0 | 2 | High |
| Testing | 0 | 0 | 0 | 3 | High |
| **Total** | **0** | **0** | **0** | **6** | High |

<sub>Suggestions are informational and do not block a CLEAN verdict.</sub>

src/components/Settings/integrations/useOauthUrl.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+const Wrapper = ({ children }: PropsWithChildren) => (
+  <TestRouter router={{ query: { accountListId } }}>{children}</TestRouter>
+);
+
+const renderUseOauthUrl = () =>
+  renderHook(() => useOauthUrl(), { wrapper: Wrapper });
+
+beforeEach(() => {
+  process.env.OAUTH_URL = 'https://auth.mpdx.org';
```
zweatshirt, line 23: @canac Hi Caleb, I had a quick question if that's okay. I assume this wouldn't cause cross-file leakage of `process.env.OAUTH_URL`? In this case, is it best to add an `afterEach` anyway?

dr-bizz (approved):
Looks great! thank you for fixing this

src/components/Settings/integrations/useOauthUrl.test.tsx
dr-bizz: This is fine

PR #1830, opened 2026-06-09
[no-Jira] Show all pages to developers during development

canac:
## Description

When `DEVELOPMENT_ENV=true` is set, disable page visibility checks, except in production.

I set `DEVELOPMENT_ENV=true` in Amplify to all environments, and made a `DEVELOPMENT_ENV=false` override for `main`.

When impersonating someone, you will only see the pages they have access to. I think this is good.

Sorry for the opportunistic cleanup of session mocking code. I let Claude follow a rabbit trail 🐰

## Testing

- Login with the test account (senior staff)
- Check that the PDS goal calculator page shows up in the menu and via direct navigation

## Checklist:

### DIFF
--- a/src/components/Shared/UserTypeAccess/UserTypeAccess.tsx
+++ b/src/components/Shared/UserTypeAccess/UserTypeAccess.tsx
@@ -1,6 +1,7 @@
 import Loading from 'src/components/Loading';
 import { UserTypeEnum } from 'src/graphql/types.generated';
 import { useIneligibleByGroup } from 'src/hooks/useIneligibleByGroup';
+import { useRequiredSession } from 'src/hooks/useRequiredSession';
 import { LimitedAccess } from '../LimitedAccess/LimitedAccess';
 
 export enum RequiredUserGroupEnum {
@@ -39,6 +40,7 @@ export const UserTypeAccess: React.FC<UserTypeAccessProps> = ({
     userLoading,
     userError,
   } = useIneligibleByGroup();
+  const { developer } = useRequiredSession();
 
   const ineligibleByGroup: Record<RequiredUserGroupEnum, boolean> = {
     [RequiredUserGroupEnum.Asr]: inAsrIneligibleGroup,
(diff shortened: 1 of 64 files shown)


src/components/Shared/UserTypeAccess/UserTypeAccess.tsx
```diff
@@ -51,7 +53,7 @@ export const UserTypeAccess: React.FC<UserTypeAccessProps> = ({
     (requireUserGroups && ineligibleByGroup[requireUserGroups]);
 
   // Once HCM is ready to go live and DISABLE_NEW_REPORTS is removed, we can remove the alwaysAllow prop
-  if (alwaysAllow) {
+  if (alwaysAllow || developerBypass) {
```
kegrimes: praise: this is great and very simple!

kegrimes (approved):
This looks good Caleb, thanks for doing this! I agree that impersonation should still have restriction so that is amazing!

One thing I did notice was that even with this change, any test account without a staff account id will have access to these reports in the nav bar, however, they will get a "no staff account" page or an error from HCM. This is what my personal account is showing but since we have the test account it shouldn't be a big deal.

canac:
> One thing I did notice was that even with this change, any test account without a staff account id will have access to these reports in the nav bar, however, they will get a "no staff account" page or an error from HCM.

@kegrimes That's true. I was thinking that developers would see the GraphQL error toast and be able to figure out what went wrong. But I could hide those staff account required pages if you think that would be helpful. Since they're not going to be able to do much with them on an user without a staff account.

kegrimes:
@canac Yeah, I was thinking maybe we could adjust `UserTypeAccess` to check for staff account id first? That way developers can still have access to all reports and understand why they can't view them. I feel like that may be a better solution because then there won't be all these exceptions that we have to remember (i.e. you can view all reports unless you don't have a staff account id). It can just be developers can see all reports and non-developers cannot. What do you think?

PR #1831, opened 2026-06-09
MPDX-9703 Fix Google Calendar sync for To Do / Letter-Card activity types

jbirdjavi:
## Summary

Turning on Google Calendar sync for tasks of certain activity types throws a graphql-js error and blocks sync:

```
Enum "ActivityTypeEnum" cannot represent value: "INITIATION_TO_DO"
Enum "ActivityTypeEnum" cannot represent value: "FOLLOW_UP_LETTER_CARD"
Enum "ActivityTypeEnum" cannot represent value: "PARTNER_CARE_LETTER_CARD"
Enum "ActivityTypeEnum" cannot represent value: "FOLLOW_UP_TO_DO"
```

## Root cause

The Google integration is a REST endpoint in mpdx_api, which this app stitches into GraphQL as its own subgraph. That subgraph keeps a hand-maintained copy of `ActivityTypeEnum` in `googleAccountIntegrations.graphql`, and it drifted out of sync with the canonical enum when the **To Do** and **Letter/Card** activity types were added (mpdx_api, March 2025).

`GoogleAccountIntegration.calendarIntegrations: [ActivityTypeEnum!]!` is resolved against this incomplete enum. `parse.ts` uppercases the REST values (`follow_up_to_do` → `FOLLOW_UP_TO_DO`), and graphql-js then fails to serialize them against the subgraph enum that's missing those values.

The merged client schema (`src/graphql/schema.graphql`, `types.generated.ts`) already had all values, which is why this hid for so long. The runtime subgraph schema is built from this typedef alone, so it threw.

## Change

Add the four missing values to the subgraph `ActivityTypeEnum` so it matches the canonical list:

- `FOLLOW_UP_LETTER_CARD`
- `FOLLOW_UP_TO_DO`
- `INITIATION_TO_DO`
- `PARTNER_CARE_LETTER_CARD`

The subgraph enum is now byte-identical to `src/graphql/schema.graphql`. No codegen artifacts change (they were already complete from the merged schema).

## Follow-up

Longer term, this enum should be sourced from the shared schema rather than maintained as a second hand-written copy, to prevent future drift.

Jira: MPDX-9703 (surfaced during QA of MPDX-8537)


canac (approved):
Looks good! Is there an account we can impersonate and test this on?

jbirdjavi:
@canac Scott Mitchell said he encountered it on staging.

canac:
I'm having trouble reproducing the issue and proving the fix works, but I trust you. And I know we want those two enum definitions to be sychronized.

jbirdjavi:
@canac Oops I forgot about this PR. Is it still good to merge?

canac:
> Is it still good to merge?

I don't see why not!

PR #1834, opened 2026-06-10
MPDX-9705 - Staff expense aggregates as default

zweatshirt:
## Description

Scott would like categories to be aggregated by default in the staff expense report. This begs the question of whether we want to keep this filtering logic at all. The codebase would benefit from removing it, but after my meeting with Scott earlier he hinted that some users may still like the option to remove this aggregate behavior and that he would follow up with @rguinee

This PR also has some needed test file refactors to DRY things up.

https://jira.cru.org/browse/MPDX-9705

## Testing

- Go to Staff Expense Report
- Test that all transaction and income rows are already pre-aggregated into their base categories.
- Test that going to the filter settings allows a user to be able to deselect aggregation on whichever categories they choose, and that this behavior works as expected.

## Checklist:

### DIFF
--- a/src/components/Reports/StaffExpenseReport/Helpers/filterTransactions.tsx
+++ b/src/components/Reports/StaffExpenseReport/Helpers/filterTransactions.tsx
@@ -41,16 +41,20 @@ interface FilterTransactionsParams {
  */
 const groupTransactionsByCategory = (
   transactions: Transaction[],
-  groupedCategories: string[],
+  groupedCategories: string[] | null,
   fundType: string,
   t: TFunction,
 ): (Transaction | GroupedTransaction)[] => {
-  // Group transactions by category if the category is in groupedCategories
+  // Group a transaction when its category is selected. `null` means no explicit
+  // selection, so every category is consolidated.
   const grouped: Map<string, Transaction[]> = new Map();
   const ungrouped: Transaction[] = [];
 
   transactions.forEach((transaction) => {
-    if (groupedCategories.includes(transaction.category)) {
+    if (
+      groupedCategories === null ||
+      groupedCategories.includes(transaction.category)
+    ) {
       const key = transaction.category;
       if (!grouped.has(key)) {
         grouped.set(key, []);
@@ -108,7 +112,6 @@ export const filterTransactions = ({
   tableType,
 }: FilterTransactionsParams): (Transaction | GroupedTransaction)[] => {
   const isInDateRange = createDateRangeFilter(filters, targetTime);
-  const selectedCategories = filters?.categories ?? [];
 
   const filteredTransactions =
     fund.categories?.flatMap((category) =>
@@ -152,9 +155,11 @@ export const filterTransactions = ({
       ),
     ) ?? [];
 
+  const groupedCategories = filters?.categories ?? null;
+
   return groupTransactionsByCategory(
     filteredTransactions,
-    selectedCategories,
+    groupedCategories,
     fund.fundType,
     t,
   );
--- a/src/components/Reports/StaffExpenseReport/SettingsDialog/SettingsDialog.test.tsx
+++ b/src/components/Reports/StaffExpenseReport/SettingsDialog/SettingsDialog.test.tsx
@@ -93,10 +93,17 @@ const TestComponent: React.FC<
   </TestRouter>
 );
 
+const baseFilters: Filters = {
+  selectedDateRange: null,
+  startDate: null,
+  endDate: null,
+  categories: [],
+};
+
 describe('SettingsDialog', () => {
   const defaultProps = {
     isOpen: true,
-    onClose: jest.fn(),
+    onClose: mutationSpy,
     selectedFundType: 'Primary',
   };
 
@@ -152,9 +159,8 @@ describe('SettingsDialog', () => {
 
   it('should populate form with selectedFilters when provided', async () => {
     const selectedFilters: Filters = {
+      ...baseFilters,
       selectedDateRange: DateRange.MonthToDate,
-      startDate: null,
-      endDate: null,
       categories: [
         StaffExpenseCategoryEnum.Benefits,
         StaffExpenseCategoryEnum.Salary,
@@ -181,10 +187,9 @@ describe('SettingsDialog', () => {
 
   it('should preserve custom dates when no selectedDateRange is set', () => {
     const selectedFilters: Filters = {
-      selectedDateRange: null,
+      ...baseFilters,
       startDate: DateTime.fromISO('2025-01-01'),
       endDate: DateTime.fromISO('2025-01-31'),
-      categories: [],
     };
 
     const { getByLabelText } = render(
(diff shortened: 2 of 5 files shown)


zweatshirt (commented):
## 🤖 Multi-Agent Review — PR #1834 (Staff expense aggregates)

**Verdict: ✅ APPROVED WITH SUGGESTIONS** · Risk: **LOW (3/10)** · Blockers: **0**

6 agents reviewed (Architecture, Data Integrity, Financial Reporting, Testing, UX, Standards). No critical/high findings — **no data-integrity or financial-correctness issues** (grouping is provably total-preserving; the report is single-currency USD, so the larger default summation set introduces no currency mixing), and **all Standards checklist items pass**. Everything below is a non-blocking suggestion (all severity < 7, all dismissible via `/dismiss: <reason>`).

### Themes
- **One thing worth a look before merge:** the dialog and the report derive "all categories" two different ways (dialog: all-funds `getAvailableCategories`; report: per-fund inline `new Set`). Harmless today, but a latent grouping drift in multi-fund reports — see the inline comment on `filterTransactions.tsx`.
- A couple of test-robustness nits (`toHaveLength(4)` brittleness, `mutationSpy` reused as `onClose` + `onCall`) and a UX copy tweak.

### Financial Checklist (domain agent)
- Arithmetic on money safe: **Yes** (rounding only at the `currencyFormat` display boundary)
- Currency mixing prevented: **N/A** (single-currency USD report; no `currencyCode` on transactions)
- Rounding at display boundary only: **Yes**
- Null/undefined amounts handled: **Yes**
- Luxon used (no `new Date()`): **Yes** (none introduced)
- Totals unchanged by default grouping: **Yes** (grouping is presentational; section totals reduce the same underlying transactions)

Inline comments below carry machine-readable severity tags.

src/components/Reports/StaffExpenseReport/Helpers/filterTransactions.tsx
```diff
@@ -152,6 +151,15 @@ export const filterTransactions = ({
       ),
     ) ?? [];
 
+  // When no explicit category selection exists, default to
+  // consolidating every category present. An explicit empty array means the
+  // user deselected all categories, so nothing is grouped.
+  const selectedCategories =
```
zweatshirt: [Medium] This default-category derivation (`new Set` over a single fund's `filteredTransactions`) is a second, divergent definition of "all categories" — the dialog seeds toggles from `getAvailableCategories` (SettingsDialog.tsx:221), which scans **all funds** and sorts. Both mean "all available categories" but over different inputs.

Harmless today because `groupTransactionsByCategory` only groups categories actually present. But in a multi-fund report the *same visual state* (all boxes checked) can produce *different grouping* before vs. after a no-op check→uncheck→recheck, since one path is per-fund-derived and the other is dialog/all-funds-derived. This is the duplicated-calculation drift CLAUDE.md warns about, applied to category aggregation.

Suggested: extract "categories present in a transaction list" into one helper and reuse it for both the dialog default and this report default (single source of truth).

src/components/Reports/StaffExpenseReport/SettingsDialog/SettingsDialog.tsx
```diff
@@ -213,6 +217,16 @@ export const SettingsDialog: React.FC<SettingsDialogProps> = ({
           validateForm,
           setTouched,
         }) => {
+          const handleCategoryToggle = (category: string, checked: boolean) => {
```
zweatshirt, line 220: [Suggestion] Two small notes on `handleCategoryToggle`:

1. It seeds from `availableCategories` (all funds) — the other half of the "two definitions of all categories" concern flagged in `filterTransactions.tsx`.
2. The re-check branch (`checked ? [...selectedCategories, category]`) is currently untested — tests only ever uncheck. Worth an uncheck-then-recheck test asserting the submitted array contains the category exactly once. (The theoretical `null`+check duplicate is unreachable via UI, since the default is all-checked so the first toggle is always an uncheck.)

src/components/Reports/StaffExpenseReport/SettingsDialog/SettingsDialog.tsx
```diff
@@ -40,7 +40,11 @@ export interface Filters {
   selectedDateRange: DateRange | null;
   startDate?: DateTime | null;
   endDate?: DateTime | null;
-  categories?: string[];
+  /**
+   * `null` means no explicit selection: every available category is shown as
+   * checked and the report aggregates all of them.
+   */
+  categories?: string[] | null;
```
zweatshirt: [Suggestion] `categories?: string[] | null` is simultaneously optional and nullable, giving three states (`undefined`, `null`, `string[]`), but only two are meaningful — every read site collapses `undefined` and `null` together via `??` / `!values.categories`. Dropping the `?` (→ `categories: string[] | null`) most accurately models the two real states (default vs. explicit), since `initialValues` already normalizes to `null`. Minor — behavior is correct.

src/components/Reports/StaffExpenseReport/SettingsDialog/SettingsDialog.tsx
```diff
@@ -310,8 +324,8 @@ export const SettingsDialog: React.FC<SettingsDialogProps> = ({
 
                 <Typography sx={{ mt: 2, whiteSpace: 'pre-line' }}>
                   {t(
-                    `You can combine certain categories of data into single rows. This may be useful for long date ranges (e.g., "Year to Date").
-                    Select which categories to consolidate. Each category remains separate.`,
+                    `Income and expenses are combined by categories by default. This may be useful for long date ranges (e.g., "Year to Date").
+                    Select which categories to keep consolidated.`,
```
zweatshirt, line 328: [Medium] Copy clarity: with all categories checked by default, the interaction is *subtractive* (uncheck to split a category into separate rows), but "Select which categories to keep consolidated" reads as *additive* and doesn't cue that unchecking is what produces separate rows. Consider: "All categories are consolidated by default. Uncheck a category to show its income and expenses as separate rows."

(Pre-existing, optional: this multi-line template literal inside `t()` bakes source indentation into the translation key — since you're re-wording the string anyway, a single-line string would give a cleaner extracted key.)

src/components/Reports/StaffExpenseReport/SettingsDialog/SettingsDialog.test.tsx
```diff
@@ -93,10 +93,17 @@ const TestComponent: React.FC<
...
+  startDate: null,
+  endDate: null,
+  categories: [],
+};
+
 describe('SettingsDialog', () => {
   const defaultProps = {
     isOpen: true,
-    onClose: jest.fn(),
+    onClose: mutationSpy,
```
zweatshirt, line 106: [Medium] `mutationSpy` now doubles as the `onClose` spy here *and* the GraphQL `onCall` spy in the query tests. Safe today (the `toHaveBeenCalledWith({...filters})` tests don't pass `onCallMock`), but fragile: if a future test ever passes `onCallMock={mutationSpy}` *and* asserts `toHaveBeenCalledWith` on a filter object, GraphQL onCall payloads would land in the same call list and a *missing* `onClose` could be silently masked (the matcher only needs one matching call). Consider separate `onCloseSpy` / `onCallSpy`.

src/components/Reports/StaffExpenseReport/StaffExpenseReport.test.tsx
```diff
@@ -247,33 +247,28 @@ describe('StaffExpenseReport', () => {
...
+      <TestComponent />,
+    );
 
     await findByRole('heading', { name: 'Primary' });
-    expect(await findByText('Currently Viewing')).toBeInTheDocument();
 
-    waitFor(() => {
-      expect(getAllByRole('row')).toHaveLength(3);
-    });
+    expect(await findAllByText('-$200')).toHaveLength(4);
```
zweatshirt: [Medium] `toHaveLength(4)` is brittle — the count of 4 is an emergent property of internal structure (screen + print renderers × row + section-total, each emitting the amount). The fund-switch is already proven by the present-then-absent `queryByText` assertions; the exact `4` adds nothing and breaks on any layout change (dropping the print table, adding a footer, etc.). Recommend asserting presence (`length >= 1`) or presence-then-absence only, and dropping the exact count. Same applies to the `-$100` assertion on line 270.

PR #1836, opened 2026-06-11
[no-Jira] Add custom `toHaveTableStructure` jest matcher

canac:
## Description

Create a custom `expect(table).toHaveTableStructure(...)` just matcher for verifying the headings and cells of a table in a single `expect`.

## Testing

- Verify that tests still pass

## Checklist:

### DIFF
--- a/__tests__/extensions/toHaveTableStructure.test.tsx
+++ b/__tests__/extensions/toHaveTableStructure.test.tsx
@@ -59,6 +59,8 @@ describe('toHaveTableStructure', () => {
     const { getByRole } = render(<TestTable />);
 
     expect(getByRole('table')).toHaveTableStructure({
+      columnHeaders: ['Category', expect.stringContaining('Jo'), 'Jane'],
+      rowHeaders: [expect.stringContaining('Sal'), 'MHA'],
       cells: [
         [expect.stringContaining('10,001'), '$20,001.00'],
         ['$10,002.00', '$20,002.00'],
@@ -86,6 +88,51 @@ describe('toHaveTableStructure', () => {
     expect(getByRole('table')).not.toHaveTableStructure({
       rowHeaders: ['Salary', 'MHA', 'Total'],
     });
+  });
+
+  it('matches an empty table', () => {
+    const { getByRole } = render(
+      <table>
+        <tbody></tbody>
+      </table>,
+    );
+
+    expect(getByRole('table')).toHaveTableStructure({
+      cells: [],
+      columnHeaders: [],
+      rowHeaders: [],
+    });
+  });
+
+  it('throws when no structure properties are provided', () => {
+    const { getByRole } = render(<TestTable />);
+
+    expect(() => expect(getByRole('table')).toHaveTableStructure({})).toThrow(
+      'toHaveTableStructure requires at least one of cells, columnHeaders, or rowHeaders',
+    );
+  });
+
+  it('describes the mismatched properties in the failure message', () => {
+    const { getByRole } = render(<TestTable />);
+
+    expect(() =>
+      expect(getByRole('table')).toHaveTableStructure({
+        columnHeaders: ['Category', 'John'],
+        rowHeaders: ['Salary', 'MHA'],
+      }),
+    ).toThrow(
+      /Expected table to match the expected structure[\s\S]*columnHeaders:[\s\S]*Expected columnHeaders[\s\S]*Received columnHeaders/,
+    );
+  });
+
+  it('explains the failure when the table unexpectedly matches', () => {
+    const { getByRole } = render(<TestTable />);
+
+    expect(() =>
+      expect(getByRole('table')).not.toHaveTableStructure({
+        rowHeaders: ['Salary', 'MHA'],
+      }),
+    ).toThrow('Expected table not to match the expected structure');
   });
 
   it('only inspects the received table', () => {
--- a/__tests__/extensions/toHaveTableStructure.test.tsx
+++ b/__tests__/extensions/toHaveTableStructure.test.tsx
@@ -1,0 +1,112 @@
+import { render } from '@testing-library/react';
+
+const TestTable: React.FC = () => (
+  <table>
+    <thead>
+      <tr>
+        <th scope="col">Category</th>
+        <th scope="col">John</th>
+        <th scope="col">Jane</th>
+      </tr>
+    </thead>
+    <tbody>
+      <tr>
+        <th scope="row">Salary</th>
+        <td>$10,001.00</td>
+        <td>$20,001.00</td>
+      </tr>
+      <tr>
+        <th scope="row">MHA</th>
+        <td>$10,002.00</td>
+        <td>$20,002.00</td>
+      </tr>
+    </tbody>
+  </table>
+);
+
+describe('toHaveTableStructure', () => {
+  it('matches the full table structure', () => {
+    const { getByRole } = render(<TestTable />);
+
+    expect(getByRole('table')).toHaveTableStructure({
+      columnHeaders: ['Category', 'John', 'Jane'],
+      rowHeaders: ['Salary', 'MHA'],
+      cells: ['$10,001.00', '$20,001.00', '$10,002.00', '$20,002.00'],
+    });
+  });
+
+  it('matches 2-dimensional arrays of cells', () => {
+    const { getByRole } = render(<TestTable />);
+
+    expect(getByRole('table')).toHaveTableStructure({
+      columnHeaders: [['Category'], ['John', 'Jane']],
+      cells: [
+        ['$10,001.00', '$20,001.00'],
+        ['$10,002.00', '$20,002.00'],
+      ],
+    });
+  });
+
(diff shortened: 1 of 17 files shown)


canac (commented):
## Multi-Agent Code Review

**Verdict: APPROVED WITH SUGGESTIONS** — no blockers. 1 medium-priority improvement (document the matcher's flatten contract), 6 informational suggestions.

**Risk:** 5/10 (MEDIUM) — `.claude/rules/code-review.md` change (verified strictly additive and review-strengthening) + new shared test infrastructure registered in the global jest setup (~708 test files blast radius, purely additive change).

**Mode:** standard — 5 agents (Security, Architecture, Testing, Standards, Financial Reporting) + dependency impact analysis, with cross-examination debate round.

### Highlights

- **Migration fidelity verified value-by-value across all 10 test files: zero transcription errors.** Every monetary string, percentage, and `–` empty-state marker is byte-identical old→new.
- Net coverage **increased**: `RequestedSalaryCard` input cells previously excluded via `.slice(0, -2)` are now asserted; `SalarySummaryCard` now scopes assertions per-table via `within()`, catching cross-table value bleed the old document-wide assertion missed.
- The matcher mirrors the existing `toHaveGraphqlOperation` pattern exactly (placement, typing, registration, declaration merge). Lint, lint:ts, and the matcher's own 7-test suite all pass.
- Review-rules change verified line-by-line: single additive bullet, no weakening of risk scoring, thresholds, or checks.
- Debate round converged: the flattening finding was rated 6.0 by two agents initially, revised to 3.0–3.5 by its originators after proof that the old code had identical flat semantics (zero regression), while the Financial domain agent raised it to 5.0 on forward-looking grounds (new shared API invites future authors to trust row grouping that isn't checked). Highest post-debate severity governs: 5.0.

These are informational and do not count toward the verdict.

**[Suggestion] jest.config.js** — `collectCoverageFrom` references `__tests__/util/extensions/**` but matchers live at `__tests__/extensions/` (no `util/`), so the new matcher escapes the coverage glob. Pre-existing path inconsistency; consider fixing in a follow-up.
- Severity: 3.0/10
- Flagged by: Dependency Impact Analysis
- Recommended Action: update the glob to `__tests__/extensions/**` (or both paths) in a separate PR.

---
*Tip: reply `/dismiss: <reason>` on any non-blocking comment (severity < 7) to acknowledge it on future re-runs.*

__tests__/extensions/toHaveTableStructure.ts
```diff
@@ -0,0 +1,64 @@
...
+  expected,
+) {
+  const mismatches: string[] = [];
+  structureRoles.forEach(([property, role]) => {
+    const expectedContents = expected[property];
+    if (!expectedContents) {
+      return;
+    }
+
+    const expectedFlattened = expectedContents.flat();
```
canac, line 47: [Medium] **2D array grouping is decorative, not enforced.** Both the expected value (`.flat()` here) and the actual cells (flat `queryAllByRole` in DOM order) are compared as 1D lists — row boundaries are never verified. Proved empirically: `cells: [['a','b','c'],['d']]` matches the same table as `[['a','b'],['c','d']]`. The `Array<ExpectedContent | ExpectedContent[]>` type and the "matches 2-dimensional arrays" test advertise grouping as meaningful when it's cosmetic.

Debate consensus (4 agents): zero regression vs the old `.flat()` + `toEqual` assertions, and per-row enforcement is the *wrong* fix (it would break legitimate ragged rows like `['13. 403(b) Contribution']`). The agreed remediation is documentation:

```ts
// Note: nested arrays are flattened before comparison — row grouping is for
// readability only. Contents are compared as one flat list in DOM order;
// row boundaries are NOT verified.
const expectedFlattened = expectedContents.flat();
```

Optionally add a test pinning that regrouping without reordering still matches, so the contract is explicit.

__tests__/extensions/toHaveTableStructure.ts
```diff
@@ -0,0 +1,64 @@
...
+  ['cells', 'cell'],
+  ['columnHeaders', 'columnheader'],
+  ['rowHeaders', 'rowheader'],
+] as const;
+
+export const toHaveTableStructure: TableStructureMatcher = function (
+  received,
+  expected,
+) {
+  const mismatches: string[] = [];
```
canac, line 37: [Suggestion] **`toHaveTableStructure({})` passes vacuously** — all three properties undefined → every role check is skipped → `pass: true`. Mitigation already in place: a typo'd key (e.g. `cell:`) in an inline object literal is a compile error under `strict: true` excess-property checking (verified — the runtime probe required `@ts-expect-error`). The residual gap is a literal `{}` or a typo inside a pre-typed variable. Cheap hardening if desired:

```ts
if (!expected.cells && !expected.columnHeaders && !expected.rowHeaders) {
  throw new Error(
    'toHaveTableStructure requires at least one of cells, columnHeaders, or rowHeaders',
  );
}
```

Also worth noting in a doc comment: `cells: []` means "assert this section is empty," which is textually identical to "forgot to populate."

__tests__/extensions/toHaveTableStructure.ts
```diff
@@ -0,0 +1,64 @@
+import { within } from '@testing-library/dom';
+
+interface AsymmetricMatcher {
```
canac, line 3: [Suggestion] Locally redeclared `AsymmetricMatcher` interface duplicates jest's built-in asymmetric-matcher shape. Standards review confirmed this is fully compliant (`unknown`, not `any`; jest exports no stable public type for this), so this is awareness-only: the local structural type can drift from jest's actual shape across major upgrades.

__tests__/extensions/toHaveTableStructure.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  it('matches when only some structure properties are provided', () => {
+    const { getByRole } = render(<TestTable />);
+
+    expect(getByRole('table')).toHaveTableStructure({
+      rowHeaders: ['Salary', 'MHA'],
+    });
+  });
+
+  it('supports asymmetric matchers', () => {
```
canac, line 58: [Suggestion] Asymmetric matchers are only exercised for `cells` — `columnHeaders`/`rowHeaders` accept the same `ExpectedContent` union but have no asymmetric-matcher test. A debate-round probe confirmed they work (same code path), so this is a cheap test addition, not a bug.

__tests__/extensions/toHaveTableStructure.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+    expect(getByRole('table')).toHaveTableStructure({
+      cells: [
+        [expect.stringContaining('10,001'), '$20,001.00'],
+        ['$10,002.00', '$20,002.00'],
+      ],
+    });
+  });
+
+  it('does not match cells in a different order', () => {
```
canac, line 71: [Suggestion] No test asserts the failure-*message* content (the `printDiffOrStringify` diff, the `Expected ${property}` labels, or the bare `.not` branch message). Since this matcher is itself test infrastructure whose failure diff is its main value over `getAllByRole`, one test that catches a failing match and asserts `message()` contains the property name and a diff would lock both branches.

__tests__/extensions/toHaveTableStructure.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+    expect(getByRole('table')).not.toHaveTableStructure({
+      columnHeaders: ['Category', 'John'],
+    });
+    expect(getByRole('table')).not.toHaveTableStructure({
+      rowHeaders: ['Salary', 'MHA', 'Total'],
+    });
+  });
+
+  it('only inspects the received table', () => {
```
canac, line 138: [Suggestion] No explicit empty-table happy-path test (`cells: []` against a table with no body rows). This test covers `columnHeaders: []`/`rowHeaders: []` indirectly; an explicit empty-table case would pin the "empty array asserts absence" semantics.

src/components/HrTools/SalaryCalculator/SalaryCalculation/RequestedSalaryCard/RequestedSalaryCard.test.tsx
```diff
@@ -74,68 +74,55 @@ As you set your salary level, the amount you receive should reflect the amount o
...
+            'Maximum Allowable Salary (CAP)',
+            'Requested Salary',
+          ],
+          cells: [
+            ['$11,111.00', '$22,222.00'],
+            ['$10,003.00', '$20,003.00'],
+            ['$10,004.00', '$20,004.00'],
+            // The requested salary inputs
+            [
+              expect.stringContaining('Requested salary'),
```
canac, line 95: [Suggestion] Net coverage improvement — the old `.slice(0, -2)` excluded these input cells entirely, and they're now asserted. One caveat for future readers: `expect.stringContaining('Requested salary')` matches the `AutosaveTextField` *label* text in the cell's `textContent`; it asserts the field is present, not its numeric value (input values don't appear in `textContent`). If the field were ever replaced by static text containing the same label, this would still pass.

PR #1839, opened 2026-06-12
MPDX-9681 Remove total ASR calculation

kegrimes:
## Description
The frontend is solely calculating the ASR total which can lead to the client setting it to a value that doesn't match the real sum. Instead, we want to calculate this value in the [backend](https://github.com/CruGlobal/mpdx_api/pull/3358), so we need to remove any instance where a calculation occurs.

Jira ticket: [MPDX-9681](https://jira.cru.org/browse/MPDX-9681) 

## Testing

- Go to `/hrTools/additionalSalaryRequest`
- Create a request and open Dev Tools > Network
- Input a random value
- Ensure the update mutation does NOT calculate the total

## Checklist:

### DIFF
--- a/src/components/HrTools/AdditionalSalaryRequest/RequestPage/Helper/CapSubContent.test.tsx
+++ b/src/components/HrTools/AdditionalSalaryRequest/RequestPage/Helper/CapSubContent.test.tsx
@@ -24,7 +24,9 @@ const FormikWrapper: React.FC<{ children: React.ReactNode }> = ({
   const formik = useFormik<CompleteFormValues>({
     initialValues: {
       ...defaultCompleteFormValues,
-      totalAdditionalSalaryRequested: '5000',
+      currentYearSalaryNotReceived: '5000',
+      previousYearSalaryNotReceived: '3000',
+      additionalSalaryWithinMax: '2000',
     },
     onSubmit: jest.fn(),
   });
@@ -87,5 +89,24 @@ describe('CapSubContent', () => {
       queryByText(/You have a Board approved Maximum Allowable Salary/),
     ).not.toBeInTheDocument();
     expect(queryByText(/Progressive Approvals/)).not.toBeInTheDocument();
+  });
+
+  it('displays the correct total from form values', () => {
+    mockUseAdditionalSalaryRequest.mockReturnValue({
+      requestData: {
+        latestAdditionalSalaryRequest: {
+          progressiveApprovalTier: {
+            approver: 'Division Head',
+            approvalTimeframe: '1-2 weeks',
+          },
+          progressiveApprovalTierReason:
+            ProgressiveApprovalTierReasonEnum.OverUserCap,
+        },
+      },
+    } as unknown as ReturnType<typeof useAdditionalSalaryRequest>);
+
+    const { getByText } = renderCapSubContent();
+
+    expect(getByText(/\$10,000.00/)).toBeInTheDocument();
   });
 });
--- a/src/components/HrTools/AdditionalSalaryRequest/RequestPage/Helper/CapSubContent.test.tsx
+++ b/src/components/HrTools/AdditionalSalaryRequest/RequestPage/Helper/CapSubContent.test.tsx
@@ -107,6 +107,6 @@ describe('CapSubContent', () => {
 
     const { getByText } = renderCapSubContent();
 
-    expect(getByText(/\$10,000.00/)).toBeInTheDocument();
+    expect(getByText(/\$10,000\.00/)).toBeInTheDocument();
   });
 });
--- a/src/components/HrTools/AdditionalSalaryRequest/RequestPage/Helper/CapSubContent.tsx
+++ b/src/components/HrTools/AdditionalSalaryRequest/RequestPage/Helper/CapSubContent.tsx
@@ -10,6 +10,7 @@ import { currencyFormat } from 'src/lib/intlFormat';
 import theme from 'src/theme';
 import { CompleteFormValues } from '../../AdditionalSalaryRequest';
 import { useAdditionalSalaryRequest } from '../../Shared/AdditionalSalaryRequestContext';
+import { getTotal } from '../../Shared/Helper/getTotal';
 
 export const CapSubContent: React.FC = () => {
   const { t } = useTranslation();
@@ -21,6 +22,8 @@ export const CapSubContent: React.FC = () => {
   const reason = latestRequest?.progressiveApprovalTierReason;
 
   const { values } = useFormikContext<CompleteFormValues>();
+
+  const total = getTotal(values);
 
   if (reason === ProgressiveApprovalTierReasonEnum.BoardCapException) {
     return null;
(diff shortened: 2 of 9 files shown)


src/components/HrTools/AdditionalSalaryRequest/Shared/AutoSave/useSaveField.ts
```diff
@@ -25,20 +24,12 @@ export const useSaveField = ({ formValues }: UseSaveFieldOptions) => {
...
-      } as CompleteFormValues;
-      const totalAdditionalSalaryRequested = getTotal(updatedValues);
-
       await trackMutation(
         updateAdditionalSalaryRequest({
           variables: {
             id: requestId,
             attributes: {
               ...attributes,
-              totalAdditionalSalaryRequested,
```
canac, line 41: You might want to keep this for the optimistic update. Otherwise, we'll have to wait for the response to come back from the server to get the new total. So the line items will update immediately but the total will lag behind briefly.

src/components/HrTools/AdditionalSalaryRequest/RequestPage/Helper/CapSubContent.tsx
```diff
@@ -22,6 +23,8 @@ export const CapSubContent: React.FC = () => {
 
   const { values } = useFormikContext<CompleteFormValues>();
 
+  const total = getTotal(values);
```
kegrimes, line 26: [Medium] **Client-computed total vs server-authoritative total** (flagged by Financial Reporting, Architecture, and UX).

This PR makes the server own `totalAdditionalSalaryRequested`, but `CapSubContent` now displays `getTotal(values)` (a client-side `.reduce`) instead of reading the server value. Previously it read `values.totalAdditionalSalaryRequested`, which reflected the server total after refetch — so this is a mild source-of-truth regression. UX also noted a temporal mismatch: the live `amount` can pair with a stale, server-derived `approver`/`approvalTimeframe` in the same sentence until autosave resolves.

Field parity is exact today (verified: `getTotal`'s 14 `numericFields` match the form fields 1:1), so this only bites if the server applies a cap/rounding/spouse adjustment the client doesn't.

**Suggested fix** — prefer the server value, fall back to the client total only before the first response:
```tsx
const total =
  latestRequest?.totalAdditionalSalaryRequested ?? getTotal(values);
```

src/components/HrTools/AdditionalSalaryRequest/Shared/AutoSave/useSaveField.ts
```diff
@@ -20,25 +20,29 @@ export const useSaveField = ({ formValues }: UseSaveFieldOptions) => {
...
       await trackMutation(
         updateAdditionalSalaryRequest({
           variables: {
             id: requestId,
-            attributes: {
-              ...attributes,
-              totalAdditionalSalaryRequested,
+            attributes,
+          },
+          optimisticResponse: {
```
kegrimes, line 38: [Medium] **The new `optimisticResponse` has no test coverage** (Testing agent).

This is the PR's central new behavior, but `grep optimistic` across the feature returns nothing, and `mockTrackMutation = jest.fn((m) => m)` discards the mutation options — so even a unit assertion on the optimistic object isn't possible as currently wired. Note: the Data Integrity agent **verified the shape is correct and type-checked**, so this is a coverage gap, not a latent bug.

Consider a test that asserts the optimistic total renders before the network resolves (delayed `UpdateAdditionalSalaryRequest` mock + `findByText`).

_Also (Architecture, severity ~4.5):_ this optimistic write depends on `latestAdditionalSalaryRequest` being read with the **same** `AdditionalSalaryRequestDetailsFields` fragment the mutation returns. Correct today; a future fragment edit could silently break it. A one-line comment noting that coupling would help.

src/components/HrTools/AdditionalSalaryRequest/RequestPage/Helper/CapSubContent.test.tsx
```diff
@@ -88,4 +90,23 @@ describe('CapSubContent', () => {
...
+          },
+          progressiveApprovalTierReason:
+            ProgressiveApprovalTierReasonEnum.OverUserCap,
+        },
+      },
+    } as unknown as ReturnType<typeof useAdditionalSalaryRequest>);
+
+    const { getByText } = renderCapSubContent();
+
+    expect(getByText(/\$10,000.00/)).toBeInTheDocument();
```
kegrimes: [Suggestion] The unescaped `.` in `/\$10,000.00/` matches any character, so `$10,000X00` would also pass. Minor — `currencyFormat` always emits `.` — but tighten to `/\$10,000\.00/`. Consider also adding a zero-state assertion (all line items `'0'` → `$0.00`) since this is currently the only total-display test.

kegrimes (commented):
## 🔍 Multi-Agent Code Review — Verdict: APPROVED WITH SUGGESTIONS

6 agents (Financial Reporting, Data Integrity, Architecture, Testing, UX, Standards) · standard mode · **Risk 5/10 (MEDIUM)**.

**No blockers.** The change direction is sound and net-reduces debt: it removes the client-computed `totalAdditionalSalaryRequested` from both mutation paths so the **server is the single source of truth**, and adds an Apollo `optimisticResponse` for instant UI feedback. The Data Integrity agent verified (High confidence) that the optimistic shape is type-checked, carries `__typename`/`id`, matches the `AdditionalSalaryRequestDetailsFields` fragment, follows two sibling-hook conventions, and is backstopped by `refetchQueries`. `tsc` and `lint:ci` pass clean.

### One decision worth settling before merge (verification, not a code defect)
Three agents independently flagged that `CapSubContent` and the optimistic response still compute the total **client-side** via `getTotal` while the persisted total is now server-authoritative. Field parity is exact today (the 14 `numericFields` match the form fields 1:1), so divergence is only possible **if the server applies a cap, rounding, or spouse adjustment the client doesn't**. None of the agents could verify the server logic from the frontend repo.

👉 **Please confirm `totalAdditionalSalaryRequested` is a plain sum of exactly those 14 line items.** If yes, the inline suggestions below are optional polish.

### Summary
| Agent | Critical | High | Important | Suggestions |
|---|:--:|:--:|:--:|:--:|
| Financial Reporting | 0 | 0 | 0 | 1 |
| Data Integrity | 0 | 0 | 0 | 2 |
| Architecture | 0 | 0 | 0 | 3 |
| Testing & Quality | 0 | 0 | 1 | 3 |
| UX | 0 | 0 | 0 | 3 |
| Standards | 0 | 0 | 0 | 0 |

---

**`src/components/HrTools/AdditionalSalaryRequest/Shared/Helper/getTotal.ts` — Severity 4.0/10** — Flagged by Testing & Financial. `getTotal` is now load-bearing for display, but its test has only one happy path. `Number(values[key] || 0)` coerces falsy values to 0, but a non-numeric string (e.g. `'abc'`) yields `NaN` and would propagate to the displayed total. Add `null`/`undefined`/`''`/non-numeric cases and confirm Yup guarantees numeric input upstream of `CapSubContent`.

---
_Reply `/dismiss: <reason>` on any inline comment (severity < 7) to acknowledge it. To dispute the verdict, fix and re-run `/quality:agent-review`._

src/components/HrTools/AdditionalSalaryRequest/RequestPage/Helper/CapSubContent.tsx
kegrimes: /dismiss Before, this component was using the calculated formik value. I changed it to calculate in this component.

PR #1840, opened 2026-06-12
Create scenarios for future NSO Questionnaire form logic (New Staff, Senior Staff, non-RMO spouse, etc)

zweatshirt:
## Description

The NSO MPD Questionnaire has different questions depending on the status of the couple or single user filling out the form. 

Instead of creating complex conditionals per each form view, I was looking into creating scenarios that are determined in the context. Hopefully this will reduce duplication in the long run, and not increase it. 

There is a chance I am getting ahead of myself, as this could probably be done in the backend and not the frontend.

Maybe we can collapse Enums like `SeniorUserNewSpouse` and `NewUserSeniorSpouse`... 

Here is a table of the scenarios:

Scenario | loading | hasSpouse | hasNonRmoSpouse | user group | spouse group
-- | -- | -- | -- | -- | --
Unknown | true | — | — | — | —
NewUserSingle | false | false | — | New | —
NewSosaUser | false | true | true | New | (any)
NewUserNewSpouse | false | true | false | New | New
NewUserSeniorSpouse | false | true | false | New | Senior
SeniorUserNewSpouse | false | true | false | Senior | New
Unknown (fallback) | false | (any combination not matched above) |   |   |  

<br class="Apple-interchange-newline">

## Testing

No UI testing as of yet.

## Checklist:

### DIFF
--- a/src/components/HrTools/NsoMpdQuestionnaire/Shared/useHouseholdStaff.test.tsx
+++ b/src/components/HrTools/NsoMpdQuestionnaire/Shared/useHouseholdStaff.test.tsx
@@ -1,0 +1,110 @@
+import { ReactElement } from 'react';
+import { renderHook } from '@testing-library/react-hooks';
+import { DeepPartial } from 'ts-essentials';
+import { GqlMockedProvider } from '__tests__/util/graphqlMocking';
+import { HcmQuery } from 'src/components/HrTools/Shared/HcmData/Hcm.generated';
+import { GetUserQuery } from 'src/components/User/GetUser.generated';
+import {
+  UsStaffGroupEnum,
+  UserPersonTypeEnum,
+} from 'src/graphql/types.generated';
+import { NsoMpdScenarioEnum, useHouseholdStaff } from './useHouseholdStaff';
+
+type HcmPersonMock = DeepPartial<HcmQuery['hcm'][number]>;
+type UserMock = DeepPartial<GetUserQuery['user']>;
+
+const staff: HcmPersonMock = {
+  staffInfo: { userPersonType: UserPersonTypeEnum.EmployeeStaff },
+};
+// In a SOSA household the non-RMO marker sits on the user's own (hcm[0]) record.
+const userWithNonRmoSpouse: HcmPersonMock = {
+  staffInfo: { userPersonType: UserPersonTypeEnum.EmployeeStaffNonRmoSpouse },
+};
+
+const renderHousehold = (hcm: HcmPersonMock[], user: UserMock) =>
+  renderHook(() => useHouseholdStaff(), {
+    wrapper: ({ children }: { children: ReactElement }) => (
+      <GqlMockedProvider<{ Hcm: HcmQuery; GetUser: GetUserQuery }>
+        mocks={{ Hcm: { hcm }, GetUser: { user } }}
+      >
+        {children}
+      </GqlMockedProvider>
+    ),
+  });
+
+describe('useHouseholdStaff', () => {
+  it('reports a single new-staff user (NewUserSingle)', async () => {
+    const { result, waitForNextUpdate } = renderHousehold([staff], {
+      usStaffGroup: UsStaffGroupEnum.NewStaff,
+      spouseUsStaffGroup: null,
+    });
+    await waitForNextUpdate();
+
+    expect(result.current.hasSpouse).toBe(false);
+    expect(result.current.hcmSpouse).toBeNull();
+    expect(result.current.userStaffGroup).toBe(UsStaffGroupEnum.NewStaff);
+    expect(result.current.spouseStaffGroup).toBeNull();
+    expect(result.current.hasNonRmoSpouse).toBe(false);
+    expect(result.current.scenario).toBe(NsoMpdScenarioEnum.NewUserSingle);
+  });
+
+  it('reports two new-staff spouses (NewUserNewSpouse)', async () => {
+    const { result, waitForNextUpdate } = renderHousehold([staff, staff], {
+      usStaffGroup: UsStaffGroupEnum.NewStaff,
+      spouseUsStaffGroup: UsStaffGroupEnum.NewStaff,
+    });
+    await waitForNextUpdate();
+
+    expect(result.current.hasSpouse).toBe(true);
+    expect(result.current.scenario).toBe(NsoMpdScenarioEnum.NewUserNewSpouse);
+  });
+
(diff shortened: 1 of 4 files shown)


src/components/HrTools/NsoMpdQuestionnaire/Shared/useHouseholdStaff.ts
```diff
@@ -0,0 +1,128 @@
...
+ * derived `scenario` is what forms switch on so they don't re-combine the statuses themselves.
+ */
+export const useHouseholdStaff = (): HouseholdStaff => {
+  const { data: hcmData, loading: hcmLoading } = useHcmQuery();
+  const { data: userData, loading: userLoading } = useGetUserQuery();
+
+  return useMemo(() => {
+    const hcmUser = hcmData?.hcm[0] ?? null;
+    const hcmSpouse = hcmData?.hcm[1] ?? null;
+    const hasSpouse = hcmSpouse !== null;
```
zweatshirt, line 103: [Medium] Cross-source consistency: `hasSpouse` is derived from HCM (`hcm[1]` presence) while `spouseStaffGroup` comes from GetUser (`spouseUsStaffGroup`) — two independent backends. If they diverge (e.g. HCM returns 2 records but `spouseUsStaffGroup` is `null`), a married new-staff user silently falls through every branch to `Unknown` with no signal distinguishing it from a genuine out-of-audience case. Worth confirming the two sources are guaranteed to co-occur for married households, or surfacing an explicit 'incomplete data' state instead of folding into `Unknown`.

src/components/HrTools/NsoMpdQuestionnaire/Shared/useHouseholdStaff.ts
```diff
@@ -0,0 +1,128 @@
...
+};
+
+/**
+ * Pulls the household's HCM records and derives each person's staffing status for the NSO MPD
+ * Questionnaire. New-vs-Senior comes from the API-computed `usStaffGroup`/`spouseUsStaffGroup` from
+ * GetUser. The non-RMO-spouse (SOSA) marker is read from the user's own HCM `userPersonType`. The
+ * derived `scenario` is what forms switch on so they don't re-combine the statuses themselves.
+ */
+export const useHouseholdStaff = (): HouseholdStaff => {
+  const { data: hcmData, loading: hcmLoading } = useHcmQuery();
```
zweatshirt, line 97: [Suggestion] Neither `useHcmQuery` nor `useGetUserQuery` `error` is read. On a query error, `data` is `undefined` → `scenario` collapses to `Unknown` with `loading: false`, identical to a legitimate out-of-audience user. Consider surfacing `error` (or an explicit errored state) so consumers can tell 'errored' apart from 'not the audience'.

src/components/HrTools/NsoMpdQuestionnaire/Shared/useHouseholdStaff.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+      },
+    );
+    await waitForNextUpdate();
+
+    expect(result.current.hasSpouse).toBe(true);
+    expect(result.current.hasNonRmoSpouse).toBe(true);
+    expect(result.current.scenario).toBe(NsoMpdScenarioEnum.NewSosaUser);
+  });
+
+  it('falls back to Unknown for audiences outside the questionnaire (senior single)', async () => {
```
zweatshirt, line 101: [Medium] Coverage gap: every test `await`s past loading, so the `loading → Unknown` branch and the `loading` field itself are never asserted — a regression returning a concrete scenario mid-load (when forms switch on `scenario`) would render the wrong form silently. Also two of the three `Unknown` *branches* are untested: SOSA-but-not-new (`hasNonRmoSpouse && !userIsNew`) and both-senior; only senior-single is covered. Add a pre-`waitForNextUpdate` loading assertion plus those two cases.

zweatshirt (commented):
## Multi-Agent Code Review — household/scenario foundation

**Verdict: ✅ APPROVED WITH SUGGESTIONS** — no blockers. Reviewed by 4 agents (Architecture, Data Integrity, Testing, Standards) + dependency analysis. (Security/UX/Financial skipped — no auth/UI/money in the diff.)

The derivation logic is correct: all agents independently confirmed `deriveScenario` branch precedence (`loading → !hasSpouse → SOSA → staff-group combos → Unknown`), the deliberate SOSA-on-`hcm[0]` design, null handling (`staffInfo` is non-nullable so `hcmUser?.staffInfo.userPersonType` can't throw), and complete `useMemo` deps. Dependency impact is additive / non-breaking — the `& HouseholdStaff` context growth doesn't touch existing consumers, and `GqlMockedProvider` auto-mocks the two new queries for all existing tests.

### Verified
- The Testing agent's only ≥7 concern (7.0 — possible `act()` warnings from adding `GqlMockedProvider` to the test wrapper) was **refuted by running the suite**: `yarn test src/components/HrTools/NsoMpdQuestionnaire` → 32/32 pass, zero `act()`/`console.error` warnings. Not counted as a finding.

### Suggestions
Three inline comments below (all non-blocking). Plus two minor, acceptable-as-is notes on `NsoMpdQuestionnaireContext.tsx`:
- The context `useMemo` spreads `householdStaff` and lists it in the deps array — correct because the hook is memoized, but it couples context stability to the hook's internals. (Architecture, 6.5)
- `... & HouseholdStaff` re-exposes the raw fields alongside the derived `scenario`, inviting ad-hoc recombination — but `SalaryCalculatorContext` sets the same precedent. (Architecture, 4.0)

### Design note for form-build time
`scenario === Unknown` currently conflates *loading*, *not-in-audience*, and *query error* (neither query's `error` is surfaced). `loading` is exposed so consumers can distinguish loading; consider exposing `error` too when forms start consuming this.

### Standards: clean PASS
Named exports, feature-local hook placement (36 precedents), typed `GqlMockedProvider` + `DeepPartial` mocks (13 precedents), no `any`, no i18n needed (no UI). Zero violations.

canac:
Thanks for thinking through this! I think we'll want to calculate the form type on the backend though, and expose it to the client via GraphQL: MPDX-9505. I'm sure we'll need the form type for backend validation, so it makes sense to calculate it once there. Do you want to take that backend ticket? You could make a service for this even though there's not anything to use it yet.

zweatshirt:
@canac Ah I missed that ticket, I'll try to give it a go

zweatshirt:
Closing: To be done in the backend instead.

PR #1842, opened 2026-06-15
MPDX-9706 Add filtering to MPGA report

kegrimes:
## Description

Currently, MPGA does not have the functionality to combine rows together in the income and expenses tables. We want something similar to Staff expenses report that allows users to choose which categories they want to combine into a single row. 

Jira ticket: [MPDX-9706](https://jira.cru.org/browse/MPDX-9706)

Designs approved by Joanna and Scott

## Testing

- Check that the rows are summarized by default
- Click "Report Settings" and uncheck a category
- Verify the unchecked category has multiple rows

## Checklist:


src/hooks/useFilteredFunds.ts
```diff
@@ -67,7 +70,12 @@ export function useFilteredFunds(funds: Funds[]) {
       const base = fund.fundType;
       fund.categories?.forEach((category) => {
         const baseId = `${base}-${category.category}`;
-        if (category.subcategories?.length) {
+        const isSelected =
+          selectedCategories === null ||
+          selectedCategories.includes(category.category.toUpperCase());
```
kegrimes: [Critical] **Category filter never matches in production.** By the time data reaches this hook, `MPGAIncomeExpensesReport.tsx:98` has replaced `category.category` with its localized label via `getLocalizedCategory` (e.g. `"Ministry Reimbursement"`). `selectedCategories` holds raw enums (`"MINISTRY_REIMBURSEMENT"`). `.includes(category.category.toUpperCase())` compares `"MINISTRY REIMBURSEMENT"` (space) against `"MINISTRY_REIMBURSEMENT"` (underscore) → always false for multi-word categories; single-word ones coincidentally match. Checking a category to consolidate it silently does nothing for most categories.

The unit test passes raw enums as `category.category`, bypassing the localization transform — so CI stays green while production is broken.

**Fix:** match on the stable enum key, not the localized label. Carry the raw enum through to the hook (localize only for `description`), compare `selectedCategories.includes(category /* enum */)`, and drop `.toUpperCase()`. Add a test that routes data through `getLocalizedCategory` before calling the hook.

src/components/Reports/StaffExpenseReport/SettingsDialog/SettingsDialog.tsx
```diff
@@ -230,103 +232,117 @@ export const SettingsDialog: React.FC<SettingsDialogProps> = ({
...
-                    Select which categories to keep consolidated.`,
-                  )}
+                <Typography
+                  sx={{ mt: isMpgaReport ? 0 : 2, whiteSpace: 'pre-line' }}
+                >
+                  {isMpgaReport
+                    ? t(
+                        `Income and expenses are combined by categories by default. Select which categories to keep consolidated.`,
+                      )
+                    : `Income and expenses are combined by categories by default. This may be useful for long date ranges (e.g., "Year to Date").
```
kegrimes: [High] **i18n regression.** This `else` branch is a raw template literal — it lost the `t()` wrapper the string had before this PR (the MPGA branch above keeps it). The translation key still exists in `translation.json:1522` but is no longer reached, so non-English StaffExpenseReport users now see hard-coded English, and `yarn extract` will eventually prune the orphaned key.

**Fix:** wrap this branch in `t()` (static literal, extracts fine), e.g. `: t(\`Income and expenses are combined…\`)`. Better: assign the chosen string to a variable and call `t()` once to avoid the trap recurring.

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesReport.tsx
```diff
@@ -92,7 +111,10 @@ export const MPGAIncomeExpensesReport: React.FC<
     [reportData],
   );
 
-  const { incomeData, expenseData } = useFilteredFunds(transformedData);
+  const { incomeData, expenseData } = useFilteredFunds(
```
kegrimes, line 118: [Important] **The PR's core feature has no behavioral test.** No test exercises the `filters` state, `onFiltersChange`, or this filters→hook→table data flow. This is why the Critical filter bug shipped with green CI. Add an interaction test: render the report, click `Report Settings`, toggle a category, click `Apply Filters`, and assert the rendered rows change. (Important-tier: strongly recommended, can't be `/dismiss`ed — ideally fix in this PR since it's the gap that hid the blocker.)

src/hooks/useFilteredFunds.ts
```diff
@@ -77,30 +85,59 @@ export function useFilteredFunds(funds: Funds[]) {
...
               monthly,
-              average,
-              total,
+              average: subcategory.averagePerMonth,
+              total: subcategory.total,
+            });
+          });
+        } else if (category.subcategories?.length) {
+          // Checked category with subcategories: combine its subcategories into one row
+          const monthCount = category.subcategories[0].breakdownByMonth.length;
```
kegrimes: [Medium] **`monthCount` assumes all subcategories share the first one's month window.** It sizes `incomeMonthly`/`expenseMonthly` from `subcategories[0]`, then accumulates every subcategory by `index`. If a later subcategory has more months, `incomeMonthly[index] += …` writes to an uninitialized slot → `undefined + number = NaN`, and the `.some(amount !== 0)` guard treats `NaN` as truthy, pushing a row of `NaN` totals. Latent today (uniform windows). **Fix:** derive `monthCount` from the category-level `breakdownByMonth.length`, or align by `month.month` into a map.

src/hooks/useFilteredFunds.ts
```diff
@@ -77,30 +85,59 @@ export function useFilteredFunds(funds: Funds[]) {
...
+          const incomeMonthly = new Array(monthCount).fill(0);
+          const expenseMonthly = new Array(monthCount).fill(0);
+
+          category.subcategories.forEach((subcategory) => {
+            subcategory.breakdownByMonth.forEach((month, index) => {
+              if (month.total >= 0) {
```
kegrimes: [Medium] **Unguarded `month.total`.** This new accumulation reads `month.total.toFixed(2)` with no nullish guard; a `null`/`undefined` total (possible from the REST proxy) throws a `TypeError` and crashes the report render, and `null >= 0` would misclassify as income. **Fix:** `const value = Number((month.total ?? 0).toFixed(2));` then branch on `value`.

src/hooks/useFilteredFunds.ts
```diff
@@ -77,30 +85,59 @@ export function useFilteredFunds(funds: Funds[]) {
...
+              total: subcategory.total,
+            });
+          });
+        } else if (category.subcategories?.length) {
+          // Checked category with subcategories: combine its subcategories into one row
+          const monthCount = category.subcategories[0].breakdownByMonth.length;
+          const incomeMonthly = new Array(monthCount).fill(0);
+          const expenseMonthly = new Array(monthCount).fill(0);
+
+          category.subcategories.forEach((subcategory) => {
```
kegrimes: [Medium] **Gross vs. net semantics — confirm with product.** Subcategories are split by sign per-month independently, so a category whose subcat A is `+500` and subcat B is `-300` in the same month produces both a `+500` income cell and a `-300` expense cell rather than a net `+200`. This matches existing per-subcategory behavior and appears intentional, but it's a reporting-semantics choice (gross consolidation). If staff expect *net* consolidation, the displayed totals mislead. Confirm intent.

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesReport.tsx
```diff
@@ -51,22 +55,37 @@ export const MPGAIncomeExpensesReport: React.FC<
...
-
   const last12Months = useGetLastTwelveMonths(locale, true);
 
   const start = convertMonths(last12Months[0], locale);
-  const end = DateTime.now().toISODate();
+  const end = DateTime.now();
+
+  const startDate = DateTime.now().minus({ months: 11 }).startOf('month');
+
+  const [filters, setFilters] = useState<Filters>(() =>
```
kegrimes, line 65: [Medium] **Dead date state.** `filters` is seeded with `startDate`/`endDate`, but the report query (`MPGAIncomeExpensesReport.tsx:84-90`) uses its own `start`/`end.toISODate()` window and the MPGA dialog hides the date UI — so only `filters.categories` is ever live. Two parallel notions of the date window can silently drift, and a future maintainer will reasonably (but wrongly) assume changing these dates re-queries. **Fix:** store only `{ categories }` for MPGA, or actually drive `startMonth`/`endMonth` from `filters`.

src/components/Reports/StaffExpenseReport/SettingsDialog/SettingsDialog.tsx
```diff
@@ -230,103 +232,117 @@ export const SettingsDialog: React.FC<SettingsDialogProps> = ({
...
-                  label={t('Select Date Range')}
-                  fullWidth
-                  value={values.selectedDateRange ?? ''}
-                  onChange={(e) => {
-                    const value = e.target.value === '' ? null : e.target.value;
-                    setFieldValue('selectedDateRange', value);
-                    if (value !== null) {
-                      setFieldValue('startDate', null);
-                      setFieldValue('endDate', null);
+                {!isMpgaReport && (
```
kegrimes: [Medium] **`isMpgaReport` boolean mode flag forces two divergent dialogs into one component** (hidden date section, branched spacing, branched copy). This smell tends to grow combinatorially. Consider extracting the shared category-selection block into a `CategoryFilterSection` consumed by two thin wrappers, or pass a capability-named `showDateRange?: boolean` rather than encoding a feature name into a reusable component.

src/components/Reports/MPGAIncomeExpensesReport/DisplayModes/ScreenOnlyReport.tsx
```diff
@@ -1,27 +1,42 @@
-import { HourglassDisabled } from '@mui/icons-material';
+import { useState } from 'react';
+import { HourglassDisabled, Settings } from '@mui/icons-material';
 import { Box, Container, Grid } from '@mui/material';
 import { useTranslation } from 'react-i18next';
 import { EmptyTable } from '../../../HrTools/Shared/EmptyTable/EmptyTable';
+import {
```
kegrimes: [Medium] **Cross-feature coupling.** The MPGA feature now imports `Filters`, `SettingsDialog`, `getFiltersWithCalculatedDates` (here) and `StyledFilterButton` (from `../StaffExpenseReport/StaffExpenseReport`) out of a sibling feature. The repo rule: components used across multiple features belong in `src/components/Shared/` (or a shared Reports folder). `StyledFilterButton` being exported from a 400-line page module is especially fragile — it pulls the entire StaffExpenseReport module into MPGA's import graph. **Fix:** promote the genuinely shared pieces to a shared location; at minimum extract `StyledFilterButton` out of the page component.

kegrimes (commented):
## 🤖 Multi-Agent Code Review — PR #1842

**Verdict: 🔴 BLOCKERS FOUND** · Risk **6/10 (MEDIUM)** · 6 specialized agents + dependency analysis (standard mode)

Two blockers must be resolved before merge:

1. **The headline feature doesn't work in production.** `useFilteredFunds` compares `selectedCategories` (raw enums like `MINISTRY_REIMBURSEMENT`) against `category.category`, which `MPGAIncomeExpensesReport.tsx:98` has already localized to a display label (`Ministry Reimbursement`). `.includes(category.category.toUpperCase())` → `"MINISTRY_REIMBURSEMENT"` vs `"MINISTRY REIMBURSEMENT"` (underscore vs space) → never matches for multi-word categories. Single-word categories (`Donation`, `Salary`) coincidentally match, so the filter is *inconsistently* broken. CI is green because the unit test feeds raw enums and bypasses the `getLocalizedCategory` transform that production always applies. (Verified end-to-end by Data Integrity + Financial agents and by the reviewer directly.)
2. **i18n regression.** The non-MPGA branch of the new help-text ternary lost its `t()` wrapper and now renders raw English on the existing StaffExpenseReport.

See inline comments for these plus the Important (untested filter wiring) and Medium findings.

### Dependency Impact — no breaking changes
`useFilteredFunds`'s new required 2nd arg: both call sites updated. New exports (`getFiltersWithCalculatedDates`, `StyledFilterButton`) and the optional `isMpgaReport` prop are backward-compatible. `DataFields.id` is consumed only as an opaque React/DataGrid key — the new `-income`/`-expense` suffixes actually *fix* a prior row-id collision.

### Suggestions (informational, not inlined)
- **Hook sign-splitting duplication** (`useFilteredFunds.ts:96-129`): the combine branch pre-splits by sign (`>= 0`), then `pushData` re-splits (`> 0`/`< 0`). Consider summing subcategories into one signed `monthly[]` and letting `pushData` own all sign logic.
- **Redundant `.toUpperCase()`** (`useFilteredFunds.ts:75`): no-op on enums; remove once the match is fixed.
- **Missing hook test branches**: `selectedCategories === null` (default path), unchecked-per-subcategory rows, and the zero-suppression guard are all unexercised.
- **`'Filter Settings'` → `'Report Settings'`** rename is consistent across buttons/tests; old key now orphaned (harmless).

### Standards / Financial checklist highlights
✅ Named exports · ✅ no `new Date()` (Luxon used) · ✅ no `any` · ✅ currency mixing prevented (hardcoded USD) · ⚠️ `.toFixed()` inside aggregation (matches existing convention) · ❌ null `month.total` unguarded (see inline).

src/hooks/useFilteredFunds.ts
kegrimes: /dismiss this is the intent. we want to separate the values into income and expenses.

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesReport.tsx
kegrimes: /dismiss I need those fields to get the last 12 month snapshot

src/hooks/useFilteredFunds.ts
```diff
@@ -67,40 +76,80 @@ export function useFilteredFunds(funds: Funds[]) {
...
+
+        if (category.subcategories?.length && !isSelected) {
+          // Unchecked category with subcategories: one row per subcategory
+          const categoryName = getLocalizedCategory(category.category, t);
           category.subcategories.forEach((subcategory) => {
+            const subcategoryName = getLocalizedSubCategory(
+              subcategory.subCategory,
+              t,
+            );
             const id = `${baseId}-${subcategory.subCategory}`;
```
kegrimes: [Critical][Pre-existing] **Duplicate row `id`s.** When a category has two subcategories with the same `subCategory` enum, this id is identical for both rows (e.g. `Primary-MINISTRY_REIMBURSEMENT-MINISTRY_REIMBURSEMENT`). `id` is the MUI DataGrid row id (`TableCard.tsx:147` `getRowId`) and the React key (`PrintTables.tsx:142`) — duplicates make DataGrid drop/merge a row and break reconciliation. The `id` line itself is unchanged (pre-existing → not a merge blocker), **but this PR routes user-uncheck actions through this branch and the unit mock (`useFilteredFunds.test.ts:76,131`) already produces the collision.** Strongly recommend fixing here: `const id = `${baseId}-${subcategory.subCategory}-${index}`;` (add `index` from the `forEach`).

src/hooks/useFilteredFunds.ts
```diff
@@ -67,40 +76,80 @@ export function useFilteredFunds(funds: Funds[]) {
...
+              const value = Number((month.total ?? 0).toFixed(2));
+              if (value >= 0) {
+                incomeMonthly[index] += value;
+              } else {
+                expenseMonthly[index] += value;
+              }
             });
           });
+
+          if (incomeMonthly.some((amount) => amount !== 0)) {
```
kegrimes: [Medium] **Zero-suppression asymmetry.** This checked-category branch suppresses all-zero income/expense rows via `.some(amount => amount !== 0)`, but the unchecked-subcategory branch and the no-subcategory rollup call `pushData` unconditionally — so an all-zero category renders as `$0.00` rows when unchecked but disappears when checked. Same data, different output by mode. Pick one policy: move a `data.monthly.every(m => m === 0)` guard inside `pushData` (then these `.some()` guards become redundant), or drop the guards for symmetry. *Also worth a comment here noting these income/expense totals are summed client-side and net to `category.total` (no server field exists for the sign-split), so a future refactor doesn't "fix" it back to `category.total`.*

src/hooks/useFilteredFunds.ts
```diff
@@ -67,40 +76,80 @@ export function useFilteredFunds(funds: Funds[]) {
...
             const id = `${baseId}-${subcategory.subCategory}`;
             const description =
-              category.category === subcategory.subCategory
-                ? category.category
-                : `${category.category} - ${subcategory.subCategory}`;
+              categoryName === subcategoryName
+                ? categoryName
+                : `${categoryName} - ${subcategoryName}`;
             const monthly = subcategory.breakdownByMonth.map((month) =>
               Number(month.total.toFixed(2)),
```
kegrimes: [Suggestion] **Inconsistent null guard.** The checked branch uses `(month.total ?? 0).toFixed(2)` (line 115) but this branch and the rollup (line 145) call `month.total.toFixed(2)` unguarded. Type-safe today (`BreakdownByMonth.total` is non-nullable), so this is a consistency nit — either apply `?? 0` on all three or drop it from line 115 to match the type contract.

src/hooks/useFilteredFunds.test.ts
```diff
@@ -185,31 +246,40 @@ const mockData: Funds[] = [
 
 describe('useFilteredFunds', () => {
   it('should filter funds correctly', () => {
-    const { result } = renderHook(() => useFilteredFunds(mockData));
+    const { result } = renderHook(() =>
+      useFilteredFunds(mockData, selectedCategories, t),
```
kegrimes, line 112: [Important] **Four new branches are unit-untested.** This is the only `useFilteredFunds` call and both categories are *selected*, so only the checked-combine + no-subcategory branches run. Uncovered: the **unchecked → one-row-per-subcategory** branch (the headline feature), `selectedCategories === null` (all-selected default), and the **zero-suppression** guards. Add cases: unchecked-split with *distinct* subcategory enums asserting distinct ids; `null` all-selected; an expense-only (all-negative) category proving suppression; and empty `funds: []`. (Important findings can't be `/dismiss`ed — address here or file a follow-up.)

src/hooks/useFilteredFunds.test.ts
```diff
@@ -6,66 +6,127 @@ import {
...
           },
           {
             month: '2024-12-01',
-            total: 1200,
+            total: 2400,
           },
         ],
         subcategories: [
+          {
+            subCategory: StaffExpensesSubCategoryEnum.MinistryReimbursement,
```
kegrimes: [Medium] **Unit mock can't catch the duplicate-id bug.** Both subcategories here use the same enum (`StaffExpensesSubCategoryEnum.MinistryReimbursement`), so even if the unchecked branch were tested with this mock it would generate two identical ids and not surface a key/DataGrid collision. Use two *distinct* subcategory enums (as the MPGA integration mock correctly does).

src/components/Reports/Shared/SettingsDialog/SettingsDialog.tsx
```diff
@@ -22,18 +22,19 @@ import * as yup from 'yup';
 import { CustomDateField } from 'src/components/Shared/DateTimePickers/CustomDateField';
 import { Fund, StaffExpenseCategoryEnum } from 'src/graphql/types.generated';
 import i18n from 'src/lib/i18n';
-import { getLocalizedCategory } from '../../Shared/Helpers/transformStaffExpenseEnums';
-import { useReportsStaffExpensesQuery } from '../GetStaffExpense.generated';
-import { DateRange } from '../Helpers/StaffReportEnum';
-import { getAvailableCategories } from '../Helpers/filterTransactions';
-import { getStaffExpenseMonthRange } from '../Helpers/getMonthRange';
+import { useReportsStaffExpensesQuery } from '../../StaffExpenseReport/GetStaffExpense.generated';
```
kegrimes, line 25: [Medium] **"Shared" component depends back on a feature folder.** This `Reports/Shared/` dialog imports four symbols from `StaffExpenseReport/` (the `ReportsStaffExpenses` query, `DateRange`, `getAvailableCategories`, `getStaffExpenseMonthRange`). A grep of other `Reports/Shared/` subtrees shows no other shared→feature import — this inverts the dependency direction, so `Shared/` can no longer be reasoned about in isolation and StaffExpenseReport can't be refactored without breaking it. Lift the StaffExpense-specific query/helpers into `Shared/`, or inject category-loading + month-range as props.

src/components/Reports/Shared/SettingsDialog/SettingsDialog.tsx
```diff
@@ -230,103 +232,117 @@ export const SettingsDialog: React.FC<SettingsDialogProps> = ({
...
-                  label={t('Select Date Range')}
-                  fullWidth
-                  value={values.selectedDateRange ?? ''}
-                  onChange={(e) => {
-                    const value = e.target.value === '' ? null : e.target.value;
-                    setFieldValue('selectedDateRange', value);
-                    if (value !== null) {
-                      setFieldValue('startDate', null);
-                      setFieldValue('endDate', null);
+                {!showDateRange && (
```
kegrimes: [Medium] **Inverted boolean prop.** `{!showDateRange && (…date UI…)}` means passing `showDateRange={true}` *hides* the date range — the name reads as the opposite of its effect, a maintenance landmine. Rename to `hideDateRange` (and flip the conditionals), or make `showDateRange === true` actually render the date UI.

src/components/Reports/Shared/SettingsDialog/SettingsDialog.tsx
```diff
@@ -230,103 +232,117 @@ export const SettingsDialog: React.FC<SettingsDialogProps> = ({
...
-                    Select which categories to keep consolidated.`,
-                  )}
+                <Typography
+                  sx={{ mt: showDateRange ? 0 : 2, whiteSpace: 'pre-line' }}
+                >
+                  {showDateRange
+                    ? t(
+                        `Income and expenses are combined by categories by default. Select which categories to keep consolidated.`,
+                      )
+                    : t(`Income and expenses are combined by categories by default. This may be useful for long date ranges (e.g., "Year to Date").
```
kegrimes: [Medium] **i18n key carries a newline + ~20 spaces of indentation.** With `whiteSpace: 'pre-line'` (line 338) the source indentation renders as a visible gap before "Select which categories…" and pollutes the extracted CrowdIn key (`translation.json` contains the literal `\n` + spaces). Pre-existing text, but this PR re-wraps these exact lines in the ternary — collapse to a single line as the new MPGA branch (line 342) already does.

src/components/Reports/MPGAIncomeExpensesReport/DisplayModes/ScreenOnlyReport.tsx
```diff
@@ -1,27 +1,42 @@
...
 
 export const ScreenOnlyReport: React.FC<ScreenOnlyReportProps> = ({
   data,
   last12Months,
   currency,
+  selectedFilters,
+  onFiltersChange,
 }) => {
   const { t } = useTranslation();
+  const [isSettingsOpen, setIsSettingsOpen] = useState(false);
```
kegrimes: [Medium] **Split state ownership.** `isSettingsOpen` lives here in the child while `filters` lives in the parent (`MPGAIncomeExpensesReport`), forcing the dialog to render in the child while its data source is owned a level up and `selectedFilters` to be prop-drilled through. The sibling `StaffExpenseReport` colocates both `isSettingsOpen` and `filters`. Render `<SettingsDialog>` + own the open-state in the parent (mirror StaffExpense), passing only an "open settings" callback down — removes a prop-drill hop.

src/components/Reports/MPGAIncomeExpensesReport/DisplayModes/ScreenOnlyReport.test.tsx
```diff
@@ -26,6 +26,13 @@ const TestComponent: React.FC = () => (
             data={mockData}
             last12Months={months}
             currency={currency}
+            selectedFilters={{
+              selectedDateRange: null,
+              startDate: null,
+              endDate: null,
+              categories: null,
+            }}
+            onFiltersChange={() => {}}
```
kegrimes: [Medium] **Dialog wiring untested here.** The new Report Settings button, dialog open/close, and `onFiltersChange` are only stubbed (`() => {}`) — this component's interactive boundary is verified only transitively by `MPGAIncomeExpensesReport.test.tsx`. Add a direct test: click the button, assert the dialog opens, toggle a category + Apply, assert `onFiltersChange` fires with the expected payload.

src/components/Reports/Shared/SettingsDialog/SettingsDialog.test.tsx
```diff
@@ -461,4 +463,32 @@ describe('SettingsDialog', () => {
       });
     });
   });
+
+  describe('MPGA report specific behavior', () => {
```
kegrimes, line 467: [Medium] **MPGA-mode apply flow untested.** This block asserts only that `showDateRange=true` hides the date UI. The existing apply-flow tests all run with `showDateRange=false`. Add a `showDateRange=true` case that toggles a category and clicks Apply, asserting `onClose` receives the right `categories` payload with the date fields omitted from `values`.

kegrimes (commented):
## 🤖 Multi-Agent Review — Verdict: ✅ APPROVED WITH SUGGESTIONS

6 agents (Architecture, Testing, Standards, UX, Data Integrity, Financial Reporting). Security skipped — no auth/api/env/dependency changes. **Risk: 7/10 HIGH** (display-side financial aggregation in `useFilteredFunds`).

**No merge blockers.** The aggregation math is sound: correct round-then-sum order, rounding at the `pushData` display boundary, Luxon dates throughout, and enum-based filter matching is consistent end-to-end. Net technical debt is slightly positive (dedup of `StyledFilterButton` + `SettingsDialog`).

### Summary
- **1 Important (7.5)** — `useFilteredFunds` unit test exercises only the all-selected path; 4 new branches (unchecked-split, `null`=all-selected, zero-suppression) are unit-untested. *Cannot be `/dismiss`ed; recommended to address here or via a follow-up ticket.*
- **~7 Medium (5.0–6.9)** — zero-suppression asymmetry between branches; "Shared" dialog depends back on `StaffExpenseReport/`; split state ownership; inverted `showDateRange` prop; i18n key whitespace leak; two untested wiring paths.
- **Pre-existing (informational, but strongly recommended here):** duplicate row `id`s when a category has two subcategories with the same `subCategory` enum — the `id` line is unchanged so it's not a blocker, but this PR's own test data triggers it and it breaks MUI DataGrid `getRowId` / React keys. One-line fix: append the loop `index`.

### Dismissed findings acknowledged (@kegrimes)
- `useFilteredFunds.ts:113` gross-vs-net sign-split — *intentional*; Financial agent independently confirmed the split nets correctly. ✅
- `MPGAIncomeExpensesReport.tsx:61` inert date seed — *needed for the 12-month snapshot*; excluded from verdict. Note: the seeded `start/endDate` never reach the `MpgaTransactions` query — only `categories` is live. Wire them into the query variables if you ever want the (currently hidden) date picker to drive the window.

Severity-tagged inline comments below; findings <7.0 can be dismissed via `/dismiss: <reason>` as a reply.

src/hooks/useFilteredFunds.ts
kegrimes: /dismiss the data should not return duplicate subcategories for the same category.

src/hooks/useFilteredFunds.test.ts
kegrimes: /dismiss there shouldn't be duplicated subcategories, but I will update the test to reflect that

src/components/Reports/Shared/SettingsDialog/SettingsDialog.tsx
kegrimes: /dismiss not too important for this task

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesReport.test.tsx
```diff
@@ -85,4 +156,36 @@ describe('MPGAIncomeExpensesReport', () => {
...
+      const applyButton = await findByRole('button', { name: 'Apply Filters' });
+      await waitFor(() => expect(applyButton).not.toBeDisabled());
+      userEvent.click(applyButton);
+
+      expect(
+        (await findAllByText('Benefits - Workers Compensation')).length,
+      ).toBeGreaterThan(0);
+      expect(
+        (await findAllByText('Benefits - Program Based')).length,
+      ).toBeGreaterThan(0);
```
zweatshirt: Optional: I tend to lean toward checking for exact lengths to be precise `toHaveLength(x)` but this is probably good as is. What do you think?

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesReport.tsx
```diff
@@ -50,23 +51,43 @@ export const MPGAIncomeExpensesReport: React.FC<
...
-  };
+  const [isSettingsOpen, setIsSettingsOpen] = useState(false);
 
   const last12Months = useGetLastTwelveMonths(locale, true);
 
   const start = convertMonths(last12Months[0], locale);
-  const end = DateTime.now().toISODate();
+  const end = DateTime.now();
+
+  const startDate = DateTime.now().minus({ months: 11 }).startOf('month');
```
zweatshirt: Minor, but I think we can keep this consistent by doing `const startDate = DateTime.fromISO(start);` or similar. I'm sure it's not a problem now, but it may be good practice to ensure that start and startDate have the same source.

src/hooks/useFilteredFunds.ts
```diff
@@ -67,40 +80,76 @@ export function useFilteredFunds(funds: Funds[]) {
...
+          const expenseMonthly = new Array(monthCount).fill(0);
+
+          category.subcategories.forEach((subcategory) => {
+            subcategory.breakdownByMonth.forEach((month, index) => {
+              const value = Number((month.total ?? 0).toFixed(2));
+              if (value >= 0) {
+                incomeMonthly[index] += value;
+              } else {
+                expenseMonthly[index] += value;
+              }
```
zweatshirt: Optional:
```suggestion
            subcategory.breakdownByMonth.forEach((month, index) => {
              const value = Number((month.total ?? 0).toFixed(2));
              const bucket = value >= 0 ? incomeMonthly : expenseMonthly;
               bucket[index] += value;
```

src/hooks/useFilteredFunds.ts
```diff
@@ -11,12 +16,20 @@ const sum = (data: number[]) => {
   return data.reduce((acc, item) => acc + item, 0);
 };
 
-export function useFilteredFunds(funds: Funds[]) {
+export function useFilteredFunds(
```
zweatshirt, line 11: This function has become pretty big. I am wondering if we could split it into multiple smaller functions for readability? 

I think a possible idea could be to create functions per each branch body: 

```
fund.categories?.forEach((category) => {
  if (hasSubcategories && !isSelected) {
    addRowPerSubcategory(category);      // unchecked: one row each
  } else if (hasSubcategories) {
    addCombinedIncomeExpenseRows(category); // checked: gross-split into 2 rows
  } else {
    addRollupRow(category);              // no subcategories
  }
});
```

src/hooks/useFilteredFunds.ts
```diff
@@ -67,40 +80,76 @@ export function useFilteredFunds(funds: Funds[]) {
...
             });
           });
+
+          pushData({
+            id: `${baseId}-income`,
+            description: getLocalizedCategory(category.category, t),
+            monthly: incomeMonthly,
+            average: average(incomeMonthly),
+            total: sum(incomeMonthly),
+          });
```
zweatshirt: I think we can reduce lines of code here by making a local wrapper function in this branch around pushData, then we can call it like 
```
pushAggregateRow(`${baseId}-income`, incomeMonthly);
pushAggregateRow(`${baseId}-expense`, expenseMonthly);
```
Feel free to ignore if you think this is overkill, it's a bit of a readability gain

zweatshirt (changes requested):
Looks great Katelyn. I think my only hope is that we can improve the readability of `useFilteredFunds` a little bit. I haven't tested the UI/UX yet, so I'll do that now!

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesReport.test.tsx
kegrimes: Sounds good!

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesReport.tsx
kegrimes: I just combined them both because I realized they were getting the same exact start date!

src/hooks/useFilteredFunds.ts
kegrimes: I like this, thanks!

src/hooks/useFilteredFunds.ts
kegrimes: I like!

zweatshirt (approved):
Looks great from what I can tell!

PR #1841, opened 2026-06-12
MPDX-9713,9714 - NSO Questionnaire - Contact and settings subsections

zweatshirt:
## Description

- Adds two of the three sections in the first step of the NSO Questionnaire
- These sections will not differ across the different user types.

https://jira.cru.org/browse/MPDX-9713
https://jira.cru.org/browse/MPDX-9714

## Testing

- Go to NSO MPD Questionnaire in HR Tools
- Check that the settings section looks good, as well as the contact information section.

## Checklist:

- [x] I have given my PR a title with the format "MPDX-(JIRA#) (summary sentence max 80 chars)"
- [x] I have applied the appropriate labels (_Add the label "Preview" to automatically create a preview environment_)
- [x] I have run the Claude Code `/pr-review` command locally and fixed any relevant suggestions
- [x] I have requested a review from another person on the project
- [x] I have tested my changes in preview or in staging
- [x] I have cleaned up my commit history

PR #1844, opened 2026-06-15
[no-Jira] Add transform support to `useAutoSave`

canac:
## Description

Add `transform` schema support to `useAutoSave` to handle raw input value processing in a flexible way.

Builds on #1841

## Testing

- Go to the NSO questionnaire
- Type letters into the phone number input
- Check that they are stripped out

## Checklist:

### DIFF
--- a/src/components/Shared/Autosave/useAutosave.test.tsx
+++ b/src/components/Shared/Autosave/useAutosave.test.tsx
@@ -55,6 +55,21 @@ const SelectTestComponent: React.FC = () => {
   );
 };
 
+const TransformTestComponent: React.FC = () => {
+  const schema = yup.object({
+    field: yup.string().transform((val) => val.replace(/\D/g, '')),
+  });
+
+  const props = useAutoSave({
+    value: '',
+    saveValue,
+    fieldName: 'field',
+    schema,
+  });
+
+  return <TextField label="Field" {...props} />;
+};
+
 describe('AutosaveTextField', () => {
   it('initializes with value and no errors', () => {
     const { getByRole } = render(<TestComponent />);
@@ -124,6 +139,15 @@ describe('AutosaveTextField', () => {
     userEvent.clear(input);
     userEvent.type(input, 'abc');
     expect(input).toHaveAccessibleDescription('Field must be a number');
+  });
+
+  it('transforms the input value', () => {
+    const { getByRole } = render(<TransformTestComponent />);
+
+    const input = getByRole('textbox', { name: 'Field' });
+    userEvent.type(input, 'a1b2c3');
+
+    expect(input).toHaveValue('123');
   });
 
   describe('select input', () => {
--- a/src/components/Shared/Autosave/useAutosave.ts
+++ b/src/components/Shared/Autosave/useAutosave.ts
@@ -50,15 +50,14 @@ export const useAutoSave = <Value extends string | number>({
 
   const transformValue = useCallback(
     (rawValue: string): string => {
-      try {
-        // Apply any schema transforms defined on the field
-        const transformed = yup
-          .reach(schema, fieldName)
-          .cast(rawValue, { assert: false });
-        return typeof transformed === 'string' ? transformed : rawValue;
-      } catch {
-        return rawValue;
-      }
+      // Apply any schema transforms defined on the field as the user types (e.g. stripping
+      // disallowed characters). Transforms must be idempotent.
+      const transformed = yup
+        .reach(schema, fieldName)
+        .cast(rawValue, { assert: false });
+      // Only apply string transforms. Number/select schemas cast to non-strings, so keep the raw
+      // keystrokes and let parseValue coerce them on save.
+      return typeof transformed === 'string' ? transformed : rawValue;
     },
     [schema, fieldName],
   );
--- a/src/components/Shared/Autosave/useAutosave.ts
+++ b/src/components/Shared/Autosave/useAutosave.ts
@@ -48,6 +48,21 @@ export const useAutoSave = <Value extends string | number>({
     [fieldName, schema],
   );
 
+  const transformValue = useCallback(
+    (rawValue: string): string => {
+      try {
+        // Apply any schema transforms defined on the field
+        const transformed = yup
+          .reach(schema, fieldName)
+          .cast(rawValue, { assert: false });
+        return typeof transformed === 'string' ? transformed : rawValue;
+      } catch {
+        return rawValue;
+      }
+    },
+    [schema, fieldName],
+  );
+
   const { parsedValue, errorMessage } = useMemo(
     () => parseValue(internalValue),
     [parseValue, internalValue],
@@ -69,7 +84,7 @@ export const useAutoSave = <Value extends string | number>({
   return {
     value: internalValue,
     onChange: (event: React.ChangeEvent<HTMLInputElement>) => {
-      const newValue = event.target.value;
+      const newValue = transformValue(event.target.value);
       setInternalValue(newValue);
 
       if (saveOnChange) {
--- a/src/lib/yupHelpers.ts
+++ b/src/lib/yupHelpers.ts
@@ -17,13 +17,18 @@ export const sanitizePhoneNumber = (value: string) =>
  * - Must contain between 7 and 15 digits (after removing non-digit characters)
  */
 export const phoneNumber = (t: TFunction) =>
-  yup.string().test('is-phone-number', t('Invalid phone number'), (val) => {
-    if (!val) {
-      return false;
-    }
-    const cleaned = val.replace(/\D/g, '');
-    return sanitizePhoneNumber(val) === val && /^\d{7,15}$/.test(cleaned);
-  });
+  yup
+    .string()
+    .transform((val) =>
+      typeof val === 'string' ? sanitizePhoneNumber(val) : val,
+    )
+    .test('is-phone-number', t('Invalid phone number'), (val) => {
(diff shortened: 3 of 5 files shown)


canac (commented):
## 🤖 Multi-Agent Code Review — PR #1844

**Verdict: ✅ APPROVED WITH SUGGESTIONS**  ·  Risk: **MEDIUM (5/10)**  ·  Mode: standard (6 agents + dependency impact)

Clean, well-scoped refactor: sanitization moves out of a component-local `onChange` handler into the Yup schema `.transform()`, and `useAutoSave` applies it centrally. Net technical debt **reduced** (removes duplicated logic, single source of truth). No blockers found.

### Top thing to confirm before merge
**Validation weakening reaches a real mutation.** `phoneNumber` changed from *rejecting* malformed input to *silently sanitizing and accepting* it. Via `useAutoSave` this now persists through `useUpdateSalaryCalculationMutation` (SalaryCalculator), and the same loosened helper affects 3 callers not updated in this PR (`AdditionalSalaryRequest`, `MinisterHousingAllowance`, `SalaryCalculator/ContactInfoForm`). Input like `123@456#7890` now saves `1234567890` with no error. Confirm (a) this is intended for those callers and (b) the server re-validates phone format (client-validation parity).

### Findings
| Sev | Location | Issue |
|-----|----------|-------|
| 6.5 | yupHelpers.ts | Validation weakening → silent sanitize-and-save, hits real mutation + 3 un-updated callers |
| 5.0 | useAutosave.ts:87 | Caret jumps to end on mid-string strip (not new; promoted into shared hook) |
| 4.0 | useAutosave.ts:51 | Double-transform safe only because transforms are idempotent — undocumented |
| 4.0 | useAutosave.ts:59 | `catch {}` swallows `reach()` errors, hides a mis-wired fieldName |
| 4.0 | useAutosave.test.tsx:60 | Test gaps: empty input, number fall-through, catch branch |
| 2.5 | useAutosave.ts:58 | Non-string-cast fall-through correct but intent implicit |

### Dependency impact
- `useAutoSave` — **CRITICAL blast radius**: 13+ leaf form fields via 4 wrappers (PdsGoalCalculator, SalaryCalculator, GoalCalculator, NsoMpdQuestionnaire). The transform now runs on every keystroke for all of them; benign for number/select fields (guarded by `typeof === 'string'`), but verify perf and that no schema has a surprising string transform.
- `phoneNumber` — HIGH: 5 dependents, behavior change to 3 not updated here.
- `sanitizePhoneNumber` — now has **zero production callers** (internal-only after this PR).

*Financial agent: N/A — path-based trigger false positive (contact info, no money math). Coverage: all changed files reviewed by 2+ agents. No dismissed findings.*

src/lib/yupHelpers.ts
```diff
@@ -17,13 +17,18 @@ export const sanitizePhoneNumber = (value: string) =>
...
-  yup.string().test('is-phone-number', t('Invalid phone number'), (val) => {
-    if (!val) {
-      return false;
-    }
-    const cleaned = val.replace(/\D/g, '');
-    return sanitizePhoneNumber(val) === val && /^\d{7,15}$/.test(cleaned);
-  });
+  yup
+    .string()
+    .transform((val) =>
```
canac, line 22: [Medium] **Validation weakening — silent sanitize-and-accept reaches a real mutation.** Previously `phoneNumber` rejected input where `sanitizePhoneNumber(val) !== val` (letters/symbols). Now it `.transform()`s first, so `123@456#7890` → `1234567890` validates and saves with no error shown. This helper is shared by 4 callers — including `SalaryCalculator/Summary/ContactInfoForm` which persists via `useUpdateSalaryCalculationMutation`, plus `AdditionalSalaryRequest` and `MinisterHousingAllowance` (Formik), none updated in this PR. Confirm: (a) silent acceptance is intended for those forms, and (b) the server re-validates the phone format (client-validation parity). The `yupHelpers.test.ts` `.toThrow()` → `.toBe(...)` flips confirm the change is deliberate; the open question is server parity + the un-updated callers.

src/components/Shared/Autosave/useAutosave.ts
```diff
@@ -69,7 +84,7 @@ export const useAutoSave = <Value extends string | number>({
   return {
     value: internalValue,
     onChange: (event: React.ChangeEvent<HTMLInputElement>) => {
-      const newValue = event.target.value;
+      const newValue = transformValue(event.target.value);
```
canac, line 86: [Medium] **Caret jumps to end on mid-string sanitization.** The input is fully controlled (`value: internalValue`); when `transformValue` returns a shorter string (a char was stripped), React re-renders and the browser resets the caret to the end — `selectionStart`/`selectionEnd` are never restored. Repro: existing `555-1234`, click after `555-`, type `x` → `x` stripped → caret lands after the final `4`. Blast radius is narrow today (phone strip set `[^\d+() -]` is permissive, so digits/`-`/`(` typed mid-string aren't transformed), but this becomes a constant annoyance the moment the hook is reused with a formatting transform (the test's own `\D` strip would jump on every keystroke). Not introduced here — the deleted ContactInformation handler had the same bug — but it's now promoted into a shared hook. Fix: capture `selectionStart`, compute chars removed before caret, restore via `setSelectionRange` in `requestAnimationFrame`; or document that transforms must be position-stable character filters.

src/components/Shared/Autosave/useAutosave.ts
```diff
@@ -48,6 +48,21 @@ export const useAutoSave = <Value extends string | number>({
     [fieldName, schema],
   );
 
+  const transformValue = useCallback(
```
canac, line 51: [Suggestion] **Transform is applied twice — safe only because current transforms are idempotent.** `transformValue` casts the value here, and `parseValue` (`schema.validateSyncAt`) casts it again before save. For `sanitizePhoneNumber` and `\D`-strip this is fine (idempotent), so displayed == saved. A future non-idempotent transform (pad, prefix, reformat `1234567 → (123) 4567`) would diverge the displayed value from the saved value, or double-apply. No invariant documents this. Suggest a one-line comment noting field transforms must be idempotent, or route display + save through a single cast.

src/components/Shared/Autosave/useAutosave.ts
```diff
@@ -48,6 +48,21 @@ export const useAutoSave = <Value extends string | number>({
...
+    (rawValue: string): string => {
+      try {
+        // Apply any schema transforms defined on the field
+        const transformed = yup
+          .reach(schema, fieldName)
+          .cast(rawValue, { assert: false });
+        return typeof transformed === 'string' ? transformed : rawValue;
+      } catch {
```
canac: [Suggestion] **`catch {}` swallows `reach()` failures silently.** `yup.reach(schema, fieldName)` throws on a bad/unreachable `fieldName` — a real wiring bug — and this returns the raw value with no signal, so a mis-wired field looks like it just has no transform. The fallback-on-failure is the right behavior for typing, but consider a dev-mode `console.warn` (or rethrow non-cast errors, mirroring how `parseValue` rethrows non-`ValidationError`) so a misconfigured schema isn't hidden.

src/components/Shared/Autosave/useAutosave.test.tsx
```diff
@@ -55,6 +55,21 @@ const SelectTestComponent: React.FC = () => {
   );
 };
 
+const TransformTestComponent: React.FC = () => {
+  const schema = yup.object({
+    field: yup.string().transform((val) => val.replace(/\D/g, '')),
```
canac, line 60: [Suggestion] **Test-fixture transform isn't empty-safe, and edge branches are uncovered.** `val.replace(/\D/g, '')` throws if `val` is `undefined`/non-string — it only survives because `transformValue` wraps `.cast()` in try/catch (an untested branch). The production `phoneNumber` helper guards with `typeof val === 'string'`; mirror that here so the fixture represents safe usage. Missing coverage: (a) empty/cleared input with a transform schema, (b) a `number()`-schema field leaving keystrokes untransformed (locks the `typeof transformed === 'string'` fall-through), (c) the `catch` fallback. Also consider `await userEvent.type(...)` for consistency.

src/components/Shared/Autosave/useAutosave.ts
```diff
@@ -48,6 +48,21 @@ export const useAutoSave = <Value extends string | number>({
...
   );
 
+  const transformValue = useCallback(
+    (rawValue: string): string => {
+      try {
+        // Apply any schema transforms defined on the field
+        const transformed = yup
+          .reach(schema, fieldName)
+          .cast(rawValue, { assert: false });
+        return typeof transformed === 'string' ? transformed : rawValue;
```
canac: [Suggestion] The `typeof transformed === 'string'` guard correctly leaves number/select fields' raw keystrokes untransformed (a `number()` cast returns a number, so it falls back to `rawValue`) — this is the right call, but the intent is implicit. A short comment ("number/select casts return non-strings; keep the raw text and let parseValue coerce") would prevent a future maintainer from "fixing" it into `String(transformed)`.

src/lib/yupHelpers.ts
canac: This is an acceptable UX change

zweatshirt (approved):
Works well from what I can tell!

src/components/Shared/Autosave/useAutosave.test.tsx
canac: /dismiss This is incorrect. Val isn't null/undefined in `transform`.

PR #1846, opened 2026-06-16
MPDX-9757 - NSO Questionnaire Step Two: Ministry Information

zweatshirt:
## Description

- This introduces the basic UI for the Ministry Information section of the NSO MPD calculator 

https://jira.cru.org/browse/MPDX-9757

## Testing

- Go to NSO MPD Questionnaire
- Test the behavior of the new step, ensuring that the continue button does not become enabled until all steps are selected
- Compare the UI design to the Figma design and ensure no discrepancies

## Checklist:

### DIFF
--- a/src/components/HrTools/NsoMpdQuestionnaire/MinistryInformation/MinistryDetails.test.tsx
+++ b/src/components/HrTools/NsoMpdQuestionnaire/MinistryInformation/MinistryDetails.test.tsx
@@ -1,0 +1,108 @@
+import React from 'react';
+import { ThemeProvider } from '@mui/material/styles';
+import { render, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { GqlMockedProvider } from '__tests__/util/graphqlMocking';
+import { GoalCalculatorConstantsQuery } from 'src/hooks/goalCalculatorConstants.generated';
+import theme from 'src/theme';
+import { MinistryDetails } from './MinistryDetails';
+
+const TestComponent: React.FC = () => (
+  <ThemeProvider theme={theme}>
+    <GqlMockedProvider<{
+      GoalCalculatorConstants: GoalCalculatorConstantsQuery;
+    }>
+      mocks={{
+        GoalCalculatorConstants: {
+          constant: {
+            mpdGoalGeographicConstants: [
+              { location: 'Atlanta, GA' },
+              { location: 'Miami, FL' },
+            ],
+          },
+        },
+      }}
+    >
+      <MinistryDetails />
+    </GqlMockedProvider>
+  </ThemeProvider>
+);
+
+describe('MinistryDetails', () => {
+  it('renders the four questions', async () => {
+    const { getByRole, getByPlaceholderText, findByRole } = render(
+      <TestComponent />,
+    );
+
+    expect(
+      getByRole('combobox', {
+        name: 'What ministry are you expecting to serve with?',
+      }),
+    ).toBeInTheDocument();
+    expect(
+      getByPlaceholderText(
+        'What is your expected ministry assignment location?',
+      ),
+    ).toBeInTheDocument();
+    expect(
+      await findByRole('combobox', {
+        name: 'Is your ministry assignment location within 50 miles of one of these cities?',
+      }),
+    ).toBeInTheDocument();
+    expect(
+      getByRole('radiogroup', {
+        name: 'What type of assignment are you expecting?',
+      }),
+    ).toBeInTheDocument();
+  });
+
+  it('offers the dummy ministry options', () => {
+    const { getByRole } = render(<TestComponent />);
+
+    userEvent.click(
+      getByRole('combobox', {
+        name: 'What ministry are you expecting to serve with?',
+      }),
+    );
+
+    expect(getByRole('option', { name: 'Cru' })).toBeInTheDocument();
+    expect(
+      getByRole('option', { name: 'Athletes in Action' }),
+    ).toBeInTheDocument();
+  });
+
+  it('populates the nearest-city options from the geographic constants', async () => {
+    const { getByRole, findByRole } = render(<TestComponent />);
+
+    const cityCombobox = await findByRole('combobox', {
+      name: 'Is your ministry assignment location within 50 miles of one of these cities?',
+    });
+    await waitFor(() =>
+      expect(cityCombobox).not.toHaveAttribute('aria-disabled', 'true'),
+    );
+    userEvent.click(cityCombobox);
+
+    expect(
+      await findByRole('option', { name: 'Atlanta, GA' }),
+    ).toBeInTheDocument();
+    expect(getByRole('option', { name: 'Miami, FL' })).toBeInTheDocument();
+  });
+
+  it('offers Field and Office assignment types', () => {
+    const { getByRole } = render(<TestComponent />);
+
+    expect(getByRole('radio', { name: 'Field' })).toBeInTheDocument();
+    expect(getByRole('radio', { name: 'Office' })).toBeInTheDocument();
+  });
+
+  it('shows a loading indicator in place of the city field until the constants load', () => {
+    const { getByRole, queryByRole } = render(<TestComponent />);
+
+    expect(getByRole('progressbar')).toBeInTheDocument();
+    expect(
+      queryByRole('combobox', {
+        name: 'Is your ministry assignment location within 50 miles of one of these cities?',
+      }),
+    ).not.toBeInTheDocument();
+  });
+});
--- a/src/components/HrTools/NsoMpdQuestionnaire/MinistryInformation/MinistryDetails.tsx
+++ b/src/components/HrTools/NsoMpdQuestionnaire/MinistryInformation/MinistryDetails.tsx
@@ -1,0 +1,185 @@
+import React, { useMemo } from 'react';
+import LocationOn from '@mui/icons-material/LocationOn';
+import MuseumSharp from '@mui/icons-material/MuseumSharp';
+import {
+  CircularProgress,
+  FormControl,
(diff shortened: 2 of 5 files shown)


src/components/HrTools/NsoMpdQuestionnaire/MinistryInformation/MinistryDetails.tsx
```diff
@@ -0,0 +1,175 @@
...
+  'Cru',
+  'Cru High School',
+  'Cru City',
+  'Athletes in Action',
+  'Bridges International',
+];
+
+export const MinistryDetails: React.FC = () => {
+  const { t } = useTranslation();
+  const { goalGeographicConstantMap, loading } = useGoalCalculatorConstants();
```
zweatshirt, line 34: [Important] `useGoalCalculatorConstants()` also returns `error`, but only `{ goalGeographicConstantMap, loading }` are destructured. Since the hook computes `loading = !data`, a failed constants query keeps `loading === true` forever → the **required** city select at line 134 stays `disabled` with no options and no error surfaced → `Continue` can never enable, with no recovery path. (UX 7.5 / Testing 7.0 / Architecture 6.5.) Surface the error (e.g. an `<Alert severity="error">` with an i18n message) and/or stop gating `disabled` purely on `loading`, then add an error-path test. Note: SalaryCalculator's `PersonalInformationSection` also ignores `error`, but there the field is optional — here it's a required, Continue-gating control, so the coupling is new to this PR.

src/components/HrTools/NsoMpdQuestionnaire/MinistryInformation/MinistryDetails.tsx
```diff
@@ -0,0 +1,175 @@
...
+        placeholder={t('What is your expected ministry assignment location?')}
+        size="small"
+        InputLabelProps={{ shrink: true }}
+        {...locationProps}
+      />
+
+      <TextField
+        select
+        label={t(
+          'Is your ministry assignment location within 50 miles of one of these cities?',
```
zweatshirt: [Medium] The required "nearest city" field no longer shows the guidance SalaryCalculator displays ("…If not, select 'None.'"). The field is satisfiable — "None" exists in the geographic constants (confirmed; SalaryCalculator relies on it too) — but the only on-screen mention of it is now buried in the validation message, which appears only after leaving the field empty. Consider restoring a short `FormHelperText`/`Typography` hint near the field for parity and discoverability.

src/components/HrTools/NsoMpdQuestionnaire/MinistryInformation/MinistryDetails.tsx
```diff
@@ -0,0 +1,175 @@
...
+        </MenuItem>
+        {ministries.map((ministry) => (
+          <MenuItem key={ministry} value={ministry}>
+            {ministry}
+          </MenuItem>
+        ))}
+      </TextField>
+
+      <TextField
+        placeholder={t('What is your expected ministry assignment location?')}
```
zweatshirt, line 119: [Suggestion] This field puts the question in `placeholder` with no `label`, unlike the other three fields and `ContactInformation`. Once the user types, the question disappears and the input has no accessible name (the test queries it via `getByPlaceholderText` rather than role+name). `InputLabelProps={{ shrink: true }}` is also inert without a `label`. Consider a short `label` (e.g. `t('Ministry assignment location')`) with the full question as placeholder.

src/components/HrTools/NsoMpdQuestionnaire/MinistryInformation/MinistryDetails.tsx
```diff
@@ -0,0 +1,175 @@
...
+export const MinistryDetails: React.FC = () => {
+  const { t } = useTranslation();
+  const { goalGeographicConstantMap, loading } = useGoalCalculatorConstants();
+
+  const cities = useMemo(
+    () => Array.from(goalGeographicConstantMap.keys()),
+    [goalGeographicConstantMap],
+  );
+
+  const requiredSchema = useMemo(
```
zweatshirt: [Suggestion] `requiredSchema` is memoized with `[]`, but it's a factory invoked 4× per render, each call building a fresh `yup.object` — so the `useMemo` stabilizes the factory identity, not the schemas passed into `useQuestionnaireAutoSave` (those are new every render, recreating each field's `parseValue`). Either drop the `useMemo` or memoize each schema like `ContactInformation.tsx` does. Low priority (in-memory dummy data today).

src/components/HrTools/NsoMpdQuestionnaire/MinistryInformation/MinistryDetails.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+    >
+      <AutosaveForm>
+        <MinistryDetails />
+        <ValidProbe />
+      </AutosaveForm>
+    </GqlMockedProvider>
+  </ThemeProvider>
+);
+
+describe('MinistryDetails', () => {
```
zweatshirt, line 31: [Medium] Coverage gaps: no test for the constants-query **loading** (city select disabled before resolution — current tests await *past* it via `waitFor(aria-disabled)`), **error**, or **empty-cities** states, and no test for any required-error message or `Continue` re-disabling after a field is cleared. The `assignmentType` `FormHelperText` render path has zero coverage. Add an error-path test (this would have caught the permanent-disable trap) and at least one validation-error assertion. Compare `ContactInformation.test.tsx`, which tests the validation-error path.

src/components/HrTools/NsoMpdQuestionnaire/MinistryInformation/MinistryInformation.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+          },
+        },
+      }}
+    >
+      <MinistryInformation />
+    </GqlMockedProvider>
+  </NsoMpdQuestionnaireTestWrapper>
+);
+
+describe('MinistryInformation', () => {
```
zweatshirt, line 13: [Medium] This file duplicates ~4 near-identical cases from `MinistryDetails.test.tsx` (same field-presence/option assertions, same mock). The only unique value here is exercising the **real** `ContinueButton` wired through the layout's `AutosaveForm` (vs. the hand-rolled `ValidProbe` in the isolated test). Recommendation: keep field/option assertions in `MinistryDetails.test.tsx` only, and reduce this file to the integration-only delta. Note the sibling `PersonalInformation` has no full-layout test file — the convention is subsection-only.

zweatshirt (commented):
## 🤖 Multi-Agent Code Review — PR #1846

**Verdict: ✅ APPROVED WITH SUGGESTIONS** · Risk: **LOW (3/10)** · 5 agents (Architecture, Testing, Standards, UX, Financial-domain) · Dependency impact LOW, no breaking changes.

No blockers (nothing ≥ 8.0). One **Important** finding (7.5) is strongly recommended before merge but does not block and cannot be `/dismiss`ed.

### Top finding
**Unhandled `error` from `useGoalCalculatorConstants`** — flagged independently by UX (7.5), Testing (7.0), Architecture (6.5). Because `loading = !data` in the hook, a failed constants query leaves the **required** city select disabled forever with no message, so `Continue` can never enable. Handle `error` (e.g. an `<Alert>`) and/or stop gating `disabled` solely on `loading`, and add a test.

### Also worth addressing
- **Test gaps** — loading / error / empty-cities states and required-error message paths are untested (the `assignmentType` `FormHelperText` path has no coverage).
- **Removed "None" guidance** — the field is satisfiable ("None" exists in the constants, same as SalaryCalculator relies on), but the only mention is now buried in the validation message. SalaryCalculator shows persistent helper text.
- **Duplicate tests** — `MinistryDetails.test.tsx` and `MinistryInformation.test.tsx` share ~4 near-identical cases; keep field/option assertions in the isolated test and reduce the parent test to the real-`ContinueButton` integration delta.

### Suggestions
- Location field uses `placeholder` not `label` (loses accessible name once typed; `InputLabelProps={{shrink:true}}` is inert without a label).
- `requiredSchema` `useMemo` memoizes the factory, not the four schemas it builds each render.

### Standards
All checklist items PASS. Two accepted WARNs: uses the established `useQuestionnaireAutoSave` (Yup, not Formik) convention; Yup `required` rules have no server-side parity yet because the autosave hook only stores locally (forward-looking, until the backing API lands).

### Not applicable
Security and Data-Integrity agents skipped (no auth/API/secret surface; no `.graphql`/mutation/cache changes). Financial-domain agent: keyword match on "goal" (`goalGeographicConstantMap`) is a false positive — keys are used only as dropdown labels, no money/calculation code.

<sub>Note: this review condensed the formal debate round — findings were well-aligned with no severity ≥ 8 to adjudicate, and the one factual crux (does the constants list include a "None" entry — yes) was verified directly.</sub>

src/components/HrTools/NsoMpdQuestionnaire/MinistryInformation/MinistryDetails.tsx
zweatshirt: I see this as optional, the current behavior is that the validation error text appears as soon as the user navigates to the step.

src/components/HrTools/NsoMpdQuestionnaire/MinistryInformation/MinistryDetails.tsx
zweatshirt: /dismiss The current implementation matches the Figma design.

src/components/HrTools/NsoMpdQuestionnaire/MinistryInformation/MinistryDetails.tsx
zweatshirt: /dismiss The error text should be sufficient for now.

PR #1851, opened 2026-06-18
MPDX-9766 - Enforce error deferment until fields are touched in useAutosave

zweatshirt:
## Description

- This improves UX when going through the NSO form and other forms, making it so users are not bombarded with errors upon rendering each step's view. The default functionality in useAutosave remains the same.
https://jira.cru.org/browse/MPDX-9766

## Testing

- Go to NSO Questionnaire, and test that the behavior works as expected.
    - Continue button should remain disabled, but the fields are still required.
 
Considerations:
- We may still want a way to make it apparent to users that all fields in a step are required, since we no longer have the clear indication of immediate errors.

## Checklist:

### DIFF
--- a/src/components/HrTools/NsoMpdQuestionnaire/MinistryInformation/MinistryDetails.tsx
+++ b/src/components/HrTools/NsoMpdQuestionnaire/MinistryInformation/MinistryDetails.tsx
@@ -75,6 +75,7 @@ export const MinistryDetails: React.FC = () => {
     <Stack spacing={4}>
       <TextField
         select
+        required
         label={t('What ministry are you expecting to serve with?')}
         size="small"
         SelectProps={{ displayEmpty: true }}
@@ -101,9 +102,9 @@ export const MinistryDetails: React.FC = () => {
       </TextField>
 
       <TextField
+        required
         label={t('What is your expected ministry assignment location?')}
         size="small"
-        InputLabelProps={{ shrink: true }}
         {...locationProps}
       />
 
@@ -112,6 +113,7 @@ export const MinistryDetails: React.FC = () => {
       ) : (
         <TextField
           select
+          required
           label={t(
             'Is your ministry assignment location within 50 miles of one of these cities?',
           )}
--- a/src/components/HrTools/NsoMpdQuestionnaire/Shared/NumberQuestion.tsx
+++ b/src/components/HrTools/NsoMpdQuestionnaire/Shared/NumberQuestion.tsx
@@ -34,6 +34,7 @@ export const NumberQuestion: React.FC<NumberQuestionProps> = ({
       label={question}
       helperText={error ? errorText : helperText}
       error={error}
+      required
       size="small"
       type="number"
       inputProps={{ min: 0, inputMode: 'numeric' }}
--- a/src/components/Shared/Autosave/useAutosave.ts
+++ b/src/components/Shared/Autosave/useAutosave.ts
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useMemo } from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { TextFieldProps } from '@mui/material';
 import { prepareDataForValidation } from 'formik';
 import * as yup from 'yup';
@@ -25,6 +25,7 @@ export const useAutoSave = <Value extends string | number>({
   const [internalValue, setInternalValue] = useSyncedState(
     value?.toString() ?? '',
   );
+  const [touched, setTouched] = useState(false);
   const { markValid, markInvalid } = useOptionalAutosaveForm() ?? {};
 
   const parseValue = useCallback(
@@ -80,6 +81,8 @@ export const useAutoSave = <Value extends string | number>({
     };
   }, [fieldName, errorMessage, markValid, markInvalid]);
 
+  const showError = !disabled && errorMessage !== null && touched;
+
   return {
     value: internalValue,
     onChange: (event: React.ChangeEvent<HTMLInputElement>) => {
@@ -87,6 +90,7 @@ export const useAutoSave = <Value extends string | number>({
       setInternalValue(newValue);
 
       if (saveOnChange) {
+        setTouched(true);
         const { parsedValue, errorMessage } = parseValue(newValue);
         if (errorMessage === null && parsedValue !== value) {
           saveValue(parsedValue);
@@ -94,13 +98,12 @@ export const useAutoSave = <Value extends string | number>({
       }
     },
     onBlur: () => {
+      setTouched(true);
       if (!saveOnChange && errorMessage === null && parsedValue !== value) {
         saveValue(parsedValue);
       }
     },
     disabled,
-    ...(!disabled && errorMessage !== null
-      ? { error: true, helperText: errorMessage }
-      : {}),
+    ...(showError ? { error: true, helperText: errorMessage } : {}),
   } satisfies Partial<TextFieldProps>;
 };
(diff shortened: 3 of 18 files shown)


src/components/HrTools/NsoMpdQuestionnaire/MinistryInformation/MinistryDetails.tsx
```diff
@@ -75,6 +75,7 @@ export const MinistryDetails: React.FC = () => {
     <Stack spacing={4}>
       <TextField
         select
+        required
```
zweatshirt, line 78: [Suggestion] The ministry and city `select` fields now set `required`, but unlike the location field (`toBeRequired`) and the `RadioQuestion`/`NumberQuestion` components (which own and test their own required state), the two selects have no explicit test asserting their required/`aria-required` state. Optional: add one assertion on the ministry combobox to regression-guard the `required` prop on selects. Non-blocking.

src/components/Shared/Autosave/useAutosave.ts
```diff
@@ -12,6 +12,7 @@ interface UseAutoSaveOptions<Value extends string | number> {
   schema: yup.Schema;
   disabled?: boolean;
   saveOnChange?: boolean;
+  deferErrorsUntilTouched?: boolean;
```
zweatshirt: [Suggestion] Two display-only caveats of the touched flag, both intentional/acceptable, noted for awareness: (1) `touched` is set on the first `onChange`, so a blur-saved number field can briefly flash an error mid-typing (this matches pre-change behavior, not a regression — but setting `touched` only `onBlur` for non-`saveOnChange` fields would be smoother); (2) `touched` never resets if the field value is reset externally — irrelevant for the questionnaire's in-memory state, and display-only regardless. No action required.

zweatshirt (commented):

**Risk: 6/10 (MEDIUM)** · standard mode · agents across the review: Architecture, Testing, Standards, Data Integrity, UX, Financial + dependency analysis

No blockers, no Important findings — only two sub-5 suggestions (inline). The feature is sound and independently verified.

### What this PR does
Adds an opt-in `deferErrorsUntilTouched` flag to the shared `useAutoSave` hook (default `false`), enabled in the questionnaire wrapper so validation errors stay hidden until a field is touched — avoiding a wall of errors on arrival. Every questionnaire field is also marked `required` (asterisk + `aria-required`) so mandatory-ness stays visible even while errors are deferred.

### Verified safe
- **Validity gating preserved (Data Integrity, High confidence):** error *display* is deferred, but `markValid`/`markInvalid` is unchanged and independent of `touched` — a user **cannot** Continue past hidden-but-invalid fields. Only the visual error is gated.
- **Backward compatible (dependency analysis):** `useAutoSave` has 5 consumers; only the questionnaire wrapper opts in. With the flag defaulting `false`, the 4 other forms (PdsGoalCalculator, SalaryCalculator, GoalCalculator ×2) are byte-identical in behavior. `showError` reduces exactly to the original condition when the flag is off.
- **Accessibility (UX, High confidence):** `aria-required` is valid on `role=radiogroup` and necessary (MUI's `FormControl required` does not propagate it to the group). The `required` work resolves the earlier concern that deferring errors left untouched required fields with no programmatic signal.
- **Standards:** clean pass — named exports, localization, no `any`, no unused imports, MUI `sx` conventions.

### Suggestions (severity < 5 — informational, non-blocking)
See inline comments. In brief: (1) the two `MinistryDetails` selects have `required` but no explicit test assertion; (2) minor display caveats of the touched-state flag. Neither blocks merge.

src/components/Shared/Autosave/useAutosave.ts
```diff
@@ -80,9 +83,13 @@ export const useAutoSave = <Value extends string | number>({
     };
   }, [fieldName, errorMessage, markValid, markInvalid]);
 
+  const showError =
+    !disabled && errorMessage !== null && (!deferErrorsUntilTouched || touched);
+
   return {
     value: internalValue,
     onChange: (event: React.ChangeEvent<HTMLInputElement>) => {
+      setTouched(true);
```
canac: For text inputs (`saveOnChange === false`), this will mark as touched as soon as the user types a character. I think we want to only mark as touched in `onChange` when `saveOnChange === true`. Always marking as touched in `onBlur` is good.

src/components/Shared/Autosave/useAutosave.ts
```diff
@@ -21,10 +22,12 @@ export const useAutoSave = <Value extends string | number>({
   schema,
   disabled = false,
   saveOnChange = false,
+  deferErrorsUntilTouched = false,
```
canac: This seems like a good thing to always have on to improve the UX so users don't see a wall of errors in any new forms we make. It would improve salary calc too, for example. I'd even lean towards making it always on and not configurable unless/until we need it configurable. What do you think?

src/components/HrTools/NsoMpdQuestionnaire/Shared/NumberQuestion.tsx
```diff
@@ -34,6 +34,7 @@ export const NumberQuestion: React.FC<NumberQuestionProps> = ({
       label={question}
       helperText={error ? errorText : helperText}
       error={error}
+      required
```
canac, line 37: question: Are all number fields required? Like debt?

canac (approved):
A couple minor thoughts, but thanks for noticing this issue and improving it!

src/components/HrTools/NsoMpdQuestionnaire/Shared/NumberQuestion.tsx
zweatshirt: I assumed if they are displayed to the user, then yes, unless we would prefer to populate empty field values with 0 even if displayed? I tend to lean toward required in general because it makes the user think through each question rather than rush through forms, but I can change this if we'd like. Maybe it's to refer to follow the original Google Form?

canac: You're right, the Google Form makes them required (if they said they have debt) and tells them to input 0. The "If you are a parent with children in Childcare, please enter the number as a value between 0 and 9" question isn't required though. It's the only non-required question I see.

Screenshot 2026-06-18 at 9 52 37 AM

src/components/Shared/Autosave/useAutosave.ts
zweatshirt: @canac We can do that, assuming it won't cause any regressions to prior forms using useAutosave.

canac: I'd consider it a UX improvement as opposed to a regression.

zweatshirt:
@canac I realized from working on this that Nearest Geographic Multiplier Location and the MHA inputs in "Your Information" step of Salary Calculator aren't strictly enforced/required in the UI (even before this PR). I'm thinking we probably want both of these enforced. [I've made a follow up ticket in the backlog](https://jira.cru.org/browse/MPDX-9768).

canac:
> I realized from working on this that Nearest Geographic Multiplier Location and the MHA inputs in "Your Information" step of Salary Calculator aren't strictly enforced/required in the UI

Are you pointing out that they're optional fields? I believe the calculations treat missing a geo multiplier as None and missing MHA as 0 (which is a valid choice). Do stakeholders want us to force people to choose a value as opposed to leaving them blank?

zweatshirt:
@canac Ah if they truly are optional, that makes sense, the MHA Request section is a quick change then. The fields display an asterisk (hinting required). Will be a quick no-jira.

canac:
Yeah, I think removing the asterisk is a good call. Unless we want to make the MHA fields required.

PR #1855, opened 2026-06-23
MPDX-9749 & MPDX-9749 - Build out MPD Supervisor Report page

dr-bizz:
In this PR, I added the skeleton of the MPD Supervisor Report page under HR Tools.

## What's included
- Infinite, virtualised staff list with header search and a left filter panel (quick filters, team, employment type).
- Click a staff row to open a right-hand drawer with staff/spouse details and tabbed reports (Monthly Summary, Quarterly, Payroll, MPGA, Staff Expense Report), each lazily loaded.
- Custom sticky header with nav/filter toggles and a "showing X of Y" count.

## Notes
- Backed by **mock data only** — there is no backend yet. `useMockInfiniteStaff` mirrors an Apollo paginated connection so it can be swapped for a real query when the API is ready.
- The detail tab components are placeholders pending real reports.

## Test
- Navigate to: `/hrTools/mpdSupervisorReport`
- Click around and see what you think

## Related Links
[MPDX-9750](https://jira.cru.org/browse/MPDX-9750)
[MPDX-9749](https://jira.cru.org/browse/MPDX-9749)

## Checklist
- [x] 🔤 PR title prefixed with Jira ticket number (Ex: `MPDX-123: #{insert_title_here}`)
- [x] 👀 Assign myself to the PR
- [x] ✅ Tests are passing
- [x] 🤖 Perform AI code review (Claude command: `/pr-review`)
- [x] 👨‍💻 Human code review requested
- [x] ✍️ Senior dev written approval before adding `On Staging`
- [ ] 🎉 Passes QA


dr-bizz:
> 🤖 **AI multi-agent code review** — config-driven agent selection + MPDX rules. Pre-flight: risk **46/CRITICAL** → senior reviewer; 6 specialists (financial, data-integrity, ux, architecture, testing, standards); import-graph blast radius 166; 21 findings synthesized. _Advisory — not a substitute for human review._

---

# MPD Supervisor Report (PR #1855) — Consolidated Review

## Executive Summary

PR #1855 adds a new MPD Supervisor Report page that is **entirely mock-data-driven** — there are no live GraphQL queries, mutations, or backend aggregation yet (acknowledged via `TODO(MPDX)` comments). Overall risk is **low for the current diff**: the architecture is idiomatic, money is formatted only at the display boundary, the shared `MultiPageHeader` change is purely additive (zero blast radius across 29 dependents), and the code is largely a11y-conscious. The headline concerns are **forward-looking data-modeling decisions that will ossify into real bugs once the backend lands** (hardcoded currency, quarter columns derived from a single row, missing-data rendered as `$0.00`) plus a runtime-reactive **localization bug in the quick-filter labels** that is live today.

## BLOCKING (8-10)

None.

## IMPORTANT (5-7)

### 1. Quick-filter labels frozen at import-time language; never react to language change — sev 6
**`Filters/mpdSupervisorReportFilters.ts:498-515`** (ux)
`quickFilters` is a module-level constant whose `label` values come from `i18n.t('All people')` etc., evaluated once at import. Unlike every other string in the PR (which uses the `t` from `useTranslation()` and re-renders), these chip labels stay in the import-time/fallback language when the user switches languages. This is a **live bug**, not forward-looking.
**Fix:** Store only enum ids in the constant; translate at render time inside `MpdSupervisorReportFilterPanel` (e.g. a `quickFilterLabel(t, id)` helper or id→`t('...')` record). Keep literals as `t('...')` so `yarn extract` still captures them.

### 2. Team/Employment-type option text untranslated and conflated with enum values — sev 5
**`Filters/MpdSupervisorReportFilterPanel.tsx:451-472`** (ux)
Team options render raw mockData strings (`'Campus'`, `'FamilyLife'`) with no `t()`, and `MpdSupervisorReportEmploymentTypeEnum.FullTime = 'Full time'` doubles the stored filter value as the English display label (compared against `user.userPersonType`). This bakes display text into data/enum values rather than mapping codes to labels — it will propagate untranslated labels and value/label coupling when wired to the backend.
**Fix:** Define employment types as opaque codes (e.g. `'full_time'`) separate from a `t('Full time')` label; map team/option codes to translated labels at render. At minimum, comment that team names are intentionally untranslated proper nouns.

### 3. `any` types in `jest.mock` factories violate no-any-in-test-types rule — sev 5
**`MpdSupervisorReport.test.tsx:555-601`** (testing)
The `InfiniteList`/`useMockInfiniteStaff` mocks type `jest.requireActual` and props/item as `any` with eslint-disable. `testing.md` forbids `any` in test types; an `any`-typed mock cannot catch prop/return-shape drift from the real module — the exact regression these mocks should guard.
**Fix:** Type the react require with `typeof import(...)`, type `InfiniteList` props with the real generic over `EmployeeData`, and type `item` as `EmployeeData`.

## MINOR (1-4)

**Currency / financial correctness**
- **`StaffMemberRow/StaffMember.tsx` (FiscalYearQuartersBase) — sev 4** (financial): Payroll currency hardcoded to `'USD'`; `QuarterStatus` has no `currencyCode`. Harmless on mock, but will mislabel non-USD payroll once wired. Add `currencyCode` to the model now and pass `quarter.currencyCode` into `currencyFormat`.
- **`StaffMemberRow/StaffMember.tsx` (FiscalYearQuartersBase) — sev 3** (financial): Missing/NaN payroll renders as a confident `$0.00` (`currencyFormat` coerces NaN→0). When backed by real data, allow `payroll?: number | null` and render `—`/`No data` so `$0.00` only ever means a real zero.

**Data integrity (forward-looking, surface when wiring backend)**
- **`MpdSupervisorReport.tsx:851, 911-927` — sev 4** (data-integrity): Quarter column headers derived from `data.nodes[0]` — header row vanishes on empty results, and assumes every row shares row 0's quarter set/order. Derive columns from a stable source (fixed constant or server-provided top-level field), independent of page rows.
- **`mockData.ts:2347-2351` — sev 3** (data-integrity): `EmployeeData` node has no top-level `id` (only nested `user.id`/`spouse.id`). The real connection query must select `id` + `__typename` for Apollo normalization, or add a `keyFields` type policy.
- **`MpdSupervisorReport.tsx:832-847, 884-887` — sev 2** (data-integrity): Client-side filtering and the "Showing X of Y" count run over the full in-memory array; against a paginated server this will hide unloaded rows and understate totals. Push filters into query variables and source counts from server `totalCount`.

**Architecture / technical debt**
- **`MpdSupervisorReport.tsx:832-847, 762, 766` — sev 4** (architecture): Production components import and filter `mockData` directly (FilterPanel too); there is no data-layer seam, so the whole data layer must change across files when the backend lands. Introduce a single `useStaffMembers({...})` hook (incl. team-option derivation) so the contract is stable across the swap.
- **`Filters/mpdSupervisorReportFilters.ts:517-521` — sev 3** (architecture): `MpdSupervisorReportTeamsEnum` (`'Team A'`/`'Team B'`) doesn't match the real data-driven team strings — dead/misleading enum members. Type `team` as `string | typeof ALL_TEAMS` or generate from a single source of truth.

**Type safety — URL-seeded tab lookup (flagged by architecture + standards, cross-agent consensus)**
- **`MpdSupervisorReportContext.tsx:1264-1269` — sev 2-4**: `StaffDetailTabEnum[query.tab.toString()] ?? ...` indexes a string enum with arbitrary URL input; the index signature hides that the access is `string | undefined`, and `query.tab` may be `string[]` (joined silently by `.toString()`). Replace with an explicit allow-list guard: take `Array.isArray(query.tab) ? query.tab[0] : query.tab`, then `Object.values(StaffDetailTabEnum).includes(...)` before casting.
- **`MpdSupervisorReportContext.tsx:1264-1269` — sev 2** (architecture): Tab state seeds from `?tab=` once but never syncs back to the URL (read-only deep-link). Either fully wire it (push tab to `router.query` on change) or drop the URL seed and keep tab state purely local.

**UX / MUI**
- **`MpdSupervisorReport.tsx:890-907` — sev 3** (ux): Search field is placeholder-only (has `aria-label`, so a11y/tests pass) with no visible `label`; sighted users lose the cue after typing. Add `label={t('Search name')}` or accept placeholder-only deliberately.
- **`StaffMember.tsx:2107, 2124-2143, 2168-2182, 2225-2228` (also `StaffMemberDrawer.tsx`, `MpdSupervisorReport.tsx`) — sev 3** (ux): Several `styled()`/`sx` usages close over the module-level `import theme` instead of the callback theme, bypassing `ThemeProvider` context. Use `styled(X)(({ theme }) => ...)` and `sx={(theme) => ...)}`.

**Loading / error states (deferred until data exists — flagged by ux + architecture + testing)**
- **`useMockInfiniteStaff.ts:2730-2764` (ux) / `index.page.tsx:78-103` (architecture) / `MpdSupervisorReport.test.tsx:555-732` (testing) — sev 2-4**: The mock hook exposes no `error` field and `loading` is permanently `false`; there is no error branch or error boundary, and no loading/error test. N/A today (no query), but ensure the backend-wiring follow-up adds an `error` to the hook result, an `<Alert severity="error">` path distinct from the empty state, an error boundary around the view, and loading/error tests.

**Testing — minor**
- **`MpdSupervisorReport.test.tsx:698, 711` — sev 4** (testing): Two `userEvent.click` calls are not awaited (every other click in the file is), risking act warnings / flaky races. Await both.
- **`StaffDetailsTabs/.../MonthlySummary.tsx` (+ Payroll, MPGA, StaffExpenseReport) — sev 2** (standards): New placeholder tab components lack colocated `*.test.tsx`. Add a minimal render test each (or one shared test), expanding as they gain content.
- **`mockData.ts:1-263` — sev 2** (standards): Mock generator has non-trivial logic (modulo spouse assignment, health cycling, pseudo-random payroll) with no colocated test. Add a smoke test for per-record invariants, or relocate to a test-only fixture.

## What's Solid

- **Financial discipline:** money is never float-summed, no client-side `.reduce` over paginated nodes, no `.toFixed` in calc logic, no currency mixing across rows; payroll is rendered one-at-a-time through `currencyFormat()` at the display boundary — the correct pattern.
- **Mock pagination hook (`useMockInfiniteStaff`):** cleanly mimics a Relay connection (`nodes`/`pageInfo.endCursor`/`hasNextPage`), guards `fetchMore` past the end, and resets to page 1 when filters change — already modeling the correct "re-fetch from first page on filter change" invariant.
- **Shared changes are backward-compatible:** `MultiPageHeader.tsx` is purely additive (`const` → `export const` on 3 styled components, zero behavior change across 29 dependents); `useHrToolsNavItems` adds one nav item with a correct memo dependency array.
- **Standards/a11y:** named exports throughout (only `export default` is the required Pages Router page), correct file naming and colocation, no `any`/`@ts-ignore`/non-null `!`/`console.*`/`new Date()` in source, `TODO(MPDX)` ticket-prefixed TODOs; icon buttons have `aria-label`s, the card uses `role="button"` + descriptive label, and quarter chips encode health in the `aria-label` so color is not the sole signal.

src/components/HrTools/MpdSupervisorReport/Filters/mpdSupervisorReportFilters.ts
```diff
@@ -0,0 +1,47 @@
+import i18n from 'src/lib/i18n';
+
+export enum MpdSupervisorReportQuickFilterEnum {
+  AllPeople = 'allPeople',
+  NegativeLastMonth = 'negativeLastMonth',
+  ThreeMonthsNegative = 'threeMonthsNegative',
+  TrendingDown = 'trendingDown',
```
kegrimes: I remember you saying that we weren't going to include "Trending down" filter. Is that still true?

src/components/HrTools/MpdSupervisorReport/Filters/MpdSupervisorReportFilterPanel.tsx
```diff
@@ -0,0 +1,138 @@
...
+          alignItems: 'center',
+          justifyContent: 'space-between',
+          px: 2,
+          py: 1.5,
+          borderBottom: '1px solid',
+          borderBottomColor: 'divider',
+        }}
+      >
+        <Typography id="left-panel-header" variant="h6">
+          {t('Filters')}
```
kegrimes, line 77: optional: Could we leave a "TODO" comment so we know to come back to add the functionality for the filter chips?

src/components/HrTools/MpdSupervisorReport/StaffDetailsTabs/MPGA/MPGA.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import { Typography } from '@mui/material';
+import { useTranslation } from 'react-i18next';
+
+export const StaffTabMPGA: React.FC = () => {
+  const { t } = useTranslation();
+
+  return <Typography>{t('MPGA')}</Typography>;
+};
```
kegrimes, line 9: Question: Is this going to be just the MPGA bar chart and then we have a link taking the supervisor to their MPGA report but in review mode only?

src/components/HrTools/MpdSupervisorReport/StaffDetailsTabs/Quarterly/Quarterly.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+    personNumber: '10000001',
+    staffAccountID: '1000000001',
+    userPersonType: 'Full time',
+    team: 'Campus',
+  },
+  quarters: [
+    { label: 'FQ4 25', health: QuarterHealthEnum.Green, payroll: 15000 },
+    { label: 'FQ1 26', health: QuarterHealthEnum.Yellow, payroll: 16000 },
+    { label: 'FQ2 26', health: QuarterHealthEnum.Red, payroll: 17000 },
+    { label: 'FQ3 26', health: QuarterHealthEnum.Green, payroll: 18000 },
```
kegrimes, line 27: Good idea on making the enum for MPD health! I think we should add that to the backend when we get to the GraphQL part.

src/components/HrTools/MpdSupervisorReport/StaffMemberDrawer/StaffMemberDrawer.tsx
```diff
@@ -0,0 +1,193 @@
...
+            variant="subtitle2"
+            fontWeight="bold"
+            sx={{ width: '100%' }}
+          >
+            {t('Spouse')}: {`${spouse.preferredName} ${spouse.lastName}`}
+          </Typography>
+          <DetailRow label={t('Person Number')} value={spouse.personNumber} />
+          <DetailRow
+            label={t('Staff Account Number')}
+            value={spouse.staffAccountID}
```
kegrimes, line 139: Question: Do we care about a spouse's employment type or team?

src/components/HrTools/MpdSupervisorReport/StaffMemberRow/StaffMember.tsx
```diff
@@ -0,0 +1,236 @@
...
+};
+
+const healthColor = (
+  health: QuarterHealthEnum,
+): { bg: string; color: string } => {
+  switch (health) {
+    case QuarterHealthEnum.Green:
+      return {
+        bg: theme.palette.chipGreenLight.main,
+        color: theme.palette.chipGreenDark.main,
```
kegrimes: This logic is duplicated in `Quarterly.tsx`! Maybe we should extract a helper function?

src/components/HrTools/MpdSupervisorReport/StaffMemberRow/StaffMember.tsx
```diff
@@ -0,0 +1,236 @@
...
+    minWidth: '80px',
+    '& .MuiChip-label': {
+      paddingInline: theme.spacing(1),
+    },
+  };
+});
+
+const initials = (name: string): string =>
+  name
+    .split(' ')
```
kegrimes: Optional: `StaffMemberDrawer` also needs to sort out the initials. A helper could be nice here too

src/components/HrTools/MpdSupervisorReport/Filters/mpdSupervisorReportFilters.ts
```diff
@@ -0,0 +1,47 @@
...
+  {
+    id: MpdSupervisorReportQuickFilterEnum.TrendingDown,
+    label: i18n.t('Trending down'),
+  },
+];
+
+export enum MpdSupervisorReportTeamsEnum {
+  All = 'all',
+  TeamA = 'Team A',
+  TeamB = 'Team B',
```
kegrimes, line 35: Is this enum needed at all? Do we have a list of possible teams a user could be on?

pages/accountLists/[accountListId]/hrTools/mpdSupervisorReport/index.page.tsx
```diff
@@ -0,0 +1,97 @@
...
+        panelOpen === Panel.Navigation ? (
+          <MultiPageMenu
+            isOpen
+            selectedId="mpdSupervisorReport"
+            onClose={() => setPanelOpen(null)}
+            navType={NavTypeEnum.HrTools}
+          />
+        ) : panelOpen === Panel.Filters ? (
+          <DynamicMpdSupervisorReportFilterPanel
+            onClose={() => setPanelOpen(null)}
```
Comment (kegrimes) [inline pages/accountLists/[accountListId]/hrTools/mpdSupervisorReport/index.page.tsx]: There is a hydration error. I asked Claude:

> What went wrong: Your filter panel was set to be open the moment the page loads. But it was also a "load-it-lazily" component (dynamic), meaning React is told "don't load this right away — show a spinner first."
>
> 
> So the two versions disagreed:
> 
> Server rendered the real filter panel (the dropdowns and chips).
> Browser rendered the spinner first, because it hadn't loaded the real panel yet.
> Real panel ≠ spinner → mismatch → error, on every refresh.
> 
> Why it happened only here: The other lazy components in the app only appear after you click something. By then hydration is long done, so there's nothing to mismatch. Yours was the only one showing immediately on load, so it got caught in the middle of hydration.
> 
> The fix: We told the page to just load the filter panel normally instead of lazily. Now both the server and the browser build the same thing — the real panel — so they match, no spinner flash, no error. And since the panel is always visible anyway, lazy-loading it never helped; we were paying for the problem and getting nothing back.


```suggestion
          <MpdSupervisorReportFilterPanel
            onClose={() => setPanelOpen(null)}
```

src/components/HrTools/MpdSupervisorReport/StaffMemberRow/StaffMember.tsx
```diff
@@ -0,0 +1,236 @@
...
+      };
+  }
+};
+
+const StyledCard = styled(Card)(({ theme }) => ({
+  marginBottom: theme.spacing(1),
+  boxShadow: theme.shadows[1],
+  cursor: 'pointer',
+  width: '100%',
+}));
```
kegrimes, line 39: Optional: I like the idea of each row having a border so it's easily identified but that might be a personal preference so feel free to ignore!

kegrimes (approved):
This looks great Daniel! Thanks for getting us started.

I think once you fix the hydration error this PR is good to go!

src/components/HrTools/MpdSupervisorReport/StaffDetailsTabs/MPGA/MPGA.tsx
dr-bizz: I was thinking, we could try to fit the page here on the right side. But if it doesn't work out, we can make the tab open a new tab to the MPGA or Staff Expense Report.

I didn't put too much thought into it, so I'm good with sticking with each report having its own page.

src/components/HrTools/MpdSupervisorReport/StaffDetailsTabs/Quarterly/Quarterly.test.tsx
dr-bizz: Yes, all the filtering will be backend

src/components/HrTools/MpdSupervisorReport/StaffMemberDrawer/StaffMemberDrawer.tsx
dr-bizz: I don't think it matters. We can always add it in later.

src/components/HrTools/MpdSupervisorReport/Filters/mpdSupervisorReportFilters.ts
dr-bizz: We will do, once we get the HCM data synced

dr-bizz: We will need it for the filters

kegrimes (approved):
This looks awesome! The UI looks great and everything seems to be working, I'm going to approve!

PR #1863, opened 2026-06-25
[No-Jira] Make fund type cards responsive

kegrimes:
## Description

Make fund type cards in Staff expenses report responsive to differing screen size.

## Testing

- Shrink the size of the screen
- Check that the cards move down a row

## Checklist:

### DIFF
--- a/src/components/Reports/StaffExpenseReport/BalanceCardList/BalanceCardList.tsx
+++ b/src/components/Reports/StaffExpenseReport/BalanceCardList/BalanceCardList.tsx
@@ -13,10 +13,10 @@ const StyledCardsBox = styled(Box, {
   shouldForwardProp: (prop) => prop !== 'isSelected',
 })<{ isSelected: boolean }>(({ theme, isSelected }) => ({
   flex: isSelected ? 2 : 1,
-  minWidth: 0,
+  minWidth: isSelected ? 250 : 200,
   display: 'flex',
   gap: theme.spacing(4),
-  transition: 'flex 0.3s ease-in-out',
+  transition: 'flex 0.3s ease-in-out, min-width 0.3s ease-in-out',
 }));
 
 export interface BalanceCardListProps {


src/components/Reports/StaffExpenseReport/BalanceCardList/BalanceCardList.tsx
```diff
@@ -13,7 +13,7 @@ const StyledCardsBox = styled(Box, {
   shouldForwardProp: (prop) => prop !== 'isSelected',
 })<{ isSelected: boolean }>(({ theme, isSelected }) => ({
   flex: isSelected ? 2 : 1,
-  minWidth: 0,
+  minWidth: 200,
```
kegrimes: [Medium] This card floor (minWidth: 200) diverges from the skeleton floor minWidth: 250 in BalanceCard/BalanceCardSkeleton.tsx:38, so the wrap threshold and card width shift slightly between the loading and loaded states. Consider using one value for both.

kegrimes: [Suggestion] Could we extract a shared BALANCE_CARD_MIN_WIDTH constant imported by both this component and BalanceCardSkeleton so the two floors can't drift apart again?

kegrimes: [Suggestion] Optional: a toHaveStyle({ minWidth: '200px' }) guard test would catch an accidental revert to 0. Note jsdom can't verify actual wrapping, so it only guards the value.

kegrimes (commented):
## Multi-Agent Review — Verdict: APPROVED WITH SUGGESTIONS

**Risk:** LOW (1/10) · **Blockers:** 0 · 5 agents (Architecture, Testing, Standards, UX, Financial Reporting).

The one-line change (`minWidth: 0` -> `minWidth: 200`) is the correct, minimal way to let the parent's existing `flexWrap` wrap the balance cards when squished (previously `minWidth:0` let them shrink forever and never wrap). It matches the repo's pervasive hardcoded-`minWidth` convention (40+ instances; no theme token exists for card min-width). No security/data/financial impact — Financial Reporting agent confirmed no calculation code is touched. 1 dependent (`StaffExpenseReport.tsx`), no breaking changes.

Non-blocking suggestions are posted as inline comments.

---

**[Pre-existing] BalanceCardList.tsx:41-45** — The loading branch renders all three `<BalanceCardSkeleton>` inside a single `StyledCardsBox` (no `flexWrap`), so the responsive wrapping this PR adds applies to the loaded state but not the loading state. Predates this PR; consider a follow-up. (Informational — does not affect the verdict.)

src/components/Reports/StaffExpenseReport/BalanceCardList/BalanceCardList.tsx
kegrimes: /dismiss the skeleton is just to show when loading. A user might not have all 5 fund types so it's okay for now

kegrimes: /dismiss

PR #1866, opened 2026-06-26
MPDX-9779 Surface Income & Expenses CSV export as a labeled header menu

wjames111:
## Description

- Replaced the easy-to-miss, icon-only CSV download buttons (one buried in each table's data-grid toolbar) with a single, clearly labeled **Export CSV** button in the report header, next to **Print**.
- The button opens a menu to choose **Income** or **Expenses**; each downloads the same CSV file as before (export logic unchanged).
- Styled to match the existing `StyledPrintButton` so it stays on the MPDX theme; added `aria-haspopup`/`aria-controls`/`aria-expanded` and localized strings (`Export CSV`, `Income`, `Expenses`).
- Removed the per-table download button from `CustomToolbar` and the now-unused `data`/`type` threading through `TableCard`.

Related ticket: [MPDX-9779](https://jira.cru.org/browse/MPDX-9779)

## Testing

- Go to **Reports → MPGA Income & Expenses** (`/accountLists/[accountListId]/reports/mpgaIncomeExpenses`)
- Check the report header shows an **Export CSV** button immediately left of **Print**
- Click **Export CSV** and confirm a menu appears with **Income** and **Expenses**
- Click **Income** → `MPGA Income Monthly Report.csv` downloads; click **Expenses** → `MPGA Expenses Monthly Report.csv` downloads
- Confirm the Income and Expenses table toolbars no longer show the standalone download icon

## Checklist:

### DIFF
--- a/src/components/Reports/MPGAIncomeExpensesReport/ExportCsvButton/ExportCsvButton.test.tsx
+++ b/src/components/Reports/MPGAIncomeExpensesReport/ExportCsvButton/ExportCsvButton.test.tsx
@@ -5,16 +5,16 @@ import userEvent from '@testing-library/user-event';
 import theme from 'src/theme';
 import { exportToCsv } from '../CustomExport/CustomExport';
 import { ReportTypeEnum } from '../Helper/MPGAReportEnum';
-import { mockData, months } from '../mockData';
+import { AllData, mockData, months } from '../mockData';
 import { ExportCsvButton } from './ExportCsvButton';
 
 jest.mock('../CustomExport/CustomExport', () => ({
   exportToCsv: jest.fn(),
 }));
 
-const TestComponent: React.FC = () => (
+const TestComponent: React.FC<{ data?: AllData }> = ({ data = mockData }) => (
   <ThemeProvider theme={theme}>
-    <ExportCsvButton data={mockData} months={months} />
+    <ExportCsvButton data={data} months={months} />
   </ThemeProvider>
 );
 
@@ -52,7 +52,7 @@ describe('ExportCsvButton', () => {
       mockData.income,
       ReportTypeEnum.Income,
       months,
-      expect.any(String),
+      'en-US',
     );
   });
 
@@ -66,8 +66,42 @@ describe('ExportCsvButton', () => {
       mockData.expenses,
       ReportTypeEnum.Expenses,
       months,
-      expect.any(String),
+      'en-US',
     );
+  });
+
+  it('disables an export option when its dataset is empty', async () => {
+    const { getByRole, findByRole } = render(
+      <TestComponent data={{ income: mockData.income, expenses: [] }} />,
+    );
+
+    userEvent.click(getByRole('button', { name: 'Export CSV' }));
+
+    expect(
+      await findByRole('menuitem', { name: 'Income' }),
+    ).not.toHaveAttribute('aria-disabled');
+
+    const expenses = await findByRole('menuitem', { name: 'Expenses' });
+    expect(expenses).toHaveAttribute('aria-disabled', 'true');
+    expect(exportToCsv).not.toHaveBeenCalled();
+  });
+
+  it('disables both export options when all datasets are empty', async () => {
+    const { getByRole, findByRole } = render(
+      <TestComponent data={{ income: [], expenses: [] }} />,
+    );
+
+    userEvent.click(getByRole('button', { name: 'Export CSV' }));
+
+    expect(await findByRole('menuitem', { name: 'Income' })).toHaveAttribute(
+      'aria-disabled',
+      'true',
+    );
+    expect(await findByRole('menuitem', { name: 'Expenses' })).toHaveAttribute(
+      'aria-disabled',
+      'true',
+    );
+    expect(exportToCsv).not.toHaveBeenCalled();
   });
 
   it('closes the menu after an export is selected', async () => {
--- a/src/components/Reports/MPGAIncomeExpensesReport/ExportCsvButton/ExportCsvButton.test.tsx
+++ b/src/components/Reports/MPGAIncomeExpensesReport/ExportCsvButton/ExportCsvButton.test.tsx
@@ -1,0 +1,85 @@
+import React from 'react';
+import { ThemeProvider } from '@mui/material/styles';
+import { render, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import theme from 'src/theme';
+import { exportToCsv } from '../CustomExport/CustomExport';
+import { ReportTypeEnum } from '../Helper/MPGAReportEnum';
+import { mockData, months } from '../mockData';
+import { ExportCsvButton } from './ExportCsvButton';
+
+jest.mock('../CustomExport/CustomExport', () => ({
+  exportToCsv: jest.fn(),
+}));
+
+const TestComponent: React.FC = () => (
+  <ThemeProvider theme={theme}>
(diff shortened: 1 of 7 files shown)


src/components/Reports/MPGAIncomeExpensesReport/ExportCsvButton/ExportCsvButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+        {t('Export CSV')}
+      </StyledPrintButton>
+      <Menu
+        id="export-csv-menu"
+        anchorEl={anchorEl}
+        open={open}
+        onClose={handleClose}
+        MenuListProps={{ 'aria-labelledby': 'export-csv-button' }}
+      >
+        <MenuItem onClick={() => handleExport(ReportTypeEnum.Income)}>
```
wjames111: [Medium] **Empty-state export is now reachable.** The old download icon lived inside `TableCard`, which only renders when `data.length` is truthy (`TableCard.tsx:137`), so an empty Income/Expenses export was previously unreachable. This header menu always renders both items, so selecting one with no rows (or before the query resolves) calls `exportToCsv([], …)` and downloads a header + zero "Overall Total" CSV that looks like real data. Per the repo's zero-state rule, disable the item when its dataset is empty:

```tsx
<MenuItem disabled={!data.income.length} onClick={() => handleExport(ReportTypeEnum.Income)}>{t('Income')}</MenuItem>
<MenuItem disabled={!data.expenses.length} onClick={() => handleExport(ReportTypeEnum.Expenses)}>{t('Expenses')}</MenuItem>
```

If exporting empty tables is intended product behavior, make it explicit. Non-blocking.

src/components/Reports/MPGAIncomeExpensesReport/ExportCsvButton/ExportCsvButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+  return (
+    <>
+      <StyledPrintButton
+        id="export-csv-button"
+        aria-controls={open ? 'export-csv-menu' : undefined}
+        aria-haspopup="true"
+        aria-expanded={open ? 'true' : undefined}
+        startIcon={
+          <SvgIcon fontSize="small">
+            <FileDownloadIcon titleAccess={t('Export CSV')} />
```
wjames111: [Suggestion] The button already has the visible text label "Export CSV" (line 55), so this `titleAccess` injects a redundant SVG `<title>` into the accessible-name computation. For a labeled button the start icon should be decorative — drop `titleAccess` so the icon is `aria-hidden`. Note the sibling Print button does the same thing, so fix both for consistency or leave both. Non-blocking.

src/components/Reports/MPGAIncomeExpensesReport/ExportCsvButton/ExportCsvButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const handleOpen = (event: React.MouseEvent<HTMLButtonElement>) => {
+    setAnchorEl(event.currentTarget);
+  };
+
+  const handleClose = () => {
+    setAnchorEl(null);
+  };
+
+  const handleExport = (type: ReportTypeEnum) => {
+    const rows = type === ReportTypeEnum.Income ? data.income : data.expenses;
```
wjames111: [Suggestion] `type === ReportTypeEnum.Income ? data.income : data.expenses` silently routes any future `ReportTypeEnum` member to `data.expenses`. Not a defect today (only Income/Expenses items exist), but a `switch` with an exhaustive/never check would be more robust if the enum grows. Non-blocking.

src/components/Reports/MPGAIncomeExpensesReport/ExportCsvButton/ExportCsvButton.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const { getByRole, findByRole } = render(<TestComponent />);
+
+    userEvent.click(getByRole('button', { name: 'Export CSV' }));
+    userEvent.click(await findByRole('menuitem', { name: 'Income' }));
+
+    expect(exportToCsv).toHaveBeenCalledWith(
+      mockData.income,
+      ReportTypeEnum.Income,
+      months,
+      expect.any(String),
```
wjames111: [Suggestion] Assert the locale concretely as `'en-US'` instead of `expect.any(String)`. The `UserPreferenceContext` default is `{ locale: 'en-US' }`, so this is deterministic and every sibling test (`CustomExport.test.tsx`, etc.) pins the concrete value. `expect.any(String)` would pass even if the wrong (but stringy) locale were forwarded. Also consider adding an empty-data test (`data={{ income: [], expenses: [] }}`) to pair with the empty-state finding. Non-blocking.

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesReport.tsx
```diff
@@ -139,7 +140,8 @@ export const MPGAIncomeExpensesReport: React.FC<
                   {t('Income & Expenses Analysis: Last 12 Months')}
                 </Typography>
               </SimplePrintOnly>
-              <SimpleScreenOnly>
+              <SimpleScreenOnly display="flex" alignItems="center">
+                <ExportCsvButton data={allData} months={last12Months} />
```
wjames111, line 148: [Suggestion] Spacing between Export CSV and the Print button currently relies on `StyledPrintButton`'s hardcoded `marginLeft: theme.spacing(2)` rather than the flex container. An explicit `gap` on this `SimpleScreenOnly` flex row would be more intentional (the first child also carries a stray leading margin). Also worth verifying the header doesn't overflow at `xs`/`sm` now that there are two labeled buttons next to the title — consider `flexWrap`. Non-blocking.

wjames111 (commented):
## 🤖 Multi-Agent Code Review — APPROVED WITH SUGGESTIONS

**5 specialized agents** (Architecture, Testing, Standards, UX, Financial Reporting) + dependency analysis · standard mode.

**Verdict:** ✅ No blockers, no breaking changes. Clean, net-debt-reducing refactor that moves CSV export from the grid toolbar into a labeled header dropdown. One Medium suggestion worth addressing before merge; the rest are minor polish.

**Risk:** 5/10 (MEDIUM) — UI refactor confined to one report feature; no auth/API/GraphQL/data-layer surface.

### Findings
- 🟡 **Medium (1):** Export menu items are always enabled — an empty table now exports a zeros-only CSV that wasn't previously reachable (the old toolbar download only rendered when `data.length` was truthy). Consider disabling each menu item when its dataset is empty.
- 🔵 **Suggestions (~5):** pin the test locale to `'en-US'`, add an empty-data test, drop the redundant icon `titleAccess`, use an explicit flex `gap` for header spacing, exhaustive enum handling.

### Verified clean
- **Dependency impact:** no breaking changes — `CustomToolbar`/`createToolbar`/`exportToCsv` callers all migrated cleanly; no orphaned props or unused imports (`TableCard` still uses `DataFields`/`ReportTypeEnum`).
- **Financial:** pure plumbing — income/expenses row selection and `months` argument are provably identical to the old call path; no calculation logic changed.
- **Standards:** named export, full i18n coverage (keys already in `en/translation.json`), no `any`/`new Date()`/dynamic `t()` keys. The `AllData`-from-`mockData` import is the established feature-wide convention, not introduced here.

Inline comments below. Nothing here blocks merge.

PR #1870, opened 2026-06-30
[MPDX-9779] - Make download button more visible on staff expense report

wjames111:
## Description

Jira: **MPDX-9779**

Standardizes the CSV download experience across the financial reports and fixes the year timeline on the MPGA Income & Expenses report.

- **Staff Expense Report** – replaced the row of `Income Report` / `Expense Report` / `Combined Report` buttons with a single **Export CSV** dropdown in the page header (next to **Print**), matching the Ministry Partner Giving Analysis report. Each menu option disables when its data is empty.
- **Shared `CsvExportMenu`** – extracted the dropdown shell (button + menu + accessibility wiring) into `src/components/Reports/Shared/CsvExportMenu`, now used by both the MPGA and Staff Expense `ExportCsvButton`s. Removed the old `DownloadButtonGroup`.
- **MPGA timeline alignment fix** – the year timeline was a separate table floating in the grid toolbar with hard-coded widths that drifted out of alignment with the month columns. Replaced it with the DataGrid `columnGroupingModel`, so each year header spans exactly its own months and the old `TableCardHead` component was removed.
- **MPGA header styling** – each year/Summary group header is split vertically into a label on top and a full-width colored timeline bar on the bottom; the **Columns**/**Filters** buttons were moved into the header cell above the Description column; and the header borders were tidied (top border on the Description/month/Average/Total cells, no divider under the year cells, default separator kept beside the controls cell).

## Testing

**Staff Expense Report**

- Go to Reports → Staff Expenses (Income and Expenses)
- Confirm a single **Export CSV** dropdown appears in the top header next to **Print**
- Open it and confirm the **Income Report**, **Expense Report**, and **Combined Report** options each download a CSV, and that an option is disabled when there is no matching data

**MPGA Income & Expenses report**

- Go to Reports → Ministry Partner Giving Analysis
- Confirm the year timeline above the month columns lines up: the first year spans exactly its months (e.g. Jul–Dec) and the second spans its months (e.g. Jan–Jun), each with a colored bar beneath the year label
- Confirm the **Columns** and **Filters** buttons sit in the header cell to the left of the year cells (above Description) and still open their panels
- Confirm the **Export CSV** and **Print** buttons still work

## Checklist:


src/components/Reports/StaffExpenseReport/StaffExpenseReport.tsx
```diff
@@ -189,6 +189,13 @@ export const StaffExpenseReport: React.FC<StaffExpenseReportProps> = ({
     return newTransactions;
   }, [allFunds, time, t, filters]);
 
+  const selectedFundTransactions = selectedFundType
```
wjames111: [Medium] `selectedFundTransactions` is rebuilt (and spreads two arrays) on every render, whereas this component memoizes its other derived arrays (`transactions`, `transferTotals`, `overallBalance`, …). It then becomes a prop to `ExportCsvButton`, whose own `.filter()` and the `CsvExportMenu` `items` array also rebuild each render.

This is behaviorally identical to the inline expression it replaced (previously passed inline to `DownloadButtonGroup`), so it's **not a regression** — just a consistency/perf nit. Consider:
```tsx
const selectedFundTransactions = useMemo(
  () =>
    selectedFundType
      ? [
          ...(transactions[selectedFundType]?.income ?? []),
          ...(transactions[selectedFundType]?.expenses ?? []),
        ]
      : [],
  [selectedFundType, transactions],
);
```

src/components/Reports/MPGAIncomeExpensesReport/Tables/TableCard.tsx
```diff
@@ -126,6 +131,93 @@ export const TableCard: React.FC<TableCardProps> = ({
     ];
   }, [months]);
 
+  const columnGroupingModel = useMemo<GridColumnGroupingModel>(() => {
```
wjames111, line 176: [Suggestion] The year-group and Summary-group `renderHeaderGroup` blocks are ~50 lines of near-identical JSX (a column-flex `Box` → bold `Typography` label → 2px colored underline `Box`); only `color` and label text differ. Extracting a small local component keeps this memo focused on structure and removes the drift risk if the header style changes:
```tsx
const GroupHeader = ({ label, color }: { label: string; color: string }) => (
  <Box sx={{ display: 'flex', flexDirection: 'column', width: '100%', height: '100%' }}>
    <Box component="span" sx={{ alignSelf: 'flex-start' }}>
      <Typography sx={{ color, fontSize: '14px', fontWeight: 'bold' }}>{label}</Typography>
    </Box>
    <Box component="span" sx={{ width: '100%', mt: '7px' }}>
      <Box sx={{ width: '100%', height: '2px', backgroundColor: color }} />
    </Box>
  </Box>
);
```
Reuse it for both the year groups and the summary group.

src/components/Reports/MPGAIncomeExpensesReport/Tables/TableCard.tsx
```diff
@@ -126,6 +131,93 @@ export const TableCard: React.FC<TableCardProps> = ({
...
+          <Box
+            sx={{
+              display: 'flex',
+              flexDirection: 'column',
+              width: '100%',
+              height: '100%',
+            }}
+          >
+            <Box component="span" sx={{ alignSelf: 'flex-start' }}>
+              <Typography sx={{ color, fontSize: '14px', fontWeight: 'bold' }}>
```
wjames111: [Suggestion] Raw px literals (`fontSize: '14px'`, `mt: '7px'`, `height: '2px'`) for the grouped header. These are copied verbatim from the now-deleted `TableCardHead` and match an existing convention across this report (8+ instances), and the colors correctly use `theme.palette.*` tokens — so this is a low-priority, pre-existing nit, not a defect. Noted for awareness only.

src/components/Reports/Shared/CsvExportMenu/CsvExportMenu.tsx
```diff
@@ -0,0 +1,87 @@
...
+      <Menu
+        id={menuId}
+        anchorEl={anchorEl}
+        open={open}
+        onClose={handleClose}
+        MenuListProps={{ 'aria-labelledby': buttonId }}
+      >
+        {items.map((item) => (
+          <MenuItem
+            key={item.label}
```
wjames111: [Suggestion] `key={item.label}` assumes labels are unique. Both current callers (MPGA: Income/Expenses; Staff: Income/Expense/Combined Report) use distinct labels, so this is safe today — but a future caller with duplicate labels would hit React key collisions. Consider adding an optional `id`/`value` field to `CsvExportMenuItem` and keying on that, or documenting the uniqueness requirement.

src/components/Reports/MPGAIncomeExpensesReport/CustomToolbar/CustomToolbar.tsx
```diff
@@ -9,17 +9,14 @@ import {
...
-}
-
-export const CustomToolbar: React.FC<CustomToolbarProps> = ({ months }) => {
+export const CustomToolbar: React.FC = () => {
   const { t } = useTranslation();
 
   return (
-    <Toolbar style={{ display: 'flex', justifyContent: 'left' }}>
+    <Toolbar
+      style={{ display: 'flex', justifyContent: 'left', borderBottom: 'none' }}
```
wjames111, line 18: [Suggestion] Per the repo MUI v5 convention, prefer `sx` over inline `style`. This PR only appended `borderBottom: 'none'` to a `style` prop that already existed on `main`, and the x-data-grid `Toolbar` slot forwards `style` reliably — so this is a pre-existing, low-priority consistency nit rather than something introduced here:
```tsx
<Toolbar sx={{ display: 'flex', justifyContent: 'left', borderBottom: 'none' }}>
```

src/components/Reports/StaffExpenseReport/ExportCsvButton/ExportCsvButton.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+    id: '2',
+    fundType: 'Expense',
+    amount: -50,
+    category: StaffExpenseCategoryEnum.Assessment,
+    transactedAt: '2025-01-01',
+    displayCategory: 'Assessment',
+  },
+];
+
+const incomeOnlyTransactions: Transaction[] = [mockTransactions[0]];
```
wjames111, line 45: [Suggestion] The deleted `DownloadButtonGroup.test` covered an `amount === 0` transaction (excluded by both `amount < 0` and `amount > 0`, so it counts as neither income nor expense). The new component preserves the identical filters but no test exercises that boundary now. Cheap to restore:
```tsx
it('treats a zero-amount transaction as neither income nor expense', async () => {
  const zero: Transaction = { ...mockTransactions[0], id: '0', amount: 0 };
  const { getByRole, findByRole } = render(
    <TestComponent transactions={[zero, ...incomeOnlyTransactions]} />,
  );
  userEvent.click(getByRole('button', { name: 'Export CSV' }));
  expect(await findByRole('menuitem', { name: 'Income Report' })).not.toHaveAttribute('aria-disabled', 'true');
  expect(await findByRole('menuitem', { name: 'Expense Report' })).toHaveAttribute('aria-disabled', 'true');
  expect(await findByRole('menuitem', { name: 'Combined Report' })).toHaveAttribute('aria-disabled', 'true');
});
```

wjames111 (commented):
## 🤖 Multi-Agent Code Review — Verdict: ✅ APPROVED WITH SUGGESTIONS

**Mode:** standard · **Risk:** 7/10 (HIGH — volume + breadth) · **Agents:** Architecture, Testing, Standards, UX, Financial Reporting + Dependency Impact · **Blockers:** 0

A clean, well-tested consolidation refactor: two divergent CSV-export UIs collapse into one shared `CsvExportMenu`, and the MPGA year/Summary headers move from a separate `TableCardHead` table into the DataGrid `columnGroupingModel`.

**Independently verified by the agents:**
- 💰 **Financial:** income/expense classification (`amount < 0` / `amount > 0`), per-item disabled-state gating, and export payloads (`createCsvReport` / `exportToCsv`) are **byte-for-byte equivalent** to the deleted components. No calculation logic touched.
- 🔗 **Dependency:** **zero dangling references** — `TableCardHead` / `DownloadButtonGroup` deletions, `downloadReport` rename, and `CustomToolbar`'s removed `months` prop are all clean. No file with 10+ dependents touched.
- 🧪 **Testing:** the deleted `TableCardHead.test` / `DownloadButtonGroup.test` did **not** lose coverage — color/count logic is unit-tested in `useMonthHeaders.test.ts`, year headers in `TableCard.test.tsx`, and the disable matrix is re-asserted (more strongly) in the new `ExportCsvButton.test.tsx`. 23/23 affected tests pass; `lint:ts` clean.
- 📐 **Standards:** all checklist items PASS/N-A — named exports, colocated tests, i18n via callers, no `any`/`new Date()`/debug output.

### Findings (all severity < 7 — non-blocking)

| Sev | File | Note |
|----|------|------|
| 5.5 Medium | `StaffExpenseReport.tsx:192` | `selectedFundTransactions` unmemoized (file's other derived arrays use `useMemo`). Behaviorally identical to what it replaced — not a regression. |
| 4.0 | `TableCard.tsx:134` | Year-group & Summary `renderHeaderGroup` are ~50 lines of near-identical JSX — extract a local `GroupHeader({label,color})`. |
| 4.0 | `ExportCsvButton.test.tsx` | Re-add the `amount === 0` edge case the deleted `DownloadButtonGroup.test` covered. |
| 4.0 | `CustomToolbar.tsx:18` | Inline `style` → `sx` (pre-existing prop; PR only appended `borderBottom`). |
| 3.5 | `CsvExportMenu.tsx:77` | `key={item.label}` assumes label uniqueness (safe for current callers). |
| 3.0 | `TableCard.tsx:158` | Hardcoded px literals (copied verbatim from deleted `TableCardHead`; matches existing report convention). |

Lower-severity hardening also suggested: open-empty-menu test for `CsvExportMenu`, and `CustomToolbar` has no colocated test (now covered indirectly via `TableCard`).

---

Identified during review but on a file this PR does not modify — cannot be posted as a line comment:

**[Suggestion] `src/hooks/useMonthHeaders.ts:29-34`** — `getBorderColor` lists `[colors, months]` as `useCallback` deps but never reads `months`, and `colors` is a fresh object literal at each call site, so `getBorderColor` gets a new identity every render. This now flows into `TableCard`'s new `columnGroupingModel` memo deps (`[monthCount, getBorderColor, t]`), effectively defeating that memo.
- Severity: 4.0/10 · Flagged by: Architecture
- Recommended Action: memoize the colors object at the call site (or pass the two color strings as primitives) and drop `months` from the `useCallback` deps. Pre-existing hook behavior the PR newly depends on — informational, not a blocker.

src/components/Reports/StaffExpenseReport/StaffExpenseReport.tsx
```diff
@@ -282,10 +293,10 @@ export const StaffExpenseReport: React.FC<StaffExpenseReportProps> = ({
               ) ? (
                 <SimpleScreenOnly
                   display="flex"
-                  flexDirection="column"
-                  alignItems="flex-end"
-                  gap={1}
+                  alignItems="center"
+                  sx={{ gap: 2, '& > button': { ml: 0 } }}
                 >
+                  <ExportCsvButton transactions={selectedFundTransactions} />
```
wjames111, line 300: [Medium] Replacing `DownloadButtonGroup` with this dropdown drops the old component's mobile/responsive handling — it used `useMediaQuery('(max-width:600px)')` to switch to vertical orientation + `width:100%` and scaled down font/padding/icon at `down('sm')`. `StyledPrintButton` (Reports/styledComponents.ts) has no breakpoint styling, so on narrow viewports the control no longer adapts. Downloads also now take 2 interactions (open menu → pick) vs 1. The collapse-to-dropdown itself is defensible (it matches the MPGA report), but please confirm mobile is out of scope or add e.g. `sx={{ width: { xs: '100%', sm: 'auto' } }}`.

src/components/Reports/MPGAIncomeExpensesReport/Tables/TableCard.tsx
```diff
@@ -55,6 +93,11 @@ export const TableCard: React.FC<TableCardProps> = ({
   const locale = useLocale();
   const { incomeTotal, expensesTotal } = useTotals();
 
+  const { monthCount, getBorderColor } = useMonthHeaders(months, {
```
wjames111: [Medium] `useMonthHeaders` is called with an inline object literal `{ first, second }` (fresh reference every render). `getBorderColor` is a `useCallback` keyed on `[colors, months]`, so it's recreated each render — which means the new `columnGroupingModel` `useMemo` (deps `[monthCount, getBorderColor, t]`) never actually caches and rebuilds every render. Perf/intent mismatch, not a correctness bug. Suggested in-diff fix:
```tsx
const monthColors = useMemo(
  () => ({ first: theme.palette.primary.main, second: theme.palette.chartOrange.main }),
  [],
);
const { monthCount, getBorderColor } = useMonthHeaders(months, monthColors);
```
(`theme` is a module singleton, so `[]` deps are sound.)

src/components/Reports/MPGAIncomeExpensesReport/Tables/TableCard.tsx
```diff
@@ -25,6 +31,44 @@ export interface TableCardProps {
   loading?: boolean;
 }
 
+// Visual styling for the grouped-column headers, matching the report's table
+// styling. The font size follows the report-wide 14px convention; the gap and
+// underline thickness are named here rather than inlined as magic numbers.
+const groupHeaderFontSize = '14px';
```
wjames111, line 37: [Medium] The deleted `TableCardHead.test.tsx` asserted each year group's border color (`primary.main` for the first year, `chartOrange.main` for the second, `chartGray.main` for Summary). `TableCard.test.tsx` now asserts only the year/Summary *text*, so a regression swapping group colors or pointing Summary at the wrong palette token would pass. Mitigated by `useMonthHeaders.test.ts` covering `getBorderColor`'s source, but the wiring into `GroupHeader` here is untested. Recommend one `toHaveStyle(color: ...)` assertion per group in `TableCard.test.tsx`.

src/components/Reports/MPGAIncomeExpensesReport/Tables/TableCard.tsx
```diff
@@ -126,6 +169,48 @@ export const TableCard: React.FC<TableCardProps> = ({
     ];
   }, [months]);
 
+  const columnGroupingModel = useMemo<GridColumnGroupingModel>(() => {
+    let monthIndex = 0;
```
wjames111: [Suggestion] `let monthIndex = 0` mutated via `monthIndex++` inside the `.map` callback is a readability smell (mutable counter in a map). It's correct here because `.map` runs synchronously in order and the counter is scoped inside the `useMemo`, but it's fragile if the mapping is ever reordered. Optional: derive the offset, e.g. `monthCount.slice(0, index).reduce((a, g) => a + g.count, 0)`.

src/components/Reports/Shared/CsvExportMenu/CsvExportMenu.tsx
```diff
@@ -0,0 +1,93 @@
...
+      <Menu
+        id={menuId}
+        anchorEl={anchorEl}
+        open={open}
+        onClose={handleClose}
+        MenuListProps={{ 'aria-labelledby': buttonId }}
+      >
+        {items.map((item) => (
+          <MenuItem
+            key={item.value ?? item.label}
```
wjames111, line 83: [Suggestion] The `item.value ?? item.label` key fallback is documented as the duplicate-label safety net, but no test renders two items with the same `label` and distinct `value`. A future regression reverting to `key={label}` wouldn't be caught (React only warns at runtime). Consider a test rendering two same-labeled items with distinct `value`s and asserting both render.

src/components/Reports/MPGAIncomeExpensesReport/CustomToolbar/CustomToolbar.tsx
wjames111, line 18: [Suggestion] Project UX convention prefers `sx` over inline `style` for theme-awareness. Note this is largely pre-existing — only `borderBottom: 'none'` is added this PR, the `style=` attribute predates it — and this is the x-data-grid `Toolbar` slot where `sx` support can be inconsistent. Minor; migrate to `sx` if the slot supports it, otherwise leave as-is.

src/components/Reports/MPGAIncomeExpensesReport/ExportCsvButton/ExportCsvButton.tsx
```diff
@@ -20,72 +17,24 @@ export const ExportCsvButton: React.FC<ExportCsvButtonProps> = ({
...
-      </Menu>
-    </>
+    <CsvExportMenu
+      label={t('Export CSV')}
+      items={[
+        {
+          label: t('Income'),
+          disabled: !data.income.length,
+          onClick: () =>
+            exportToCsv(data.income, ReportTypeEnum.Income, months, locale),
```
wjames111, line 29: [Suggestion] The previous `switch` had a `never`-based exhaustiveness guard on `ReportTypeEnum` (adding a member became a compile error). Converting to this inline `items` array drops that compile-time enforcement — acceptable, since the options are now enumerated explicitly as data, but noting the lost guard for awareness.

wjames111 (commented):
## 🤖 Multi-Agent Code Review — PR #1870

**Verdict: ✅ APPROVED WITH SUGGESTIONS** · Risk: MEDIUM (5/10) · 5 agents (Architecture, Testing, Standards, UX, Financial Reporting)

Clean, net-debt-reducing refactor: ~400 lines of bespoke/duplicated code removed (`TableCardHead` faux-header + `DownloadButtonGroup`) in favor of a tested shared `CsvExportMenu` and DataGrid's native `columnGroupingModel`. The **Financial Reporting agent verified all money logic is preserved byte-for-byte** — income/expense sign split (`amount < 0` / `> 0`), zero-amount handling, combined-report disable logic, and totals rendering are unchanged. No financial-correctness impact.

**No blockers.** 3 medium-priority items + suggestions below; none block merge.

### Medium Priority (5.0–6.9)
- **[6.5] Mobile/responsive download UX dropped** — the deleted `DownloadButtonGroup` had `useMediaQuery` mobile handling (vertical orientation, full-width, scaled font/padding); the replacement `StyledPrintButton` has no breakpoint styling. Also 2 clicks vs 1 to download. Collapse-to-dropdown is defensible (matches MPGA pattern), but confirm mobile is out of scope or add responsive `sx`.
- **[6.5] Group-header colors no longer directly tested** — old `TableCardHead.test.tsx` asserted per-group `data-color`; `TableCard.test.tsx` now asserts only text. Mitigated: `useMonthHeaders.test.ts` unit-tests `getBorderColor` source, but the wiring into `GroupHeader` is untested.
- **[6.0] Unstable `getBorderColor` defeats the `columnGroupingModel` memo** — `useMonthHeaders` is called with an inline object literal, so `getBorderColor` is a fresh ref each render and the new `useMemo` never caches. Perf/intent mismatch, not a correctness bug.

### Suggestions (informational)
- `monthIndex++` mutable counter inside `.map` (TableCard) — readability smell, behavior correct.
- `CsvExportMenu` duplicate-label `value ?? label` key fallback is untested.
- `CustomToolbar` inline `style` should be `sx` (pre-existing; x-data-grid Toolbar `sx` support is inconsistent).
- Named pixel consts `'14px'/'7px'/'2px'` in `TableCard` aren't theme tokens (named + commented, `14px` matches existing convention — acceptable).
- MPGA `ExportCsvButton` dropped its `never`-based exhaustiveness check when the `switch` became a data array (acceptable).
- `CsvExportMenuItem.value` is a public API field no caller currently uses (intentional future-proofing).

### Cross-Cutting Check
Both report export paths (MPGA + Staff) now share `CsvExportMenu` — a consistency improvement. `DownloadButtonGroup` removed with zero remaining references. Disabled-state differs slightly between callers but is justified (Staff already hides the control when no fund has data).

<sub>Standards ✅ · Financial ✅ no logic change · all findings severity < 7, no inter-agent contradictions, debate skipped.</sub>

PR #1873, opened 2026-06-30
[MPDX-9522] Use new staff goal calculations from API

canac:
## Description

Use server-side calculations in the admin view and the presenting your goal step.

Jira ticket: MPDX-9522
~~Depends on https://github.com/CruGlobal/mpdx_api/pull/3422~~ deployed 🚀

## Testing

- Test 1
  - Go to the coaching page of a new staff goal calculation
  - Check that the 403b amounts are calculated from the percentages
  - Change the percentages and save
  - Check that the 403b amounts change
- Test 2
  - Open a new staff goal calculations
  - Go to step 2
  - Check that the table populates with real data

## Checklist:


src/components/HrTools/NsGoalCalculator/PresentingYourGoal/PresentingYourGoalStep.tsx
```diff
@@ -67,90 +56,101 @@ export const PresentingYourGoalStep: React.FC = () => {
...
 
+  const { calculations } = goalCalculation;
+  const married =
+    goalCalculation.maritalStatus ===
+    NewStaffQuestionnaireMaritalStatusEnum.Married;
+  const monthlyNeeds = useMemo(
+    () => ({
+      married,
+      salary: calculations.salary,
+      ministryExpenses: calculations.totalMinistryExpenses,
```
canac: [High] **Presentation "Total Support Goal" diverges from the authoritative `monthlyGoal`.**

`ministryExpenses: calculations.totalMinistryExpenses` maps only lines 2–7, and this `monthlyNeeds` mapping has **no attrition term** — the fragment doesn't even query `attrition`. `useMonthlyNeedsRows` reduces these 6 fields into the row labeled **"Total Support Goal"** (`MonthlyNeedsCard.tsx:25`), which is a printable/shareable donor-facing figure. Since `monthlyGoal` = `subtotalWithAdminCharge × 1.06` (attrition, always applied) and also includes lines 8/10/11/12/13, this total **systematically understates** the goal the admin view shows (`GoalSettingsForm.tsx:123  calculation.calculations.monthlyGoal`).

The established sibling reconciles deliberately:
```ts
// GoalCalculator/SummaryReport/Steps/PresentingYourGoalStep/PresentingYourGoal.tsx:41
ministryExpenses: goalTotals.ministryExpensesTotal + goalTotals.attrition,
```
Tests don't catch it because the mock is rigged so the 6 fields sum to exactly `monthlyGoal` (8774+1492+990+898+1911+1795 = 15860) — values that don't satisfy the real ×1.06 relationship.

**Fix (either):** (a) add `attrition` (+ the missing subtotal lines) to the fragment and fold attrition into a row like the sibling so the reduce reconciles to `monthlyGoal`; or (b) display `calculations.monthlyGoal` directly as the total (rules prefer server totals over client `.reduce`). Then add a test with non-rigged values asserting the presentation total equals `monthlyGoal`. If the subset is intentional, please confirm with product.

src/components/HrTools/NsGoalCalculator/PresentingYourGoal/PresentingYourGoalStep.tsx
```diff
@@ -36,29 +38,16 @@ const PrintableContent = styled('div')(({ theme }) => ({
...
-  married: mockPersonalInfo.married,
-  salary: 8774,
-  ministryExpenses: 898,
-  benefits: 1911,
-  socialSecurityAndTaxes: 1492,
-  voluntaryRetirement: 990,
-  adminCharge: 1795,
-};
+// TODO(MPDX-9801): Special needs are not available yet.
+const specialNeedsPlaceholder = 3624;
```
canac, line 41: [Medium] Hardcoded `specialNeedsPlaceholder = 3624` renders as "Total Special Needs Goal — $3,624" on the donor-facing presentation. Now that every other figure is real API data, this lone fabricated number is indistinguishable from real financial data to staff and donors. `TODO(MPDX-9801)` tracks it (good), but consider **hiding** the Special Needs card/value until the API provides it, rather than shipping a plausible-looking fake dollar amount on a printable page.

src/components/HrTools/NsGoalCalculator/GoalSettings/Sections/FinancialInformationSection.tsx
```diff
@@ -60,6 +71,27 @@ export const FinancialInformationSection: React.FC<
         )}
       </FieldRow>
 
+      <FieldRow
+        label={t('403(b) Amount')}
+        helperText={t('Calculated monthly amount')}
+      >
+        <Typography variant="body1">
+          {currencyFormat(calculations.contributing403bAmount, 'USD', locale, {
```
canac, line 83: [Suggestion] (a11y) The primary and spouse 403(b) amounts render as two bare `<Typography>` nodes in a CSS grid (not a table), and the person column headers are plain text — so a screen reader hears "403(b) Amount … $600.00 … $390.00" with no way to tell which value is the primary's vs. the spouse's. The **input** rows in this same section already solve this via `useGoalSettingsField` (`aria-label` = `` `${label} — ${personName}` ``). Mirror it here, e.g. `aria-label={t('403(b) Amount — {{name}}', { name: primaryName })}` / `{ name: spouseName }`.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/ReviewYourCalculationStep.tsx
```diff
@@ -1,20 +1,22 @@
 import { Box, Typography } from '@mui/material';
 import { useTranslation } from 'react-i18next';
-import { NsGoalCalculatorLayout } from '../Shared/NsGoalCalculatorLayout';
+import { NsGoalCalculation } from '../Shared/NsGoalCalculatorContext';
 
-export const ReviewYourCalculationStep: React.FC = () => {
+interface ReviewYourCalculationStepProps {
+  goalCalculation: NsGoalCalculation;
```
canac, line 6: [Suggestion] (4-agent consensus: Architecture, Standards, Testing, Data Integrity) `ReviewYourCalculationStep` declares a required `goalCalculation` prop and `NsGoalCalculator.tsx:41` passes it, but the component body is `= () => {` — the prop is never destructured/used (the step only renders a "coming soon" placeholder). This passes lint/tsc (no unused *binding*), so CI won't flag it, but it's dead wiring. Drop the prop from the interface and the call site until the step consumes it (sibling `NextStepsStep` takes none), or add a one-line comment marking it reserved.

src/components/HrTools/NsGoalCalculator/NsGoalCalculator.tsx
```diff
@@ -1,18 +1,46 @@
...
+  const accountListId = useAccountListId();
+
+  const { data, loading, error } = useNewStaffGoalCalculationQuery({
+    variables: { accountListId: accountListId ?? '' },
+    skip: !accountListId,
+  });
+  const goalCalculation = data?.newStaffGoalCalculation ?? null;
+
+  if (loading) {
+    return <Skeleton variant="rectangular" height={400} />;
```
canac: [Suggestion] The loading `Skeleton`, error `Alert`, and empty `Alert` are returned directly into the layout's `mainContent` slot, which has no padding — while each resolved step supplies its own `spacing(4)` horizontal inset (`ReviewYourCalculationStep` `Box mx={4}`, `PresentingYourGoalStep` `PrintableContent marginInline`). So these three states render edge-to-edge and then the content jumps inward on load. Consider wrapping them in a `<Box mx={4} my={2}>` for visual continuity.

src/components/HrTools/NsGoalCalculator/NsGoalCalculator.tsx
```diff
@@ -1,18 +1,46 @@
...
+  }
+
+  if (error) {
+    return <Alert severity="error">{error.message}</Alert>;
```
canac: [Suggestion] (informational) At runtime the global Apollo error link already calls `snackNotifications.error(...)` for every GraphQL/network error, so a query failure surfaces **both** a snackbar and this inline `<Alert>`. This is **not** a rule violation — an inline error state is required and this matches sibling `GoalSettingsForm.tsx:82` — noting only so the team is aware the double-surface is app-wide. Separately, rendering raw `error.message` as user copy is the local convention; a friendly fallback with the raw message as detail would read better. No change required.

src/components/HrTools/NsGoalCalculator/PresentingYourGoal/PresentingYourGoalStep.test.tsx
```diff
@@ -19,7 +22,7 @@ describe('PresentingYourGoalStep', () => {
     beforeTestResizeObserver();
   });
 
-  afterEach(() => {
+  afterAll(() => {
```
canac, line 45: [Suggestion] This `afterEach → afterAll` change is a genuine **fix** (it now pairs symmetrically with `beforeAll(beforeTestResizeObserver)`; the old `beforeAll`+`afterEach` tore the ResizeObserver mock down after the first test). One residual footgun: `afterTestResizeObserver()` calls `jest.restoreAllMocks()` and now runs only once, so the `jest.spyOn(window, 'print')` from the print test is no longer restored between tests — harmless today, but a future test reading `window.print` without re-spying would inherit the stale spy. Optional to guard.

canac (commented):
## 🤖 Multi-Agent Code Review — PR #1873

**Verdict: APPROVED WITH SUGGESTIONS** — no hard blockers, but one confirmed financial-display divergence should be **resolved or product-confirmed before merge**. Human review recommended.

6 specialized agents (Financial, Data Integrity, Architecture, Testing, Standards, UX) reviewed in parallel. Testing agent ran the suite: **45/45 pass, `tsc` exit 0**.

### Headline
**The "Presenting Your Goal" total re-sums a subset of fields client-side and omits the always-applied ×1.06 attrition factor, so the donor-facing "Total Support Goal" systematically understates the admin's authoritative `monthlyGoal`.** See the inline comment on `PresentingYourGoalStep.tsx`. The established sibling `GoalCalculator` deliberately folds attrition in (`…/PresentingYourGoal.tsx:41  ministryExpenses: ministryExpensesTotal + attrition`); this new step drops it and doesn't even query `attrition`. Tests miss it because the mock is rigged so the 6 fields sum to exactly `monthlyGoal` (15860).

### Verified clean / refuted (no action)
- **Null `calculations`** — schema is `NewStaffGoalCalculationCalculations!` with all `Float!` subfields; dropping the old `?? 0` guard is correct. Parent is nullable but guarded in both consumers.
- **Coaching vs hrTools layout** — coaching page renders `GoalSettingsForm` directly (intended split); hrTools wraps `NsGoalCalculator` in exactly one `NsGoalCalculatorLayout` (no double/lost wrap).
- **Apollo cache / codegen / mutation refresh** — `id` present, embedded value-object needs none, all 9 fields exist in `schema.graphql`, mutation re-selects the fragment. Clean.
- **Standards** — full checklist PASS, no violations. All new strings localized; named exports; no `any`; TODOs ticket-referenced.

### Risk
Mechanical score 10/10 (CRITICAL) is **breadth-driven** — 12 small sibling files each +1, +2 for the `.graphql`. Substance is display-only (no auth, no mutations, non-null types), so effective operational risk ≈ MEDIUM, elevated to **HIGH** by the confirmed financial finding. Auto-approve intentionally does not apply.

### Lower-priority items (posted inline)
Fake `$3,624` special-needs placeholder shown as real money · 403(b) values missing per-person `aria-label` · unused `goalCalculation` prop on `ReviewYourCalculationStep` (4-agent consensus) · loading/error states render flush to edge · inline error Alert + global snackbar double-surface (informational, not a violation).

### Minor (not posted inline)
- Low-value `useMemo` for `monthlyNeeds` (`PresentingYourGoalStep.tsx:63-74`) — consumers don't benefit from a stable ref; optional.
- Hardcoded `'USD'` in the 403(b) format — reasonable US-only assumption, consistent with siblings.
- Non-null `!` in `NsGoalCalculatorTestWrapper.tsx:56` — safe test infra.
- `afterEach→afterAll` in `PresentingYourGoalStep.test.tsx` is a correct fix; only residual is the `window.print` spy no longer restored between tests (harmless today).

<sub>Generated by /quality:agent-review · standard mode · this is a COMMENT review, not an approval.</sub>

src/components/HrTools/NsGoalCalculator/PresentingYourGoal/PresentingYourGoalStep.tsx
```diff
@@ -67,90 +55,102 @@ export const PresentingYourGoalStep: React.FC = () => {
   });
   const supportRaised = data?.accountList.receivedPledges ?? null;
 
+  const { calculations } = goalCalculation;
+  const married =
+    goalCalculation.maritalStatus ===
+    NewStaffQuestionnaireMaritalStatusEnum.Married;
+  const monthlyNeeds = useMemo(
```
canac, line 62: **[Important]** The presented **"Total Support Goal" is re-derived client-side** here (this `monthlyNeeds` object feeds `MonthlyNeedsCard` → `useMonthlyNeedsRows.ts:80`, which does `rows.reduce((sum, row) => sum + row.amount, 0)`), instead of using the authoritative `calculations.monthlyGoal` — which this same PR now uses in `GoalSettingsForm.tsx:123`.

The 6 displayed buckets omit line items 8/10/11/12/13 (`staffConferenceTransfer`, `medicalExpenses`, `accountTransfers`, `advocacyTransfers`, `otherExpenses`). Verified against `schema.graphql`: `totalMinistryExpenses` = *"sum of lines 2-7"*, `monthlyGoal` = subtotal(1-13) + adminCharge + attrition. Therefore `monthlyGoal − displayedTotal` = the five omitted lines. When any is non-zero (medical/other expenses are **not** senior-only), the staffer is shown a goal **lower than their real `monthlyGoal`**, and the presentation silently disagrees with the Goal Settings header. The fixture reconciles only because all five omitted lines are 0.

**Fix:** pass `calculations.monthlyGoal` through as the Total Support Goal (add an optional `total`/`monthlyGoalTotal` override prop to `MonthlyNeedsCard`) and surface the missing lines (or a residual "Other" row) so the visible breakdown still sums. Confirm with product which lines the NS presentation should expose.

_Flagged by: Financial Reporting (7.5), Data Integrity (7.0), Architecture (6.0); crux verified against schema.graphql._

src/components/HrTools/NsGoalCalculator/PresentingYourGoal/PresentingYourGoalStep.tsx
```diff
@@ -67,90 +55,102 @@ export const PresentingYourGoalStep: React.FC = () => {
...
+    () => ({
+      married,
+      salary: calculations.salary,
+      ministryExpenses:
+        calculations.totalMinistryExpenses + calculations.attrition,
```
canac: **[Medium]** `ministryExpenses: totalMinistryExpenses + attrition` folds the attrition gross-up into the "Ministry Expenses" row. Per schema, `attrition` is a top-of-goal factor (line 16 − line 15), **not** a ministry expense (lines 2-7). The row displays `$1,838` when actual ministry expenses are `$898`. Mitigated by an identical pre-existing convention in the PDS calculator, and it's currently only there to force the client-side total to approach `monthlyGoal` (see the Important finding above). Consider a distinct "Attrition" row; this resolves naturally if the total is sourced from `monthlyGoal`.

_Flagged by: Financial Reporting, Data Integrity._

src/components/HrTools/NsGoalCalculator/PresentingYourGoal/PresentingYourGoalStep.test.tsx
canac, line 45: **[Medium]** Changing `afterEach(afterTestResizeObserver)` to `afterAll(...)` while keeping `beforeEach(beforeTestResizeObserver)` breaks per-test isolation: `afterTestResizeObserver` calls `jest.restoreAllMocks()`, so the `jest.spyOn(window, 'print')` in the print test now leaks to later tests. It also masks unhandled async in the three sync render tests (deferred `useAccountListSupportRaisedQuery` updates), and is inconsistent with the sibling `NsGoalCalculator.test.tsx:23` which still uses `afterEach`.

**Fix:** revert to `afterEach(afterTestResizeObserver)`, and `await findByRole(...)` a post-load assertion in the three sync tests so no state update escapes the test body.

_Flagged by: Testing._

src/components/HrTools/NsGoalCalculator/PresentingYourGoal/PresentingYourGoalStep.test.tsx
```diff
@@ -79,7 +84,7 @@ describe('PresentingYourGoalStep', () => {
     expect(getByText('Administrative Charge')).toBeInTheDocument();
     expect(getByText('$1,795')).toBeInTheDocument();
     expect(getByText('Total Support Goal')).toBeInTheDocument();
-    expect(getByText('$15,860')).toBeInTheDocument();
+    expect(getByText('$16,800')).toBeInTheDocument();
```
canac: **[Suggestion]** Asserting the hardcoded `$16,800` won't catch drift between the displayed total and the API's `monthlyGoal`. Assert against the mock's authoritative value instead, so a grouping change breaks the test:
```tsx
expect(
  getByText(currencyFormat(defaultGoalCalculation.calculations.monthlyGoal, 'USD', 'en-US')),
).toBeInTheDocument();
```
_Flagged by: Testing._

src/components/HrTools/NsGoalCalculator/GoalSettings/Sections/FinancialInformationSection.tsx
```diff
@@ -60,6 +71,33 @@ export const FinancialInformationSection: React.FC<
         )}
       </FieldRow>
 
+      <FieldRow
+        label={t('403(b) Amount')}
+        helperText={t('Calculated monthly amount')}
+      >
+        <Typography
+          variant="body1"
+          aria-label={t('403(b) Amount — {{name}}', { name: primaryName })}
```
canac: **[Medium]** `aria-label` on `<Typography variant="body1">` (renders a `<p>`, ARIA role `paragraph`) is spec-non-conformant: `paragraph` is a name-prohibited role, so screen readers won't reliably announce whose 403(b) amount this is — SR users hear "$600.00" with no owner. The `findByLabelText` test passes because Testing Library reads the DOM attribute directly, not the ARIA name-computation rule, so green ≠ announced.

**Fix:** use a visually-hidden span so the label becomes real, announced text content:
```tsx
import { visuallyHidden } from '@mui/utils';
<Typography variant="body1">
  <Box component="span" sx={visuallyHidden}>
    {t('403(b) Amount — {{name}}', { name: primaryName })}
  </Box>
  {currencyFormat(calculations.contributing403bAmount, 'USD', locale, { showTrailingZeros: true })}
</Typography>
```
(Then update the test's `findByLabelText` to a text/context query.)

_Flagged by: UX._

src/components/HrTools/NsGoalCalculator/NsGoalCalculator.tsx
canac: **[Suggestion]** This loading/error/empty block (`useNewStaffGoalCalculationQuery` + `<Skeleton variant="rectangular" height={400} />` + `<Alert severity="error">{error.message}</Alert>` + the info-alert copy) is a near-verbatim duplicate of `GoalSettingsForm.tsx:71-91`, introduced by this PR. They run on different routes (not a duplicate fetch), but the two copies will drift. Consider extracting a shared `useNewStaffGoalCalculation(accountListId)` hook or a shared render wrapper for the three states.

_Flagged by: Architecture._

canac (commented):
## 🤖 Multi-Agent Code Review — Verdict: APPROVED WITH SUGGESTIONS

6 specialized agents (Architecture, Testing, Standards, Data Integrity, UX, Financial Reporting) + dependency-impact analysis, standard mode. Solid PR: removes mock data, adds real loading/error/empty states, improves type safety, and hoists a remounting layout. Net debt reduced.

**No hard blockers.** One **Important (7.5)** finding is strongly recommended before merge (see the inline comment on `PresentingYourGoalStep.tsx`).

> ⚠️ **1 Important finding (severity 7.0–7.9) cannot be dismissed via `/dismiss`.** Fix it in this PR or open a follow-up ticket. Medium/Suggestion findings (<7) can be dismissed by replying `/dismiss: <reason>` to the inline comment.

### Risk
Rubric score **10/10 (CRITICAL)** — but capped purely by *breadth* (17 files, 11 medium-risk components) of one cohesive feature. No auth / mutations / cache-policy / server code touched. **Practical risk is MEDIUM-HIGH**, concentrated in financial-display correctness.

### Top finding (Important, 7.5) — agreed by 3 agents + verified against schema
The presentation's **"Total Support Goal" is re-derived client-side** (`useMonthlyNeedsRows` `rows.reduce`) instead of using the authoritative server value `calculations.monthlyGoal` — which *this same PR* switched to in `GoalSettingsForm.tsx:123`. The 6 displayed buckets omit line items 8/10/11/12/13 (`staffConferenceTransfer`, `medicalExpenses`, `accountTransfers`, `advocacyTransfers`, `otherExpenses`). Verified from `schema.graphql`: `totalMinistryExpenses` = *"sum of lines 2-7"*, `monthlyGoal` = subtotal(1-13) + adminCharge + attrition. So `monthlyGoal − displayedTotal` = those five omitted lines. When any is non-zero (medical/other are not senior-only), the staffer is shown a goal **lower than reality**, and the presentation disagrees with the Goal Settings header. The fixture reconciles only because all five are 0.

### Findings summary
| # | Sev | Tier | Location |
|---|----|------|----------|
| 1 | 7.5 | Important | PresentingYourGoalStep.tsx — client `reduce` vs server `monthlyGoal` |
| 2 | 6.0 | Medium | PresentingYourGoalStep.test.tsx — `afterEach`→`afterAll` isolation regression |
| 3 | 5.5 | Medium | FinancialInformationSection.tsx — `aria-label` on `<p>` won't announce |
| 4 | 5.0 | Medium | PresentingYourGoalStep.tsx — `attrition` folded into Ministry Expenses row |
| 5 | 4.5 | Suggestion | Test: assert total == `monthlyGoal` (would catch #1) |
| 6 | 4.0 | Suggestion | NsGoalCalculator.tsx — duplicated loading/error/empty block |
| 7 | 4.0 | Suggestion | Test: single-person 403(b) hide path |
| 8 | 3.5 | Suggestion | ReviewYourCalculationStep.tsx — dead `goalCalculation` prop |
| 9 | 3.0 | Suggestion | NsGoalCalculatorTestWrapper.tsx — non-null assertion (test-only) |

Additional low-severity suggestions (not line-anchored): loading `Skeleton height={400}` doesn't approximate content (kept identical to sibling — consistency nit); support-raised query now waterfalls after goal-calc load on the Presenting step (minor latency).

### Dependency impact
No unhandled breaking changes — every consumer updated in-PR or compatible. `calculatedResults` removed with **0 remaining references** repo-wide. Two operational notes: **`yarn gql` codegen must succeed** for the fragment change; the default test mock flipped `spouseJoining false→true` and added fields — **run the 4 wrapper-dependent test files** (`NextStepsStep`, `NsGoalCalculatorContext`, `NsGoalCalculatorStepsList`, `NsGoalCalculatorLayout`) to confirm no assertion drift.

### Standards
All checklist items PASS (2 low-severity WARN). Named exports ✓ · i18n `{{interpolation}}` ✓ · `<Trans t={t}>` ✓ · no `any`/`@ts-ignore`/`new Date()` ✓ · GraphQL `id` correctly present (`calculations` is a non-normalizable value object — no `id` needed) · codegen fields verified present in schema ✓.

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsForm.test.tsx
```diff
@@ -117,6 +117,15 @@ describe('GoalSettingsForm', () => {
     ).not.toBeInTheDocument();
   });
 
+  it('shows the calculated 403(b) contribution amount for each person', async () => {
```
canac, line 121: **[Suggestion]** The new 403(b) amount display is only tested for the couple case. Add a test asserting the spouse 403(b) amount is **not** rendered for a single person (`hasSpouse === false`, using `singleMock`), so a regression that always renders the spouse row is caught.

_Flagged by: Testing._

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/ReviewYourCalculationStep.tsx
```diff
@@ -1,20 +1,22 @@
-import { Box, Typography } from '@mui/material';
+import { Typography } from '@mui/material';
 import { useTranslation } from 'react-i18next';
-import { NsGoalCalculatorLayout } from '../Shared/NsGoalCalculatorLayout';
+import { NsGoalCalculation } from '../Shared/NsGoalCalculatorContext';
 
-export const ReviewYourCalculationStep: React.FC = () => {
+interface ReviewYourCalculationStepProps {
+  goalCalculation: NsGoalCalculation;
```
canac, line 6: **[Suggestion]** `ReviewYourCalculationStepProps` declares a required `goalCalculation` and `NsGoalCalculator.tsx:41` passes it, but the component signature is `= () =>` — the prop is accepted and silently ignored (step is still "coming soon"). Not a lint error (never destructured), but it reads as wired-up while doing nothing. Either drop the prop until the step consumes it, or add a `// consumed once this step is built (MPDX-XXXX)` comment to signal intent.

_Flagged by: Architecture, Testing, Standards._

src/components/HrTools/NsGoalCalculator/NsGoalCalculatorTestWrapper.tsx
```diff
@@ -1,36 +1,68 @@
...
+        attrition: 940,
+        monthlyGoal: 16800,
+      },
+    },
   },
-};
+});
+
+export const defaultGoalCalculation =
+  defaultGoalCalculationMock.newStaffGoalCalculation!;
```
canac, line 45: **[Suggestion]** Non-null assertion `defaultGoalCalculationMock.newStaffGoalCalculation!` on a schema-nullable field. Acceptable in test infra, but the sibling `PdsGoalCalculatorTestWrapper` avoids `!` by consuming the gqlMock result directly — consider matching that to keep the project's "no non-null assertion" convention.

_Flagged by: Standards._

src/components/HrTools/NsGoalCalculator/PresentingYourGoal/PresentingYourGoalStep.tsx
canac: This intentionally matches the old PDF we're recreating.

src/components/HrTools/Shared/GoalPresentation/useMonthlyNeedsRows.ts
```diff
@@ -77,9 +73,7 @@ export const useMonthlyNeedsRows = ({
       },
     ];
 
-    const total = rows.reduce((sum, row) => sum + row.amount, 0);
-
-    return { rows, total };
+    return rows;
```
zweatshirt, line 76: I noticed now, there is about a ~$305 discrepancy between the sum of rows here and `monthlyNeeds.monthlyGoal` under your account list in the test account. Do we want to display the goal subtotals that contribute to this discrepancy to avoid confusion? Or I am guessing that may be later work?

zweatshirt (approved):
Looks good from what I can tell. I only had one real concern. Since this PR will effect https://github.com/CruGlobal/mpdx-react/pull/1884 I'll wait for your PR to merge

src/components/HrTools/Shared/GoalPresentation/useMonthlyNeedsRows.ts
canac: Good catch! Healthcare and transfers needed to be part of the ministry expenses row.

PR #1886, opened 2026-07-08
[MPDX-9673] Add review calculations page

canac:
## Description

Implement the new staff goal calculator review calculations page.

~~Branches off of #1873 (merged)~~

MPDX-9673

## Testing

- Go to a new staff goal calculation (I'm using `/accountLists/90935daa-0e49-4b31-a41b-d6b46854ebca/hrTools/nsGoalCalculator`. You can ask Claude to help make yourself coach that account list if you like)
- Check that the calculations render on step 1

## Checklist:


canac (commented):

6 specialized agents (Architecture, Data Integrity, Testing, UX, Standards, Financial Reporting) + 1 adversarial verifier. Standard mode. **No blockers** — nothing reached the Important (7.0) or blocker (8.0) tier. This is a display-only page over **server-computed, non-nullable** money fields, which is the right architecture and keeps real risk low.

**Risk: 10/10 CRITICAL** — driven by breadth (6 new financial components) + the financial-reporting domain, per repo rules; **real runtime risk is materially lower** (display-only, server-computed, well-tested). Reviewer level: **SENIOR** (financial/HR-tools domain, not severity). Not auto-approve eligible (CRITICAL risk → human review).

### Top 3 to address before merge
1. **Line 1D shows primary-only 403(b) — worksheet won't foot for dual-403(b) couples** (5.5, CONFIRMED). `MonthlyNeedsCard.tsx:71` binds `contributing403bAmount` (primary), but line 1 (`totalSalary`) embeds the *combined* 1D. Sibling `PresentingYourGoalStep.tsx:70` correctly uses `totalContributing403bAmount`.
2. **"Special Needs Goal $2,801.00" headline contradicts the $0.00 detail table** on the same screen (5.5). The hardcoded `specialNeedsGoalPlaceholder = 2801` is refuted by the all-zero `SpecialNeedsCard` below it.
3. **Raw rate factors interpolated into labels can render float artifacts** e.g. `0.9299999999999999` (5.0, 4-agent consensus). `secaRate` / `1-adminRate` / `1+attritionRate` bypass the display-boundary formatting used elsewhere.

### Confirmed good
- `gqlMock` refactor of `goalSettingsApiMapping.test.ts` verified sound (every asserted value explicitly mocked; seeded/deterministic; −60 lines). No weakening.
- Fragment expansion is additive; `id` present on the normalizable type; codegen consistent; no `any`/`new Date()`; named exports; i18n mechanics correct.
- `NeedsTable.tsx` having no test is acceptable (styled-module convention, verified across 10+ peers).

Inline comments below are severity-tagged. Findings below 7.0 can be dismissed by replying `/dismiss: <reason>` on the comment.

---

_Informational — cannot be posted as line comments._

- **`PresentingYourGoalStep.tsx`** uses a **third** special-needs placeholder (`specialNeedsPlaceholder = 3624`) for the same concept this PR renders as `2801`. A follow-up (MPDX-9801) should consolidate the special-needs figure to one source across both NS steps.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/MonthlyNeedsCard.tsx
```diff
@@ -0,0 +1,237 @@
...
+      {
+        line: '1C',
+        category: t('Subtotal Salary'),
+        description: t('Add lines 1A-B'),
+        amount: calculations.salarySubtotal,
+      },
+      {
+        line: '1D',
+        category: t('403(b) Voluntary Retirement Plan'),
+        amount: calculations.contributing403bAmount,
```
canac: [Medium] **Line 1D uses primary-only 403(b); the worksheet won't foot for dual-403(b) couples.** (CONFIRMED by adversarial verification.)

`contributing403bAmount` is documented in the schema as *"Line 1D. **Primary applicant's** monthly 403(b)..."*. But line 1 (`totalSalary`, "Add lines 1C-D") is computed server-side as `salarySubtotal (1C) + totalContributing403bAmount` — the **combined** 1D. So for a married couple where the spouse also contributes, displayed `1C + 1D` is short of displayed line 1 by the spouse's amount, and the visible column doesn't add up. The sibling worksheet `PresentingYourGoalStep.tsx:70` correctly binds the combined field. (The test masks this: `contributing403bAmount: 990` is actually the combined wrapper value 600+390.)

```suggestion
        amount: calculations.totalContributing403bAmount,
```
Also add a test with distinct primary/spouse 403(b) values so footing is enforced.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/ReviewYourCalculationStep.tsx
```diff
@@ -1,22 +1,103 @@
...
+  NsGoalCalculation,
+  useNsGoalCalculator,
+} from '../Shared/NsGoalCalculatorContext';
+import { AccountBalanceCard } from './AccountBalanceCard';
+import { GoalSummaryCard } from './GoalSummaryCard';
+import { MonthlyNeedsCard } from './MonthlyNeedsCard';
+import { SpecialNeedsCard } from './SpecialNeedsCard';
+
+// TODO(MPDX-9801): Special needs are not available yet.
+const specialNeedsGoalPlaceholder = 2801;
```
canac: [Medium] **Fabricated Special Needs headline contradicts its own detail table.** This hardcoded `2801` renders as "Special Needs Goal $2,801.00" in `GoalSummaryCard`, while `SpecialNeedsCard` shows every line (incl. subtotal) as $0.00 directly below — two inconsistent numbers on one screen. Flagged by Financial, UX, and Architecture agents.

Until MPDX-9801 lands, drive the summary figure from the same (zero) source as the detail card, or hide/"coming soon" it — don't display a hardcoded non-zero goal the rest of the page refutes.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/MonthlyNeedsCard.tsx
```diff
@@ -0,0 +1,237 @@
...
+export const MonthlyNeedsCard: React.FC<MonthlyNeedsCardProps> = ({
+  calculations,
+  supportRaised,
+  columnLabel,
+}) => {
+  const { t } = useTranslation();
+  const locale = useLocale();
+
+  const rows = useMemo((): NeedsRow[] => {
+    const secaMultiplier = calculations.secaRate;
```
canac: [Medium] **Raw computed rates interpolated into labels can surface float artifacts** (4-agent consensus: Financial, UX, Data Integrity, Standards). `secaMultiplier` (`secaRate`), `adminDivisor = 1 - adminRate`, and `attritionFactor = 1 + attritionRate` are passed unformatted into `t('Line 1A x {{multiplier}}')` / `t('Divide line 14 by {{divisor}}')` / `t('Line 15 x {{factor}} attrition factor')`. Verified in Node that e.g. `1 - 0.07 = 0.9299999999999999`. Today's constants render clean, but rates are year-versioned server values. The adjacent line-15 label already does it right via `percentageFormat(...)`. Format these three through `numberFormat`/`percentageFormat` before interpolation.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/GoalSummaryCard.test.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { GoalSummaryCard } from './GoalSummaryCard';
```
canac, line 3: [Suggestion] **Misnamed test file — standards FAIL** (Standards, Testing, Architecture). This file tests and imports `GoalSummaryCard` (`describe('GoalSummaryCard')`), but there is no `ScenarioGoalSummaryCard` component and no colocated `GoalSummaryCard.test.tsx`. Breaks the `Foo.test.tsx`-alongside-`Foo.tsx` convention. Fix: `git mv ScenarioGoalSummaryCard.test.tsx GoalSummaryCard.test.tsx` (no content change).

src/components/HrTools/NsGoalCalculator/GoalSettings/NewStaffGoalCalculation.graphql
```diff
@@ -4,14 +4,34 @@ fragment NewStaffGoalCalculationFields on NewStaffGoalCalculation {
...
+    meals
+    mpd
+    supplies
+    staffConferenceTransfer
     benefitsCharge
+    medicalExpenses
+    accountTransfers
+    advocacyTransfers
+    otherExpenses
+    totalMinistryExpenses
```
canac: [Suggestion] **Duplicate `totalMinistryExpenses` selection** (also present at line 11). Harmless (GraphQL/codegen dedupe, `yarn gql` runs clean), but redundant — and it isn't consumed by the new cards. Remove this line.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/GoalSummaryCard.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  return (
+    <PresentationCard title={t('Your MPD Goal Calculation')}>
+      <Stack direction={{ xs: 'column', md: 'row' }} spacing={3}>
+        {figures.map(({ label, amount }) => (
+          <Stack key={label} sx={{ flex: 1 }}>
+            <Typography
+              variant="subtitle2"
+              fontWeight="bold"
+              noWrap
```
canac: [Suggestion] **`noWrap` truncates the longest label even on desktop.** "MIN STAFF ACCOUNT BALANCE TO REPORT" (uppercased, ~35 chars) won't fit a third of the card and will be silently ellipsized with no way to recover the full text. Drop `noWrap` and let the label wrap (the `Stack` already stacks label above value), or add a `title`/`Tooltip`.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/AccountBalanceCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+export const AccountBalanceCard: React.FC<AccountBalanceCardProps> = ({
+  minAccountBalance,
+  columnLabel,
+}) => {
+  const { t } = useTranslation();
+  const locale = useLocale();
+
+  return (
+    <PresentationCard title={t('Min Staff Account Balance Upon Reporting')}>
+      <NeedsTable size="small">
```
canac: [Suggestion] **Missing `overflowX` scroll wrapper.** `MonthlyNeedsCard`, `SpecialNeedsCard`, and the shared `SupportNeedsTable` all wrap their `<Table>` in `<Box sx={{ width: '100%', overflowX: 'auto' }}>` (established convention); this card drops `<NeedsTable>` straight in, so a narrow viewport can trigger horizontal page scroll. Wrap it the same way (add `Box` to the import).

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/ReviewYourCalculationStep.tsx
```diff
@@ -1,22 +1,103 @@
...
   ReviewYourCalculationStepProps
-> = () => {
+> = ({ goalCalculation }) => {
   const { t } = useTranslation();
+  const { handleContinue } = useNsGoalCalculator();
+  const accountListId = useAccountListId() ?? '';
+  const { data } = useAccountListSupportRaisedQuery({
+    variables: { accountListId },
+  });
+  const supportRaised = data?.accountList.receivedPledges ?? 0;
```
canac: [Suggestion] **`?? 0` flashes a real-looking `$0` before the query resolves.** While `useAccountListSupportRaisedQuery` loads, line 17 shows $0.00 and line 18 shows the full goal, then both jump. Sibling `PresentingYourGoalStep` uses `?? null`. _Contested:_ Data Integrity/Architecture note `?? 0` matches 2+ siblings (established convention). Consider passing `null` + a `Skeleton` for lines 17/18 to avoid the misleading transient.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/ReviewYourCalculationStep.tsx
```diff
@@ -1,22 +1,103 @@
...
+              'Call MPD Coach to verify support goal has been raised (monthly and special gifts).',
+            )}
+          </li>
+          <li>{t('First checks have been collected.')}</li>
+        </Box>
+      </Alert>
+
+      <Button
+        variant="contained"
+        onClick={handleContinue}
```
canac, line 90: [Suggestion] **The Continue button's only behavior is untested.** `ReviewYourCalculationStep.test.tsx` only asserts the button exists; nothing clicks it, so a broken/omitted `onClick={handleContinue}` would pass CI. Add a `userEvent` click test that proves the step advances (assert the next step renders). Per convention, do not `await` the `userEvent.click`.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/MonthlyNeedsCard.tsx
```diff
@@ -0,0 +1,237 @@
...
+      {
+        line: '17',
+        category: t('Solid Monthly Support Developed'),
+        amount: supportRaised,
+      },
+      {
+        line: '18',
+        category: t('Monthly Support to be Developed'),
+        description: t('Subtract line 17 from line 16'),
+        amount: calculations.monthlyGoal - supportRaised,
```
canac, line 174: [Suggestion] **Line 18 is the only client-computed value and has no over-raised (negative) test, and is unclamped vs the sibling.** `PdsSummaryTable.tsx:178` clamps the equivalent line with `Math.max(..., 0)`; this renders a negative `-$X.00` when `supportRaised > monthlyGoal`. Decide whether the negative is intended (surplus) or should clamp, make the siblings agree, and add a test case for `supportRaised > monthlyGoal`.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/NeedsTable.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export interface NeedsRow {
+  line: string;
+  category: string;
+  description?: string;
+  amount: number;
+  /** Bolds the entire row (totals and the final lines). */
+  bold?: boolean;
+}
+
```
canac, line 28: [Suggestion] **`MonthlyNeedsCard` and `SpecialNeedsCard` duplicate ~45 lines of identical table markup; `NeedsTable` only styles.** Both render a byte-identical `PresentationCard > Box(overflowX) > NeedsTable > head[line/Category/columnLabel] > rows.map` body — only the title and `rows` source differ. Precedent: `Shared/GoalPresentation/SupportNeedsTable.tsx` takes `rows[]` and owns the markup. Promote `NeedsTable` (or add `NeedsWorksheet({ columnLabel, rows })`) so both cards collapse to `<PresentationCard><NeedsWorksheet .../></PresentationCard>`.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/GoalSummaryCard.tsx
```diff
@@ -0,0 +1,62 @@
...
+              sx={{ textTransform: 'uppercase' }}
+            >
+              {label}
+            </Typography>
+            <StyledHeaderValue
+              variant="h3"
```
canac: [Suggestion] `variant="h3"` is then hard-overridden to the h4 token via `sx`, which discards `StyledHeaderValue`'s built-in responsive font size (h2 → h4 on `down('sm')`). Use a matching variant or accept the size as a prop, so the `variant` prop isn't misleading.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/MonthlyNeedsCard.tsx
```diff
@@ -0,0 +1,237 @@
...
+    ];
+  }, [t, locale, calculations, supportRaised]);
+
+  return (
+    <PresentationCard title={t('Monthly Needs')}>
+      <Box sx={{ width: '100%', overflowX: 'auto' }}>
+        <NeedsTable size="small">
+          <TableHead>
+            <TableRow>
+              <TableCell className="line" />
```
canac: [Suggestion] a11y: the empty first `<TableCell className="line" />` header leaves the line-number column unnamed for screen readers (the test even asserts `columnHeaders: ['', ...]`). Optionally add a visually-hidden label, e.g. `<Typography sx={visuallyHidden}>{t('Line')}</Typography>`. Same applies in `SpecialNeedsCard`.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/SpecialNeedsCard.tsx
```diff
@@ -0,0 +1,137 @@
...
+    {
+      line: '5',
+      category: t('Subtotal'),
+      description: t('Add lines 1-4'),
+      amount: 0,
+    },
+    {
+      line: '6',
+      category: t('Subtotal with Admin Assessment'),
+      description: t('Divide line 5 by .88'),
```
canac: [Suggestion] Hardcoded `.88` in the description drifts from `MonthlyNeedsCard`, which derives the equivalent from `1 - calculations.adminRate`. Inside the MPDX-9801 placeholder card (all $0), so low impact — wire to the server-provided rate when the card is completed.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/SpecialNeedsCard.tsx
```diff
@@ -0,0 +1,137 @@
...
+      description: t(
+        'This amount comes from what you inputted on the MPD Questionnaire.',
+      ),
+      amount: 0,
+      bold: true,
+    },
+    {
+      line: '8',
+      category: t('Special Needs to be Developed'),
+      description: t('Subtract line 6 from line 5'),
```
canac: [Suggestion] Line 8 description `Subtract line 6 from line 5` looks mislabeled — line 6 (with admin) > line 5 (subtotal), yielding a negative; likely should reference line 7 (Developed to Date). Placeholder card (MPDX-9801); fix when wiring real data.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/ReviewYourCalculationStep.tsx
```diff
@@ -1,22 +1,103 @@
...
   goalCalculation: NsGoalCalculation;
 }
 
 export const ReviewYourCalculationStep: React.FC<
   ReviewYourCalculationStepProps
-> = () => {
+> = ({ goalCalculation }) => {
   const { t } = useTranslation();
+  const { handleContinue } = useNsGoalCalculator();
+  const accountListId = useAccountListId() ?? '';
```
canac: [Suggestion] `accountListId ?? ''` fires the query with an empty id when absent; the parent `NsGoalCalculator.tsx` uses `skip: !accountListId`. Unreachable in practice here (steps render only after `goalCalculation` resolves) and matches the sibling pattern — informational. Add `skip: !accountListId` if tightening.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/ReviewYourCalculationStep.tsx
canac: /dismiss This page isn't user facing and will be fixed soon

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/GoalSummaryCard.tsx
canac: /dismiss The card converts to a vertical stack when the screen is narrow.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/ReviewYourCalculationStep.tsx
canac: /dismiss We will be removing this waterfall soon by loading the support raised with the calculations.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/MonthlyNeedsCard.tsx
```diff
@@ -0,0 +1,191 @@
...
+ */
+export const MonthlyNeedsCard: React.FC<MonthlyNeedsCardProps> = ({
+  calculations,
+  columnLabel,
+}) => {
+  const { t } = useTranslation();
+  const locale = useLocale();
+
+  const rows = useMemo((): NeedsRow[] => {
+    const supportRaised = calculations.supportRaised ?? 0;
```
canac, line 29: [Medium] **`supportRaised ?? 0` null branch is untested.** Both tests pass a numeric `supportRaised` (1200, 20000), so the `null → $0.00` fallback (and the resulting full-goal line 18) is never exercised. A regression that dropped the coalescing would silently break the "no support raised yet" display for a real financial figure — the exact null-amount invariant the project's Financial rules call out. Add a case: `buildCalculations({ monthlyGoal: 16960, supportRaised: null })` asserting line 17 = `$0.00` and line 18 = `$16,960.00`.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/MonthlyNeedsCard.tsx
```diff
@@ -0,0 +1,191 @@
...
+  calculations: Calculations;
+  /** Column header naming the staff member (and spouse when married). */
+  columnLabel: string;
+}
+
+/**
+ * The "Monthly Needs" card on the Review Your Calculation step, containing the
+ * 18-line needs worksheet.
+ */
+export const MonthlyNeedsCard: React.FC<MonthlyNeedsCardProps> = ({
```
canac, line 21: [Medium] **Duplicate component name `MonthlyNeedsCard`.** A second `MonthlyNeedsCard` now sits beside the pre-existing `src/components/HrTools/Shared/GoalPresentation/MonthlyNeedsCard.tsx` — different props and tables, but both imported into sibling NsGoalCalculator steps. Identical import identifiers for two unrelated components is a maintenance trap. Consider renaming this one `MonthlyNeedsWorksheetCard` (matches the `NeedsWorksheetTable` it wraps). (The underlying `NeedsWorksheetTable` vs `SupportNeedsTable` split was judged justified — genuinely different worksheets — not a defect.)

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/NeedsTable.tsx
```diff
@@ -0,0 +1,118 @@
...
+}));
+
+export interface NeedsWorksheetTableProps {
+  rows: NeedsRow[];
+  /** Column header naming the staff member (and spouse when married). */
+  columnLabel: string;
+  /** Accessible name for the table. */
+  ariaLabel: string;
+}
+
```
canac, line 54: [Medium] **No colocated `NeedsTable.test.tsx`.** Convention (and the sibling `SupportNeedsTable`) is one test per component. All functional branches here — `amount === null → '—'`, description sub-line, currency formatting — are covered transitively via `MonthlyNeedsCard.test.tsx` and `SpecialNeedsCard.test.tsx`, so practical risk is low. Add a thin dedicated test or accept the transitive coverage with a note. Two agents (Standards, Testing) flagged this as minor.

src/components/HrTools/Shared/GoalPresentation/PresentationCard.tsx
```diff
@@ -30,6 +38,10 @@ export const PresentationCard: React.FC<PresentationCardProps> = ({
 
     <Divider sx={{ margin: `${theme.spacing(2)} ${theme.spacing(-3)}` }} />
 
-    {children}
+    {horizontalScroll ? (
```
canac, line 41: [Suggestion] **`horizontalScroll` generalization is only half-applied (fix-one-fix-all).** `PersonalInfoCard` correctly drops its own overflow Box to rely on this new default-`true` wrapper — but `SupportNeedsTable.tsx` still wraps its content in an identical `<Box overflowX:'auto'>` and wasn't touched. The two Shared cards rendering it (`Shared/GoalPresentation/MonthlyNeedsCard`, `SpecialNeedsCard`) don't pass `horizontalScroll={false}`, so they now double-wrap. Not a visual regression (single scrollbar), but redundant DOM the refactor should sweep — remove the inner Box from `SupportNeedsTable`, mirroring the `PersonalInfoCard` cleanup.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/NeedsTable.tsx
```diff
@@ -0,0 +1,118 @@
...
+              <Typography variant="body1">{category}</Typography>
+              {description && (
+                <Typography variant="body2" color="text.secondary">
+                  {description}
+                </Typography>
+              )}
+            </TableCell>
+            <TableCell className={clsx('amount', { bold })}>
+              <Typography variant="body1">
+                {amount === null
```
canac, line 106: [Suggestion] **Lone `—` placeholder is ambiguous to screen readers.** Color isn't the only indicator (good — it's a text char), but a bare em dash announces inconsistently ("em dash" / "dash" / skipped) and doesn't convey "not yet available." The sibling `GoalSummaryCard` uses the clearer, translated `t('Coming soon')` for the same case. Add a visually-hidden label, e.g. `aria-label={t('Not yet available')}`. Low priority — temporary MPDX-9801 state.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/ReviewYourCalculationStep.tsx
```diff
@@ -1,22 +1,97 @@
...
   return (
-    <>
+    <Stack spacing={3}>
       <Typography variant="h6">{t('Review Your Calculation')}</Typography>
-      <Typography variant="body1" color="textSecondary">
-        {t('This step is coming soon.')}
+
+      <Typography variant="body1">
+        <Trans t={t}>
+          Please review your MPD Goal Calculation and contact your coach if you
```
canac, line 41: [Suggestion] **Copy inconsistency.** This intro refers to the next step as "Presenting **your** Goal" (lowercase y), but the actual step title is "Presenting **Your** Goal" (`PresentingYourGoalStep.tsx:90`, asserted in tests). Align the capitalization so the reference matches the destination label.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/ReviewYourCalculationStep.tsx
```diff
@@ -1,22 +1,97 @@
...
 
   return (
-    <>
+    <Stack spacing={3}>
       <Typography variant="h6">{t('Review Your Calculation')}</Typography>
-      <Typography variant="body1" color="textSecondary">
-        {t('This step is coming soon.')}
+
+      <Typography variant="body1">
+        <Trans t={t}>
```
canac, line 40: [Suggestion] **`<Trans>` is heavier than needed here.** This block has no interpolation and no embedded components, so a plain `t('…')` is simpler and the more standard react-i18next choice. Not wrong (many HrTools files use `<Trans t={t}>`), just unnecessary for pure static text.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/ReviewYourCalculationStep.tsx
```diff
@@ -1,22 +1,97 @@
...
+  const married =
+    goalCalculation.maritalStatus ===
+    NewStaffQuestionnaireMaritalStatusEnum.Married;
+  const columnLabel =
+    married && goalCalculation.spouseFirstName
+      ? t('{{firstName}} & {{spouseFirstName}}', {
+          firstName: goalCalculation.firstName,
+          spouseFirstName: goalCalculation.spouseFirstName,
+        })
+      : (goalCalculation.firstName ?? '');
```
canac, line 33: [Suggestion] **Empty column header when single and `firstName` is null.** `columnLabel` falls back to `''`, producing a blank `columnheader` across all three worksheet tables. Edge case (a real staff record has a first name, and this mirrors the `firstName ?? ''` pattern elsewhere), but a fallback like `t('Amount')` would avoid an empty header cell. Not covered by a test.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/MonthlyNeedsCard.tsx
```diff
@@ -0,0 +1,191 @@
...
+      },
+      {
+        line: '14',
+        category: t('Subtotal'),
+        description: t('Add lines 1-13'),
+        amount: calculations.goalSubtotal,
+      },
+      {
+        line: '15',
+        category: t('Subtotal for {{admin}} Admin charge', {
```
canac, line 146: [Suggestion] **Line 15 label vs divisor precision could disagree.** The category uses `percentageFormat(adminRate, { fractionDigits: 0 })` ("12%") while the divisor description uses full-precision `numberFormat(1 - adminRate)` ("0.88"). For the real `adminRate` of 0.12 these agree; for a hypothetical non-round rate they'd cosmetically mismatch ("12%" vs "0.885"). Both are display-only strings — the actual `subtotalWithAdminCharge` is server-computed and unaffected. Purely a description-consistency nit.

src/components/HrTools/Shared/SummaryHeaderCard.tsx
```diff
@@ -17,4 +17,4 @@ export const StyledHeaderValue = styled(Typography)(({ theme }) => ({
   [theme.breakpoints.down('sm')]: {
     fontSize: theme.typography.h4.fontSize,
   },
-}));
+})) as typeof Typography;
```
canac, line 20: [Suggestion] **Document the `as typeof Typography` cast.** This is the idiomatic MUI+TS workaround — `styled(Typography)` erases the polymorphism, so `<StyledHeaderValue component="p">` (needed by the new `GoalSummaryCard`) wouldn't typecheck without re-asserting `typeof Typography`. Sound and matches 8+ existing repo usages, and it doesn't break the two `variant="h2"` consumers. Add a one-line comment explaining why so a future reader doesn't "clean it up" and reintroduce the `component`-prop error.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/ReviewYourCalculationStep.tsx
```diff
@@ -1,22 +1,97 @@
...
+        columnLabel={columnLabel}
+        adminRate={goalCalculation.calculations.adminRate}
+      />
+
+      <AccountBalanceCard
+        minAccountBalance={goalCalculation.calculations.minimumAccountBalance}
+        columnLabel={columnLabel}
+      />
+
+      <Alert severity="info" icon={<InfoOutlinedIcon />}>
```
canac, line 72: [Suggestion] **Info Alert + its list items aren't asserted.** `ReviewYourCalculationStep.test.tsx` covers the heading, intro, and four card titles, but not the net-new "In order to report to your assignment…" alert or its two `<li>` steps. Low value (static localized text), but it is new user-facing content.

canac (commented):
## 🤖 Multi-Agent Review — ✅ APPROVED WITH SUGGESTIONS

**6 agents** (Financial Reporting, Data Integrity, UX, Architecture, Testing, Standards) · **0 blockers.**

The 18-line worksheet math was traced line-by-line against the server schema and is correct. Cache normalization (value-object `calculations` needs no `id`), single-server routing, nullability, and i18n all pass. All remaining items are low-severity polish.

### Risk
`10/10` **CRITICAL** *(formula cap)* — but this is inflated by **breadth**: 13 small feature-component files + the GraphQL fragment + volume ×1.3 scope. Substantive risk is MEDIUM: display-only, no mutations, no auth surface, strong test coverage. Reviewer level: **SENIOR** (financial display page). Not auto-approve eligible.

### Most actionable (cross-cutting)
The `horizontalScroll` generalization on `PresentationCard` was only half-applied — `PersonalInfoCard` was cleaned up, but `SupportNeedsTable` still wraps its own `overflowX` Box, so the two Shared cards that render it now double-wrap. See inline comment on `PresentationCard.tsx`.

### Standards
All mandatory checklist items **PASS** (named exports, i18n interpolation form, `<Trans t={t}>`, no `any`/`@ts-ignore`/`new Date()`/`fireEvent`/`await userEvent`/bare TODO, GraphQL `id` normalization; both type casts match established repo conventions). Only WARN: confirm `yarn gql` + `lint` run clean in CI (`.generated.ts` not committed).

### Dismissed (acknowledged, not counted)
3 prior-review findings dismissed by @canac all reference earlier code that has since changed; none match an active finding.

| Agent | Crit | High | Important | Suggestions |
|-------|:---:|:---:|:---:|:---:|
| Financial Reporting | 0 | 0 | 0 | 3 |
| Data Integrity | 0 | 0 | 0 | 0 |
| UX | 0 | 0 | 0 | 6 |
| Architecture | 0 | 0 | 0 | 4 |
| Testing | 0 | 0 | 1 | 3 |
| Standards | 0 | 0 | 0 | 1 |

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/GoalSummaryCard.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+  it('renders the headline figures and "Coming soon" for special needs', () => {
+    const { getByText } = render(
+      <GoalSummaryCard
+        monthlyGoal={6430.25}
+        specialNeedsGoal={null}
+        minStaffAccountBalance={9580}
+      />,
+    );
+
+    expect(getByText('Your MPD Goal Calculation')).toBeInTheDocument();
```
zweatshirt: Optional: We can use getByRole here. Maybe not worth it, to continue matching the other assertions' use of getByText

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/MonthlyNeedsCard.tsx
```diff
@@ -0,0 +1,191 @@
...
+      {
+        line: '7',
+        category: t('Supplies and Materials (ministry purposes only)'),
+        description: t('Ministry purposes only'),
+        amount: calculations.supplies,
+      },
+      {
+        line: '8',
+        category: t('Summer Expenses'),
+        description: t('Summer assignments, U.S. Staff Conf'),
```
zweatshirt: I know this is identical to the Figma, but would it be less confusing here to fully spell out `Conference`?

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/MonthlyNeedsCard.tsx
```diff
@@ -0,0 +1,191 @@
...
+      },
+      {
+        line: '10',
+        category: t('Reimbursable Medical Expenses'),
+        description: t('Deductible and non-covered medical expenses'),
+        amount: calculations.medicalExpenses,
+      },
+      {
+        line: '11',
+        category: t('Transfers to other staff, EMAF, etc.'),
```
zweatshirt: Wondering if down the line we will want to extend upon some of these. I'm not sure a new staff would know what 'EMAF' is. I guess they could ask their coach.

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/AccountBalanceCard.tsx
```diff
@@ -0,0 +1,73 @@
...
+                {columnLabel}
+              </Typography>
+            </TableCell>
+          </TableRow>
+        </TableHead>
+        <TableBody>
+          <TableRow className="bold">
+            <TableCell component="th" scope="row">
+              <Typography variant="body1">
+                {t('Min Account Balance Needed to Report')}
```
zweatshirt: Same situation here mentioned previously, this matches the Figma but it seems spelling out 'Minimum' would be friendlier to users.

zweatshirt (approved):
The code itself looks great. I'm not sure if this browser specific, but I noticed this issue at the bottom of the form where you can continue scrolling. I noticed this in the Admin View as well, so I believe this might be due to the MUI upgrade.

Image

src/components/HrTools/NsGoalCalculator/ReviewYourCalculation/MonthlyNeedsCard.tsx
canac: Yeah, maybe they can ask their coach? This doesn't seem like the place to explain EMF. (I just realized this has an extra A. It stands for Ethnic Minority Fund.)

I think I'm going to fix the abbreviation but otherwise leave this for now, if that's OK. We can get more feedback from stakeholders during UAT on labels.

canac:
> I noticed this issue at the bottom of the form where you can continue scrolling.

Thanks for catching that! Yeah, it's probably from the MUI upgrade.

PR #1888, opened 2026-07-08
MPDX-9812 - Add validation for future start dates in SettingsDialog schema in StaffExpenseReport

zweatshirt:
## Description

Fixes an issue where an SAA side error occurs if a start date is after the end date (automatically set to the end of current month)

## Testing

- Go to StaffExpenseReport
- Test setting the start and end dates
- Check that no SAA side errors occur, and validation is appropriately handled.

## Checklist:

### DIFF
--- a/src/components/Reports/Shared/SettingsDialog/SettingsDialog.test.tsx
+++ b/src/components/Reports/Shared/SettingsDialog/SettingsDialog.test.tsx
@@ -9,7 +9,12 @@ import { GqlMockedProvider } from '__tests__/util/graphqlMocking';
 import { StaffExpenseCategoryEnum } from 'src/graphql/types.generated';
 import { ReportsStaffExpensesQuery } from '../../StaffExpenseReport/GetStaffExpense.generated';
 import { DateRange } from '../../StaffExpenseReport/Helpers/StaffReportEnum';
-import { Filters, SettingsDialog, SettingsDialogProps } from './SettingsDialog';
+import {
+  Filters,
+  SettingsDialog,
+  SettingsDialogProps,
+  getValidationSchema,
+} from './SettingsDialog';
 
 const mutationSpy = jest.fn();
 const TestComponent: React.FC<
@@ -253,6 +258,32 @@ describe('SettingsDialog', () => {
     expect(
       queryByText('Start date must be earlier than end date'),
     ).not.toBeInTheDocument();
+  });
+
+  it('rejects a start date after the viewed month when no end date is set', async () => {
+    await expect(
+      getValidationSchema(DateTime.fromISO('2019-06-01')).validateAt(
+        'startDate',
+        {
+          startDate: DateTime.fromISO('2019-09-01'),
+          endDate: null,
+        },
+      ),
+    ).rejects.toThrow(
+      'Select an end date when the start date is later than the month being viewed',
+    );
+  });
+
+  it('accepts a start date within the viewed month when no end date is set', async () => {
+    await expect(
+      getValidationSchema(DateTime.fromISO('2019-06-01')).validateAt(
+        'startDate',
+        {
+          startDate: DateTime.fromISO('2019-06-30'),
+          endDate: null,
+        },
+      ),
+    ).resolves.toBeTruthy();
   });
 
   it('checks all available categories by default when no selection exists', async () => {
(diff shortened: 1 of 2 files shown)


src/components/Reports/Shared/SettingsDialog/SettingsDialog.test.tsx
```diff
@@ -255,6 +260,17 @@ describe('SettingsDialog', () => {
     ).not.toBeInTheDocument();
   });
 
+  it('rejects a future month when no end date is set', async () => {
```
zweatshirt: [Suggestion] **Failure-path-only coverage for the new validator.** The `start-date-not-future-without-end` rule has three branches — (a) no value → valid, (b) `endDate` present → valid, (c) future start, no `endDate` → invalid. Only (c) is tested here. A predicate that got inverted or broken so it rejected *every* no-end-date start date (including valid past dates) would still pass this test. Adding one positive-path assertion is the only thing guarding the over-reject side:

```ts
it('accepts a non-future start date with no end date', async () => {
  await expect(
    validationSchema.validateAt('startDate', {
      startDate: DateTime.fromISO('2019-06-01'), // before pinned now (2020-01-01)
      endDate: null,
    }),
  ).resolves.toBeTruthy();
});
```

Optional bonus: the inclusive boundary (a start date landing exactly on end-of-current-month with no end date should be *valid*) is also untested, so an off-by-one flip to `<` would go unnoticed.

zweatshirt (commented):
## Multi-Agent Code Review — Verdict: CLEAN ✅

5 specialized agents (Architecture, Testing, Standards, UX, Financial Reporting) reviewed this PR in standard mode. **No blockers and no Important-tier findings** — every finding is a low-severity, optional suggestion (max 3.5/10).

**Risk: LOW (1/10)** — 14 logic lines in a Formik/Yup schema, one colocated unit test.

### What's good
- Uses Luxon `DateTime.now().endOf('month')` (not `new Date()`), per convention.
- New `.test()` matches the exact `function(value)` / `this.parent` idiom of the two sibling schema tests.
- Module-level `i18n.t(...)` for the schema message — correct, since the schema is defined outside the component where `useTranslation`'s `t` isn't available.
- Named export only; error message surfaces correctly through `CustomDateField` → MUI `TextField` `error`/`helperText` (UX-traced).
- Financial Reporting agent: N/A — no money math; the rule actually prevents a degenerate open-ended future reporting window.

### Suggestions (all optional, none block merge)
See the 3 inline comments. The only one with real value is the failure-path-only test coverage on the new validator — a positive-path assertion guards against a predicate that over-rejects.

src/components/Reports/Shared/SettingsDialog/SettingsDialog.tsx
```diff
@@ -64,6 +64,18 @@ const validationSchema = yup.object({
 
         return value <= endDate;
       },
+    )
+    .test(
+      'start-date-not-future-without-end',
+      i18n.t('Select an end date when the start date is a month in the future'),
```
zweatshirt: [Suggestion] Two minor notes on this message:

1. **Wording:** the rule (line 77) fires for *any* start date after `DateTime.now().endOf('month')` — i.e. anything in a later month, not literally "a month" ahead. "a month in the future" can read as "~a month away." Consider "Select an end date when the start date is beyond the current month." (still actionable either way — nit only).
2. **i18n key:** this new key is not yet present in `public/locales/en/translation.json`. `yarn extract` will add it; flagged for completeness.

src/components/Reports/Shared/SettingsDialog/SettingsDialog.tsx
```diff
@@ -64,6 +64,18 @@ const validationSchema = yup.object({
...
+      function (value) {
+        const { endDate } = this.parent;
+        if (!value || endDate) {
+          return true;
+        }
+
+        return value <= DateTime.now().endOf('month');
```
zweatshirt: [Suggestion] The module-level `validationSchema` hardcodes real `DateTime.now()`, while the component accepts an injectable `time` prop (`currentTime = time ?? DateTime.now().startOf('month')`) so callers/tests can pin "now." This validation branch ignores that injected clock. No action needed — a Yup schema can't easily read component props, and the test relies on Luxon's global `Settings.now` override (2020-01-01), so determinism holds. Only relevant if clock-injection parity ever becomes a requirement (would need a `makeValidationSchema(now)` factory), which is out of proportion here.

PR #1885, opened 2026-07-08
MPDX-9819 Relax customized login for partner reminders

kegrimes:
## Description

Ministry Partner Reminders broke in staff web, so we want to launch this report before go-live on October 1st. This means we need to relax the restrictions around the customized login work:
- User type = Cru US staff
- Staff group = Senior staff, new staff, PTFS, paid with designation, and interns
- Staff account (staff account numbers will NOT be in production yet)

But, we still want developers to be able to see partner reminders in "HR Tools" dropdown menu.

_**This PR should be reverted when HCM go-live is ready!**_

## Testing

Developer test: ✅
- Using your personal account, click on "HR Tools"
- Check that "Ministry Partner Reminders" is present

Non-developer/Senior staff test: ✅
- Impersonate: courtney.campbell@cru.org
- Click on "HR Tools"
- Check that "Ministry Partner Reminders" is NOT present
- Navigate to `/hrTools/partnerReminders` and ensure user can access page

New Staff test: ✅
- Impersonate: timothy.ethington@cru.org
- Click on "HR Tools"
- Check that "Ministry Partner Reminders" is NOT present
- Navigate to `/hrTools/partnerReminders` and ensure user can access page

PTFS test: ✅
- Impersonate: christina.garber@athletesinaction.org 
- Click on "HR Tools"
- Check that "Ministry Partner Reminders" is NOT present
- Navigate to `/hrTools/partnerReminders` and ensure user can access page

No staff account/Paid with designation test: ✅
- Impersonate: kathryn.hougham@cru.org
- Click on "HR Tools"
- Check that "Ministry Partner Reminders" is NOT present
- Navigate to `/hrTools/partnerReminders` and ensure user can access page

Intern test: ✅
- Impersonate: nataley.guetherman@cru.org
- Click on "HR Tools"
- Check that "Ministry Partner Reminders" is NOT present
- Navigate to `/hrTools/partnerReminders` and ensure user can access page

## Checklist:

- [x] I have given my PR a title with the format "MPDX-(JIRA#) (summary sentence max 80 chars)"
- [x] I have applied the appropriate labels (_Add the label "Preview" to automatically create a preview environment_)
- [x] I have run the Claude Code `/pr-review` command locally and fixed any relevant suggestions
- [ ] I have requested a review from another person on the project
- [x] I have tested my changes in preview or in staging
- [x] I have cleaned up my commit history

PR #1889, opened 2026-07-08
[No Jira] - Convert raw-committed yarn cache zips to Git LFS pointers

wjames111:
## Description

- The webpack upgrade (#1885) committed 23 `.yarn/cache/*.zip` files as raw binaries instead of Git LFS pointers (the committing environment's LFS hooks were inactive), even though `.gitattributes` maps them to LFS.
- For every developer with Git LFS enabled, git re-runs the LFS clean filter on these files, gets a pointer, compares it to the raw blob in the commit, and permanently reports them as locally modified — which blocks every `git switch`/`checkout` touching them (`error: Your local changes to the following files would be overwritten by checkout`).
- This PR re-adds the 23 zips through the LFS clean filter so they are stored as proper pointers (~130 bytes each) with the archive content in LFS, matching every other zip in `.yarn/cache`. No package contents change — the pointer OIDs hash the exact bytes that were committed raw.

## Testing

- Check out this branch with Git LFS enabled (`git lfs install`)
- Run `git status` — the working tree should be clean, with no phantom-modified `.yarn/cache/*.zip` entries
- Switch to another branch and back — no "local changes would be overwritten" errors
- Run `git show HEAD:.yarn/cache/tapable-npm-2.3.3-b058b6534f-21fb64a7ae.zip | head -3` — output should be an LFS pointer (`version https://git-lfs.github.com/spec/v1 ...`), not binary
- Run `yarn install` — completes normally (cache zips resolve identically)

## Checklist:


wjames111:
Closing — will be re-pushed and re-opened by the author.

PR #1892, opened 2026-07-09
[No Jira] - Fix dimmed address text in contact list

wjames111:
## Description

- Restores the dimmed appearance of the address text in the contact list rows.
- During the MUI v7 upgrade the `Hidden` wrapper around the address `Typography` was removed. In v5 that wrapper's `ListItemText` secondary context implicitly applied the `text.secondary` color; without it the address rendered in the default (full-strength) text color.
- Fix: explicitly set `color="textSecondary"` on the address `Typography` in `ContactRow` so it renders dimmed again.
- Adds a test asserting the address renders with `theme.palette.text.secondary`.

## Testing

- Go to the Contacts page (contact list view).
- Look at a contact row that has an address.
- Check that the address text (e.g. "1111 Test Street Any City, TT Test") appears dimmed/greyed (secondary text color), matching the pre-v7 appearance, rather than the full-strength primary text color.

## Checklist:

### DIFF
--- a/src/components/Contacts/ContactRow/ContactRow.test.tsx
+++ b/src/components/Contacts/ContactRow/ContactRow.test.tsx
@@ -115,6 +115,14 @@ describe('ContactsRow', () => {
     ).toBeInTheDocument();
   });
 
+  it('renders the address dimmed', () => {
+    const { getByText } = render(<Components />);
+
+    expect(getByText('1111 Test Street Any City, TT Test')).toHaveStyle(
+      `color: ${theme.palette.text.secondary}`,
+    );
+  });
+
   it('should render check event', async () => {
     const { getByRole } = render(<Components />);
 
--- a/src/components/Contacts/ContactRow/ContactRow.tsx
+++ b/src/components/Contacts/ContactRow/ContactRow.tsx
@@ -126,6 +126,7 @@ export const ContactRow: React.FC<Props> = ({ contact, useTopMargin }) => {
                 <Typography
                   component="span"
                   variant="body2"
+                  color="textSecondary"
                   sx={{ display: { xs: 'none', sm: 'block' } }}
                 >
                   {[


wjames111 (commented):

**4 agents** (Architecture, Testing, Standards, UX) · standard mode · **Risk 1/10 (LOW)** · **0 blockers**

This is a correct, well-scoped MUI v7 bugfix. The removed `Hidden` wrapper used to supply `text.secondary` implicitly via the `ListItemText` secondary slot; adding `color="textSecondary"` restores it. Verified by consensus:

- **Correctly layered.** The address `Typography` is a *custom element* in `ListItemText`'s `secondary` slot — MUI only auto-applies `text.secondary` to a plain-string secondary, so a per-instance `color` prop is the only place the fix can live.
- **Test is sound, not brittle.** `theme` is imported; the rendered address resolves to exactly the asserted string; both sides of `toHaveStyle` derive from the same theme constant (`#9C9FA1`).
- **No standards violations, no a11y regression, no breaking change** (single rendering consumer, unchanged public API).

The two inline comments below are **non-blocking suggestions** (severity < 5) — informational only. Nothing needs to change to merge.

src/components/Contacts/ContactRow/ContactRow.tsx
```diff
@@ -126,6 +126,7 @@ export const ContactRow: React.FC<Props> = ({ contact, useTopMargin }) => {
                 <Typography
                   component="span"
                   variant="body2"
+                  color="textSecondary"
```
wjames111, line 129: [Suggestion] `color="textSecondary"` is the string-shorthand color form. It is not removed in MUI v7 and is the dominant convention in this repo (~41 uses vs ~27 of `color="text.secondary"`), so this is consistent with surrounding code and fine as-is. Since this branch is the v7 migration, you *could* optionally switch to `color="text.secondary"` to lean into the newer form — but purely optional. No change required.

src/components/Contacts/ContactRow/ContactRow.test.tsx
```diff
@@ -115,6 +115,14 @@ describe('ContactsRow', () => {
     ).toBeInTheDocument();
   });
 
+  it('renders the address dimmed', () => {
+    const { getByText } = render(<Components />);
+
+    expect(getByText('1111 Test Street Any City, TT Test')).toHaveStyle(
```
wjames111, line 121: [Suggestion] This asserts against a hardcoded address literal (`'1111 Test Street Any City, TT Test'`), while the sibling `'default'` test derives the same string from the contact mock. Minor duplication — if the mock changes, this literal must be hand-updated. Reusing the derived expression would be marginally more robust. Low priority.

src/components/Contacts/ContactRow/ContactRow.tsx
wjames111: 🤖 Fixed in 675fbdf: switched `color="textSecondary"` to the v7 form `color="text.secondary"` on the address Typography. Colocated test still passes (both resolve to the same `theme.palette.text.secondary` value).

src/components/Contacts/ContactRow/ContactRow.test.tsx
wjames111: 🤖 Fixed in 2dc070c: the "renders the address dimmed" test now derives the expected address from `contact.primaryAddress` (same expression as the sibling `default` test) instead of the hardcoded literal. Full file: 5/5 tests pass.

PR #1895, opened 2026-07-09
Allow API_URL env var to override .env.local in codegen

canac:
## Summary

Makes `API_URL=xxx yarn gql` override an `API_URL` set in `.env.local`.

`dotenv.config()` never overwrites a variable that is already set, so precedence is decided by load order (first writer wins) plus `override`. Previously `.env.local` was loaded last with `override: true`, which clobbered an inline `API_URL=xxx` passed on the command line — so you couldn't point codegen at a different schema ad hoc without editing `.env.local`.

Now `.env.local` is loaded first *without* `override`, so an inline env var wins.

Resulting precedence: **CLI env > `.env.local` > `.env`**.

## Test plan

- `yarn gql` still uses `API_URL` from `.env.local` when no inline var is set.
- `API_URL=https://api.stage.mpdx.org/graphql yarn gql` uses the inline value instead of `.env.local`.

### DIFF
--- a/codegen.ts
+++ b/codegen.ts
@@ -1,8 +1,8 @@
 import dotenv from 'dotenv';
 import type { CodegenConfig } from '@graphql-codegen/cli';
 
+dotenv.config({ path: '.env.local' });
 dotenv.config();
-dotenv.config({ path: '.env.local', override: true });
 
 const apiUrl = process.env.API_URL ?? 'https://api.stage.mpdx.org/graphql';
 const schema = [apiUrl, './pages/api/Schema/**/*.graphql'];


canac (commented):
## 🤖 Multi-Agent Code Review — Quick Mode

**Verdict: ✅ CLEAN** · Risk: LOW (3/10) · Blockers: 0

3 agents (Standards, Architecture, Testing) reviewed the one-line `codegen.ts` change. The dotenv precedence logic was verified independently against dotenv's `populate()` semantics (keys already in `process.env` are never overwritten unless `override: true`):

- **CLI env > `.env.local` > `.env`** — matches the stated goal (`API_URL=xxx yarn gql` now overrides `.env.local`).
- Regression checks pass: `.env.local` still wins over `.env`; `.env` still supplies unset keys; the old `override: true` clobbering of an inline CLI var cannot recur.

No blockers, no important issues. One optional suggestion (below) — informational only.

| Standards | 0 | 0 | 0 | 1 | High |
| Architecture | 0 | 0 | 0 | 0 | High |
| Testing | 0 | 0 | 0 | 1 | High |

codegen.ts
```diff
@@ -1,8 +1,8 @@
 import dotenv from 'dotenv';
 import type { CodegenConfig } from '@graphql-codegen/cli';
 
+dotenv.config({ path: '.env.local' });
```
canac, line 4: [Suggestion] Optional: the precedence rule (CLI env > `.env.local` > `.env`) relies on dotenv's implicit no-overwrite default and lives only in the commit message. A one-line comment here would stop a future maintainer from re-adding `override: true` and re-breaking `API_URL=xxx yarn gql`. Informational — not a blocker.

PR #1904, opened 2026-07-13
[no-Jira] Make account list id required

canac:
## Description

Make `useAccountListId` hook return `string` instead of `string | undefined` since most pages have an account list id. Have special logic guards for setup pages with an account list id.

Removing `account_list_id` from Google and organization OAuth URLs is intentional because the API doesn't use them.

## Testing

- Go to `/setup/connect`
- Add a new organization
- Check that there are no errors

## Checklist:

### DIFF
--- a/src/components/Settings/integrations/useOauthUrl.ts
+++ b/src/components/Settings/integrations/useOauthUrl.ts
@@ -1,11 +1,13 @@
 import { useEffect, useState } from 'react';
 import { IntegrationAccordion } from 'src/components/Shared/Forms/Accordions/AccordionEnum';
-import { useAccountListId } from 'src/hooks/useAccountListId';
+import { useOptionalAccountListId } from 'src/hooks/useAccountListId';
 import { useRequiredSession } from 'src/hooks/useRequiredSession';
 
 export const useOauthUrl = () => {
   const { apiToken } = useRequiredSession();
-  const accountListId = useAccountListId();
+  // Optional because the setup Connect flow reuses ConnectOrganization on
+  // /setup/connect, which has no accountListId in the route.
+  const accountListId = useOptionalAccountListId();
 
   const [origin, setOrigin] = useState('');
   useEffect(() => {
@@ -14,13 +16,15 @@ export const useOauthUrl = () => {
 
   const getRedirectUrl = (accordion: IntegrationAccordion) =>
     encodeURIComponent(
-      `${origin}/accountLists/${accountListId}/settings/integrations?selectedTab=${accordion}`,
+      accountListId
+        ? `${origin}/accountLists/${accountListId}/settings/integrations?selectedTab=${accordion}`
+        : `${origin}/setup/connect`,
     );
 
   return {
     getGoogleOauthUrl: () =>
-      `${process.env.OAUTH_URL}/auth/user/google?account_list_id=${accountListId}` +
-      `&redirect_to=${getRedirectUrl(IntegrationAccordion.Google)}` +
+      `${process.env.OAUTH_URL}/auth/user/google` +
+      `?redirect_to=${getRedirectUrl(IntegrationAccordion.Google)}` +
       `&access_token=${apiToken}`,
 
     getMailChimpOauthUrl: () =>
@@ -29,8 +33,8 @@ export const useOauthUrl = () => {
       `&access_token=${apiToken}`,
 
     getOrganizationOauthUrl: (organizationId: string) =>
-      `${process.env.OAUTH_URL}/auth/user/donorhub?account_list_id=${accountListId}` +
-      `&redirect_to=${getRedirectUrl(IntegrationAccordion.Organization)}` +
+      `${process.env.OAUTH_URL}/auth/user/donorhub` +
+      `?redirect_to=${getRedirectUrl(IntegrationAccordion.Organization)}` +
       `&access_token=${apiToken}` +
       `&organization_id=${organizationId}`,
 
--- a/src/components/Task/Modal/TaskModal.tsx
+++ b/src/components/Task/Modal/TaskModal.tsx
@@ -68,7 +68,7 @@ const TaskModal = ({
   const { t } = useTranslation();
   const { data, loading } = useGetTaskForTaskModalQuery({
     variables: {
-      accountListId: accountListId ?? '',
+      accountListId,
       taskId: taskId ?? '',
       includeComments: view === TaskModalEnum.Comments,
     },
--- a/src/hooks/useAccountListId.test.tsx
+++ b/src/hooks/useAccountListId.test.tsx
@@ -17,7 +17,7 @@ describe('useOptionalAccountListId', () => {
       wrapper: makeWrapper({ isReady: false, query: {} }),
     });
 
-    expect(result.current).toBeFalsy();
+    expect(result.current).toBeNull();
   });
 
   it('returns null when the route has no accountListId', () => {
@@ -25,7 +25,7 @@ describe('useOptionalAccountListId', () => {
       wrapper: makeWrapper({ isReady: true, query: {} }),
     });
 
-    expect(result.current).toBeFalsy();
+    expect(result.current).toBeNull();
   });
 
   it('returns the accountListId from the route', () => {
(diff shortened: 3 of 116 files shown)


src/hooks/useAccountListId.ts
```diff
@@ -1,12 +1,37 @@
...
+};
+
+/**
+ * Read the accountListId from the route, throwing if it is missing. Use this on
+ * pages nested under an [accountListId] route (all of which use
+ * getServerSideProps, so the router is ready on the first render and the param
+ * is always present). On pages that may render without an accountListId, use
+ * useOptionalAccountListId instead.
+ */
+export const useAccountListId = (): string => {
```
canac, line 27: [Medium] **The whole PR rests on this hook throwing only where an `accountListId` is guaranteed** — and that guarantee is enforced *purely structurally* (every throwing call site sits under a GSSP `[accountListId]` route or an `accountListId &&` gate). It holds today across all 58 `[accountListId]` pages + the layout/context/setup trees, but:

1. There's **no regression test** for it. `Primary.test.tsx` / `TopBar.test.tsx` only exercise the happy path. A future refactor that moves a throwing-hook component out from behind an `accountListId &&` guard (or a statically-optimized `[accountListId]` page where `router.isReady` is false on first render) would crash with nothing to catch it.
2. The only safety net — the `_app.page.tsx` Rollbar `<ErrorBoundary>` — has **no `fallbackUI`**, so a missed site is a blank white page.

Recommend: add a `TopBar`/`Primary` test rendering with `query: {}` asserting no throw + menus gated out, and give the ErrorBoundary a themed fallback. Not a blocker — the code is correct today.

src/hooks/useAccountListId.ts
```diff
@@ -1,12 +1,37 @@
...
+export const useOptionalAccountListId = (): string | null => {
   const router = useRouter();
 
   if (!router.isReady) {
-    return undefined;
+    return null;
   }
 
-  return getQueryParam(router.query, 'accountListId');
+  return getQueryParam(router.query, 'accountListId') ?? null;
```
canac, line 17: [Suggestion] Edge case: `getQueryParam` returns `''` for an empty-string param, and `'' ?? null` is `''` (not nullish) — so `useAccountListId()` returns `''` **without throwing**, quietly violating its "throws when missing" contract. Not reachable via real `[accountListId]` routes, but the new `useAccountListId.test.tsx` has no `''` case. Decide whether `''` should be treated as missing (throw) and add a test either way.

src/components/Settings/integrations/useOauthUrl.ts
```diff
@@ -14,13 +16,15 @@ export const useOauthUrl = () => {
...
   const getRedirectUrl = (accordion: IntegrationAccordion) =>
     encodeURIComponent(
-      `${origin}/accountLists/${accountListId}/settings/integrations?selectedTab=${accordion}`,
+      accountListId
+        ? `${origin}/accountLists/${accountListId}/settings/integrations?selectedTab=${accordion}`
+        : `${origin}/setup/connect`,
     );
 
   return {
     getGoogleOauthUrl: () =>
```
canac, line 25: [Suggestion] Confirmed the Google + DonorHub `account_list_id` removal is intentional (API doesn't use them; `access_token` unchanged so auth isn't weakened). The residual, flagged by 5/6 agents: this leaves the four builders **inconsistent** — `getMailChimpOauthUrl` (L31) and `getPrayerlettersOauthUrl` (L42) still emit `&account_list_id=${accountListId}`. Since the hook is now `useOptionalAccountListId()` (`string | null`), those two would serialize the literal `account_list_id=null` if ever called without an id. Not reachable today (only the org/DonorHub URL is reused on `/setup/connect`). Recommend a one-line comment on why the asymmetry is deliberate, and/or guarding the two remaining interpolations against `null` to kill the future footgun.

canac (commented):
## 🤖 Multi-Agent Review — APPROVED WITH SUGGESTIONS

**6 agents** (Security, Architecture, Data Integrity, Testing, Standards, UX) + a Financial-domain check. Model: opus · Mode: standard · Risk: **8/10 HIGH** (breadth: 137 call sites / 115 files — not a defect signal). All agents **High** confidence, each after reading real source and running `tsc` + `eslint` + tests.

### The core question — answered ✅
*Does any component that switched to the throwing `useAccountListId()` and dropped its `skip: !accountListId` guard ever render on a route without an `accountListId`?* → **No.** Verified across all 58 `[accountListId]` pages (all use `getServerSideProps` → router ready with the param on first render), the `Primary` → `TopBar`/`NavBar` gating chain, all three context providers, and the `/setup/connect` + chooser + `acceptInvite` no-id routes. On `/setup/connect`, `ConnectOrganization` imports only helpers (not the component) from `OrganizationAccordion` and reaches the id solely via `useOauthUrl` — which was correctly switched to the optional hook.

### Context note — adjudicated
Dropping `account_list_id` from the Google and organization (DonorHub) OAuth URLs **checks out**: the auth credential (`access_token`) is unchanged so authorization isn't weakened, and the org URL still carries `organization_id`. Accepted. See the inline note on the residual per-provider asymmetry.

### Findings (none are blockers)
- **F1 · Medium (5.0)** — the gating invariant is enforced only structurally; no regression test, and the `_app` Rollbar `<ErrorBoundary>` has no `fallbackUI` (blank screen if a site is ever missed). *Inline on `useAccountListId.ts`.*
- **F2 · Suggestion (3.0, 5/6 agents)** — `useOauthUrl` `account_list_id` handling is now inconsistent across the 4 providers. *Inline on `useOauthUrl.ts`.*
- **F3 · Suggestion (3.0)** — `useAccountListId()` returns `''` (doesn't throw) for an empty-string param. *Inline on `useAccountListId.ts`.*
- **F5 · Cleanup** — `NavMenu.tsx` dead `null` ternary; `TaskModalProvider` global-reach note; `Announcements.tsx` `/accountLists/null` guard.
- Financial domain: **No financial calculation code in this PR** — HrTools/Reports files only swap the hook; zero amount/currency/rounding/date changes.

### Technical debt
Net **reduction**: removes ~100 `?? ''` fallbacks, dozens of `skip: !accountListId` guards, and ~9 `jest.mock('…/useAccountListId')` anti-patterns (replaced with `TestRouter`). `tsc` exit 0; `eslint` 0 errors across all 87 changed source files.

---

### Findings on Related Files (not line-anchored)

**[Suggestion 4.0 · Pre-existing] `pages/_app.page.tsx` — `<ErrorBoundary>` has no `fallbackUI`.** On a caught error `@rollbar/react` renders `null` (blank page). Pre-existing, but this PR raises the stakes since a missed throwing site now crashes instead of silently becoming `?? ''`. Give it a themed, translated fallback with a link back to `/accountLists`.

**[Suggestion 4.5 · Pre-existing] `useOauthUrl.ts` — `access_token` in the OAuth URL query string.** Present on `main`; not introduced or worsened here. Tokens in URLs can leak via `Referer`/history/proxy logs. Worth a separate ticket.

**[Cleanup · Pre-existing, out of scope] `src/components/HrTools/Shared/CalculationReports/DirectionButtons/DirectionButtons.test.tsx`** still `jest.mock`s `useAccountListId` — the one straggler of the anti-pattern this PR otherwise removes. (Being addressed alongside this review.)

src/hooks/useAccountListId.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  it('throws when the route has no accountListId', () => {
+    expect(() =>
+      renderHook(() => useAccountListId(), {
+        wrapper: makeWrapper({ isReady: true, query: {} }),
+      }),
+    ).toThrow(/accountListId/);
+  });
+
+  it('throws when the route has no blank accountListId', () => {
```
zweatshirt:
```suggestion
  it('throws when the route has a blank accountListId', () => {
```
Do we want this?

src/components/Task/Modal/TaskModal.tsx
zweatshirt, line 165: This path is dead right?

zweatshirt (approved):
Code looks great. I didn't notice any issues doing the suggested user testing.

src/components/Task/Modal/TaskModal.tsx
canac: It's imported by `TaskModalProvider.tsx`

PR #1905, opened 2026-07-13
MPDX-9838 Add HR Tools and partner reminders to prod

kegrimes:
## Description

**_Set to merge on Monday, July 27th, 2026_**

Since Ministry Partner Reminders broke in staff web, we want to release this report before HCM go-live. We need to add HR Tools and partner reminders to production.

Production: `DISABLE_NEW_REPORTS=true`, `DEVELOPMENT_ENV=false`, `user_type_verified="false"` (no modal), and `us_staff_group=null` (no HCM data)

## Testing

User type has NOT been verified in the modal (this is what will happen in production)
```
user_type_verified: 'false'
```

US staff production test: ✅
- Set `DISABLE_NEW_REPORTS` to `true` in your local
- Impersonate: dave.cook@crumilitary.org
- Verify that "HR Tools" is present with only Ministry Partner Reminders
- Go to `/hrTools/partnerReminders`
- Check that the page is accessible

Non-US staff production test: ✅
- Set `DISABLE_NEW_REPORTS` to `true` in your local
- Log in to the test account (it is seen as Global right now)
- Verify that "HR Tools" is NOT present

US staff non-production test: ✅
- Set `DISABLE_NEW_REPORTS` to `false` in your local
- Impersonate: dave.cook@crumilitary.org
- Verify that "HR Tools" is present with only Ministry Partner Reminders (user group is `null` and user type verified is false - no HCM data and no modal)
- Go to `/hrTools/partnerReminders`
- Check that the page is accessible

User type has been verified in the modal...
```
user_type_verified: 'true'
```
... AND HCM data exists (this is similar to what devs should see locally)

US staff non-production test: ✅
- Set `DISABLE_NEW_REPORTS` to `false` in your local
- Verify that "HR Tools" is present with all eligible reports
- Go to `/hrTools/partnerReminders`
- Check that the page is accessible

## Checklist:

### DIFF
--- a/src/hooks/useHrToolsNavItems.test.tsx
+++ b/src/hooks/useHrToolsNavItems.test.tsx
@@ -4,6 +4,7 @@ import { GqlMockedProvider } from '__tests__/util/graphqlMocking';
 import { mockSession } from '__tests__/util/mockSession';
 import { GetUserQuery } from 'src/components/User/GetUser.generated';
 import { UsStaffGroupEnum, UserTypeEnum } from 'src/graphql/types.generated';
+import { UserOptionQuery } from './UserPreference.generated';
 import { useHrToolsNavItems } from './useHrToolsNavItems';
 
 // A user in a group that is ineligible for every HR Tool and with no staff account
@@ -31,8 +32,11 @@ const Wrapper = ({ children }: { children: ReactElement }) => (
 );
 
 const MpdGoalEligibleWrapper = ({ children }: { children: ReactElement }) => (
-  <GqlMockedProvider<{ GetUser: GetUserQuery }>
-    mocks={{ GetUser: { user: mpdGoalEligibleUser } }}
+  <GqlMockedProvider<{ GetUser: GetUserQuery; UserOption: UserOptionQuery }>
+    mocks={{
+      GetUser: { user: mpdGoalEligibleUser },
+      UserOption: { userOption: { key: 'user_type_verified', value: 'true' } },
+    }}
   >
     {children}
   </GqlMockedProvider>
@@ -42,16 +46,33 @@ describe('useHrToolsNavItems', () => {
   afterEach(() => {
     process.env.DEVELOPMENT_ENV = 'false';
     process.env.DISABLE_MPD_GOAL_ADMIN = 'false';
+    process.env.DISABLE_NEW_REPORTS = 'false';
   });
 
-  it('hides all items for an ineligible user when not in a development env', async () => {
+  it('hides all items, except partner reminders, for an ineligible user when not in a development env', async () => {
     const { result, waitForNextUpdate } = renderHook(
       () => useHrToolsNavItems(),
       { wrapper: Wrapper },
     );
     await waitForNextUpdate();
 
-    expect(result.current.items).toHaveLength(0);
+    expect(result.current.items).toHaveLength(1);
+    expect(result.current.items[0].id).toBe('partnerReminders');
+  });
+
+  it('hides every eligible item except partner reminders when DISABLE_NEW_REPORTS is on', async () => {
+    process.env.DISABLE_NEW_REPORTS = 'true';
+    mockSession({ developer: false });
+
+    const { result, waitForNextUpdate } = renderHook(
+      () => useHrToolsNavItems(),
+      { wrapper: MpdGoalEligibleWrapper },
+    );
+    await waitForNextUpdate();
+
+    expect(result.current.items.map((item) => item.id)).toEqual([
+      'partnerReminders',
+    ]);
   });
 
   it('shows all items for an ineligible developer in a development env', async () => {
@@ -111,7 +132,7 @@ describe('useHrToolsNavItems', () => {
     );
   });
 
-  it('hides all items for an ineligible non-developer in a development env', async () => {
+  it('hides all items, except partner reminders, for an ineligible non-developer in a development env', async () => {
     process.env.DEVELOPMENT_ENV = 'true';
     mockSession({ developer: false });
 
@@ -121,10 +142,11 @@ describe('useHrToolsNavItems', () => {
     );
     await waitForNextUpdate();
 
-    expect(result.current.items).toHaveLength(0);
+    expect(result.current.items).toHaveLength(1);
+    expect(result.current.items[0].id).toBe('partnerReminders');
   });
 
-  it('hides all items for an ineligible developer outside a development env', async () => {
+  it('hides all items, except partner reminders, for an ineligible developer outside a development env', async () => {
     process.env.DEVELOPMENT_ENV = 'false';
     mockSession({ developer: true });
 
@@ -134,6 +156,7 @@ describe('useHrToolsNavItems', () => {
     );
     await waitForNextUpdate();
 
-    expect(result.current.items).toHaveLength(0);
+    expect(result.current.items).toHaveLength(1);
+    expect(result.current.items[0].id).toBe('partnerReminders');
   });
 });
--- a/src/hooks/useHrToolsNavItems.ts
+++ b/src/hooks/useHrToolsNavItems.ts
@@ -3,6 +3,7 @@ import { useTranslation } from 'react-i18next';
 import { useDeveloperBypass } from './useDeveloperBypass';
 import { useIneligibleByGroup } from './useIneligibleByGroup';
 import { NavItems } from './useReportNavItems';
+import { useReportsDisabled } from './useReportsDisabled';
 
 export function useHrToolsNavItems(): {
   items: NavItems[];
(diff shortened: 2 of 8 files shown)


src/hooks/useHrToolsNavItems.ts
```diff
@@ -30,63 +33,67 @@ export function useHrToolsNavItems(): {
...
         id: 'pdsGoalCalculator',
         title: t('Paid with Designation Support Goal Calculator'),
-        hideItem: inPdsGoalCalcIneligibleGroup,
+        hideItem: reportsDisabled || inPdsGoalCalcIneligibleGroup,
       },
       {
         id: 'partnerReminders',
         title: t('Ministry Partner Reminders'),
-        hideItem: !developerBypass,
+        // TODO: Once HCM goes live, add has no staff account gate back
```
kegrimes: [Medium] `// TODO` without a Jira/MPDX ticket reference — violates the Standards Checklist ("no TODO without a ticket"). This one is load-bearing: it's the only record that `partnerReminders` is deliberately un-gated (`hideItem: false`) and must be re-gated once HCM is live. Not lint-enforced, so it won't fail CI, but it's easy to lose.

Fix: `// TODO(MPDX-XXXX): re-add hasNoStaffAccount gate once HCM is live`. Flagged by Standards, Architecture, Testing, and UX agents.

src/hooks/useHrToolsNavItems.ts
```diff
@@ -20,6 +21,8 @@ export function useHrToolsNavItems(): {
     userLoading,
   } = useIneligibleByGroup();
   const developerBypass = useDeveloperBypass();
+  // Partner Reminders is live in production; every other HR Tool is still disabled
+  const { reportsDisabled } = useReportsDisabled();
```
kegrimes, line 25: [Medium] `reportsDisabled` fails **open** while its query is loading, causing a flash-of-extra-items. In `useReportsDisabled.ts`, `reportsDisabled = env || (!!data && userGroupVerified !== 'true')` — the `!!data` guard means that while the *separate* `UserOption` query is in flight, `reportsDisabled` is `false`. But the `items` memo only waits on `GetUser` (`userLoading`), so on cold load: GetUser resolves → tab shows **all** eligible tools → UserOption resolves → list collapses to just Partner Reminders.

Dormant in prod today (with `DISABLE_NEW_REPORTS=true`, `reportsDisabled` is unconditionally true), but it surfaces the moment reports go live. Same root makes the NavMenu "unverified" test's negative assertions timing-dependent (potential flake).

Fix: make the gate fail **closed** — have `useReportsDisabled` also return `loading` and treat loading as disabled (default `reportsDisabled` true until the value is known). _(UX + Testing agents)_

src/hooks/useHrToolsNavItems.test.tsx
```diff
@@ -44,14 +48,15 @@ describe('useHrToolsNavItems', () => {
...
-  it('hides all items for an ineligible user when not in a development env', async () => {
+  it('hides all items, except partner reminders, for an ineligible user when not in a development env', async () => {
     const { result, waitForNextUpdate } = renderHook(
       () => useHrToolsNavItems(),
       { wrapper: Wrapper },
     );
     await waitForNextUpdate();
 
-    expect(result.current.items).toHaveLength(0);
+    expect(result.current.items).toHaveLength(1);
```
kegrimes, line 59: [Medium] The production kill-switch is untested and the new gate isn't isolated. Two points:

1. `process.env.DISABLE_NEW_REPORTS === 'true'` — the actual prod protection per the PR title — executes in **zero tests** repo-wide.
2. These "only partner reminders" cases use `ineligibleUser` (PartTimeFieldStaff, `staffAccountId: null`), for whom every non-partner item is already hidden by eligibility alone. Mutation check: delete every new `reportsDisabled ||` prefix in the source and these tests still pass — so the core change has no hook-level test that would fail if reverted.

Add a hook test with an **eligible** user (e.g. `MpdGoalEligibleWrapper`) + `DISABLE_NEW_REPORTS='true'` asserting `items` === `['partnerReminders']`. _(Testing agent)_

kegrimes (commented):
## 🤖 Multi-Agent Review — PR #1905

**Verdict: ✅ APPROVED WITH SUGGESTIONS** · 0 blockers · Risk **3/10 (LOW)**
Agents: Security, Architecture, Testing, Standards, UX (standard mode)

The refactor is sound: moving the `reportsDisabled` gate from tab-level to item-level is a net tech-debt reduction (removes duplicated gating between `useNavPages` and `useHrToolsNavItems`), dependency arrays are correct, no consumer breaks, and server-side authorization (`blockImpersonatingNonDevelopers` + API account-list scoping) is unchanged and consistent with all 11 sibling HR Tools pages.

### Medium-priority findings (see inline comments)
1. **`// TODO` without a ticket ref** — load-bearing: it's the only record that `partnerReminders` is intentionally un-gated and must be re-gated post-HCM. _(Standards, Architecture, Testing, UX)_
2. **`reportsDisabled` fails _open_ while the separate `UserOption` query loads** → flash-of-extra-items when reports go live (dormant today under `DISABLE_NEW_REPORTS=true`); also makes the NavMenu unverified test timing-sensitive. _(UX, Testing)_
3. **Prod kill-switch (`DISABLE_NEW_REPORTS=true`) is untested**, and the "only partner reminders" hook tests use an _ineligible_ user, so deleting the new `reportsDisabled ||` prefixes would still pass them — the gate isn't isolated. _(Testing)_

### Suggestions (non-blocking)
- No colocated tests for `useNavPages.tsx` and `useReportsDisabled.ts` (the latter covers finding #3 cheapest at the source).
- Item-level `reportsDisabled ||` diverges from `useReportNavItems`'s conditional-spread — **intentional and correct** (one item must be exempted); noted so it isn't mistaken for inconsistency.
- `hrToolsItems.length === 0` in `hideTab` is now dead post-load — harmless safety net.

**Recommended before merge:** the two cheap wins that pay off at HCM launch — add a ticket to the TODO, and make `reportsDisabled` fail _closed_ (expose `loading`, treat as disabled until known), which kills both the flash and the test flakiness in one change.

src/hooks/useHrToolsNavItems.ts
kegrimes: /dismiss I am not seeing a flash

kegrimes:
@canac Can you glance over this for me? I think I have everything covered since there are certain things that will always be true in production:
- `DISABLE_NEW_REPORTS=true`
- `DEVELOPMENT_ENV=false`
- `user_type_verified="false"` because there is no modal (although there may be a few people who have it set to true when the modal was out accidentally but it should not matter)
- `us_staff_group=null` because there is no HCM data

I'm also wondering if we should add the UserTypeAccess gate back on without requiring the staff account id. This way only US staff can see the report which is what is happening in the nav bar anyway.

src/hooks/useHrToolsNavItems.ts
```diff
@@ -30,63 +33,67 @@ export function useHrToolsNavItems(): {
...
         id: 'pdsGoalCalculator',
         title: t('Paid with Designation Support Goal Calculator'),
-        hideItem: inPdsGoalCalcIneligibleGroup,
+        hideItem: reportsDisabled || inPdsGoalCalcIneligibleGroup,
       },
       {
         id: 'partnerReminders',
         title: t('Ministry Partner Reminders'),
-        hideItem: !developerBypass,
+        // TODO (MPDX-9822): Once HCM goes live, add has no staff account gate back
```
canac, line 90: praise: Thanks for making a TODO for this!

canac (approved):
The logic looks great to me! I assume we're not merging this PR until Scott is ready for it to go out.

kegrimes:
@canac Sorry, did you see this question?

> I'm also wondering if we should add the UserTypeAccess gate back on without requiring the staff account id. This way only US staff can see the report which is what is happening in the nav bar anyway.

canac:
> I'm also wondering if we should add the UserTypeAccess gate back on without requiring the staff account id. This way only US staff can see the report which is what is happening in the nav bar anyway.

Sorry, I did miss that question! Yes, I think you can add the `UserTypeAccess` check back if people's user types are pretty accurate! I know you and Justin were working on fixing some issues with that on the backend. This would mean global staff would see the menu item, but when they go to the page, they will get a limited access message, right?

kegrimes:
> This would mean global staff would see the menu item, but when they go to the page, they will get a limited access message, right?

@canac Actually I've made it where global and non-Cru staff won't see the HR tools tab. Ministry Partner Reminders will only be available to US and hybrid staff members!

EJTison:
@kegrimes Thanks for doing this!  I'll inform our users they can now access this on MPDX :)

PR #1906, opened 2026-07-13
MPDX-9836 Preview changes to new staff goals before saving

canac:
## Description

Previews the recalculated MPD goal in the New Staff Goal Calculator header so coaches and admins can see the impact of unsaved edits before saving.

- Add `MpdGoalPreview` component that renders the `MPD Goal:` figure in the header. On field blur, if the form is dirty and valid, it calls the new `previewNewStaffGoalCalculation` mutation and shows the recalculated total plus the signed difference from the saved goal (e.g. `+$200.00` / `-$75.00`).
- Preview runs on blur (via a document `focusout` listener, since the edited fields are siblings of the header), not on every keystroke, and coalesces rapid blurs with a 500ms debounce.
- While recalculating, the amount is replaced by a spinner and the difference is hidden so no stale number is shown mid-update.
- Falls back to the saved goal (no difference) when the preview request fails; errors are suppressed so a failed preview doesn't surface a snackbar.
- Scenario goals have no account list, so `accountListId` is passed as `null` from `GoalSettingsForm` and omitted from the preview request; the API previews the scenario goal without it.
- Add the `PreviewNewStaffGoalCalculation` mutation to `NewStaffGoalCalculation.graphql`.

Jira: MPDX-9836

## Testing

- Open the New Staff Goal Calculator for a staff goal (HR Tools → Goal Calculator).
- Note the `MPD Goal:` total in the header.
- Edit a goal-affecting field (e.g. requested salary) and tab/click away to blur it.
- Check that the header briefly shows a spinner, then updates to the recalculated total with a signed difference (e.g. `+$200.00`) next to it.
- Make a change that doesn't affect the goal and confirm no difference is shown.
- Save (or cancel) and confirm the difference clears and the header shows the saved goal.
- Repeat for a scenario goal (no account list) and confirm the preview still updates.

## Checklist:

### DIFF
--- a/src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.test.tsx
+++ b/src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.test.tsx
@@ -1,0 +1,265 @@
+import React from 'react';
+import { ThemeProvider } from '@mui/material';
+import { render, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { Field, Form, Formik } from 'formik';
+import { ApolloErgonoMockMap } from 'graphql-ergonomock';
+import { SnackbarProvider } from 'notistack';
+import { GqlMockedProvider } from '__tests__/util/graphqlMocking';
+import theme from 'src/theme';
+import { defaultGoalCalculation } from '../NsGoalCalculatorTestWrapper';
+import { MpdGoalPreview } from './MpdGoalPreview';
+import { PreviewNewStaffGoalCalculationMutation } from './NewStaffGoalCalculation.generated';
+import { calculationToFormValues } from './goalSettingsApiMapping';
+
+const accountListId = 'account-list-1';
+const calculationId = 'goal-calculation-1';
+
+const previewGoalMock = (
+  monthlyGoal: number,
+): {
+  PreviewNewStaffGoalCalculation: PreviewNewStaffGoalCalculationMutation;
+} => ({
+  PreviewNewStaffGoalCalculation: {
+    previewNewStaffGoalCalculation: {
+      newStaffGoalCalculation: {
+        id: calculationId,
+        calculations: { monthlyGoal },
+      },
+    },
+  },
+});
+
+interface TestComponentProps {
+  accountListId?: string | null;
+  savedMonthlyGoal?: number;
+  mocks?: ApolloErgonoMockMap;
+  onCall?: jest.Mock;
+}
+
+const TestComponent: React.FC<TestComponentProps> = ({
+  accountListId: accountListIdProp = accountListId,
+  savedMonthlyGoal = 5000,
+  mocks,
+  onCall,
+}) => (
+  <ThemeProvider theme={theme}>
+    <SnackbarProvider>
+      <GqlMockedProvider<{
+        PreviewNewStaffGoalCalculation: PreviewNewStaffGoalCalculationMutation;
+      }>
+        mocks={mocks}
+        onCall={onCall}
+      >
+        <Formik
+          initialValues={calculationToFormValues(defaultGoalCalculation)}
+          onSubmit={jest.fn()}
+        >
+          <Form>
+            {/* Two fields so tabbing off the first blurs it (focusout). */}
+            <Field
+              name="annualRequestedSalary"
+              type="number"
+              aria-label="Salary"
+            />
+            <Field name="tenure" type="number" aria-label="Tenure" />
+            <MpdGoalPreview
+              accountListId={accountListIdProp}
+              calculationId={calculationId}
+              savedMonthlyGoal={savedMonthlyGoal}
+            />
+          </Form>
+        </Formik>
+      </GqlMockedProvider>
+    </SnackbarProvider>
+  </ThemeProvider>
+);
+
+/** Edit the salary field and blur it (tab away) to commit a preview. */
+const editSalaryAndBlur = (field: HTMLElement, value: string) => {
+  userEvent.clear(field);
+  userEvent.type(field, value);
+  userEvent.tab();
+};
+
+describe('MpdGoalPreview', () => {
+  it('shows the saved goal total when there are no unsaved changes', async () => {
+    const onCall = jest.fn();
+    const { findByText, queryByText } = render(
+      <TestComponent onCall={onCall} />,
+    );
+
+    expect(await findByText('MPD Goal: $5,000.00')).toBeInTheDocument();
+    // No signed difference is shown while the form is untouched.
+    expect(queryByText(/[+-]\$/)).not.toBeInTheDocument();
+    expect(onCall).not.toHaveGraphqlOperation('PreviewNewStaffGoalCalculation');
+  });
+
+  it('does not preview while typing, before the field is blurred', () => {
+    jest.useFakeTimers();
+    try {
+      const onCall = jest.fn();
+      const { getByRole } = render(
+        <TestComponent mocks={previewGoalMock(5200)} onCall={onCall} />,
+      );
+
+      const salary = getByRole('spinbutton', { name: 'Salary' });
+      userEvent.clear(salary);
+      userEvent.type(salary, '80000');
+      // No blur yet: even past the debounce window, typing triggers no preview.
+      jest.advanceTimersByTime(1000);
+
+      expect(onCall).not.toHaveGraphqlOperation(
+        'PreviewNewStaffGoalCalculation',
+      );
+    } finally {
+      jest.useRealTimers();
+    }
+  });
+
+  it('previews the new goal total and a positive difference when an edit raises the goal', async () => {
+    const { findByText, getByRole } = render(
+      <TestComponent mocks={previewGoalMock(5200)} />,
+    );
+
+    editSalaryAndBlur(getByRole('spinbutton', { name: 'Salary' }), '80000');
+
+    expect(await findByText('MPD Goal: $5,200.00')).toBeInTheDocument();
+    expect(await findByText('+$200.00')).toBeInTheDocument();
+  });
+
+  it('shows a negative difference when an edit lowers the goal', async () => {
+    const { findByText, getByRole } = render(
+      <TestComponent mocks={previewGoalMock(4925)} />,
+    );
+
+    editSalaryAndBlur(getByRole('spinbutton', { name: 'Salary' }), '40000');
+
+    expect(await findByText('MPD Goal: $4,925.00')).toBeInTheDocument();
+    expect(await findByText('-$75.00')).toBeInTheDocument();
+  });
+
+  it('hides the goal total and difference while recalculating after another edit', async () => {
+    const { findByText, getByRole, getByText, getByLabelText, queryByText } =
+      render(<TestComponent mocks={previewGoalMock(5200)} />);
+
+    editSalaryAndBlur(getByRole('spinbutton', { name: 'Salary' }), '80000');
+    expect(await findByText('MPD Goal: $5,200.00')).toBeInTheDocument();
+
+    // A further edit starts a new preview. Until it returns, the "MPD Goal:"
+    // label stays but the amount is replaced by a spinner and the difference is
+    // hidden — no stale number on screen.
+    editSalaryAndBlur(getByRole('spinbutton', { name: 'Salary' }), '70000');
+
+    expect(getByLabelText('Calculating new goal total')).toBeInTheDocument();
+    expect(getByText('MPD Goal:')).toBeInTheDocument();
+    expect(queryByText(/MPD Goal: \$/)).not.toBeInTheDocument();
+    expect(queryByText(/[+-]\$/)).not.toBeInTheDocument();
+  });
+
+  it('shows no difference when the change does not affect the goal', async () => {
+    const onCall = jest.fn();
+    const { findByText, queryByText, getByRole } = render(
+      <TestComponent mocks={previewGoalMock(5000)} onCall={onCall} />,
+    );
+
+    editSalaryAndBlur(getByRole('spinbutton', { name: 'Salary' }), '80000');
+
+    // Wait for the preview to run, then confirm the goal is unchanged and no
+    // signed difference is rendered.
+    await waitFor(() =>
+      expect(onCall).toHaveGraphqlOperation('PreviewNewStaffGoalCalculation'),
+    );
+    expect(await findByText('MPD Goal: $5,000.00')).toBeInTheDocument();
+    expect(queryByText(/[+-]\$/)).not.toBeInTheDocument();
+  });
+
+  it('requests a preview with the edited attributes', async () => {
+    const onCall = jest.fn();
+    const { getByRole } = render(
+      <TestComponent mocks={previewGoalMock(5200)} onCall={onCall} />,
+    );
+
+    editSalaryAndBlur(getByRole('spinbutton', { name: 'Salary' }), '99999');
+
+    await waitFor(() =>
+      expect(onCall).toHaveGraphqlOperation('PreviewNewStaffGoalCalculation', {
+        input: {
+          accountListId,
+          id: calculationId,
+          attributes: { annualRequestedSalary: 99999 },
+        },
+      }),
+    );
+  });
+
+  it('previews a scenario goal, omitting the account list', async () => {
+    const onCall = jest.fn();
+    const { findByText, getByRole } = render(
+      <TestComponent
+        accountListId={null}
+        mocks={previewGoalMock(5200)}
+        onCall={onCall}
+      />,
+    );
+
+    editSalaryAndBlur(getByRole('spinbutton', { name: 'Salary' }), '80000');
+
+    // A scenario goal has no account list; the API previews it all the same,
+    // with the account list left out of the request.
+    await waitFor(() =>
+      expect(onCall).toHaveGraphqlOperation('PreviewNewStaffGoalCalculation', {
+        input: {
+          id: calculationId,
+          attributes: { annualRequestedSalary: 80000 },
+        },
+      }),
+    );
+    const previewCall = onCall.mock.calls.find(
+      ([{ operation }]) =>
+        operation.operationName === 'PreviewNewStaffGoalCalculation',
+    );
+    expect(
+      previewCall?.[0].operation.variables.input.accountListId,
+    ).toBeUndefined();
+    expect(await findByText('MPD Goal: $5,200.00')).toBeInTheDocument();
+  });
+
+  it('shows a spinner while the preview is loading', async () => {
+    const { findByLabelText, getByRole } = render(
+      <TestComponent mocks={previewGoalMock(5200)} />,
+    );
+
+    editSalaryAndBlur(getByRole('spinbutton', { name: 'Salary' }), '80000');
+
+    expect(
+      await findByLabelText('Calculating new goal total'),
+    ).toBeInTheDocument();
+  });
+
+  it('falls back to the saved goal when the preview request fails', async () => {
+    const onCall = jest.fn();
+    const { findByText, queryByText, getByRole } = render(
+      <TestComponent
+        onCall={onCall}
+        mocks={
+          {
+            PreviewNewStaffGoalCalculation: {
+              previewNewStaffGoalCalculation: () => {
+                throw new Error('Preview failed');
+              },
+            },
+          } as ApolloErgonoMockMap
+        }
+      />,
+    );
+
+    editSalaryAndBlur(getByRole('spinbutton', { name: 'Salary' }), '80000');
+
+    await waitFor(() =>
+      expect(onCall).toHaveGraphqlOperation('PreviewNewStaffGoalCalculation'),
+    );
+    expect(await findByText('MPD Goal: $5,000.00')).toBeInTheDocument();
+    expect(queryByText(/[+-]\$/)).not.toBeInTheDocument();
+  });
+});
--- a/src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.test.tsx
+++ b/src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.test.tsx
@@ -75,8 +75,8 @@ const TestComponent: React.FC<TestComponentProps> = ({
   </ThemeProvider>
 );
 
-/** Edit the salary field and blur it (tab away) to commit a preview. */
-const editSalaryAndBlur = (field: HTMLElement, value: string) => {
+/** Edit a field and blur it (tab away) to commit a preview. */
+const editAndBlur = (field: HTMLElement, value: string) => {
   userEvent.clear(field);
   userEvent.type(field, value);
   userEvent.tab();
@@ -96,25 +96,38 @@ describe('MpdGoalPreview', () => {
   });
 
   it('does not preview while typing, before the field is blurred', () => {
-    jest.useFakeTimers();
-    try {
-      const onCall = jest.fn();
-      const { getByRole } = render(
-        <TestComponent mocks={previewGoalMock(5200)} onCall={onCall} />,
-      );
+    const onCall = jest.fn();
+    const { getByRole } = render(
+      <TestComponent mocks={previewGoalMock(5200)} onCall={onCall} />,
+    );
 
-      const salary = getByRole('spinbutton', { name: 'Salary' });
-      userEvent.clear(salary);
-      userEvent.type(salary, '80000');
-      // No blur yet: even past the debounce window, typing triggers no preview.
-      jest.advanceTimersByTime(1000);
+    const salary = getByRole('spinbutton', { name: 'Salary' });
+    userEvent.clear(salary);
+    userEvent.type(salary, '80000');
 
-      expect(onCall).not.toHaveGraphqlOperation(
-        'PreviewNewStaffGoalCalculation',
-      );
-    } finally {
-      jest.useRealTimers();
-    }
+    // Nothing is committed until a field blurs, so no preview is scheduled — the
+    // preview is gated on blur, not on keystrokes.
+    expect(onCall).not.toHaveGraphqlOperation('PreviewNewStaffGoalCalculation');
+  });
+
+  it('coalesces rapid blurs across fields into a single preview request', async () => {
+    const onCall = jest.fn();
+    const { getByRole } = render(
+      <TestComponent mocks={previewGoalMock(5200)} onCall={onCall} />,
+    );
+
+    // Tab quickly through two fields within the debounce window.
+    editAndBlur(getByRole('spinbutton', { name: 'Salary' }), '80000');
+    editAndBlur(getByRole('spinbutton', { name: 'Tenure' }), '5');
+
+    await waitFor(() =>
+      expect(onCall).toHaveGraphqlOperation('PreviewNewStaffGoalCalculation'),
+    );
+    const previewCalls = onCall.mock.calls.filter(
+      ([{ operation }]) =>
+        operation.operationName === 'PreviewNewStaffGoalCalculation',
+    );
+    expect(previewCalls).toHaveLength(1);
   });
 
   it('previews the new goal total and a positive difference when an edit raises the goal', async () => {
@@ -122,7 +135,7 @@ describe('MpdGoalPreview', () => {
       <TestComponent mocks={previewGoalMock(5200)} />,
     );
 
-    editSalaryAndBlur(getByRole('spinbutton', { name: 'Salary' }), '80000');
+    editAndBlur(getByRole('spinbutton', { name: 'Salary' }), '80000');
 
     expect(await findByText('MPD Goal: $5,200.00')).toBeInTheDocument();
     expect(await findByText('+$200.00')).toBeInTheDocument();
@@ -133,7 +146,7 @@ describe('MpdGoalPreview', () => {
       <TestComponent mocks={previewGoalMock(4925)} />,
     );
 
-    editSalaryAndBlur(getByRole('spinbutton', { name: 'Salary' }), '40000');
+    editAndBlur(getByRole('spinbutton', { name: 'Salary' }), '40000');
 
     expect(await findByText('MPD Goal: $4,925.00')).toBeInTheDocument();
     expect(await findByText('-$75.00')).toBeInTheDocument();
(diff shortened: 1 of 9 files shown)


src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.tsx
```diff
@@ -0,0 +1,185 @@
...
+  accountListId,
+  calculationId,
+  savedMonthlyGoal,
+}) => {
+  const { t } = useTranslation();
+  const locale = useLocale();
+  const { values, dirty, isValid } = useFormikContext<GoalSettingsFormValues>();
+
+  const [preview, setPreview] = useState<PreviewState | null>(null);
+  const [previewMutation] = usePreviewNewStaffGoalCalculationMutation();
```
canac: [Suggestion] Investigated as a possible CRITICAL cache-clobber and **confirmed safe** — not a bug. The preview payload's inner `newStaffGoalCalculation` has `__typename: 'NewStaffGoalCalculationPreview'` (a distinct type from the query/update's `NewStaffGoalCalculation`), so Apollo normalizes it under a separate key (`NewStaffGoalCalculationPreview:{id}`) and cannot overwrite the saved calculation's cached worksheet. Optional cleanup only: pass `{ fetchPolicy: 'no-cache' }` here to avoid persisting a throwaway preview entity in the cache at all (precedent: `ContactDetailsPartnerAccounts.tsx`).

src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.tsx
```diff
@@ -0,0 +1,185 @@
...
+}
+
+/**
+ * MPD Goal figure in the header. When a field is blurred with unsaved, valid
+ * edits it asks the API to recompute the goal (`previewNewStaffGoalCalculation`)
+ * and shows the new total plus the signed difference from the saved goal
+ * (e.g. `+$200.00` / `-$75.00`) so coaches and admins can see the impact of a
+ * change before saving. Preview runs on blur, not on every keystroke.
+ */
+export const MpdGoalPreview: React.FC<MpdGoalPreviewProps> = ({
```
canac, line 26: [Medium] Separation of concerns: this display component owns ~100 lines of side-effect orchestration (debounced key, ref-mirror, document `focusout` subscription, commit-on-blur state, mutation-firing effect with race handling, error fallback). This feature already extracts such logic into colocated hooks (`useNewStaffGoalCalculation`, `Fields/useGoalSettingsField`). Consider extracting `useMpdGoalPreview({ accountListId, calculationId, savedMonthlyGoal })` returning `{ displayGoal, diff, changed, calculating, diffLabel }`, leaving this component as pure JSX — matches the sibling-hook convention and makes the debounce/commit/race logic unit-testable in isolation.

src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.tsx
```diff
@@ -0,0 +1,185 @@
...
+  }, []);
+
+  // Drop the preview once the form is clean again (saved or cancelled).
+  useEffect(() => {
+    if (!dirty) {
+      setCommittedKey(null);
+    }
+  }, [dirty]);
+
+  const debouncedKey = useDebouncedValue(committedKey, PREVIEW_DEBOUNCE_MS);
```
canac: [Medium] Debounce coalescing is untested — the one behavior the 500ms debounce exists for. No test blurs two fields inside the window and asserts a single `PreviewNewStaffGoalCalculation` fires. The `TestComponent` already renders a second `Tenure` field for exactly this. Without the test, someone could set the debounce to 0 (N requests per tab-through) and every test still passes. Add a test that edits+blurs Salary then Tenure and asserts `previewCalls` has length 1.

src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.tsx
```diff
@@ -0,0 +1,185 @@
...
+    // listen there.
+    const handleFocusOut = () => {
+      setCommittedKey(committableKeyRef.current);
+    };
+    document.addEventListener('focusout', handleFocusOut);
+    return () => document.removeEventListener('focusout', handleFocusOut);
+  }, []);
+
+  // Drop the preview once the form is clean again (saved or cancelled).
+  useEffect(() => {
```
canac: [Medium] The `dirty`-reset path (drop the preview once the form is clean again after save/cancel) has zero coverage. A regression here would leave a stale `+$200.00` on screen after saving. Add a test that previews a change, then reverts the field to its original value, and asserts the delta disappears and the display reverts to the saved goal.

src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.tsx
```diff
@@ -0,0 +1,185 @@
...
+        if (active) {
+          setPreview({
+            attributesKey: debouncedKey,
+            monthlyGoal:
+              data?.previewNewStaffGoalCalculation?.newStaffGoalCalculation
+                .calculations.monthlyGoal ?? null,
+          });
+        }
+      })
+      .catch(() => {
```
canac: [Medium] Silent preview failure is ambiguous UX. On failure the catch sets `monthlyGoal: null`, so the header snaps back to the saved goal with no delta and no error indicator (the `suppressErrors` context intentionally suppresses the global toast too). A user who edited a goal-affecting field then sees no delta cannot distinguish "this change has no goal impact" from "the preview failed." Consider a subtle non-blocking hint (e.g. a `text.secondary` "Preview unavailable" caption) on the failed key. Suppressing the toast for a best-effort preview is otherwise reasonable; saving is unaffected.

src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.tsx
```diff
@@ -0,0 +1,185 @@
...
+  // isn't previewable. Held in a ref so the blur listener reads the latest
+  // without re-subscribing.
+  const committableKeyRef = useRef<string | null>(null);
+  committableKeyRef.current = dirty && isValid ? liveAttributesKey : null;
+
+  // Attributes committed for preview — updated only when a field blurs, so the
+  // preview reflects the last-blurred state rather than every keystroke.
+  const [committedKey, setCommittedKey] = useState<string | null>(null);
+
+  useEffect(() => {
```
canac: [Medium] Novel unscoped global listener: `document.addEventListener('focusout', ...)` is the only such usage in `src/` (verified). It fires for every focusout anywhere in the document. Correct today only because `committableKeyRef.current` is `null` unless `dirty && isValid` and identical keys are deduped, and cleanup is present (no leak). Prefer scoping the listener to a ref-wrapped container around the editable fields, or debounce directly on `liveAttributesKey` to drop the document listener + `committableKeyRef` + `committedKey` machinery entirely (the team chose on-blur deliberately, so container-scoping is the minimal-risk option).

src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  const liveAttributesKey = useMemo(
+    () => JSON.stringify(formValuesToAttributes(values)),
+    [values],
+  );
+  // The attributes key to commit when a field blurs, or `null` when the form
+  // isn't previewable. Held in a ref so the blur listener reads the latest
+  // without re-subscribing.
+  const committableKeyRef = useRef<string | null>(null);
+  committableKeyRef.current = dirty && isValid ? liveAttributesKey : null;
```
canac: [Suggestion] Ref mutated during render (`committableKeyRef.current = ...`). The codebase's ref-mirroring convention keeps this in an effect (`useDebounce.ts:12`, `GoalCalculatorContext.tsx:116`). Deterministic here (derived purely from current render values) so low risk, but writing a ref during render is a discouraged idiom under concurrent renders. Mirror it in a `useEffect`, or fold it into the extracted hook.

src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.tsx
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    let active = true;
+    previewMutation({
+      variables: {
+        input: {
+          // Omitted for a scenario goal that has no account list.
+          accountListId: accountListId ?? undefined,
+          id: calculationId,
+          attributes: JSON.parse(
```
canac: [Suggestion] `JSON.parse(debouncedKey) as NewStaffGoalCalculationAttributesInput` discards type-safety across the stringify/parse boundary, and equality relies on stable key ordering (safe today because `formValuesToAttributes` returns a fixed-order literal, but fragile). Consider keeping the typed attributes object in state alongside a separate string key, and passing the object straight to the mutation instead of round-tripping through JSON.

src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.tsx
```diff
@@ -0,0 +1,185 @@
...
+        },
+      },
+      context: { suppressErrors: true },
+    })
+      .then(({ data }) => {
+        if (active) {
+          setPreview({
+            attributesKey: debouncedKey,
+            monthlyGoal:
+              data?.previewNewStaffGoalCalculation?.newStaffGoalCalculation
```
canac: [Suggestion] Uneven optional chaining: chains through `previewNewStaffGoalCalculation?` but not `newStaffGoalCalculation`. A null `newStaffGoalCalculation` would throw — it lands in the `.catch` and degrades safely (falls back to saved goal), so this is cosmetic. Optional-chain the rest for clarity/intent.

src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.tsx
```diff
@@ -0,0 +1,185 @@
...
+    currencyFormat(Math.abs(diff), 'USD', locale, {
+      showTrailingZeros: true,
+    });
+
+  // While recalculating, swap the amount for a spinner and hide the difference,
+  // so no stale number is on screen mid-update.
+  return (
+    <Stack direction="row" spacing={1} alignItems="center" sx={{ ml: 3 }}>
+      <Typography variant="h6">
+        {calculating
```
canac: [Suggestion] Two i18n keys for one label: `t('MPD Goal:')` (calculating) and `t('MPD Goal: {{amount}}')` (settled). Both are static keys (extraction-safe), but they force translators to keep colon/spacing consistent across two strings. Cleaner: render `t('MPD Goal:')` once and place the amount / spinner / delta as separate adjacent nodes, removing the duplicate key.

src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const settledPreview =
+    committedKey !== null && preview?.attributesKey === committedKey
+      ? preview
+      : null;
+  // A blur committed new edits but their preview hasn't arrived yet.
+  const calculating = committedKey !== null && !settledPreview;
+
+  const previewGoal = settledPreview?.monthlyGoal ?? null;
+  const displayGoal = previewGoal ?? savedMonthlyGoal;
+  const diff =
```
canac: [Suggestion] `Math.round((previewGoal - savedMonthlyGoal) * 100) / 100`: Financial review confirmed this is safe (single same-currency subtraction, rounded once at the display boundary, not aggregation logic). Only a theoretical sub-cent cosmetic: if the server returned sub-cent floats, the raw diff could round to a value one cent off what subtracting the two displayed totals would show — but the two totals are never on screen simultaneously, so not user-visible. No change required; noted for completeness.

src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.test.tsx
```diff
@@ -0,0 +1,265 @@
...
+      <TestComponent onCall={onCall} />,
+    );
+
+    expect(await findByText('MPD Goal: $5,000.00')).toBeInTheDocument();
+    // No signed difference is shown while the form is untouched.
+    expect(queryByText(/[+-]\$/)).not.toBeInTheDocument();
+    expect(onCall).not.toHaveGraphqlOperation('PreviewNewStaffGoalCalculation');
+  });
+
+  it('does not preview while typing, before the field is blurred', () => {
```
canac: [Suggestion] This test is near-vacuous. With `useFakeTimers` + `advanceTimersByTime(1000)` but no blur/`tab()`, `committedKey` stays `null`, `debouncedKey` never becomes non-null, and the effect returns early — the debounce path is never entered, so the assertion passes regardless of debounce logic. It only proves "typing without blur = no request." Either drop the fake timers (proves the same thing without them) or convert it into the real coalescing test (see the debounce comment on the source).

canac (commented):
## Multi-Agent Code Review — PR #1906

**Verdict: APPROVED WITH SUGGESTIONS** — no blockers. 6 agents (Architecture, Testing, Standards, Data Integrity, UX, Financial Reporting), standard mode.

### Risk Assessment
- **Score: 9/10 — CRITICAL** (`.graphql` High-risk +2, three components Medium +1 = 5; 228 non-test lines +2; GraphQL-codegen special +2)
- Files: 6 (+505 / -21). Recommended reviewer: **senior**.

### Resolved debate — the one finding that mattered
The Data Integrity agent raised a **9.0 CRITICAL**: the `previewNewStaffGoalCalculation` mutation is fired without `fetchPolicy: 'no-cache'`, so its partial `calculations { monthlyGoal }` would normalize into the same cache entity as the saved calculation and clobber the ~30-field worksheet + the saved `monthlyGoal`.

**This was investigated and refuted.** The generated types show the preview payload's inner `newStaffGoalCalculation` has `__typename: 'NewStaffGoalCalculationPreview'` (NewStaffGoalCalculation.generated.ts:306) — a **distinct type** from the query/update's `'NewStaffGoalCalculation'` (lines 102, 203). Apollo keys normalized objects by `__typename` + `id`, so the preview lands in a **separate** cache object (`NewStaffGoalCalculationPreview:{id}`) and cannot overwrite `NewStaffGoalCalculation:{id}`. The backend deliberately returns a dedicated preview type to prevent exactly this. The Standards agent's distinct-typename reading was correct. Downgraded to an optional `no-cache` cleanup (sev 3.0, inline).

With that refuted, the highest surviving finding is 6.5 → **no blockers**.

### Themes across agents
- **Money math is sound** (Financial: all checklist Yes/N/A). Goal is server-computed; the only client arithmetic is a single same-currency (USD) subtraction, rounded once at the display boundary. No currency mixing, no float accumulation, null handled (falls back to saved goal).
- **Standards-clean.** Named exports, i18n via `t()` with interpolation vars, generated Apollo types, no `any`/`@ts-ignore`/`new Date()`, no debug output, unused imports correctly relocated out of `GoalSettingsHeader`.
- **Main improvement areas (all non-blocking):** two untested behaviors the code explicitly documents (debounce coalescing, `dirty`-reset), and a separation-of-concerns question (side-effect orchestration + a novel document-level `focusout` listener living in a display component).

### Additional minor findings (not posted inline, informational)
- **[Suggestion 3.5]** `MpdGoalPreview.test.tsx` `TestComponentProps.mocks?: ApolloErgonoMockMap` is untyped, weakening the typed `GqlMockedProvider` generic at the call boundary.
- **[Suggestion 3.5]** UX: the delta uses neutral `text.secondary`, inconsistent with other HrTools deltas that color-code direction (`success.main`/`error.main`). Defensible for a provisional preview; keep the `+`/`-` sign if color is added so color isn't the sole cue.
- **[Suggestion 3.5]** UX: swapping the amount for a 16px spinner on each blur reflows the row (contained — it's the last item in a `flexWrap` row). A min-width on the amount slot avoids the jitter.
- **[Suggestion 3.0]** Testing: no zero-saved-goal case at the `MpdGoalPreview` unit level (only via the header test).
- **[Suggestion 2.5]** `0.005` epsilon at MpdGoalPreview.tsx:143 is a magic number; a named const reads better. (Harmless — diff is already cent-rounded, so it reduces to `!= 0`.)
- **[Pre-existing 3.0]** UX: `Typography variant="h6"` renders a real `<h6>` for a field label (pre-dates this PR, carried into the new component). Adding `component="span"` would fix the heading outline while the file is open.

### Review Summary Table

|-------|----------|------|-----------|--------|-------------|------------|
| Security | — skipped (no api/auth/apollo-link changes) | | | | | |
| Architecture | 0 | 0 | 0 | 2 | 2 | High |
| Data Integrity | 0 (1 refuted) | 0 | 0 | 0 | 2 | High |
| Testing | 0 | 0 | 0 | 2 | 4 | High |
| UX | 0 | 0 | 0 | 1 | 4 | Med-High |
| Standards | 0 | 0 | 0 | 0 | 2 | High |
| Financial | 0 | 0 | 0 | 0 | 1 | High |
| **Total (active)** | **0** | **0** | **0** | **7** | **~12** | |

Note: risk is CRITICAL (9/10) so human senior review is expected regardless of the clean verdict; auto-approval does not apply.

src/components/HrTools/NsGoalCalculator/GoalSettings/useMpdGoalPreview.ts
```diff
@@ -0,0 +1,136 @@
...
+    previewMutation({
+      variables: {
+        input: {
+          // Omitted for a scenario goal that has no account list.
+          accountListId: accountListId ?? undefined,
+          id: calculationId,
+          attributes: debounced,
+        },
+      },
+      context: { suppressErrors: true },
```
canac, line 93: [Medium] **Network errors aren't actually suppressed.** `suppressErrors: true` is honored only by the `graphQLErrors` branch of the global error link (`src/lib/apollo/client.ts:57`, `else if (!suppressErrors)`). The `networkError` branch (`client.ts:62-65`) is ungated — it always calls `snackNotifications.error(networkError.message)` and `dispatch('mpdx-api-error')`.

Since a preview fires on every debounced edit, an offline/5xx/timeout condition produces a red toast **per keystroke-batch** on top of the intended inline "Preview unavailable" — defeating the suppress-and-inline design exactly when it matters. This is the first consumer of `suppressErrors`, so the gap surfaces here.

The fix lives in `client.ts` (out of this PR, affects all ops), so this is a decision to make consciously:
```ts
if (networkError && !suppressErrors) {
  dispatch('mpdx-api-error');
  snackNotifications.error(networkError.message);
}
```
Either make that change or document the known behavior. Note the failure test throws a *GraphQL* error, so the network path is untested.

canac (commented):
## 🤖 Multi-Agent Code Review — ✅ Approved with suggestions

**6 specialized agents** (Architecture, Testing, Standards, Data Integrity, UX, Financial Reporting) + dependency analysis. Risk **8/10 HIGH** (driven by change volume + a new `.graphql` operation across 3 component files — not danger).

**No blockers. Nothing reached the Important (7.0+) or blocker (8.0+) tier.** This is a solid, well-tested feature: clean hook/component split, race-safety independently verified correct, server-authoritative goal math, and 13 test cases. Standards checklist all PASS; Financial checklist all Yes/N/A.

### What to look at before merge (4 medium items)
1. **Network-error snackbar not suppressed** — `useMpdGoalPreview.ts:93`. `suppressErrors` gates only the GraphQL branch in `client.ts:57`; the `networkError` branch (`client.ts:62-65`) still toasts on every debounced edit while offline. Fix is in `client.ts` (out of this PR) — decide whether to fix or document.
2. **`isValid` gate untested** — the test `<Formik>` has no `validationSchema`, so `dirty && isValid` never sees `isValid=false`. Add a dirty-but-invalid case.
3. **Goal figure hidden on first keystroke** — `calculating` goes true before the debounce even fires; the reference total is a spinner throughout editing. Intentional "no stale number" tradeoff — worth reconsidering.
4. **"Preview unavailable" is muted grey** — HrTools convention for inline errors is `error.main`.

### Polish (suggestions)
Stale "blurred" JSDoc (flagged by 4 agents), missing `aria-live` for the recalculated total, the "zero saved goal" test never actually previews a raise, redundant `CENT_EPSILON` after cents-rounding, and manual `+`/`-` sign concat (RTL locales). See inline comments.

### Dependency impact
**No breaking changes** — `GoalSettingsHeader`'s two new required props are satisfied at both render sites; `GoalSettingsForm`'s public contract is unchanged and both page callers remain valid.

---
<sub>Findings with severity < 7 can be dismissed by replying `/dismiss: <reason>` to the inline comment. Auto-approval does not apply (risk is HIGH — human review required).</sub>

src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.test.tsx
```diff
@@ -0,0 +1,311 @@
...
+  <ThemeProvider theme={theme}>
+    <SnackbarProvider>
+      <GqlMockedProvider<{
+        PreviewNewStaffGoalCalculation: PreviewNewStaffGoalCalculationMutation;
+      }>
+        mocks={mocks}
+        onCall={onCall}
+      >
+        <Formik
+          initialValues={calculationToFormValues(defaultGoalCalculation)}
```
canac, line 66: [Medium] **The `isValid` half of the preview gate is never exercised.** This wrapper `<Formik>` has no `validationSchema`, so `isValid` is always `true` — but the real `GoalSettingsForm` uses a schema with `validateOnMount`, so `dirty && isValid` (`useMpdGoalPreview.ts:73`) genuinely toggles in production.

Two production behaviors are unverified: (a) a dirty-but-**invalid** form must NOT fire a preview request (else partial/garbage attributes hit `previewNewStaffGoalCalculation`), and (b) making an already-previewed form invalid must clear the stale preview. Add a case that threads a `validationSchema`, clears a required field, and asserts no `PreviewNewStaffGoalCalculation` fires after the debounce window.

src/components/HrTools/NsGoalCalculator/GoalSettings/useMpdGoalPreview.ts
```diff
@@ -0,0 +1,136 @@
...
+    };
+  }, [debounced, accountListId, calculationId, previewMutation]);
+
+  // The preview matches the current edits once their request has settled.
+  const settledPreview =
+    previewable !== null && preview?.attributes === previewable
+      ? preview
+      : null;
+  // Edits are pending but their preview hasn't arrived yet.
+  const calculating = previewable !== null && settledPreview === null;
```
canac, line 124: [Medium] **The goal figure vanishes on the first keystroke, before any request.** `calculating` becomes true as soon as `previewable` is non-null (`dirty && isValid`) — i.e. on the first keystroke, before the 500ms debounce even starts a request. So the reference total (`$5,000.00`) is replaced by a spinner immediately and stays one through continuous editing, rarely showing a number.

This is an intentional "no stale number" tradeoff (and it's tested), so treat as a design decision to reconsider rather than a defect. Consider keeping the last-known value visible (optionally dimmed) during the debounce window and only showing the spinner once a request is actually in flight.

src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.tsx
```diff
@@ -0,0 +1,85 @@
...
+          })}
+        >
+          {diffLabel}
+        </Typography>
+      )}
+      {!calculating && failed && (
+        <Typography
+          variant="body2"
+          component="span"
+          sx={{ color: 'text.secondary' }}
```
canac, line 74: [Medium] **"Preview unavailable" reads as muted info, not a failure.** It uses `body2` + `color: 'text.secondary'` (grey), while the HrTools convention for inline error text is `color="error"` / `error.main` (e.g. `AdditionalSalaryRequest.tsx:128`, `ElectionTypeQuestion.tsx:121`). A coach can easily miss that the recalculation failed.
```diff
-          sx={{ color: 'text.secondary' }}
+          sx={{ color: 'error.main' }}
```
(Or add a small inline warning icon if a full red treatment feels heavy for a header preview.)

src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.tsx
```diff
@@ -0,0 +1,85 @@
...
+   * account list.
+   */
+  accountListId: string | null;
+  calculationId: string;
+  /** The saved goal total; shown when there are no goal-affecting edits. */
+  savedMonthlyGoal: number;
+}
+
+/**
+ * MPD Goal figure in the header. When a field is blurred with unsaved, valid
```
canac: [Suggestion] **Stale JSDoc.** "When a field is **blurred** with unsaved, valid edits…" — the design no longer gates on blur; it debounces on Formik value changes. The sibling hook says so explicitly (`useMpdGoalPreview.ts:48-49`) and commit `3399aeafa` is exactly this change. The two docs now contradict each other.
```diff
- * MPD Goal figure in the header. When a field is blurred with unsaved, valid
- * edits it shows the recomputed goal total plus the signed difference from the
+ * MPD Goal figure in the header. While the form has unsaved, valid edits it
+ * shows the recomputed goal total plus the signed difference from the
```
(Flagged independently by 4 agents.)

src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.tsx
```diff
@@ -0,0 +1,85 @@
...
+  );
+
+  const diffLabel =
+    (diff > 0 ? '+' : '-') +
+    currencyFormat(Math.abs(diff), 'USD', locale, {
+      showTrailingZeros: true,
+    });
+
+  return (
+    <Stack direction="row" spacing={1} alignItems="center" sx={{ ml: 3 }}>
```
canac, line 44: [Suggestion] **No `aria-live` region for the recalculated result.** The updated total, the signed diff, and "Preview unavailable" all change in place with no live region, and the diff's `aria-label` sits on a roleless `<span>` (unreliably announced). A screen-reader user editing a field gets no feedback that the goal recalculated, what it is, or that it failed. Wrap the changing content in `aria-live="polite"` and/or add a `visuallyHidden` status node (precedent: `Sections/FinancialInformationSection.tsx`). Codebase-wide gap (no `aria-live` in `src`), so this is a pattern improvement rather than a convention violation.

src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.test.tsx
```diff
@@ -0,0 +1,311 @@
...
+
+    // Undo the edit → the form is pristine again → the preview is dropped and
+    // the header returns to the saved goal with no stale difference.
+    editField(salary, original);
+
+    expect(await findByText('MPD Goal: $5,000.00')).toBeInTheDocument();
+    expect(queryByText(/[+-]\$/)).not.toBeInTheDocument();
+  });
+
+  it('previews a raise from a zero saved goal', async () => {
```
canac, line 210: [Suggestion] **This test never previews a raise.** It renders `savedMonthlyGoal={0}` with `previewGoalMock(200)` but edits no field, so the form stays pristine and no preview runs — it only asserts the pristine `$0.00` zero-state (which is valuable). The `0 → +$200.00` path implied by the name is unverified. Add an `editField(...)` so the raise-from-zero diff is exercised, or rename the test to reflect that it covers the pristine zero-state only.

src/components/HrTools/NsGoalCalculator/GoalSettings/useMpdGoalPreview.ts
```diff
@@ -0,0 +1,136 @@
...
+  const failed = settledPreview !== null && settledPreview.monthlyGoal === null;
+
+  const previewGoal = settledPreview?.monthlyGoal ?? null;
+  const displayGoal = previewGoal ?? savedMonthlyGoal;
+  const diff =
+    previewGoal === null
+      ? 0
+      : Math.round((previewGoal - savedMonthlyGoal) * 100) / 100;
+  const changed = Math.abs(diff) >= CENT_EPSILON;
```
canac, line 133: [Suggestion] **`CENT_EPSILON` is redundant after the cents-round.** `diff` is already snapped to whole cents on the line above (`Math.round(... * 100) / 100`), so its smallest non-zero magnitude is `0.01` and `>= 0.005` reduces to `diff !== 0`. Keep the epsilon **or** the pre-round, not both — e.g. drop the epsilon and compare `diff !== 0`. (Rounding in the hook is otherwise fine here since the value drives both the label and the `changed` gate.)

src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.tsx
```diff
@@ -0,0 +1,85 @@
...
+  savedMonthlyGoal,
+}) => {
+  const { t } = useTranslation();
+  const locale = useLocale();
+  const { calculating, displayGoal, diff, changed, failed } = useMpdGoalPreview(
+    { accountListId, calculationId, savedMonthlyGoal },
+  );
+
+  const diffLabel =
+    (diff > 0 ? '+' : '-') +
```
canac, line 38: [Suggestion] **Manual sign placement isn't locale-aware.** `(diff > 0 ? '+' : '-') + currencyFormat(Math.abs(diff), 'USD', locale, …)` prepends the sign, so in RTL locales or locales with a trailing currency symbol the `+`/`-` may land in the wrong position. `Intl.NumberFormat`'s `signDisplay: 'exceptZero'` places the sign per-locale — consider extending `currencyFormat` with a `signDisplay` option rather than concatenating.

src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.test.tsx
zweatshirt: Kind of agree with this suggestion

src/components/HrTools/NsGoalCalculator/GoalSettings/MpdGoalPreview.test.tsx
```diff
@@ -0,0 +1,370 @@
...
+          accountListId,
+          id: calculationId,
+          attributes: { annualRequestedSalary: 99999 },
+        },
+      }),
+    );
+  });
+
+  it('previews a scenario goal, omitting the account list', async () => {
+    const onCall = jest.fn();
```
zweatshirt: onCall is initialized in multiple assertion blocks, do we want to move it to the top of the file?

zweatshirt (approved):
This is great. Thank for making this change, looks like it was fun to implement

PR #1908, opened 2026-07-14
[MPDX-9654] - Show stopped recurring transfers in Transfer History

wjames111:
## Description

- Jira: [MPDX-9654](https://jira.cru.org/browse/MPDX-9654) — stopped recurring fund transfers were disappearing from the Transfer History section; they need to remain visible with a stopped status
- Stopping a recurring transfer soft-deletes it in SAA (`active: false`). The SAA report endpoint now returns stopped recurring transfers in history (SAA-646), but the frontend never used the `active` flag, so a stopped transfer was labeled "Ongoing" with live stop/edit actions
- Adds a `Stopped` status to `StatusEnum`, derived in `getStatusLabel`: a recurring transfer whose end date has passed is still "Ended"; otherwise inactive means "Stopped" and active means "Ongoing"
- Renders the Stopped status as a gray chip (same terminal-state styling as Ended); row actions were already hidden for non-Pending/non-Ongoing statuses, so Stopped rows get no stop/edit buttons
- Bounds the missed-month scan in `filterTransfers` at the last real transaction for transfers stopped before their end date, so months after the stop don't accumulate as false "failed" transfers. Transfers that ended naturally keep the old scan window (regression-tested)
- Note: the API doesn't expose *when* a transfer was stopped, so a month that genuinely failed between the last successful transaction and the stop won't be flagged — chose no-false-positives over that edge case

## Testing

- Go to the Staff Savings Fund Transfer screen (impersonate a staff member who has a recurring transfer that has already fired at least once)
- Confirm the recurring transfer's past transactions show in Transfer History with status "ongoing"
- Stop the recurring transfer via the stop action on the history row
- Check that the transfer remains in Transfer History with a gray "stopped" chip
- Check that the stopped row has no stop or edit-end-date actions
- Check that the failed-transfers badge on the row does not count months after the stop
- Unit coverage: `yarn test getStatus.test.ts filterTransfers.test.ts TransfersPage.test.tsx`

## Checklist:

### DIFF
--- a/src/components/HrTools/SavingsFundTransfer/Helper/filterTransfers.ts
+++ b/src/components/HrTools/SavingsFundTransfer/Helper/filterTransfers.ts
@@ -62,13 +62,22 @@ export function filteredTransfers(transfers: Transactions[]) {
     const transferRow = filtered[index];
 
     const currentDate = DateTime.local().startOf('day');
-    const start = transferRow.recurringTransfer?.recurringStart.startOf('day');
-    const end = DateTime.min(
-      (transferRow.recurringTransfer?.recurringEnd ?? currentDate).startOf(
-        'day',
-      ),
-      currentDate,
-    );
+    const recurring = transferRow.recurringTransfer;
+    const start = recurring?.recurringStart.startOf('day');
+    const recurringEnd = recurring?.recurringEnd?.startOf('day') ?? null;
+    let end: DateTime = DateTime.min(recurringEnd ?? currentDate, currentDate);
+
+    // A transfer stopped before its end date creates no transactions after the
+    // stop, so only look for missed months through its last actual transaction.
+    const stoppedEarly =
+      recurring?.active === false &&
+      (!recurringEnd || recurringEnd > currentDate);
+    if (stoppedEarly) {
+      const lastTransactedAt = DateTime.max(
+        ...[...transactions.values()].map((tx) => tx.transaction!.transactedAt),
+      ).startOf('day');
+      end = DateTime.min(end, lastTransactedAt);
+    }
 
     if (!start) {
       continue;
--- a/src/components/HrTools/SavingsFundTransfer/Helper/getStatus.test.ts
+++ b/src/components/HrTools/SavingsFundTransfer/Helper/getStatus.test.ts
@@ -112,4 +112,40 @@ describe('createTable', () => {
     const status = mockData[3];
     expect(getStatusLabel(status)).toBe('ongoing');
   });
+
+  it('should return stopped status if inactive with no end date', () => {
+    const transfer: Transactions = {
+      ...mockData[2],
+      recurringTransfer: {
+        ...mockData[2].recurringTransfer!,
+        recurringEnd: null,
+        active: false,
+      },
+    };
+    expect(getStatusLabel(transfer)).toBe('stopped');
+  });
+
+  it('should return stopped status if inactive with a future end date', () => {
+    const transfer: Transactions = {
+      ...mockData[3],
+      recurringTransfer: {
+        ...mockData[3].recurringTransfer!,
+        recurringEnd: DateTime.fromISO('2024-12-31'),
+        active: false,
+      },
+    };
+    expect(getStatusLabel(transfer)).toBe('stopped');
+  });
+
+  it('should return ended status if inactive with an end date in the past', () => {
+    const transfer: Transactions = {
+      ...mockData[1],
+      recurringTransfer: {
+        ...mockData[1].recurringTransfer!,
+        recurringEnd: DateTime.fromISO('2023-01-01'),
+        active: false,
+      },
+    };
+    expect(getStatusLabel(transfer)).toBe('ended');
+  });
 });
--- a/src/components/HrTools/SavingsFundTransfer/Helper/getStatus.test.ts
+++ b/src/components/HrTools/SavingsFundTransfer/Helper/getStatus.test.ts
@@ -148,4 +148,16 @@ describe('createTable', () => {
     };
     expect(getStatusLabel(transfer)).toBe('ended');
   });
+
+  it('should return ended status if inactive with an end date of today', () => {
+    const transfer: Transactions = {
+      ...mockData[1],
+      recurringTransfer: {
+        ...mockData[1].recurringTransfer!,
+        recurringEnd: DateTime.fromISO('2023-12-15'),
+        active: false,
+      },
+    };
+    expect(getStatusLabel(transfer)).toBe('ended');
+  });
 });
--- a/src/components/HrTools/SavingsFundTransfer/Helper/getStatus.test.ts
+++ b/src/components/HrTools/SavingsFundTransfer/Helper/getStatus.test.ts
@@ -137,7 +137,7 @@ describe('createTable', () => {
     expect(getStatusLabel(transfer)).toBe('stopped');
   });
 
-  it('should return ended status if inactive with an end date in the past', () => {
+  it('should return ended status when the end date is in the past, even if inactive', () => {
     const transfer: Transactions = {
       ...mockData[1],
       recurringTransfer: {
--- a/src/components/HrTools/SavingsFundTransfer/Helper/getStatus.ts
+++ b/src/components/HrTools/SavingsFundTransfer/Helper/getStatus.ts
@@ -6,12 +6,14 @@ export function getStatusLabel(history: Transactions): StatusEnum {
     return StatusEnum.Complete;
   }
 
-  const endDate = history.recurringTransfer.recurringEnd;
+  const { recurringEnd, active } = history.recurringTransfer;
   const today = DateTime.now().endOf('day');
 
-  if (!endDate) {
-    return StatusEnum.Ongoing;
+  if (recurringEnd && recurringEnd <= today) {
+    return StatusEnum.Ended;
   }
 
-  return endDate > today ? StatusEnum.Ongoing : StatusEnum.Ended;
+  // An inactive recurring transfer that hasn't reached its end date was
+  // stopped manually rather than running its course.
+  return active ? StatusEnum.Ongoing : StatusEnum.Stopped;
 }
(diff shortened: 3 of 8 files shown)


src/components/HrTools/SavingsFundTransfer/Helper/getStatus.test.ts
```diff
@@ -112,4 +112,40 @@ describe('createTable', () => {
...
+      recurringTransfer: {
+        ...mockData[3].recurringTransfer!,
+        recurringEnd: DateTime.fromISO('2024-12-31'),
+        active: false,
+      },
+    };
+    expect(getStatusLabel(transfer)).toBe('stopped');
+  });
+
+  it('should return ended status if inactive with an end date in the past', () => {
```
wjames111: [Medium] **Test name doesn't match what it exercises.** This test sets `active: false`, but `getStatusLabel` returns `Ended` from the past-end-date branch *before* reading `active` — it passes identically with `active: true`, so the inactive path isn't actually covered here. The invariant it really proves is "a past end date takes precedence over stopped." Recommend renaming to make that explicit, e.g.:

```ts
it('should return ended status when the end date is in the past, even if inactive', () => {
```

(Flagged by: Testing agent)

src/components/HrTools/SavingsFundTransfer/Helper/getStatus.ts
```diff
@@ -6,12 +6,14 @@ export function getStatusLabel(history: Transactions): StatusEnum {
     return StatusEnum.Complete;
   }
 
-  const endDate = history.recurringTransfer.recurringEnd;
+  const { recurringEnd, active } = history.recurringTransfer;
   const today = DateTime.now().endOf('day');
 
-  if (!endDate) {
-    return StatusEnum.Ongoing;
+  if (recurringEnd && recurringEnd <= today) {
```
wjames111, line 12: [Medium] **Missing boundary test: `recurringEnd` exactly today.** The new branch structure has an untested boundary: an inactive transfer whose end date is *today* reads as `Ended` (this branch fires first), not `Stopped` — and in `filterTransfers.ts` the strict `recurringEnd > currentDate` makes `stoppedEarly` false for the same input, so the two files agree. Nothing pins that behavior. Suggested additions (mocked now is `2023-12-15` in getStatus.test.ts, `2024-01-15` in filterTransfers.test.ts):

```ts
it('should return ended status when an inactive transfer ends exactly today', () => {
  const transfer: Transactions = {
    ...mockData[3],
    recurringTransfer: {
      ...mockData[3].recurringTransfer!,
      recurringEnd: DateTime.utc(2023, 12, 15), // == mocked now
      active: false,
    },
  };
  expect(getStatusLabel(transfer)).toBe('ended');
});
```

plus a `filterTransfers` case with `recurringEnd: '2024-01-15'` asserting the scan runs through the end date (not bounded to the last transaction).

(Flagged by: Testing agent; boundary consistency verified by Financial agent and debate adjudication)

src/components/HrTools/SavingsFundTransfer/Helper/filterTransfers.ts
```diff
@@ -62,13 +62,22 @@ export function filteredTransfers(transfers: Transactions[]) {
...
-      currentDate,
-    );
+    const recurring = transferRow.recurringTransfer;
+    const start = recurring?.recurringStart.startOf('day');
+    const recurringEnd = recurring?.recurringEnd?.startOf('day') ?? null;
+    let end: DateTime = DateTime.min(recurringEnd ?? currentDate, currentDate);
+
+    // A transfer stopped before its end date creates no transactions after the
+    // stop, so only look for missed months through its last actual transaction.
+    const stoppedEarly =
```
wjames111, line 72: [Suggestion] **Duplicated "stopped" predicate with different zone normalization.** This `stoppedEarly` predicate re-derives the same business rule as `getStatusLabel` (`getStatus.ts:12-18`) with a different strategy: here `recurringEnd.startOf('day')` (in the value's *own* zone) vs local `startOf('day')`; there raw `recurringEnd` vs local `endOf('day')`. Debate adjudication confirmed the two agree for all same-zone inputs (including end-date-today), but a `recurringEnd` carrying a far-from-local offset (parsed with `setZone: true`) that straddles the local day seam can label a row **Ended** while bounding its scan as if **Stopped**. Low probability, but it's the classic drift the repo rules warn about for duplicated calculation logic.

Suggested direction: extract a shared `isStoppedEarly(recurring)` (natural home: `getStatus.ts`), normalize `recurringEnd` to local zone before day-boundary comparisons in both call sites, and add one cross-zone boundary test.

(Flagged by: Architecture agent at 4.5; adjudicated to 3.0 after the same-zone divergence example was refuted)

src/components/HrTools/SavingsFundTransfer/Helper/filterTransfers.ts
```diff
@@ -62,13 +62,22 @@ export function filteredTransfers(transfers: Transactions[]) {
...
+      recurring?.active === false &&
+      (!recurringEnd || recurringEnd > currentDate);
+    if (stoppedEarly) {
+      const lastTransactedAt = DateTime.max(
```
wjames111, line 76: [Suggestion] **Cross-timezone `startOf('day')` on transaction dates.** `transactedAt` is parsed with `setZone: true` (TransfersPage.tsx), so `lastTransactedAt.startOf('day')` is computed in the transaction's own zone while `end` descends from `DateTime.local().startOf('day')`. Luxon compares absolute instants, so an offset-straddling transaction can shift the scan boundary by a day in rare cases. This matches the pre-existing zone handling in this file (`seenMonths` keys already use own-zone `.year`/`.month`), so it's consistent — not a regression. If tightened later, normalize both sides to one zone before `startOf('day')`.

(Flagged by: Financial Reporting agent)

src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.test.tsx
```diff
@@ -254,6 +277,28 @@ describe('TransfersPage', () => {
     expect(within(tables[0]).getAllByRole('columnheader')).toHaveLength(8);
   });
 
+  it('should display stopped recurring transfers in history with a stopped status and no actions', async () => {
```
wjames111, line 280: [Suggestion] **Add one positive assertion to the stopped-row test.** The test finds the row via `queryByText('stopped')` and asserts three action buttons are absent — solid negative coverage, but a row that rendered *nothing except* the word "stopped" would still pass. One positive assertion (e.g. the $300.00 amount or the Oct 5, 2023 transfer date visible within the row) would catch a refactor that drops the row's content.

(Flagged by: Testing agent)

wjames111 (commented):
## Multi-Agent Code Review

**Verdict: APPROVED WITH SUGGESTIONS** — no blockers; 2 medium-severity test improvements recommended.

**Risk:** 4/10 MEDIUM · 7 files, +178/−11 (~40 non-test lines) · Mode: standard (5 specialized agents + gap review + debate adjudication)

### Summary
- **No critical, high, or important findings.** The Stopped-status derivation, chip styling, action gating, and missed-month scan bounding were all verified against consumers (table, print, CSV, FailedTransferModal) by the UX, Financial, Standards, and dependency-impact agents.
- **Dependency impact:** purely additive — the single exhaustive `Record<StatusEnum, …>` (`chipStyle`) was updated in this PR; no switches or positional uses of the enum exist; no breaking changes.
- **Financial checklist:** money arithmetic safe (none added), USD-only, rounding at display boundary only, null amounts handled, Luxon throughout, no client-side sums over paginated data.
- **Debate note:** an Architecture claim that `getStatus.ts` and `filterTransfers.ts` disagree on the "end date is today" boundary was adjudicated and **refuted** for same-zone inputs (strict `>` on day-normalized dates keeps them consistent); a narrower cross-zone day-seam divergence is real but low-probability — captured as a 3.0 suggestion inline.

### Suggestions not posted inline (informational, severity ≤ 3.5)
- `filterTransfers.ts:76-79` — the `stoppedEarly` bounding work runs before the `if (!start) continue` guard (dead work in a rare case); optionally move it below the guard.
- `filterTransfers.test.ts:198` — the stopped-transfer mock reuses `id: '3'`; a distinct id like `'stopped-recurring'` would read clearer (cosmetic).
- `filterTransfers.ts:77` — `tx.transaction!` was verified provably safe (map entries only created behind the `!transfer.transaction` guard) and matches existing convention; informational only.

### Pre-existing (not this PR)
- Status chip labels render un-localized enum values feature-wide (all 6 statuses, `createTableRow.tsx:163`, also print + CSV). Consider a follow-up i18n ticket for the feature.

---
Medium findings (severity < 7) can be dismissed by replying `/dismiss: <reason>` on the inline comment.

src/components/HrTools/SavingsFundTransfer/Helper/getStatus.test.ts
wjames111: 🤖 Fixed in 70ad296: renamed the test to 'should return ended status when the end date is in the past, even if inactive' so the name states the precedence invariant it actually proves.

src/components/HrTools/SavingsFundTransfer/Helper/getStatus.ts
wjames111: 🤖 Fixed in e53ed39: added boundary tests pinning end-date-equals-today behavior — getStatus.test.ts asserts an inactive transfer ending today reads 'ended', and filterTransfers.test.ts asserts the missed-month scan runs through the end date (not bounded to the last transaction).

src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.test.tsx
wjames111: 🤖 Fixed in 6c35668: the stopped-row test now also asserts the row renders its $300.00 amount and Oct 5, 2023 transfer date, not just the status text.

PR #1909, opened 2026-07-14
[MPDX-9655] - Allow one time Savings Fund Transfers to be future dated

wjames111:
## Description

- Replaces the disabled (grayed-out) Transfer Date field on one-time Savings Fund Transfers with an editable date picker, so users can schedule a one-time transfer for today or a future date
- Sends the selected date to the API via the new required `transactedAt` argument on the `CreateTransfer` mutation
- Adds a one-time-specific validation message ("Transfer date cannot be in the past") — previously a past date showed the recurring-transfer error message
- Adds tests covering the enabled date field, future-date submission (including `transactedAt` in the mutation variables), and past-date validation

**Dependent on:** CruGlobal/mpdx_api#3342 (adds the required `transactedAt` argument to `TransferCreateMutation`). GraphQL codegen in CI will fail until that PR is deployed to the staging API. Note also that since `transacted_at` is required on the API side, deploying the API change before this PR would break one-time transfers on the current UI — release order needs coordination.

Jira: [MPDX-9655](https://jira.cru.org/browse/MPDX-9655)

## Testing

- Go to HR Tools → Staff Saving Fund → Transfers
- Click a balance card to open the New Fund Transfer modal
- With the "One Time" schedule selected, check that the Transfer Date field is enabled (no longer grayed out) and defaults to today
- Pick a future date, fill in To Account, Amount, and Note, and submit — check that the transfer is created with the selected date
- Pick a past date and check that the "Transfer date cannot be in the past" validation error appears and Submit is disabled
- Switch to "Monthly" and check that the recurring transfer date/end date behavior is unchanged

## Checklist:

### DIFF
--- a/src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
@@ -469,6 +469,56 @@ describe('TransferModal', () => {
       ).toBeInTheDocument();
     });
 
+    it('should require a transfer date for one-time transfers when cleared', async () => {
+      const { getByRole, getByLabelText, findByText } = render(<Components />);
+
+      expect(getByRole('radio', { name: /one time/i })).toBeChecked();
+
+      const transferDate = getByLabelText(/transfer date/i);
+
+      userEvent.clear(transferDate);
+      expect(transferDate).toHaveValue('');
+      userEvent.tab();
+
+      expect(await findByText('Transfer date is required')).toBeInTheDocument();
+
+      await waitFor(() =>
+        expect(getByRole('button', { name: /submit/i })).toBeDisabled(),
+      );
+    });
+
+    it('should not submit a one-time transfer with an unparseable transfer date', async () => {
+      const { getByRole, getByLabelText, findByText } = render(<Components />);
+
+      const amountField = getByRole('spinbutton', { name: /amount/i });
+      const noteField = getByRole('textbox', { name: /note/i });
+
+      userEvent.click(getByRole('combobox', { name: /to account/i }));
+      userEvent.click(getByRole('option', { name: /staff savings/i }));
+
+      userEvent.clear(amountField);
+      userEvent.type(amountField, '100');
+
+      userEvent.type(noteField, 'Test note');
+
+      const transferDate = getByLabelText(/transfer date/i);
+
+      userEvent.clear(transferDate);
+      userEvent.type(transferDate, '99/99/9999');
+      expect(transferDate).toHaveValue('99/99/9999');
+      userEvent.tab();
+
+      expect(
+        await findByText('Transfer date cannot be in the past'),
+      ).toBeInTheDocument();
+
+      await waitFor(() =>
+        expect(getByRole('button', { name: /submit/i })).toBeDisabled(),
+      );
+
+      expect(mutationSpy).not.toHaveGraphqlOperation('CreateTransfer');
+    });
+
     it('should show error message when monthly schedule is selected', async () => {
       const { getByRole, getByLabelText, findByText } = render(<Components />);
 
--- a/src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
@@ -431,6 +431,44 @@ describe('TransferModal', () => {
       ).toBeInTheDocument();
     });
 
+    it('should allow editing the transfer date for one-time transfers', async () => {
+      const { getByRole, getByLabelText, queryByText } = render(<Components />);
+
+      expect(getByRole('radio', { name: /one time/i })).toBeChecked();
+
+      const transferDate = getByLabelText(/transfer date/i);
+      expect(transferDate).toBeEnabled();
+      expect(transferDate).toHaveValue('1/1/2020');
+
+      userEvent.clear(transferDate);
+      userEvent.type(transferDate, '02/15/2020');
+      expect(transferDate).toHaveValue('02/15/2020');
+      userEvent.tab();
+
+      await waitFor(() =>
+        expect(
+          queryByText('Transfer date cannot be in the past'),
+        ).not.toBeInTheDocument(),
+      );
+    });
+
+    it('should show error message when one-time transfer date is in the past', async () => {
+      const { getByRole, getByLabelText, findByText } = render(<Components />);
+
+      expect(getByRole('radio', { name: /one time/i })).toBeChecked();
+
+      const transferDate = getByLabelText(/transfer date/i);
+
+      userEvent.clear(transferDate);
+      userEvent.type(transferDate, '12/31/2019');
+      expect(transferDate).toHaveValue('12/31/2019');
+      userEvent.tab();
+
+      expect(
+        await findByText('Transfer date cannot be in the past'),
+      ).toBeInTheDocument();
+    });
+
     it('should show error message when monthly schedule is selected', async () => {
       const { getByRole, getByLabelText, findByText } = render(<Components />);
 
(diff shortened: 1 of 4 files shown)


src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
```diff
@@ -431,6 +431,44 @@ describe('TransferModal', () => {
       ).toBeInTheDocument();
     });
 
+    it('should allow editing the transfer date for one-time transfers', async () => {
```
wjames111, line 434: [Medium] **Date-edge cases untested on the newly-editable field** (Testing + Data Integrity, supported by Architecture & UX after debate)

The field went from `disabled` to an editable, clearable, required input, but no test covers:

1. **Cleared/empty date** — every `userEvent.clear(transferDate)` in the suite is immediately followed by a `type(...)`. Nothing asserts that clearing the date surfaces "Transfer date is required" and disables submit. The runtime behavior was verified correct during review, so this is a regression-exposure gap, not a bug — but a dropped `.required()` would currently be caught by nothing.
2. **Invalid/garbage date** — takes a *different* yup branch: an invalid Luxon DateTime is truthy, slips past `required()`, and is blocked only by the `start-date` comparison (`selected >= minimum` is false for invalid dates).

Suggested addition:

```tsx
it('should require a transfer date for one-time transfers when cleared', async () => {
  const { getByRole, getByLabelText, findByText } = render(<Components />);
  const transferDate = getByLabelText(/transfer date/i);
  userEvent.clear(transferDate);
  userEvent.tab();
  expect(await findByText('Transfer date is required')).toBeInTheDocument();
  await waitFor(() =>
    expect(getByRole('button', { name: /submit/i })).toBeDisabled(),
  );
});
```

plus a variant typing an unparseable date and asserting submit stays disabled. (Minor completeness add, conceded to 3.5 in debate: a user-*typed* today's date is also unasserted, though the equality boundary is already exercised by the default-submit test and typed parsing by the future-date test.)

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
```diff
@@ -219,6 +222,7 @@ export const TransferModal: React.FC<TransferModalProps> = ({
             sourceFundTypeName: transferFrom,
             destinationFundTypeName: transferTo,
             description: note.trim(),
+            transactedAt: convertedTransferDate,
```
wjames111, line 222: [Suggestion] **Type-honesty + guarded serialization** (Data Integrity, severity revised down 6.5 → ~4.0 in debate after the crash was proven unreachable)

`transferDate` is typed non-nullable (`DateTime<boolean>`, line 47), but the now-editable picker can emit `null` (clear / empty blur) or an invalid DateTime, and `convertedTransferDate = transferDate.toISO() ?? ''` (line 193) now feeds this required `ISO8601DateTime!` argument. The crash/empty-string paths are unreachable today — Formik's `submitForm` re-validates and never calls `onSubmit` with a validation error (note: `!isValid` button gating alone would *not* protect the first render, since `isValid` starts `true` without `validateOnMount`).

Hardening suggestion so correctness doesn't depend on validation wiring:

```tsx
// interface
transferDate: DateTime<boolean> | null;

// handleSubmit
const convertedTransferDate =
  transferDate && transferDate.isValid ? transferDate.toISO() : '';
```

(or throw instead of `''` so a future gating regression fails loudly rather than sending an empty string to the API). Verified in debate that widening the type does not ripple into the recurring branch — the `end>start` test already null-guards and `initialValues` defaults to `getToday()`.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
```diff
@@ -104,9 +104,12 @@ const transferSchema = (locale: string) =>
         }
 
         return this.createError({
-          message: i18n.t(
-            'Recurring transfers must start at least one day in the future',
-          ),
+          message:
+            schedule === ScheduleEnum.OneTime
```
wjames111: [Suggestion] **Three message-producing branches in one yup test** (Architecture)

The `start-date` test now carries `isEditing` → "earlier than {date}", one-time → "cannot be in the past", and recurring → "one day in the future", interleaved with the schedule-dependent `minimum` computation. Logic is sound and test-verified; optional extraction of `minimumTransferDate(schedule)` / `pastDateMessage(schedule)` helpers would keep it readable as more schedule types (the enum already has `Annually`) arrive. Not required for merge.

One behavioral note from debate worth confirming intentionally: `isEditing` short-circuits *before* the one-time/recurring branch, so an edited transfer never sees the new "cannot be in the past" message — moot today since only recurring transfers are editable, but worth remembering if one-time editing is ever added.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
```diff
@@ -455,11 +459,21 @@ export const TransferModal: React.FC<TransferModalProps> = ({
                   <Grid container spacing={2}>
                     {schedule === ScheduleEnum.OneTime ? (
                       <Grid size={12}>
-                        <TextField
-                          fullWidth
+                        <CustomDateField
```
wjames111, line 478: [Suggestion] **Two low-cost follow-ups on the new date field** (Architecture + UX)

1. This `CustomDateField` block is now byte-identical to the recurring branch's transferDate field (only the Grid `size` differs). Optionally hoist a single field above the conditional to avoid maintaining the wiring twice.
2. The calendar doesn't constrain past dates — `disablePast`/`minDate` would gray out invalid days instead of erroring after the fact. `CustomDateField`'s desktop path can't honor those props today (they'd be silently dropped), so this is follow-up-ticket material for the shared component (would benefit the recurring branch too), not this PR.

UX notes from review (no action needed): desktop surfaces the required-error only on blur (brief "disabled submit, no visible reason" window — consistent with the existing form model); mobile's explicit Clear action surfaces it immediately.

wjames111 (commented):
## Multi-Agent Code Review (standard mode)

**Verdict: APPROVED WITH SUGGESTIONS** — 0 blockers, 1 medium-priority finding, 7 informational suggestions.

**Risk:** 5/10 (MEDIUM) — mutation contract change with a cross-repo deploy dependency (CruGlobal/mpdx_api#3342, already documented in the PR description).

**Dependency impact:** LOW — `TransferModal` has one production render site; `CreateTransfer` has exactly one call site, updated in this PR. No breaking changes.

**Agents:** Architecture, Testing, Data Integrity, UX, Standards, Financial Reporting (domain) + gap review + dependency analysis, with cross-examination debate and rebuttal rounds. Standards checklist: full PASS. Debate rounds *downgraded* two findings with evidence: the initially-flagged "latent null-date crash" was proven unreachable (Formik's `submitForm` validates before `onSubmit`), and the today-boundary was shown to already be exercised end-to-end by the default-submit test.

**Timezone adjudication (asked by 4 agents, resolved with evidence):** `transferDate.toISO()` serializes local start-of-day *with* the local offset, and the backend takes `.to_date` in that offset — the user's chosen calendar date survives in every timezone. The default-today path is not a regression vs the old server-side default.

---

**[Medium] Downstream display of future-dated one-time transfers** — flagged by the Financial Reporting agent, sharpened in debate
- This PR is the first code path that lets a one-time transfer carry a future `transactedAt`, which *activates* latent assumptions in pre-existing files: `src/components/HrTools/SavingsFundTransfer/Helper/getStatus.ts:5-6` returns **Complete** for any non-recurring row regardless of date; `Helper/filterTransfers.ts` routes any row with a transaction to History; `TransfersPage/TransfersPage.tsx:186-199` hardcodes Upcoming rows as Monthly (a one-time row landing there would render amount 0 / undefined date).
- Which of these manifests depends on whether the backend materializes the transaction immediately or at the transacted date.
- Recommended action: confirm backend behavior with the mpdx_api team and file a follow-up ticket for the display pipeline. Informational — does not block this PR.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
wjames111: 🤖 Fixed in 8900f16: added tests for cleared/empty date (required error + disabled submit) and invalid date (start-date branch, submit blocked, mutation not fired).

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
wjames111: 🤖 Fixed in f6d7602: widened transferDate to DateTime<boolean> | null and guarded the ISO serialization with an isValid check.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
wjames111: 🤖 Fixed in 62aa3ad: extracted minimumTransferDate(schedule) and pastDateMessage(schedule) module-scope helpers (static i18n keys preserved); validation behavior unchanged.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
wjames111: 🤖 Fixed in 3135265: hoisted the duplicated Transfer Date CustomDateField into a single block (Grid size 12/6 by schedule); disablePast/minDate support in the shared component remains a follow-up ticket.

zweatshirt:
Seems the transfer (if future dated) works, but fails to display in the upcoming transfers table

wjames111:
@zweatshirt yeah workin on that now, I'll let you know once fixed.

PR #1910, opened 2026-07-14
MPDX-9832 Update MPGA chart designs

kegrimes:
## Description

Income & Expenses Analysis UAT chart issues:
- Make the monthly summary chart display both income and expenses
- Change "green/red" colors to a truer version on green and red
- Break assessments, benefits, and salary into separate slices in the pie chart
- Make the summary bars in the bar chart skinnier

I also added some loading skeletons to each chart.

Jira ticket: [MPDX-9832](https://jira.cru.org/browse/MPDX-9832)

## Testing

- Check that all the charts meet the criteria above

## Checklist:


src/components/Reports/MPGAIncomeExpensesReport/Charts/MonthlySummaryChart.tsx
```diff
@@ -38,137 +37,133 @@ export const MonthlySummaryChart: React.FC<MonthlySummaryChartProps> = ({
...
-  const combinedData = useMemo(() => {
-    const transformedExpenseData = expenseData.map((item) => ({
-      ...item,
-      monthly: item.monthly.map((value) => -Math.abs(value)),
-    }));
-
-    return [...incomeData, ...transformedExpenseData];
-  }, [incomeData, expenseData]);
-
   const monthlyTotals = useMemo(() => {
```
kegrimes: [Important] The rewritten `monthlyTotals` memo — income/expenses reduce, `Math.abs(item.monthly[index] ?? 0)`, and `difference` — is the core behavior change of this PR, but the only new test asserts a bar *count* (`months.length * 2`). The per-month math, the tooltip's `net`/sign-color branch, and the `Math.abs` (a no-op against the all-positive mock, so negative expenses are unverified) have no coverage. Recharts tooltips don't render in jsdom, so extract this computation into a pure helper `computeMonthlyTotals(incomeData, expenseData, months)` and unit-test it directly (mixed-sign expenses, missing/short `monthly` arrays, empty inputs). Can't be dismissed (severity ≥ 7) but does not block merge.

src/components/Reports/MPGAIncomeExpensesReport/Charts/MonthlySummaryChart.tsx
```diff
@@ -38,137 +37,133 @@ export const MonthlySummaryChart: React.FC<MonthlySummaryChartProps> = ({
...
-      </Box>
-    </>
+                  {t('Net')}: {currencyFormat(net, currency, locale)}
+                </Typography>
+              </Box>
+            );
+          }}
+        />
+        <Bar dataKey="income" name={t('Income')} fill={chartColors[0]} />
+        <Bar dataKey="expenses" name={t('Expenses')} fill={chartColors[1]} />
```
kegrimes: [Important] Print regression: this rewrite removed the `GlobalStyles` `labels-print-only` wrapper and the `<LabelList>`, so bar values are now tooltip-only — and tooltips don't print. This chart renders in `DisplayModes/PrintOnlyReport.tsx`, and sibling `SummaryBarChart.tsx` still keeps `LabelList`, so the two charts now diverge in the same printed report. `PrintTables` still prints the raw monthly numbers, which keeps this from being critical — but please confirm dropping the print-only labels was intentional. If not, restore a print-only `LabelList`.

src/components/Reports/MPGAIncomeExpensesReport/Charts/MonthlySummaryChart.tsx
```diff
@@ -26,8 +25,8 @@ interface MonthlySummaryChartProps {
 }
 
 const chartColors = [
-  theme.palette.chartGreen.main,
-  theme.palette.chartPink.main,
+  theme.palette.statusSuccess.main,
```
kegrimes: [Important] Accessibility: the two bars per month (income/expenses) and the legend swatches are distinguished only by color — `statusSuccess` (green) vs `chipRedDark` (red). Red/green is the worst-case pair for deutan/protan color-vision deficiency; on screen the hover tooltip disambiguates, but static and printed views rely entirely on the color. Consider a pattern/texture fill, a lightness contrast, or value labels. (Color-vision users can't tell income from expense bars, and the swatches look identical.) Can't be dismissed (severity ≥ 7) but does not block merge.

src/hooks/useExpenseCategories.ts
```diff
@@ -22,13 +24,12 @@ export function useExpenseCategories(data: DataFields[]) {
         ministry.push(item);
       } else if (category === 'Healthcare Reimbursement') {
         healthcare.push(item);
-      } else if (
-        category === 'Assessment' ||
-        category === 'Additional Salary' ||
-        category === 'Benefits' ||
-        category === 'Salary'
-      ) {
+      } else if (category === 'Assessment') {
```
kegrimes, line 27: [Medium] Behavior change: the old bucket matched `Assessment || 'Additional Salary' || Benefits || Salary`; these new exact-match branches no longer catch `'Additional Salary'`, so it now falls through to the final `else` → `other`. This is **not** a total error — `expensesTotal` still sums every bucket and each row lands in exactly one, so no double-count or dropped amount. But now that there's a dedicated "Salary" pie slice, a user would reasonably expect "Additional Salary" under Salary, not lumped into the ~69% "Other" slice. No mock row exercises this path, so it's invisible to the test suite. Please confirm this reclassification is intended; if it should join Salary, add `|| category === 'Additional Salary'` to the salary branch (and add a test).

src/hooks/useExpenseCategories.test.ts
```diff
@@ -10,9 +10,11 @@ describe('useExpenseCategories', () => {
 
     expect(result.current.ministryTotal).toBe(2124);
     expect(result.current.healthcareTotal).toBe(1933);
-    expect(result.current.assessmentTotal).toBe(26);
+    expect(result.current.assessmentTotal).toBe(0);
```
kegrimes: [Medium] Two coverage gaps here: (1) no test routes a `'Additional Salary'` category — the behavior change one line over (it now lands in `other`) is unguarded; add a row and assert `otherTotal`. (2) The `assessment` bucket has no true-positive test — the mock's `'Staff Assessment'` never matches `=== 'Assessment'`, so `assessmentTotal` is always `0` in tests. Add a row with `description: 'Assessment'` and assert a non-zero `assessmentTotal`, so a future rename of the branch string is caught.

src/components/Reports/MPGAIncomeExpensesReport/Charts/MonthlySummaryChart.tsx
```diff
@@ -38,137 +37,133 @@ export const MonthlySummaryChart: React.FC<MonthlySummaryChartProps> = ({
...
+    return months.map((name, index) => {
+      const income = incomeData.reduce(
+        (sum, item) => sum + (item.monthly[index] ?? 0),
+        0,
+      );
+      const expenses = expenseData.reduce(
+        (sum, item) => sum + Math.abs(item.monthly[index] ?? 0),
+        0,
+      );
+      return { name, income, expenses, difference: income - expenses };
```
kegrimes: [Suggestion] Flagged independently by all 5 review agents: `difference: income - expenses` is added to each `monthlyTotals` entry but never consumed — the two `<Bar>`s use `income`/`expenses` and the tooltip recomputes its own `net` (line 78). Two sources of truth for "net." Either drop `difference`, or have the tooltip read `payload[0]?.payload?.difference` instead of recomputing.

src/components/Reports/MPGAIncomeExpensesReport/Charts/SummaryBarChart.tsx
```diff
@@ -49,7 +52,7 @@ export const SummaryBarChart: React.FC<SummaryBarChartProps> = ({
         <YAxis
           tickFormatter={(value) => currencyFormat(value, currency, locale)}
         />
-        <Bar dataKey="total">
+        <Bar dataKey="total" barSize={125}>
```
kegrimes: [Suggestion] `barSize={125}` is a hardcoded pixel value rather than a theme token, and it's fixed rather than responsive — it won't scale down on narrow/mobile widths. Consider extracting it to a named constant or using a percentage-based width.

src/components/Reports/MPGAIncomeExpensesReport/Charts/ExpensesPieChart/ExpensesPieChart.tsx
```diff
@@ -11,23 +11,33 @@ interface ExpensesPieChartProps {
...
+    benefitsTotal,
+    salaryTotal,
+    otherTotal,
+  } = useTotals();
 
   const data = [
     { name: t('Ministry'), value: ministryTotal ?? 0 },
     { name: t('Healthcare'), value: healthcareTotal ?? 0 },
+    { name: t('Assessment'), value: assessmentTotal ?? 0 },
+    { name: t('Benefits'), value: benefitsTotal ?? 0 },
```
kegrimes, line 41: [Suggestion] With the slice count now grown from 4 to 6, the all-zero empty state is more exposed: if every category total is 0, all slices are 0% (labels suppressed) and the pie renders a degenerate empty circle while the legend still lists all six categories. Consider an explicit "no data" state for a report with no expense data. (Largely pre-existing — the 0% suppression predates this PR.)

kegrimes (commented):
## 🤖 Multi-Agent Code Review — PR #1910

**Verdict: ✅ APPROVED WITH SUGGESTIONS** · Risk 7/10 (HIGH) · Mode: standard · 5 agents (Architecture, Testing, Standards, UX, Financial Reporting) + dependency analysis

**No blockers found.** The change is financially sound — every total was recomputed by hand and reconciles exactly (`18569 + 2400 = 20969` ✓), and the rewritten `net` math is a genuine correctness *improvement* over the old sign-based bucketing (the old code misclassified any expense row whose first month was `0`). Dependency analysis found **zero breaking changes** — blast radius is fully contained in the MPGA report tree and every consumer of the changed hook/context was updated.

> ⚠️ **3 Important findings (severity 7.0–7.9)** — these don't block merge and can't be `/dismiss`ed. All three are essentially *"confirm this was intentional"* decisions. Address them here or file follow-ups.

### Important Issues (7.0–7.9)
1. **`MonthlySummaryChart` income/expenses/net math is untested (7.5)** — the core behavior change of the PR; only a bar *count* is asserted. `Math.abs` is a no-op vs the all-positive mock, so negative expenses are unverified. Extract the per-month computation to a pure helper and unit-test it.
2. **Print regression (7.0)** — removing `GlobalStyles` + `LabelList` means the Monthly Summary bars show no values in print (tooltips don't print), and it diverges from sibling `SummaryBarChart` which kept `LabelList`. `PrintTables` still prints the raw numbers, so it's not critical — but confirm the drop was intentional.
3. **Income vs Expenses distinguished only by red/green (7.0)** — worst-case pair for red/green color-vision deficiency; static/print views can't disambiguate. Consider a pattern fill, lightness contrast, or value labels.

### Medium Priority (5.0–6.9)
- **"Additional Salary" silently reclassified to "Other" (6.5)** — the rewrite's exact-match branches no longer catch `'Additional Salary'`, so it falls through to `other`. `expensesTotal` is unaffected (no double-count), but with a dedicated "Salary" slice now, users would expect it under Salary. No test exercises this. Confirm intent.
- **`useExpenseCategories` missing coverage (6.0)** — no test for `'Additional Salary'` routing, and the `assessment` bucket has no true-positive test (mock's `'Staff Assessment'` never matches `=== 'Assessment'`, so `assessmentTotal` is always 0 in tests).

### Suggestions (< 5.0)
- **Unused `difference` field** (flagged by all 5 agents) — computed in `monthlyTotals` but never read; the tooltip recomputes `net`. Drop it or read it back.
- **`barSize={125}` magic number**, **pie all-zero empty state**, **MonthlySummaryChart empty-data test**.

**Standards checklist: all PASS** (named exports, i18n on all new strings, no `any`, no `new Date()`, rounding at display boundary only).

src/hooks/useExpenseCategories.ts
kegrimes: /dismiss this is intentional. Additional salary is a completely separate category, it was wrong before.

src/components/Reports/MPGAIncomeExpensesReport/Charts/MonthlySummaryChart.tsx
kegrimes: /dismiss the design was approved

src/hooks/useExpenseCategories.ts
kegrimes, line 27: [Important] Buckets on a **localized** string — `description` comes from `getLocalizedCategory()`, so this is `'Bewertung'` in de. 22/24 locales break; everything falls to "Other". Pre-existing (zero delta), so: follow-up ticket, bucket on `StaffExpenseCategoryEnum`.

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesReport.tsx
```diff
@@ -88,7 +88,7 @@ export const MPGAIncomeExpensesReport: React.FC<
 
   const { data: staffAccountData, error } = useStaffAccountQuery();
 
-  const { data: reportData } = useMpgaTransactionsQuery({
+  const { data: reportData, loading } = useMpgaTransactionsQuery({
```
kegrimes, line 91: [Medium] This PR added `loading` here but not `error`, so on query failure the pie asserts "You have no data at this time." — new prose stating something false. `useStaffAccountQuery` at :89 already destructures `error`. One-token fix on a line you own.

src/components/Reports/MPGAIncomeExpensesReport/Charts/MonthlySummaryChart/MonthlySummaryChart.tsx
```diff
@@ -0,0 +1,231 @@
...
+            <Bar dataKey="income" name={t('Income')} fill={chartColors[0]} />
+            <Bar
+              dataKey="expenses"
+              name={t('Expenses')}
+              fill={chartColors[1]}
+            />
+            <Bar
+              dataKey="tallest"
+              xAxisId="net"
+              shape={<g />}
```
kegrimes: [Medium] `shape={<g />}` is a ReactElement, so recharts `cloneElement`s internal props onto a DOM node — verified: "React does not recognize the `tooltipPayload` prop" ×12 per render. Could we use `shape={() => <g />}`? Verified 7/7 still pass, warnings gone.

src/components/Reports/MPGAIncomeExpensesReport/TotalsContext/TotalsContext.test.tsx
```diff
@@ -72,12 +76,14 @@ describe('TotalsContext', () => {
     );
 
     expect(getByText('108856')).toBeInTheDocument();
-    expect(getByText('18569')).toBeInTheDocument();
+    expect(getByText('20969')).toBeInTheDocument();
 
     expect(getByText('2124')).toBeInTheDocument();
     expect(getByText('1933')).toBeInTheDocument();
+    expect(getByText('13779')).toBeInTheDocument();
+    expect(getByText('2400')).toBeInTheDocument();
```
kegrimes: [Medium] `getByText('2400')` only proves *some* div holds it — swap `benefitsTotal`/`salaryTotal` and this still passes. The `data-testid`s at :44-51 are declared but unused. Consider `getByTestId('benefits')).toHaveTextContent('2400')`.

src/components/Reports/MPGAIncomeExpensesReport/mockData.tsx
```diff
@@ -91,7 +98,7 @@ export const mockData: AllData = {
     },
     {
       id: crypto.randomUUID(),
-      description: 'Staff Assessment',
+      description: 'Assessment',
```
kegrimes, line 101: [Medium] Good fix — `'Staff Assessment'` is a *sub*category label, so the old fixture was unreachable data. Could we build descriptions via `getLocalizedCategory(StaffExpenseCategoryEnum.Assessment, t)` so the next drift breaks the build instead of silently re-bucketing?

src/components/Reports/MPGAIncomeExpensesReport/Charts/SummaryBarChart/SummaryBarChart.tsx
```diff
@@ -29,7 +25,7 @@ export const SummaryBarChart: React.FC<SummaryBarChartProps> = ({
   const { t } = useTranslation();
   const locale = useLocale();
 
-  const { incomeTotal, expensesTotal } = useTotals();
+  const { incomeTotal, expensesTotal, loading } = useTotals();
 
   const data = [
```
kegrimes, line 30: [Medium] No `noData` guard, so zero data renders `$0.00` bars as fact — the repo rules name this exact case. The pie got a guard (`ExpensesPieChart.tsx:45`) and the bar charts didn't. Flagged independently by 3 agents; consider hoisting `noData` into `ChartFrame`.

src/components/Reports/MPGAIncomeExpensesReport/Charts/MonthlySummaryChart/MonthlySummaryChart.tsx
```diff
@@ -0,0 +1,231 @@
...
+                    </Typography>
+                    <Typography variant="body1">
+                      {t('Expenses')}:{' '}
+                      {currencyFormat(expenses, currency, locale)}
+                    </Typography>
+                    <Typography
+                      variant="body1"
+                      sx={{
+                        fontWeight: 'bold',
+                        color:
```
kegrimes, line 139: [Medium] New tooltip: Net text uses fill colors as text colors. `statusSuccess #24C976` is 2.17:1 on white; 16px body1 needs 4.5:1 (WCAG 1.4.3, which has no redundancy exemption). Consider `moss.main` / `statusDanger.main` for text, keeping the bright fills on bars.

src/components/Reports/MPGAIncomeExpensesReport/Charts/MonthlySummaryChart/MonthlySummaryChart.tsx
```diff
@@ -0,0 +1,231 @@
...
+  expenseData,
+  months,
+  aspect,
+  width,
+  currency,
+}) => {
+  const { t } = useTranslation();
+  const locale = useLocale();
+
+  const { loading } = useTotals();
```
kegrimes: [Suggestion] This chart is fully prop-driven yet reaches into context for `loading` alone — and `useTotals()` throws, so it hard-requires a provider it otherwise doesn't use. `loading` and `data` come from one query but travel by two transports. Consider a 7th prop.

src/components/Reports/MPGAIncomeExpensesReport/Charts/ExpensesPieChart/ExpensesPieChart.tsx
```diff
@@ -0,0 +1,146 @@
...
+            }}
+            outerRadius="80%"
+            dataKey="value"
+          >
+            {data.map((entry, index) => (
+              <Cell key={entry.name} fill={chartColors[index]} />
+            ))}
+          </Pie>
+        )}
+        <Legend
```
kegrimes, line 102: [Suggestion] Confirmed bug: `<Legend>` sits outside the `noData` ternary (which closes at :99), so the empty state shows "You have no data at this time." above six swatches naming six categories. Could we move it into the else branch?

src/components/Reports/MPGAIncomeExpensesReport/Charts/ExpensesPieChart/ExpensesPieChart.tsx
```diff
@@ -0,0 +1,146 @@
...
+import theme from 'src/theme';
+import { useTotals } from '../../TotalsContext/TotalsContext';
+import { ChartFrame } from '../ChartFrame';
+
+interface ExpensesPieChartProps {
+  aspect: number;
+  width: number;
+}
+
+const chartColors = [
```
kegrimes, line 13: [Suggestion] 12/15 slice pairs fall below WCAG's 3:1 (pink/orange 1.17:1, green/secondary 1.03:1). Not a conformance failure — the adjacent Expenses TableCard gives text redundancy — but the pie is unusable *as a pie* for a deuteranope, and 4→6 slices worsened it.

src/components/Reports/MPGAIncomeExpensesReport/Charts/MonthlySummaryChart/MonthlySummaryChart.tsx
```diff
@@ -0,0 +1,231 @@
...
+                style={{
+                  fill: theme.palette.chartBlack.main,
+                  fontSize: theme.typography.body2.fontSize,
+                }}
+                formatter={(value: number) =>
+                  currencyFormat(value, currency, locale)
+                }
+              />
+            </Bar>
+            <Legend
```
kegrimes, line 177: [Suggestion] ~33 lines of Legend renderer duplicated with ExpensesPieChart.tsx:101 (differs only in gap/flexWrap); `legend-swatch` lives in exactly these 2 files, so it isn't a convention. `ChartFrame` was the right seam — a sibling `ChartLegend` would finish the job.

src/components/Reports/MPGAIncomeExpensesReport/Charts/ChartFrame.tsx
```diff
@@ -0,0 +1,39 @@
...
+export const ChartFrame: React.FC<ChartFrameProps> = ({
+  width,
+  aspect,
+  children,
+  loading,
+}) => (
+  <Box
+    sx={{
+      width: `${width}%`,
+      aspectRatio: `${aspect}`,
```
kegrimes, line 21: [Suggestion] Aspect has two sources of truth (Box `aspectRatio` + `ResponsiveContainer aspect`), and the `@media print` override exists only to reconcile them. Applying `aspectRatio` only while loading would let RC own height and drop the print hack.

src/components/Reports/MPGAIncomeExpensesReport/Charts/ChartFrame.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  it('reserves the same space while loading as when loaded', () => {
+    const { container: loadingContainer } = render(<TestComponent loading />);
+    const { container: loadedContainer } = render(<TestComponent />);
+
+    const loadingFrame = loadingContainer.firstChild as HTMLElement;
+    const loadedFrame = loadedContainer.firstChild as HTMLElement;
+
+    expect(loadingFrame).toHaveStyle({ width: '80%' });
+    expect(loadingFrame.className).toEqual(loadedFrame.className);
```
kegrimes, line 46: [Suggestion] Tautological: the Box renders outside the `loading` ternary, so both branches emit an identical emotion class by construction — this line cannot fail. (`:44`'s `toHaveStyle` is real.) Assert `aspectRatio` instead to test the space-reservation it names.

src/components/Reports/MPGAIncomeExpensesReport/Charts/ExpensesPieChart/ExpensesPieChart.tsx
```diff
@@ -0,0 +1,146 @@
...
+    <ChartFrame width={width} aspect={aspect} loading={loading}>
+      <PieChart>
+        {noData ? (
+          <text
+            x="50%"
+            y="40%"
+            textAnchor="middle"
+            dominantBaseline="central"
+            fill={theme.palette.chartBlack.main}
+          >
```
kegrimes, line 58: [Suggestion] Empty state is a raw SVG `<text>` — inherits no typography scale and won't match the sibling `EmptyTable` empty states this report already uses. Consider rendering it as a sibling of ChartFrame rather than a PieChart child.

src/components/Reports/MPGAIncomeExpensesReport/Charts/MonthlySummaryChart/MonthlySummaryChart.test.tsx
```diff
@@ -0,0 +1,170 @@
...
+      id: description,
+      description,
+      monthly,
+      average: 0,
+      total: monthly.reduce((sum, value) => sum + value, 0),
+    });
+
+    const { findByRole } = render(
+      <ThemeProvider theme={theme}>
+        <TotalsProvider data={mockData}>
```
kegrimes: [Suggestion] `<TotalsProvider data={mockData}>` wraps a chart fed entirely different props — the provider is inert (only `loading` is read) and reads as if the fixture drives the assertion. `data={{ income: [], expenses: [] }}` would make the intent obvious.

kegrimes (commented):
## Multi-Agent Review — 5 agents + dependency analysis, 2 debate rounds

**Verdict: APPROVED WITH SUGGESTIONS.** No blockers. 1 Important, 6 Medium, 14 Suggestions, 2 pre-existing, 1 dismissed.

### Lead with the good news

**This PR fixes a severe pre-existing miscalculation that nobody knew was there.** The old `MonthlySummaryChart` negated expenses via `-Math.abs(value)` and then bucketed income-vs-expense on `item.monthly[0] >= 0`. `-Math.abs(0)` is `-0`, and `-0 >= 0` is `true` — so **any expense row starting at zero was summed into income while carrying negative values**. Three of seven fixture rows start at zero (Transfer, Ministry, Healthcare), dragging reported income by up to −4257. The rewrite consumes the already-separated `incomeData`/`expenseData` arrays instead of re-deriving type from sign, which eliminates the bug class rather than patching it. Verified independently against `main` by two agents. This deserves a line in the PR description.

Also verified clean: **all 22 new numeric expectations are arithmetically correct** (recomputed by hand, twice, independently); no-drop/no-double-count holds (`20969` both ways); the `18569 → 20969` delta is exactly the new Benefits row; currency mixing is structurally impossible; no `new Date()`; `yarn lint:ts` exits 0; `lint:ci` has 0 PR-attributable warnings; 64/64 tests pass. The `'Staff Assessment' → 'Assessment'` fixture rename is a **genuine fix** — `'Staff Assessment'` is a *sub*category label, so the old fixture described data the API cannot emit.

### The one Important finding

`useExpenseCategories` buckets expenses by comparing against **localized** strings — `item.description` comes from `getLocalizedCategory()`, so `'Assessment'` is `'Bewertung'` in German. I confirmed **22 of 24 locales** translate it; only `en`/`en-US` pass through. Every named bucket collapses to "Other".

**But this PR does not cause it.** The old code compared the same English literals, so the German pie was *already* one 100% blob — this takes it from a 4-entry legend with one blob to a 6-entry legend with one blob. **The delta is zero**, verified by three agents against `main`. The fix lives in `DataFields` (which carries no category enum), not in the rewritten lines. Consensus: real, high-value, **its own ticket** — not this PR's to carry.

Severity note: 8.5 is the highest individual rating, but only 2 of 5 agents landed at 8+ *after* seeing the zero-delta evidence (Architecture 6.5, UX 9.0→6.5, Financial 7.0). Per the consensus table that classifies as **Important, not a blocker**. It is still ≥7, so it cannot be `/dismiss`ed — please file the ticket instead.

### What the debate did

The agents corrected themselves more than they corrected each other: **9 findings revised down, 1 withdrawn, zero held at 9.0.** Testing challenged both of its own 9.0s and conceded both — including discovering that its "no test can catch this" claim was false (`__tests__/util/testingLibraryReactMock.tsx:8` already ships the mechanism; a `de` test is ~1 line). UX withdrew a finding after Standards produced 8 counter-instances, conceded its headline contrast finding once it found the adjacent `TableCard` provides text redundancy, then **falsified a peer's evidence** (Standards' "`es` works by accident" — there is no `es` locale, only `es-419`, which translates fully).

### Cross-cutting: fix-one-fix-all miss

| Chart | Loading | Empty | Error |
|---|---|---|---|
| SummaryBarChart | ✓ | **✗** | ✗ |
| ExpensesPieChart | ✓ | ✓ | ✗ |
| MonthlySummaryChart | ✓ | **✗** | ✗ |

The `noData` guard landed on the pie only. Three agents found this independently. The same report now tells two stories side by side: the pie says "no data", the bars next to it render `$0.00` as fact.

### Your dismissal

✅ **`Additional Salary` → Other — dismissal accepted and upheld.** All five agents converged *toward* you once they read the diff: the old code lumped it into `assessment`, which is indefensible, so "it was wrong before" is factually correct. Architecture self-corrected 7.0→6.0 and wrote: *"I am not deflating this to accommodate the author, and I would not have inflated it to defeat him. 7.0 was simply wrong."* Final range 3.0–6.5, dismissible. The only residue is a **test-lock** (suggestion, below): nothing pins your decision, so anyone could re-add the branch and all 64 tests stay green.

⚠️ **Your other dismissal was procedurally invalid** — the accessibility finding was tagged severity 7.0, and only <7.0 can be dismissed. Good news: it's moot. This round's UX agent re-examined the same concern independently and rated it **5.5** (and the pie's color-only issue 4.5), because the adjacent Expenses `TableCard` renders every category name and amount in text — WCAG 1.4.1 is satisfied at page level. It lands in suggestion tier on the merits, not by dismissal.

---

### Pre-existing findings (informational — not anchorable to this diff)

Both sit on lines outside any diff hunk, so GitHub can't attach them inline. Neither counts toward the verdict, and neither needs `/dismiss`.

**`src/hooks/useExpenseCategories.ts:4-6`** — severity 4.0 — `sum` does `rows?.reduce(...) || 0`. A single row with `total: undefined` makes the reduce return `NaN`, and `|| 0` silently converts that to `0` — **zeroing an entire category** rather than surfacing "unknown". There's no `?? 0` on `item.total`. Pre-existing, but this PR grew the blast radius from 4 buckets to 6 and adds two new call sites. This helper is also byte-identical to the one in `TotalsContext.tsx:35-37` — two copies that can drift.

**`src/components/Reports/MPGAIncomeExpensesReport/mockData.tsx:34-39` and `:95-97`** — severity 2.0 — two fixture rows are internally inconsistent: `Contributions` has Σmonthly=108153 vs `total: 108156` (+3), and `Charge(s) for Credit Card gift(s)` has Σ=508 vs `total: 507` (−1). The resulting $4 gap is why a cross-chart consistency test — asserting `SummaryBarChart`'s `expensesTotal` agrees with `MonthlySummaryChart`'s summed monthlies — **cannot be written today**; it would fail on fixture noise rather than real behavior. Worth 2 characters to unlock that test. The rows this PR added are clean.

src/components/Reports/MPGAIncomeExpensesReport/TotalsContext/TotalsContext.tsx
```diff
@@ -8,7 +8,10 @@ export type TotalsType = {
   ministryTotal: number;
   healthcareTotal: number;
   assessmentTotal: number;
+  benefitsTotal: number;
+  salaryTotal: number;
   otherTotal: number;
+  loading: boolean;
```
kegrimes: [Suggestion] The props→context move is defensible (it fixed table loading states that were dead in production), but `loading` mixes fetch-state into an otherwise-derived-totals contract, and `useTotals().loading` doesn't say *what* is loading. Consider naming it for its query.

src/components/Reports/MPGAIncomeExpensesReport/Charts/MonthlySummaryChart/MonthlySummaryChart.test.tsx
```diff
@@ -0,0 +1,170 @@
...
+        '$7,026',
+        '$7,713',
+        '$7,312',
+        '$9,579',
+        '$16,588',
+      ]);
+    });
+  });
+
+  it('subtracts expenses from income, treating expenses as positive', async () => {
```
kegrimes, line 90: [Suggestion] `expenseData: [{ monthly: [-40, 250] }]` is unreachable — `filterFunds` guarantees expense months ≥ 0 on all 4 branches. It pins a contract the pipeline can't produce and implies a 38% cross-chart divergence that can't occur. `[0, 250]` would keep the test honest.

src/components/Reports/MPGAIncomeExpensesReport/Charts/ChartFrame.tsx
```diff
@@ -0,0 +1,39 @@
...
+      '@media print': { aspectRatio: 'auto' },
+    }}
+  >
+    {loading ? (
```
kegrimes, line 25: [Suggestion] One query now drives 5 concurrent spinners (3 × ChartFrame + 2 × TableCard), all with the same `data-testid`, flashing in unison then swapping independently. A single card-level skeleton per card would read as more deliberate.

src/components/Reports/MPGAIncomeExpensesReport/Charts/ExpensesPieChart/ExpensesPieChart.tsx
```diff
@@ -0,0 +1,146 @@
...
+              innerRadius,
+              outerRadius,
+              percent,
+            }) => {
+              const RADIAN = Math.PI / 180;
+              const radius = innerRadius + (outerRadius - innerRadius) * 1.2;
+              const x = cx + radius * Math.cos(-midAngle * RADIAN);
+              const y = cy + radius * Math.sin(-midAngle * RADIAN);
+
+              const percentage = (percent * 100).toFixed(0);
```
kegrimes, line 80: [Suggestion] `.toFixed(0)` is correctly at the display boundary, and slices aren't dropped — but the 6 rounded percentages sum to 99%, and Salary (0.12%) renders a colored sector with no label. `<1%` might read better than nothing.

src/components/Reports/MPGAIncomeExpensesReport/Charts/MonthlySummaryChart/MonthlySummaryChart.tsx
```diff
@@ -0,0 +1,231 @@
...
+                        fontWeight: 'bold',
+                        color:
+                          net === 0
+                            ? 'inherit'
+                            : net > 0
+                              ? chartColors[0]
+                              : chartColors[1],
+                      }}
+                    >
+                      {t('Net')}: {currencyFormat(net, currency, locale)}
```
kegrimes, line 147: [Suggestion] `t('Net')` and `t('You have no data at this time.')` aren't in en/translation.json yet. Both extract cleanly, and 144 keys are already missing repo-wide (extraction is batched), so this is consistent with practice — just noting for the next `yarn extract`.

src/components/Reports/MPGAIncomeExpensesReport/Charts/SummaryBarChart/SummaryBarChart.tsx
```diff
@@ -49,7 +45,7 @@ export const SummaryBarChart: React.FC<SummaryBarChartProps> = ({
         <YAxis
           tickFormatter={(value) => currencyFormat(value, currency, locale)}
         />
-        <Bar dataKey="total">
+        <Bar dataKey="total" maxBarSize={125}>
```
kegrimes, line 48: [Suggestion] `maxBarSize={125}` is a new magic number that isn't derivable from a theme token — a short comment on why 125 would help the next reader.

src/hooks/useExpenseCategories.ts
kegrimes: /dismiss this is only a Cru US staff report. It will all be in English

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesReport.tsx
kegrimes: /dismiss skip for now

src/components/Reports/MPGAIncomeExpensesReport/mockData.tsx
kegrimes: /dismiss skip for now

src/components/Reports/MPGAIncomeExpensesReport/Charts/SummaryBarChart/SummaryBarChart.tsx
kegrimes: /dismiss the design looks fine on the bar charts - the pie chart was the only one to disappear

src/components/Reports/MPGAIncomeExpensesReport/Charts/MonthlySummaryChart/MonthlySummaryChart.tsx
kegrimes: /dismiss the design was approved

src/components/Reports/MPGAIncomeExpensesReport/Charts/MonthlySummaryChart/MonthlySummaryChart.tsx
kegrimes: /dismiss leave for now

src/components/Reports/MPGAIncomeExpensesReport/Charts/ExpensesPieChart/ExpensesPieChart.tsx
kegrimes: /dismiss this was the intentional design

src/components/Reports/MPGAIncomeExpensesReport/Charts/ExpensesPieChart/ExpensesPieChart.tsx
kegrimes: /dismiss this design was approved

src/components/Reports/MPGAIncomeExpensesReport/Charts/ChartFrame.tsx
kegrimes: /dismiss they both read from the same prop

src/components/Reports/MPGAIncomeExpensesReport/Charts/ExpensesPieChart/ExpensesPieChart.tsx
kegrimes: /dismiss it is better for the empty chart to be shown instead of a one-line message

src/components/Reports/MPGAIncomeExpensesReport/Charts/ExpensesPieChart/ExpensesPieChart.tsx
kegrimes: /dismiss this is not part of the design

src/components/Reports/MPGAIncomeExpensesReport/Charts/ChartFrame.tsx
kegrimes: /dismiss

src/components/Reports/MPGAIncomeExpensesReport/Charts/MonthlySummaryChart/MonthlySummaryChart.tsx
kegrimes: /dismiss the cron job will take care of it

wjames111 (approved):
Sorry for the late review. Looks good to me, thanks for the updates!

PR #1912, opened 2026-07-14
[MPDX-9771] - MPD leader impersonation (frontend)

wjames111:
## Summary

Frontend half of MPDX-9771: lets MPD leaders — any user who coaches an account list — impersonate the staff they coach under a **restricted, server-enforced scope**, so they can help with MPD forms without seeing contacts, tasks, or contact PII.

Companion API PR (deploy prerequisite): https://github.com/CruGlobal/mpdx_api/pull/3455 — the API mints the scoped JWT (`impersonation_scope: 'mpd_leader'`) and enforces every restriction server-side. This PR wires the scope through the frontend session and adapts the UI.

## Permission model (4 tiers)

1. **Hidden** — Contacts, Tasks, and (non-HR) Tools: nav tabs hidden, routes redirect to the dashboard via `blockRestrictedImpersonation`, and the API independently blocks the data.
2. **Editable** — NS Goal Calculator, NSO MPD Questionnaire, Salary Calculator, and MHA requests remain fully editable (matches the API mutation allowlist).
3. **Read-only** — all other HR tools: enforced by the API's mutation allowlist (non-allowlisted mutations are denied server-side). Per review, this PR adds no frontend read-only affordances — attempted edits surface the API's authorization error.
4. **Visible** — all reports and the dashboard; contact PII (addresses, emails, phones, donor account numbers) is scrubbed by the API.

## Changes

- **Session plumbing**: `GetUserAccess` now selects `coachingAccountLists(first: 1) { totalCount }` and the NextAuth JWT/session derive a `coach` boolean from it (plus `impersonationScope`); eligibility is coaching-based, not an admin-granted flag; `impersonateHelper` stores the API's `impersonation_scope` in a signed `mpdx-handoff.impersonationScope` cookie which `setUserInfo` verifies and folds into the JWT.
- **Route guards** (`pagePropsHelpers`): new `enforceAdminOrCoach` (admin console access for coaches) and `blockRestrictedImpersonation` (contacts/tasks/tools pages redirect during scoped impersonation); applied to 15 pages + `admin.page.tsx`.
- **Nav**: contacts/tasks/tools tabs hidden during scoped impersonation (`useNavPages`); Admin settings nav item visible to coaches; Reset Account accordion stays admin-only.
- **Impersonate entry points**: Admin Console (existing form, now accessible to coaches) and a new `ImpersonateStaffModal` on the MPD Goal Admin table (lazy-loaded via the Dynamic modal pattern, Formik/Yup reason field).
- **Restricted-impersonation signal**: `useRestrictedImpersonation` hook + `isRestrictedImpersonation` server-side predicate (used for nav/route hiding). Per review, the earlier per-mutation frontend guards and disabled states were removed — the API is the enforcement layer for read-only tools.

## Testing

- `yarn lint`, `yarn lint:ts` clean; full `yarn test` green (16 suites that failed under parallel load all pass in isolation — known flakiness, verified with `--onlyFailures`).
- New/updated tests cover session callbacks, route guards, nav hiding, and the impersonate modal.

## Deploy notes / follow-up

- **Requires** CruGlobal/mpdx_api#3455 to be deployed first; without it the API neither issues nor honors `impersonation_scope`.
- The API denies non-allowlisted mutations by default during scoped impersonation. Background/incidental mutations (e.g. preference writes fired by shared components on reports pages) should be spot-checked on staging while impersonating — no local API run was feasible in this environment.

### DIFF
--- a/pages/api/auth/[...nextauth].page.ts
+++ b/pages/api/auth/[...nextauth].page.ts
@@ -27,11 +27,13 @@ declare module 'next-auth' {
       email: string;
       admin: boolean;
       developer: boolean;
+      mpdSupervisorAdmin: boolean;
       apiToken: string;
       userID: string;
       impersonating?: boolean;
       impersonatorApiToken?: string;
       isImpersonatorDeveloper?: boolean;
+      impersonationScope?: string;
     };
   }
 
@@ -41,6 +43,7 @@ declare module 'next-auth' {
     impersonating?: boolean;
     impersonatorApiToken?: string;
     isImpersonatorDeveloper?: boolean;
+    impersonationScope?: string;
   }
 }
 
@@ -48,11 +51,13 @@ declare module 'next-auth/jwt' {
   interface JWT {
     admin: boolean;
     developer: boolean;
+    mpdSupervisorAdmin: boolean;
     apiToken: string;
     userID?: string;
     impersonating?: boolean;
     impersonatorApiToken?: string;
     isImpersonatorDeveloper?: boolean;
+    impersonationScope?: string;
   }
 }
 
@@ -178,6 +183,7 @@ const Auth = (req: NextApiRequest, res: NextApiResponse): Promise<void> => {
           user.impersonating = userInfo.impersonating;
           user.impersonatorApiToken = userInfo.impersonatorApiToken;
           user.isImpersonatorDeveloper = userInfo.isImpersonatorDeveloper;
+          user.impersonationScope = userInfo.impersonationScope;
 
           if (cookies) {
             res.setHeader('Set-Cookie', cookies);
@@ -241,11 +247,13 @@ const Auth = (req: NextApiRequest, res: NextApiResponse): Promise<void> => {
             ...token,
             admin: data.user.admin,
             developer: data.user.developer,
+            mpdSupervisorAdmin: data.user.mpdSupervisorAdmin,
             apiToken: user.apiToken,
             userID: user.userID,
             impersonating: user.impersonating,
             impersonatorApiToken: user.impersonatorApiToken,
             isImpersonatorDeveloper: user.isImpersonatorDeveloper,
+            impersonationScope: user.impersonationScope,
           };
         } else {
           return token;
@@ -255,10 +263,12 @@ const Auth = (req: NextApiRequest, res: NextApiResponse): Promise<void> => {
         const {
           admin,
           developer,
+          mpdSupervisorAdmin,
           apiToken,
           userID,
           impersonating,
           isImpersonatorDeveloper,
+          impersonationScope,
         } = token;
 
         // Check the expiration of the API token JWT without verifying its signature
@@ -273,10 +283,12 @@ const Auth = (req: NextApiRequest, res: NextApiResponse): Promise<void> => {
             ...session.user,
             admin,
             developer,
+            mpdSupervisorAdmin,
             apiToken,
             userID,
             impersonating,
             isImpersonatorDeveloper,
+            impersonationScope,
           },
         };
       },
--- a/pages/api/auth/[...nextauth].page.ts
+++ b/pages/api/auth/[...nextauth].page.ts
@@ -27,7 +27,7 @@ declare module 'next-auth' {
       email: string;
       admin: boolean;
       developer: boolean;
-      mpdSupervisorAdmin: boolean;
+      coach: boolean;
       apiToken: string;
       userID: string;
       impersonating?: boolean;
@@ -51,7 +51,7 @@ declare module 'next-auth/jwt' {
   interface JWT {
     admin: boolean;
     developer: boolean;
-    mpdSupervisorAdmin: boolean;
+    coach: boolean;
     apiToken: string;
     userID?: string;
     impersonating?: boolean;
@@ -247,7 +247,7 @@ const Auth = (req: NextApiRequest, res: NextApiResponse): Promise<void> => {
             ...token,
             admin: data.user.admin,
             developer: data.user.developer,
-            mpdSupervisorAdmin: data.user.mpdSupervisorAdmin,
+            coach: data.coachingAccountLists.totalCount > 0,
             apiToken: user.apiToken,
             userID: user.userID,
             impersonating: user.impersonating,
@@ -263,7 +263,7 @@ const Auth = (req: NextApiRequest, res: NextApiResponse): Promise<void> => {
         const {
           admin,
           developer,
-          mpdSupervisorAdmin,
+          coach,
           apiToken,
           userID,
           impersonating,
@@ -283,8 +283,8 @@ const Auth = (req: NextApiRequest, res: NextApiResponse): Promise<void> => {
             ...session.user,
             admin,
             developer,
-            // Default to false for JWTs minted before mpdSupervisorAdmin existed
-            mpdSupervisorAdmin: mpdSupervisorAdmin ?? false,
+            // Default to false for JWTs minted before coach existed
+            coach: coach ?? false,
             apiToken,
             userID,
             impersonating,
--- a/pages/api/auth/[...nextauth].page.ts
+++ b/pages/api/auth/[...nextauth].page.ts
@@ -283,7 +283,8 @@ const Auth = (req: NextApiRequest, res: NextApiResponse): Promise<void> => {
             ...session.user,
             admin,
             developer,
-            mpdSupervisorAdmin,
+            // Default to false for JWTs minted before mpdSupervisorAdmin existed
+            mpdSupervisorAdmin: mpdSupervisorAdmin ?? false,
             apiToken,
             userID,
             impersonating,
--- a/pages/api/utils/pagePropsHelpers.ts
+++ b/pages/api/utils/pagePropsHelpers.ts
@@ -113,8 +113,8 @@ export const enforceAdmin: GetServerSideProps<PagePropsWithSession> = async (
 };
 
 // Redirect back to the dashboard unless the user is an admin, a developer, or
-// an MPD supervisor admin (MPD leader)
-export const enforceAdminOrMpdLeader: GetServerSideProps<
+// a coach (MPD leader)
+export const enforceAdminOrCoach: GetServerSideProps<
   PagePropsWithSession
 > = async (context) => {
   const session = await getSession(context);
@@ -123,11 +123,7 @@ export const enforceAdminOrMpdLeader: GetServerSideProps<
     return loginRedirect(context);
   }
 
-  if (
-    !session.user.admin &&
-    !session.user.developer &&
-    !session.user.mpdSupervisorAdmin
-  ) {
+  if (!session.user.admin && !session.user.developer && !session.user.coach) {
     return dashboardRedirect(context, RedirectReason.Unauthorized);
   }
 
--- a/pages/api/utils/pagePropsHelpers.ts
+++ b/pages/api/utils/pagePropsHelpers.ts
@@ -8,6 +8,7 @@ import { Session } from 'next-auth';
 import { getSession } from 'next-auth/react';
 import { RedirectReason } from 'pages/api/auth/redirectReasonEnum';
 import makeSsrClient from 'src/lib/apollo/ssrClient';
+import { isRestrictedImpersonation } from 'src/lib/restrictedImpersonation';
 import {
   GetDefaultAccountDocument,
   GetDefaultAccountQuery,
@@ -46,7 +47,8 @@ export const dashboardRedirect = (
  * but should be restricted when admins impersonate (e.g., staff expense data)
  * Exception: restricted-scope impersonation sessions (e.g. MPD leaders, where
  * impersonationScope is set) are allowed through because their access is
- * limited to the sections their scope permits
+ * limited to the sections their scope permits — see the access matrix
+ * documented on isRestrictedImpersonation
  */
 export const blockImpersonatingNonDevelopers: GetServerSideProps<
   PagePropsWithSession
@@ -61,7 +63,7 @@ export const blockImpersonatingNonDevelopers: GetServerSideProps<
   if (
     session.user.impersonating &&
     !session.user.isImpersonatorDeveloper &&
-    !session.user.impersonationScope
+    !isRestrictedImpersonation(session.user)
   ) {
     return dashboardRedirect(context, RedirectReason.ImpersonationBlocked);
   }
@@ -143,7 +145,8 @@ export const enforceAdminOrMpdLeader: GetServerSideProps<
 /* Block restricted-scope impersonation sessions (e.g. MPD leaders, where
  * impersonationScope is set), allow everyone else
  * Use case: Sections like Contacts, Tasks, and Tools that must be unreachable
- * during a restricted impersonation session
+ * during a restricted impersonation session — see the access matrix
+ * documented on isRestrictedImpersonation
  */
 export const blockRestrictedImpersonation: GetServerSideProps<
   PagePropsWithSession
@@ -154,7 +157,7 @@ export const blockRestrictedImpersonation: GetServerSideProps<
     return loginRedirect(context);
   }
 
-  if (session.user.impersonationScope) {
+  if (isRestrictedImpersonation(session.user)) {
     return dashboardRedirect(context, RedirectReason.ImpersonationBlocked);
   }
 
--- a/pages/api/utils/pagePropsHelpers.ts
+++ b/pages/api/utils/pagePropsHelpers.ts
@@ -112,8 +112,8 @@ export const enforceAdmin: GetServerSideProps<PagePropsWithSession> = async (
   };
 };
 
-// Redirect back to the dashboard unless the user is an admin or an MPD
-// supervisor admin (MPD leader)
+// Redirect back to the dashboard unless the user is an admin, a developer, or
+// an MPD supervisor admin (MPD leader)
 export const enforceAdminOrMpdLeader: GetServerSideProps<
   PagePropsWithSession
 > = async (context) => {
@@ -123,7 +123,11 @@ export const enforceAdminOrMpdLeader: GetServerSideProps<
     return loginRedirect(context);
   }
 
-  if (!session.user.admin && !session.user.mpdSupervisorAdmin) {
+  if (
+    !session.user.admin &&
+    !session.user.developer &&
+    !session.user.mpdSupervisorAdmin
+  ) {
     return dashboardRedirect(context, RedirectReason.Unauthorized);
   }
 
--- a/pages/api/utils/pagePropsHelpers.ts
+++ b/pages/api/utils/pagePropsHelpers.ts
@@ -44,6 +44,9 @@ export const dashboardRedirect = (
 /* Block non-developers who are impersonating, allow everyone else
  * Use case: Pages that regular users can access on their own account,
  * but should be restricted when admins impersonate (e.g., staff expense data)
+ * Exception: restricted-scope impersonation sessions (e.g. MPD leaders, where
+ * impersonationScope is set) are allowed through because their access is
+ * limited to the sections their scope permits
  */
 export const blockImpersonatingNonDevelopers: GetServerSideProps<
   PagePropsWithSession
@@ -54,8 +57,12 @@ export const blockImpersonatingNonDevelopers: GetServerSideProps<
     return loginRedirect(context);
   }
 
-  // Check if the impersonator is a developer
-  if (session.user.impersonating && !session.user.isImpersonatorDeveloper) {
+  // Check if the impersonator is a developer or has a restricted scope
+  if (
+    session.user.impersonating &&
+    !session.user.isImpersonatorDeveloper &&
+    !session.user.impersonationScope
+  ) {
     return dashboardRedirect(context, RedirectReason.ImpersonationBlocked);
   }
 
@@ -86,6 +93,69 @@ export const enforceAdmin: GetServerSideProps<PagePropsWithSession> = async (
 
   if (!session.user.admin) {
     return dashboardRedirect(context, RedirectReason.Unauthorized);
+  }
+
+  const underscoreRedirect = await handleUnderscoreAccountListRedirect(
+    session,
+    context.resolvedUrl,
+  );
+  if (underscoreRedirect) {
+    return underscoreRedirect;
+  }
+
+  return {
+    props: {
+      session,
+    },
+  };
+};
+
+// Redirect back to the dashboard unless the user is an admin or an MPD
+// supervisor admin (MPD leader)
+export const enforceAdminOrMpdLeader: GetServerSideProps<
+  PagePropsWithSession
+> = async (context) => {
+  const session = await getSession(context);
+
+  if (!session?.user.apiToken) {
+    return loginRedirect(context);
+  }
+
+  if (!session.user.admin && !session.user.mpdSupervisorAdmin) {
+    return dashboardRedirect(context, RedirectReason.Unauthorized);
+  }
+
+  const underscoreRedirect = await handleUnderscoreAccountListRedirect(
+    session,
+    context.resolvedUrl,
+  );
+  if (underscoreRedirect) {
+    return underscoreRedirect;
+  }
+
+  return {
+    props: {
+      session,
+    },
+  };
+};
+
+/* Block restricted-scope impersonation sessions (e.g. MPD leaders, where
+ * impersonationScope is set), allow everyone else
+ * Use case: Sections like Contacts, Tasks, and Tools that must be unreachable
+ * during a restricted impersonation session
+ */
+export const blockRestrictedImpersonation: GetServerSideProps<
+  PagePropsWithSession
+> = async (context) => {
+  const session = await getSession(context);
+
+  if (!session?.user.apiToken) {
+    return loginRedirect(context);
+  }
+
+  if (session.user.impersonationScope) {
+    return dashboardRedirect(context, RedirectReason.ImpersonationBlocked);
   }
 
   const underscoreRedirect = await handleUnderscoreAccountListRedirect(
--- a/src/components/HrTools/AdditionalSalaryRequest/Shared/AdditionalSalaryRequestContext.tsx
+++ b/src/components/HrTools/AdditionalSalaryRequest/Shared/AdditionalSalaryRequestContext.tsx
@@ -18,7 +18,6 @@ import {
   AdditionalSalaryRequestCalculations,
   AsrStatusEnum,
 } from 'src/graphql/types.generated';
-import { useRestrictedImpersonation } from 'src/hooks/useRestrictedImpersonation';
 import { useStepList } from 'src/hooks/useStepList';
 import { useTrackMutation } from 'src/hooks/useTrackMutation';
 import { Steps } from '../../Shared/CalculationReports/StepsList/StepsList';
@@ -201,7 +200,6 @@ export const AdditionalSalaryRequestProvider: React.FC<Props> = ({
 
   const [deleteAdditionalSalaryRequest] =
     useDeleteAdditionalSalaryRequestMutation();
-  const restrictedImpersonation = useRestrictedImpersonation();
 
   const currentStep = sections[currentIndex];
 
@@ -215,12 +213,6 @@ export const AdditionalSalaryRequestProvider: React.FC<Props> = ({
 
   const handleDeleteRequest = useCallback(
     async (id: string, isCancel: boolean) => {
-      // The tool is read-only during restricted impersonation, so deleting
-      // must not fire mutations that the API would reject
-      if (restrictedImpersonation) {
-        return;
-      }
-
       await deleteAdditionalSalaryRequest({
         variables: { id },
         refetchQueries: ['AdditionalSalaryRequest'],
@@ -243,13 +235,7 @@ export const AdditionalSalaryRequestProvider: React.FC<Props> = ({
         },
       });
     },
-    [
-      deleteAdditionalSalaryRequest,
-      restrictedImpersonation,
-      enqueueSnackbar,
-      resetSteps,
-      t,
-    ],
+    [deleteAdditionalSalaryRequest, enqueueSnackbar, resetSteps, t],
   );
 
   const [isDrawerOpen, setIsDrawerOpen] = useState(true);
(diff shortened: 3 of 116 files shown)


src/components/HrTools/AdditionalSalaryRequest/Shared/AdditionalSalaryRequestContext.tsx
```diff
@@ -213,6 +215,12 @@ export const AdditionalSalaryRequestProvider: React.FC<Props> = ({
 
   const handleDeleteRequest = useCallback(
     async (id: string, isCancel: boolean) => {
+      // The tool is read-only during restricted impersonation, so deleting
+      // must not fire mutations that the API would reject
+      if (restrictedImpersonation) {
```
wjames111: [Important] This `restrictedImpersonation` guard in `handleDeleteRequest` has zero test coverage — `AdditionalSalaryRequestContext.test.tsx` exists but was not modified, and `RequestPage.test.tsx` mocks this context (`jest.mock('../Shared/AdditionalSalaryRequestContext')`), so its restricted-impersonation tests only exercise disabled buttons, never this early-return. This guard pattern is the PR's load-bearing invariant across ~8 handlers. Recommend adding a case asserting `DeleteAdditionalSalaryRequest` is NOT fired with `impersonationScope` set, matching the pattern in the three `useSaveField` test files. (Flagged by Testing 7.5, defended in debate with strong Architecture support.)

src/components/HrTools/AdditionalSalaryRequest/Shared/useAdditionalSalaryRequestForm.ts
```diff
@@ -202,7 +204,9 @@ export const useAdditionalSalaryRequestForm = (
 
   const onSubmit = useCallback(
     async (values: CompleteFormValues) => {
-      if (!requestId) {
+      // The tool is read-only during restricted impersonation, so submitting
+      // must not fire mutations that the API would reject
+      if (restrictedImpersonation || !requestId) {
```
wjames111: [Important] Same as the context guard: this `onSubmit` early-return is fully unexercised — `useAdditionalSalaryRequestForm.test.tsx` has zero `restrictedImpersonation` references. Recommend: `mockSession({ impersonationScope: 'mpd_leader' })`, submit the form, and assert `mutationSpy` received no submit/update operation.

pages/api/utils/pagePropsHelpers.ts
```diff
@@ -54,8 +57,12 @@ export const blockImpersonatingNonDevelopers: GetServerSideProps<
     return loginRedirect(context);
   }
 
-  // Check if the impersonator is a developer
-  if (session.user.impersonating && !session.user.isImpersonatorDeveloper) {
+  // Check if the impersonator is a developer or has a restricted scope
+  if (
+    session.user.impersonating &&
+    !session.user.isImpersonatorDeveloper &&
+    !session.user.impersonationScope
```
wjames111: [Medium] Any truthy `impersonationScope` string now bypasses the developer gate for ~18 HR-tools/staff-expense pages — nowhere in the codebase is the scope compared to a value (verified: zero string comparisons at PR head). Ship-acceptable today (one scope exists, both gate directions fail safe), but a future API scope (`'coach'`, a debug value) silently inherits full HR-tools access, and this helper treats scope-truthy as ALLOW while `blockRestrictedImpersonation` treats it as BLOCK — inverted semantics of the same field selected by per-page import choice (the structural reason the `/tools` landing page was missed). Recommended consolidation: `type ImpersonationScope = 'mpd_leader'` union + one shared `isRestrictedImpersonation(session)` predicate used by both helpers and the client hook. One change collapses this, the useNavPages duplication, and the opaque typing. (Security 6.5 + Architecture 6.0 + Data Integrity expansion.)

src/hooks/useNavPages.tsx
```diff
@@ -59,6 +60,11 @@ export function useNavPages(coachingAccountCount: boolean, isSearch = false) {
   const developerBypass = useDeveloperBypass();
   const canSeeHrTools = userType === UserTypeEnum.UsStaff || developerBypass;
 
+  // During a restricted impersonation session the impersonator may only
+  // access MPD leader tools, so hide the contacts, tasks, and tools pages
+  const { impersonationScope } = useRequiredSession();
+  const isRestrictedImpersonation = !!impersonationScope;
```
wjames111: [Medium] This inlines `!!impersonationScope` instead of calling the `useRestrictedImpersonation()` hook added in this same PR — the only one of 14 consumers not using it, creating two sources of truth for an authorization signal. If the definition ever hardens to an allow-list, this site will silently drift.

```suggestion
  const isRestrictedImpersonation = useRestrictedImpersonation();
```
(plus the import; drop `useRequiredSession` if unused). (Architecture 6.5, Security support.)

wjames111 (commented):
## Multi-Agent Code Review — BLOCKERS FOUND

**Mode:** standard · **Agents:** 7 specialized reviewers + dependency-impact + 2 gap reviewers, with cross-examination debate · **Risk:** CRITICAL (10/10 — NextAuth session + impersonation authorization, 61 files)

### Verdict: 1 high-priority blocker

The core of the PR is solid — the HMAC-signed scope trust chain is sound, all 15 contacts/tasks/tools **sub**-pages are correctly blocked, read-only guards are correctly placed with correct dependency arrays, i18n is complete, the dynamic-modal pattern is followed exactly, and the Standards checklist passes across the board. One gap must be fixed before merge (below), plus one strongly-recommended test gap and a handful of medium/suggestion items posted as inline comments.

---

**[High] `pages/accountLists/[accountListId]/tools.page.tsx:46` — Tools landing page is not blocked during restricted impersonation**
- Severity: 8.0/10 — **BLOCKER** (flagged by Dependency Impact, escalated by Security, supported by all 6 debating agents; verified twice at PR head)
- All 15 contacts/tasks/tools sub-pages got `blockRestrictedImpersonation`, but the `/tools` landing page still exports `getServerSideProps = ensureSessionAndAccountList`. The nav tab is hidden, but a bookmarked/typed URL renders the full Tools home during a restricted impersonation session. `ToolsHome` unconditionally fires `GetToolNotifications`, which queries the impersonated staffer's `contacts`/`people` aggregate counts (partner-status/address/email/phone filters) — exactly the data class the block exists to withhold. This contradicts the commit "Block contacts, tasks, and tools routes during restricted impersonation."
- **Fix (one line):** swap `ensureSessionAndAccountList` → `blockRestrictedImpersonation` in `tools.page.tsx`, matching its 15 siblings. Consider also asserting the guard in `tools.page.test.tsx` the way `hrTools/mpdGoalAdmin/index.page.test.tsx` asserts its `getServerSideProps`.

_(This file is not in the PR diff, so it cannot carry a line comment — but it is counted as a blocker because it is a missing change required by the PR's stated goal: 16 routes in the access matrix, 15 guarded, 1 missed.)_

---

### Debate resolutions worth knowing
- **Canonical scope value:** the wire value is `impersonation_scope: 'mpd_leader'`. `mpdSupervisorAdmin` is the *admin capability* flag — a different concept. Test fixtures currently use `'mpd_supervisor'` and `'mpd_leader'` interchangeably for the scope (harmless today since only truthiness is checked, but worth a shared fixture before any allow-list lands).
- **TransfersTable unguarded handlers:** debate verified MUI's disabled `IconButton` genuinely blocks the child-SVG `onClick` via mouse AND keyboard — so this is a defense-in-depth/consistency gap, not a live bypass (down-revised 6.5 → 6.0 merged finding).
- **Server-side enforcement:** confirmed posture — the API rejects restricted mutations; all client guards are defense-in-depth.

### Pre-existing issues (informational, separate PR)
- **6.0** `hrTools/mpdGoalAdmin/scenario/[scenarioGoalId]/index.page.tsx` uses bare `ensureSessionAndAccountList` — no admin gate or restriction block, now more meaningful since mpdGoalAdmin hosts the impersonate action.
- **3.5** `impersonateHelper.ts:118` sets handoff cookies even when the token fetch fails (the PR's new scope cookie is correctly guarded).
- Shared `Modal` lacks explicit `aria-labelledby`.

### Verified clean
Scope trust chain (HMAC-signed cookie, tamper-rejection tested, `JWT_SECRET` server-only, no token logging) · `mpdSupervisorAdmin` exists in the schema, `yarn gql` will pass, `id` in selection set · guard placement precedes all optimistic state, dependency arrays correct · no financial calculation code touched · i18n byte-identical across all 7 read-only tooltip sites · no debug output / `any` / `new Date()` / empty catches.

src/components/HrTools/SavingsFundTransfer/Table/TransfersTable.tsx
```diff
@@ -186,6 +188,7 @@ export const TransfersTable: React.FC<TransfersTableProps> = ({
     handleFailedTransferOpen,
     t,
     locale,
+    restrictedImpersonation,
```
wjames111: [Medium] `restrictedImpersonation` is threaded here only for disabling row icons — but `handleAcceptCalendar` (fires `updateRecurringTransfer`) and the delete-modal mutation have no handler-level guard, making these the only two write paths in the PR relying solely on disabled buttons. Debate verified the disabled gating is currently airtight (MUI disabled suppresses child-SVG clicks, mouse and keyboard), so this is a defense-in-depth/consistency gap, not a live bypass — but it is one refactor away from a live write. Fix: add `if (restrictedImpersonation) return;` to both handlers for parity with the other ~8 guarded paths, and add a `TransfersTable.test.tsx` case asserting the action buttons are disabled under `impersonationScope`. (UX + Testing, merged finding.)

pages/api/auth/[...nextauth].page.ts
```diff
@@ -273,10 +283,12 @@ const Auth = (req: NextApiRequest, res: NextApiResponse): Promise<void> => {
             ...session.user,
             admin,
             developer,
+            mpdSupervisorAdmin,
             apiToken,
             userID,
             impersonating,
             isImpersonatorDeveloper,
+            impersonationScope,
```
Comment (wjames111) [inline pages/api/auth/[...nextauth].page.ts:292]: [Medium] JWTs minted before this deploys will lack `mpdSupervisorAdmin`/`impersonationScope` until the next token refresh, so consumers see `undefined`. This fails closed on both axes (falsy → fewer privileges, not restricted) and self-heals, so no action is strictly required — but a `mpdSupervisorAdmin: mpdSupervisorAdmin ?? false` default when spreading into `session.user` would make the migration window deterministic. (Data Integrity, self-revised 6.5 → 5.5 in debate.)

src/components/HrTools/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
```diff
@@ -128,7 +131,8 @@ export const HoursPerWeekGrid: React.FC<HoursPerWeekGridProps> = ({
 
   const saveHoursItem = useCallback(
     async (entry: HoursPerWeekEntry, currentEntries: HoursPerWeekEntry[]) => {
-      if (!calculation) {
+      // Read-only during restricted impersonation
+      if (restrictedImpersonation || !calculation) {
```
wjames111: [Medium] Four new restricted-impersonation guards in this file (`saveHoursItem`, `addEntry`, `deleteEntry`, `isCellEditable`) plus the disabled Add Entry button are all untested — `HoursPerWeekGrid.test.tsx` exists but was not modified. Recommend a `describe('restricted impersonation')` block with `mockSession({ impersonationScope: 'mpd_leader' })` asserting Add Entry is disabled and cell edits don't fire mutations. (Testing 6.5; UX challenged to 5.0 as branch-free prop threading — recorded at 6.0.)

src/components/Reports/Shared/GoalCard/GoalCard.tsx
```diff
@@ -63,6 +63,7 @@ export interface GoalCardProps {
   viewHref: string;
   onDelete: () => Promise<void>;
   badge?: React.ReactNode;
+  deleteDisabled?: boolean;
```
wjames111: [Medium] The new `deleteDisabled` path (disabled Delete + tooltip + `text.disabled` color) is untested at every layer — `GoalCard.test.tsx`, `MpdGoalCard.test.tsx`, and `PdsGoalCard.test.tsx` all exist and were not modified, unlike the sibling Create button in `GoalsList` which did get a test. A single `it('disables Delete when deleteDisabled is true')` in `GoalCard.test.tsx` covers all three consumers. (Testing 6.0 / UX 5.0.)

src/components/HrTools/MpdGoalAdmin/GoalsTable/GoalsTable.tsx
```diff
@@ -37,12 +42,19 @@ export const DEFAULT_ROWS_PER_PAGE = 5;
...
   const [page, setPage] = useState(0);
   const [rowsPerPage, setRowsPerPage] = useState(DEFAULT_ROWS_PER_PAGE);
   // The staff row whose coach is being assigned; null when the modal is closed.
   const [coachRow, setCoachRow] = useState<StaffGoalRow | null>(null);
+  // The staff row being impersonated; null when the modal is closed.
+  const [impersonateRow, setImpersonateRow] = useState<StaffGoalRow | null>(
+    null,
+  );
+
+  const canImpersonate = admin || mpdSupervisorAdmin;
```
wjames111: [Suggestion] `canImpersonate = admin || mpdSupervisorAdmin` doesn't exclude sessions that are already impersonating, so the Impersonate action could be offered mid-impersonation (nested impersonation UI). The server backstops the authorization and the impersonated session normally lacks these flags, so this is UI hygiene: consider `(admin || mpdSupervisorAdmin) && !impersonating`. (Security 4.5; Architecture challenged to 3.0 — recorded as suggestion.)

src/hooks/useSettingsNavItems.ts
```diff
@@ -59,7 +59,7 @@ export function useSettingsNavItems(): NavItems[] {
     {
       id: 'admin',
       title: t('Admin Console'),
-      grantedAccess: ['admin', 'developer'],
+      grantedAccess: ['admin', 'developer', 'mpdSupervisorAdmin'],
```
wjames111: [Suggestion] `grantedAccess` still includes `'developer'`, but `admin.page.tsx` now guards with `enforceAdminOrMpdLeader`, which bounces a developer-only account — leaving them a dead nav link. Partly pre-existing (the old `enforceAdmin` had the same mismatch), so no regression, but worth reconciling: either drop `'developer'` here or include developers in the guard, per intent. (Gap review.)

src/components/HrTools/Shared/CalculationReports/DirectionButtons/DirectionButtons.tsx
```diff
@@ -139,15 +139,19 @@ export const DirectionButtons: React.FC<DirectionButtonsProps> = ({
...
-              {t('Submit')}
-            </Button>
+            <Tooltip title={disableNext ? (disabledNextTooltip ?? '') : ''}>
+              <Box component="span">
+                <Button
+                  variant="contained"
+                  color="primary"
+                  endIcon={<ChevronRight />}
+                  onClick={handleSubmit}
+                  disabled={disableNext || (submitCount ? !isValid : false)}
```
wjames111: [Suggestion] The new Submit-branch `disableNext` composition is covered at the RequestPage integration layer but not in `DirectionButtons.test.tsx`, whose existing `disableNext` tests only exercise the Continue branch. A small unit case (`isSubmission + disableNext` → Submit disabled) would protect the other consumers (MHA steps, PDS calculator) that RequestPage tests don't reach. (Testing, self-revised 6.5 → 4.5 in debate.)

wjames111:
🤖 Fixed blocker in b499a6561: `tools.page.tsx` getServerSideProps now uses `blockRestrictedImpersonation` (matching its 15 sub-page siblings), with a test asserting the guard. _Posted by Claude Code via /agent-fix._

src/components/HrTools/AdditionalSalaryRequest/Shared/AdditionalSalaryRequestContext.tsx
wjames111: 🤖 Fixed in 51dce70ee: added a restricted-impersonation describe block asserting handleDeleteRequest fires no DeleteAdditionalSalaryRequest mutation (cancel and discard paths)

src/components/HrTools/AdditionalSalaryRequest/Shared/useAdditionalSalaryRequestForm.ts
wjames111: 🤖 Fixed in 3ba67c1b8: added an onSubmit test asserting neither UpdateAdditionalSalaryRequest nor SubmitAdditionalSalaryRequest fires during restricted impersonation (validation confirmed clean first so the guard is what blocks; placed last in the describe per the mockSession convention, see 2a1caf42e)

pages/api/utils/pagePropsHelpers.ts
wjames111: 🤖 Fixed in 407f1c1a0: extracted a shared isRestrictedImpersonation(user) predicate + ImpersonationScope ('mpd_leader') union in src/lib/restrictedImpersonation.ts with the documented access matrix; both helpers and the client hook now use it (behavior unchanged)

src/hooks/useNavPages.tsx
wjames111: 🤖 Fixed in 6e9b5ac6a: useNavPages now calls useRestrictedImpersonation() instead of inlining the scope check

src/components/HrTools/SavingsFundTransfer/Table/TransfersTable.tsx
wjames111: 🤖 Fixed in b9e937715: added early-return guards to handleAcceptCalendar and handleDeleteModalOpen, plus a TransfersTable test asserting the row action buttons are disabled during restricted impersonation

pages/api/auth/[...nextauth].page.ts
Comment (wjames111) [reply pages/api/auth/[...nextauth].page.ts:292]: 🤖 Fixed in e7d9a966c: session callback now defaults mpdSupervisorAdmin to false so pre-deploy JWTs yield a deterministic boolean

src/components/HrTools/PdsGoalCalculator/Setup/HoursPerWeekGrid/HoursPerWeekGrid.tsx
wjames111: 🤖 Fixed in 1b497d06e: added a restricted-impersonation describe block: Add Entry disabled, delete disabled, and cell-edit attempts open no editor and fire no mutations

src/components/Reports/Shared/GoalCard/GoalCard.tsx
wjames111: 🤖 Fixed in 5f9a9463b: added GoalCard tests asserting Delete is disabled when deleteDisabled is true and enabled by default

src/components/HrTools/MpdGoalAdmin/GoalsTable/GoalsTable.tsx
wjames111: 🤖 Fixed in 3f607eb01: canImpersonate now excludes already-impersonating sessions, with a test asserting the action is hidden while impersonating

src/hooks/useSettingsNavItems.ts
wjames111: 🤖 Fixed in ff99d0cb8: dropped 'developer' from the Admin Console grantedAccess (the server guard never admitted developer-only accounts) and updated the MultiPageMenu test

src/components/HrTools/Shared/CalculationReports/DirectionButtons/DirectionButtons.tsx
wjames111: 🤖 Fixed in 4010f6482: added a unit test asserting isSubmission + disableNext disables the Submit button and shows the tooltip

src/components/HrTools/AdditionalSalaryRequest/Shared/useAdditionalSalaryRequestForm.test.tsx
```diff
@@ -765,6 +766,40 @@ describe('useAdditionalSalaryRequestForm', () => {
       );
       expect(mockHandleNextStep).not.toHaveBeenCalled();
     });
+
+    it('should not fire mutations when submitting during restricted impersonation', async () => {
+      mockSession({ impersonationScope: 'mpd_supervisor' });
```
wjames111: [Medium] This `mockSession` return value persists for the rest of the file — `clearMocks`/`clearAllMocks` only clear calls, not `mockReturnValue`. The test is placed last in the `onSubmit` describe as mitigation, but the `formik integration` describe still runs afterward with the restricted session (benign today only because none of those tests submit). One-line hardening: add `mockSession({});` to the file's `beforeEach` so every test starts from a known session regardless of order. (Testing agent, 5.0 — non-blocking.)

src/lib/restrictedImpersonation.ts
```diff
@@ -0,0 +1,36 @@
...
+ * The impersonation scope values the API currently mints. As of now the only
+ * value produced by the impersonation handoff is 'mpd_leader' (see
+ * pages/api/auth/impersonate/impersonateHelper.ts, which reads
+ * `impersonation_scope` from the API response). The session field itself is
+ * typed as `string` in pages/api/auth/[...nextauth].page.ts because the value
+ * arrives from a signed cookie minted by the API, but any code that needs to
+ * branch on a specific scope should compare against this union so that new
+ * scope values are added here deliberately.
+ */
+export type ImpersonationScope = 'mpd_leader';
```
wjames111, line 13: [Suggestion] `ImpersonationScope` is exported but has zero consumers — the predicate uses truthiness and the session field stays `string`. The JSDoc frames it as the deliberate registry for future scopes, which is fine to land; if you want it live now, a typed comparison helper (e.g. `isScope(user, scope: ImpersonationScope)`) or narrowing the session type would give the union a job. (Architecture agent, 3.5 — optional.)

src/components/HrTools/MpdGoalAdmin/GoalsTable/GoalsTable.test.tsx
```diff
@@ -164,4 +171,62 @@ describe('GoalsTable', () => {
...
+      expect(
+        queryByRole('button', { name: 'Impersonate' }),
+      ).not.toBeInTheDocument();
+    });
+
+    it('hides the Impersonate action while already impersonating', () => {
+      mockSession({
+        mpdSupervisorAdmin: true,
+        admin: true,
+        impersonating: true,
```
wjames111, line 207: [Suggestion] This test asserts Impersonate is hidden while impersonating; adding one assertion that the disabled View/Edit link still renders would also lock in the non-empty-column invariant. (UX agent, 3.0 — optional.)

wjames111 (commented):
## Multi-Agent Code Review — Round 2 (re-review after fixes) — APPROVED WITH SUGGESTIONS

**Mode:** standard · **Agents:** 7 specialized reviewers + dependency impact · **Risk:** CRITICAL (10/10 — auth/impersonation surface; human review always required at this risk level)

### Verdict: 0 blockers — all 12 round-1 findings VERIFIED-FIXED

Every round-1 finding, including the 8.0 blocker, was independently re-verified by multiple agents:
- **Blocker (tools.page.tsx)**: now `blockRestrictedImpersonation`, identical to its 15 siblings, with an identity-assertion test. The 59-page route audit confirms contacts/tasks/tools coverage is **complete**.
- **Shared predicate** (`src/lib/restrictedImpersonation.ts`): fully adopted — whole-tree grep finds **zero** remaining inline `!!impersonationScope` checks; both SSR helpers and the client hook route through it; behavior byte-identical; server/client import boundary safe (type-only next-auth import). Net tech debt reduced.
- **All 8 test-gap fixes**: verified non-vacuous (each asserts the negative against a positive baseline); suites pass.
- **Merge of origin/main**: no auth-relevant, financial, or structural changes entered the PR diff; codegen and lint:ts now pass cleanly.

### Remaining suggestions (non-blocking, inline comments)
1. (5.0) Harden the `mockSession` leakage mitigation in `useAdditionalSalaryRequestForm.test.tsx` — reset the session in `beforeEach` instead of relying on test order.
2. (3.5) `ImpersonationScope` union is exported but unreferenced — forward-looking by design; optionally give it a live consumer.
3. (3.0) GoalsTable test could also assert View/Edit still renders while impersonating.

### Informational / pre-existing (no action in this PR)
- (4.0) [Pre-existing] Two profile menus (`ProfileMenu.tsx:364`, `ProfileMenuPanel.tsx:166`) still show the Admin Console link to `developer`-only users, who get bounced by `enforceAdminOrMpdLeader` — the same dead-link UX fixed in `useSettingsNavItems`. Suggest a follow-up ticket to align all three surfaces.
- (3.0) [Pre-existing, from main merge] `coaching/[coachingId]/nsGoalCalculator/index.page.tsx` uses bare `ensureSessionAndAccountList` with no impersonation gate — authored on main (commit 9cda1b782), not by this PR. Confirm coaching-scoped calculators are intentionally outside the restricted-impersonation surface.
- (3.0) The `?? false` default in the session callback is unreachable per the JWT type but correct as runtime defense for pre-deploy tokens — intentional, keep.
- Known items carried forward: scenario page guard, impersonateHelper failure-path cookies, fixture value `'mpd_supervisor'` vs wire value `'mpd_leader'` (truthiness-only today).

### Standards
Full checklist PASS. Lint, type-check, codegen, and all touched test suites pass (verified by orchestrator).

src/components/HrTools/AdditionalSalaryRequest/Shared/useAdditionalSalaryRequestForm.test.tsx
wjames111: 🤖 Fixed in 42dd3bf85: added mockSession({}) to the file-level beforeEach so every test starts from the default session regardless of order

src/lib/restrictedImpersonation.ts
wjames111: 🤖 Fixed in be640d5c0: exported MPD_LEADER_SCOPE: ImpersonationScope and consumed it in the restrictedImpersonation and pagePropsHelpers test fixtures — the union now has a live compile-time consumer pinning the wire value; session augmentation stays string since the cookie value is unvalidated upstream input

src/components/HrTools/MpdGoalAdmin/GoalsTable/GoalsTable.test.tsx
wjames111: 🤖 Fixed in a41626805: the impersonating test now also asserts the disabled View/Edit link renders on every row, locking the non-empty-column invariant

src/hooks/useRestrictedImpersonation.ts
```diff
@@ -0,0 +1,12 @@
+import { isRestrictedImpersonation } from 'src/lib/restrictedImpersonation';
+import { useRequiredSession } from './useRequiredSession';
+
+// During a restricted MPD-leader impersonation session the API rejects all
+// mutations except the ones powering the editable MPD leader tools, so the
+// remaining HR tools must render read-only and never fire mutations.
+// Shares its definition of "restricted" with the server-side page guards via
+// isRestrictedImpersonation — see the access matrix documented there.
+export const useRestrictedImpersonation = (): boolean => {
```
canac, line 9: It's going to be annoying and error-prone to have to check this before every mutation in our application.

pages/api/auth/[...nextauth].page.ts
```diff
@@ -273,10 +283,13 @@ const Auth = (req: NextApiRequest, res: NextApiResponse): Promise<void> => {
             ...session.user,
             admin,
             developer,
+            // Default to false for JWTs minted before mpdSupervisorAdmin existed
+            mpdSupervisorAdmin: mpdSupervisorAdmin ?? false,
```
Comment (canac) [inline pages/api/auth/[...nextauth].page.ts]: If the frontend needs to check for one more role (4+), I'd consider making this an enum.

src/hooks/useSettingsNavItems.ts
```diff
@@ -59,7 +59,7 @@ export function useSettingsNavItems(): NavItems[] {
     {
       id: 'admin',
       title: t('Admin Console'),
-      grantedAccess: ['admin', 'developer'],
+      grantedAccess: ['admin', 'mpdSupervisorAdmin'],
```
canac: Is this right? Why are we removing `developer`?

canac (commented):
It's a lot of code, so I focused on the hotspots and had Claude review the rest. I'm concerned about needing to check the impersonation scope in so many parts of the application and in lots of new code. That seems like a maintenance nightmare, so I'd like to think through alternatives before committing to that route.

The server blocks mutations, right? So what do you think allowing mutations, know that the server will reject them? We can just make sure that the server returns an understandable error message like, "Changes are not allowed during impersonation".

wjames111:
@canac allowing mutations works for simplifying things a little but. but we still have the PII leaking everywhere across the app, so either way it's kind of a nightmare. Could we just give them access to SAA? Architecturally this is a really bad request

src/hooks/useSettingsNavItems.ts
wjames111: Thanks for catching this!

src/hooks/useRestrictedImpersonation.ts
wjames111: I think your suggestion of just relying on the back-end to block it works better.

canac:
> but we still have the PII leaking everywhere across the app, so either way it's kind of a nightmare.

Doesn't your backend PR redact PII? Or are you saying there are more fields that could include PII?

src/hooks/useRestrictedImpersonation.ts
wjames111: 🤖 Fixed in 5a8888292: removed the per-mutation frontend guards and disabled states from the HR tools — the API's mutation allowlist is now the sole enforcement layer. Nav/route hiding for contacts/tasks/tools remains.

_Posted by Claude Code._

src/hooks/useSettingsNavItems.ts
wjames111: 🤖 Fixed in 9ac10be41: restored 'developer' in the Admin Console grantedAccess and admitted developers in enforceAdminOrMpdLeader so the nav item and page guard agree.

_Posted by Claude Code._

pages/api/auth/[...nextauth].page.ts
Comment (wjames111) [reply pages/api/auth/[...nextauth].page.ts]: Agreed — the plan is to finish the role column work in CruGlobal/mpdx_api#3285, migrate admins/developers/MPD supervisors to it, and have the frontend read a single role value instead of these booleans. Leaving the booleans as-is here so the swap happens once, in the role-migration PR.

_Posted by Claude Code._

PR #1927, opened 2026-07-17
MPDX-9862 - Change NSO Questionnaire to non-linear form

zweatshirt:
## Description

- Allow users to skip around the form, enforcing form completion before the submit button becomes enables
- Increase user awareness of unfilled fields through a warning about the submit button, and using error color font in the Summary tables.
- https://jira.cru.org/browse/MPDX-9862

Claude flagged this PR as critical so it cannot be merged until approved by a human.

## Testing

- Go to NSO MPD Questionnaire, preferably with an empty form
- Half fill the form
- Check the submission view and whether the design is sufficient for our needs.

## Checklist:

### DIFF
--- a/src/components/HrTools/NsoMpdQuestionnaire/NsoMpdQuestionnaire.test.tsx
+++ b/src/components/HrTools/NsoMpdQuestionnaire/NsoMpdQuestionnaire.test.tsx
@@ -65,4 +65,20 @@ describe('NsoMpdQuestionnaire', () => {
       queryByRole('heading', { name: 'Staff Information' }),
     ).not.toBeInTheDocument();
   });
+
+  it('jumps to an incomplete section from the Summary warning', async () => {
+    const { findByRole, getByText, getByRole } = render(<TestComponent />);
+
+    // Jump straight to the Summary step via the rail.
+    userEvent.click(await findByRole('button', { name: 'Summary' }));
+
+    // The default mock leaves the debt fields empty, so Financial is flagged.
+    userEvent.click(getByRole('button', { name: 'Financial Information' }));
+
+    // Assert on copy unique to the Financial step (absent from the Summary page),
+    // so a broken warning link that never navigates would fail this test.
+    expect(
+      getByText('Tell us about your financial situation.'),
+    ).toBeInTheDocument();
+  });
 });
--- a/src/components/HrTools/NsoMpdQuestionnaire/NsoMpdQuestionnaire.test.tsx
+++ b/src/components/HrTools/NsoMpdQuestionnaire/NsoMpdQuestionnaire.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render, waitFor } from '@testing-library/react';
+import { render } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { NsoMpdQuestionnaire } from './NsoMpdQuestionnaire';
 import {
@@ -27,17 +27,14 @@ describe('NsoMpdQuestionnaire', () => {
   it('renders the matching step page after continuing to the next step', async () => {
     const { findByRole } = render(<TestComponent />);
 
-    // Personal Information is a review-only step, so Continue enables once it registers.
-    const continueButton = await findByRole('button', { name: 'Continue' });
-    await waitFor(() => expect(continueButton).toBeEnabled());
-    userEvent.click(continueButton);
+    userEvent.click(await findByRole('button', { name: 'Continue' }));
 
     expect(
       await findByRole('heading', { name: 'Ministry Information' }),
     ).toBeInTheDocument();
   });
 
-  it('enables Continue on the review-only first step and gates it on the next', async () => {
+  it('keeps Continue enabled even when the current step has missing fields', async () => {
     const { findByRole, getByRole } = render(
       <TestComponent
         newStaffQuestionnaire={{
@@ -49,17 +46,13 @@ describe('NsoMpdQuestionnaire', () => {
       />,
     );
 
-    // Personal Information has nothing to fill in, so Continue enables.
     const continueButton = await findByRole('button', { name: 'Continue' });
-    await waitFor(() => expect(continueButton).toBeEnabled());
     userEvent.click(continueButton);
 
     expect(
       await findByRole('heading', { name: 'Ministry Information' }),
     ).toBeInTheDocument();
-    await waitFor(() =>
-      expect(getByRole('button', { name: 'Continue' })).toBeDisabled(),
-    );
+    expect(getByRole('button', { name: 'Continue' })).toBeEnabled();
   });
 
   it('shows the informational view when there is no open questionnaire', async () => {
--- a/src/components/HrTools/NsoMpdQuestionnaire/Summary/Summary.test.tsx
+++ b/src/components/HrTools/NsoMpdQuestionnaire/Summary/Summary.test.tsx
@@ -13,6 +13,12 @@ import { Summary } from './Summary';
 
 const mutationSpy = jest.fn();
 
+const filledDebtFields = {
+  studentLoanMonthlyPayment: 0,
+  carLoanMonthlyPayment: 0,
+  creditCardDebtMonthlyPayment: 0,
+};
+
 const TestComponent: React.FC<
   Omit<NsoMpdQuestionnaireTestWrapperProps, 'children'>
 > = (props) => (
@@ -144,9 +150,14 @@ describe('Summary', () => {
 
   it('submits via the confirmation modal', async () => {
     const { findByRole, getByRole } = render(
-      <TestComponent onCall={mutationSpy} />,
+      <TestComponent
+        onCall={mutationSpy}
+        newStaffQuestionnaire={filledDebtFields}
+      />,
     );
-    userEvent.click(await findByRole('button', { name: 'Submit' }));
+    const submitButton = await findByRole('button', { name: 'Submit' });
+    await waitFor(() => expect(submitButton).toBeEnabled());
+    userEvent.click(submitButton);
     const dialog = getByRole('dialog');
     userEvent.click(within(dialog).getByRole('button', { name: 'Submit' }));
     await waitFor(() =>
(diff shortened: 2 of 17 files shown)


zweatshirt:
@canac I'll run an AI review Monday morning, I am mostly curious to get your opinion on the user experience

zweatshirt (commented):

**6 specialized agents + dependency analysis** reviewed this PR. **No blockers.** The change is well-isolated (all dependents live within the `NsoMpdQuestionnaire` feature), additive / behavior-only (no breaking API changes — `SummaryRow.required?` is additive, `isQuestionnaireComplete` is a new export, `ContinueButton` keeps its props), and does no financial arithmetic (Financial Reporting agent: "No financial calculation code in this PR"). The shared `Autosave/**` is untouched, so SalaryCalculator / PdsGoalCalculator are unaffected.

**Risk:** 9/10 by the rules formula, but that is breadth-driven (8 feature files × +1) — there is no security / auth / GraphQL / cache / schema / dependency surface. Practical risk is low.

### Findings on Related Files (not in this PR diff)

**Orphaned autosave wiring — `src/components/HrTools/NsoMpdQuestionnaire/FinancialInformation/DebtQuestions.tsx:13,75-83`** · ~6.0 · flagged by **Architecture + Standards + Dependency** (3-agent consensus)

This file still imports `useOptionalAutosaveForm` and runs a `markValid('hasDebt')` / `markInvalid('hasDebt')` `useEffect`. This PR removed the only `<AutosaveForm>` provider, so that hook now always returns `null` and the effect is a **permanent no-op** — dead code that reads as live validity-gating. The refactor tore this wiring out of `ContinueButton`, `PersonalInformation`, and `NsoMpdQuestionnaireLayout` but left it here. Recommend folding the cleanup into this PR: delete the import + effect (keep the local `hasDebt` / `hasDebtTouched` field validation, which still drives the "Please select an answer." helper text). It cannot be posted as a line comment because the file isn't in this PR's diff.

### Inline comments

7 findings posted as line comments: **1 Important (7.0)**, 3 Medium, 3 Suggestions. The Important item is a non-discriminating navigation test.

> ⚠️ 1 Important finding (severity 7.0) — strongly recommended, does not block merge, and cannot be dismissed via `/dismiss`. Consider fixing it here or tracking a follow-up.

src/components/HrTools/NsoMpdQuestionnaire/NsoMpdQuestionnaire.test.tsx
```diff
@@ -72,4 +65,20 @@ describe('NsoMpdQuestionnaire', () => {
...
+    // Jump straight to the Summary step via the rail.
+    userEvent.click(await findByRole('button', { name: 'Summary' }));
+
+    // The default mock leaves the debt fields empty, so Financial is flagged.
+    userEvent.click(
+      await findByRole('button', { name: 'Financial Information' }),
+    );
+
+    expect(
+      await findByRole('heading', { name: 'Financial Information' }),
```
zweatshirt: [Important] This assertion doesn't actually verify the navigation. Every `SummarySection` renders an `h6` with its title, so a `heading` named "Financial Information" already exists on the Summary page *before* the warning link is clicked — a broken link `onClick` (no navigation) would still pass this test. Assert on content unique to the Financial **step** instead, e.g. `expect(await findByText('Tell us about your financial situation.')).toBeInTheDocument();` (that intro copy is not present on the Summary page).

src/components/HrTools/NsoMpdQuestionnaire/Summary/Summary.test.tsx
```diff
@@ -180,4 +200,32 @@ describe('Summary', () => {
...
+  it('disables Submit and warns when required fields are missing', async () => {
+    const { findByRole, getByRole } = render(<TestComponent />);
+
+    expect(await findByRole('alert')).toHaveTextContent(
+      'Your form is missing information.',
+    );
+    expect(getByRole('button', { name: 'Submit' })).toBeDisabled();
+  });
+
+  it('links each incomplete section in the warning to its step', async () => {
```
zweatshirt: [Medium] This test only asserts the Financial link is *present* — it never verifies the filtering, which is the actual new behavior. `useSummarySections` returns all four sections and `isStepComplete(step, null)` is `false` during the query load, so all four links render while the questionnaire is still loading; `findByRole` can resolve against that transient state. Add assertions that completed sections (Ministry / NSO / Personal) are **absent** after the questionnaire loads (`await waitFor(() => expect(queryByRole('button', { name: 'Ministry Information' })).not.toBeInTheDocument())`).

src/components/HrTools/NsoMpdQuestionnaire/Summary/Summary.tsx
```diff
@@ -17,12 +21,17 @@ export const Summary: React.FC = () => {
...
   const { enqueueSnackbar } = useSnackbar();
   const accountListId = useAccountListId();
-  const { handleStepChange, handleBack, completeQuestionnaire } =
+  const { questionnaire, handleStepChange, handleBack, completeQuestionnaire } =
     useNsoMpdQuestionnaire();
   const [confirmOpen, setConfirmOpen] = useState(false);
 
   const sections = useSummarySections();
 
+  const canSubmit = isQuestionnaireComplete(questionnaire);
```
zweatshirt: [Medium] The Submit gate uses `isQuestionnaireComplete` (which iterates `stepCompletion.ts`'s `steps`) while `incompleteSections` (next line) filters `useSummarySections()` — two independently-maintained step lists. The completeness *rules* are centralized in `isStepComplete` (good), but if the lists ever drift (a future data-entry step added to one and not the other), the gate and the actionable warning diverge (Submit stuck disabled with nothing listed, or a section listed that never blocks). Not a bug today. Consider a single `getIncompleteSteps(questionnaire)` helper in `stepCompletion.ts` as the source for both `canSubmit` and the warning list.

src/components/HrTools/NsoMpdQuestionnaire/Summary/SummarySection.tsx
```diff
@@ -80,7 +82,11 @@ export const SummarySection: React.FC<SummarySectionProps> = ({
                     component="span"
                     variant="body2"
                     fontStyle="italic"
-                    color="text.secondary"
+                    sx={{
+                      color: row.required
```
zweatshirt, line 86: [Medium] Required vs optional empty rows are distinguished by color only — identical `"No value provided"` text in `error.main` (required) vs `text.secondary` (optional). That's a WCAG 1.4.1 (use of color) concern at the row level. It's mitigated by the section-level `Alert` + disabled Submit (so non-blocking), but consider adding a non-color cue on required-empty rows — a small warning icon or a `(required)` marker — gated on `row.required`.

src/components/HrTools/NsoMpdQuestionnaire/Summary/useSummarySections.ts
```diff
@@ -216,5 +216,16 @@ export const useSummarySections = (): SummarySectionData[] => {
...
     ];
+
+    // Personal Information is a read-only review step; every other section's rows are
+    // user-enterable and required, so an empty one should be flagged in the Summary.
+    return sections.map((section) =>
+      section.step === NsoMpdQuestionnaireStepEnum.PersonalInformation
+        ? section
+        : {
+            ...section,
+            rows: section.rows.map((row) => ({ ...row, required: true })),
```
zweatshirt, line 227: [Suggestion] Blanket-marking every non-Personal row `required: true` is hand-maintained parity with `stepCompletion`'s `stepRequiredFields` + `variantRequiredFields`. It holds across all variants today, but adding a display-only row here later would render it red ("No value provided" in error color) while `isStepComplete` ignores it — so it would look like a blocking error but never block Submit or appear in the warning. Deriving `required` from the same `getRequiredFields` source would keep the two in lockstep.

src/components/HrTools/NsoMpdQuestionnaire/Summary/Summary.tsx
```diff
@@ -56,9 +65,31 @@ export const Summary: React.FC = () => {
         </Stack>
       </Box>
 
+      {!canSubmit && (
+        <Alert severity="error" sx={{ mx: 4, '& ul': { m: 0, pl: 3 } }}>
```
zweatshirt, line 67: [Suggestion] `severity="error"` for a form that is merely *incomplete* rather than errored. `severity="warning"` is arguably the more accurate semantic (and this repo uses `severity="info"` for neutral questionnaire states). That said, `error` is defensible — it blocks submission and visually echoes the red required-empty rows. Author's call.

src/components/HrTools/NsoMpdQuestionnaire/Summary/Summary.tsx
```diff
@@ -56,9 +65,31 @@ export const Summary: React.FC = () => {
...
+          </ul>
+        </Alert>
+      )}
+
       <Stack direction={{ xs: 'column', sm: 'row' }} spacing={2} mx={4}>
         <BackButton onClick={handleBack} />
-        <QuestionnaireActionButton onClick={() => setConfirmOpen(true)}>
+        <QuestionnaireActionButton
+          onClick={() => setConfirmOpen(true)}
+          disabled={!canSubmit}
```
zweatshirt, line 89: [Suggestion] This Submit gate is client-only / advisory — a user can remove the `disabled` attribute or call `completeNewStaffQuestionnaire` directly with an incomplete questionnaire. The mutation routes to the external primary API (`api.mpdx.org`), so this is a data-quality gate, not an authorization boundary (account-list scoping + bearer auth still bound *who* can complete *which* questionnaire). Worth confirming with the backend team that the primary API rejects (or is intended to accept) completion when required fields are unset. Not verifiable in this repo.

src/components/HrTools/NsoMpdQuestionnaire/Summary/Summary.tsx
```diff
@@ -56,9 +63,31 @@ export const Summary: React.FC = () => {
         </Stack>
       </Box>
 
+      {!canSubmit && (
+        <Alert severity="error" sx={{ mx: 4, '& ul': { m: 0, pl: 3 } }}>
+          {t('Your form is missing information.')}
+          <ul>
+            {incompleteSections.map((section) => (
+              <li key={section.step}>
+                <Link
```
canac, line 72: praise: This is a really nice UX!

canac (approved):
I think this is great!

PR #1929, opened 2026-07-20
MPDX-9834 Add year filter to MPGA

kegrimes:
## Description

MPGA UAT issues:
- Add "Select Date Range" on report settings for MPGA
- Add "Year to Date" option, as well as the previous years where transactions occurred
- Grayed out future months if "Year to Date" filter is applied

Extra:
- Created a shared context to remove prop drilling
- Created a shared test wrapper for all tests

Jira ticket: [MPDX-9834](https://jira.cru.org/browse/MPDX-9834)

## Testing

Impersonate: courtney.campbell@cru.org

No filter (default) test:
- Check that the data is from the last 12 months (August 2025 - July 2026)

Year to Date filter test:
- Click "Report Settings" and select "Year to Date"
- Check that the data is from this year only (January 2026 - July 2026)
- Verify the future months are grayed out

Past year filter test:
- Click "Report Settings" and select "2025"
- Check that the data is from 2025 only (January 2025 - December 2025)

## Checklist:


src/components/Reports/MPGAIncomeExpensesReport/Tables/TableCard.tsx
```diff
@@ -237,14 +268,21 @@ export const TableCard: React.FC<TableCardProps> = ({
...
-        subtitle={t('Last 12 Months')}
         styling={{ padding: 0, '&:last-child': { paddingBottom: 0 } }}
       >
         <Box>
           <StyledGrid
             rows={cardTableRows}
             columns={columns}
             columnGroupingModel={columnGroupingModel}
+            sx={{
+              '& .future-month-header .MuiDataGrid-columnHeaderTitle': {
```
kegrimes: **[Medium] Future-month state is conveyed by color only (WCAG 1.4.1).** The `future-month-header` / `future-month` styling here (and the sibling gray treatments in `TotalRow.tsx:65,127` and `PrintTables.tsx:47-52,158,186`) uses only `text.disabled` / `action.hover` — no icon, text, or aria. A screen-reader or color-blind user can't distinguish a not-yet-occurred month from a genuine `$0`. Add a non-color cue: a visually-hidden label / `title` on future cells, or a legend/footnote ("Grayed columns haven't occurred yet").

src/components/Reports/MPGAIncomeExpensesReport/Tables/TableCard.tsx
```diff
@@ -180,7 +192,7 @@ export const TableCard: React.FC<TableCardProps> = ({
         headerAlign: 'right',
       },
     ];
-  }, [months, locale, t, description, average, total]);
+  }, [months, firstFutureMonthIndex, locale, t, description, average, total]);
```
kegrimes: **[Suggestion] Dep array uses `firstFutureMonthIndex` as a proxy for `isFutureMonth`.** The memo body calls `isFutureMonth(index)` but lists `firstFutureMonthIndex` in deps (and doesn't otherwise use it here). Functionally correct today because `isFutureMonth` is a `useCallback` keyed only on `firstFutureMonthIndex`, and `react-hooks/exhaustive-deps` isn't enforced — but it goes stale silently if `isFutureMonth` ever gains another dependency. Prefer depending on the `isFutureMonth` callback directly. (Same pattern at `TotalRow.tsx:113`.)

src/components/Reports/MPGAIncomeExpensesReport/Tables/PrintTables.tsx
```diff
@@ -15,30 +15,42 @@ import { zeroAmountFormat } from 'src/lib/intlFormat';
...
+    dataLoading,
+    monthLabels: months,
+    firstFutureMonthIndex,
+  } = useReport();
 
   const overallTotal =
     type === ReportTypeEnum.Income ? incomeTotal : expensesTotal;
 
+  const grayColor = theme.palette.text.disabled;
+  const futureCellSx = {
```
kegrimes, line 46: **[Medium] Print path: future months conveyed by color only (WCAG 1.4.1).** `futureCellSx` (and the gray text/underline below) marks future/unavailable months with gray alone. In print especially there's no other signal. Pair with the a11y fix on `TableCard.tsx` — a textual/`title` marker so future months read as "not yet occurred" rather than a real `$0`.

src/components/Reports/MPGAIncomeExpensesReport/Tables/TotalRow.tsx
```diff
@@ -105,7 +110,7 @@ export const TotalRow: React.FC<TotalRowProps> = ({ data, overallTotal }) => {
         renderCell: overall,
       },
     ];
-  }, [monthlyTotals, avgSum, overallTotal]);
+  }, [monthlyTotals, avgSum, overallTotal, firstFutureMonthIndex]);
```
kegrimes: **[Medium / pre-existing, extended by this PR] `columns` useMemo is defeated.** `monthlyTotals` (line 35) and `avgSum` (line 44) are recomputed as fresh references every render and are dependencies of this memo, so it re-runs every render and provides no benefit; this PR extends the memo (adds `firstFutureMonthIndex`, `isFutureMonth`). Also `locale` is used in `renderCell` but absent from deps — harmless only because the memo never actually memoizes. Fix: wrap `monthlyTotals` / `avgSum` in their own `useMemo([data])`, then this memo works as intended.

src/components/Reports/Shared/SettingsDialog/SettingsDialog.tsx
```diff
@@ -153,6 +168,12 @@ export const SettingsDialog: React.FC<SettingsDialogProps> = ({
     [time],
   );
 
+  // TODO: Get list of all possible years
```
kegrimes: **[Medium / Standards] `// TODO` without a ticket reference.** `// TODO: Get list of all possible years` violates the CLAUDE.md Code Quality rule (no `// TODO` without a Jira/MPDX reference). Either resolve it or reference a ticket: `// TODO(MPDX-XXXX): Get list of all possible years`.

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesContext/MPGAIncomeExpensesContext.tsx
```diff
@@ -0,0 +1,266 @@
...
+}
+
+const sum = (rows?: DataFields[]): number => {
+  return rows?.reduce((acc, item) => acc + item.total, 0) || 0;
+};
+
+export const ReportProvider: React.FC<ReportContextProps> = ({ children }) => {
+  const { t } = useTranslation();
+  const locale = useLocale();
+  const currency = 'USD';
```
kegrimes, line 72: **[Suggestion / pre-existing] Hardcoded `currency = 'USD'`.** Relocated here from the old component, so not introduced by this PR — but the report's whole purpose is financial accuracy, and `MPGATransactions` returns no `currencyCode`, so any non-USD staff account shows the wrong symbol. No client-side currency *mixing* occurs (totals are server-aggregated into a single figure), so this is a display-label issue. Worth a follow-up ticket. (`BreakdownModal.tsx:41` has the same constant, un-migrated.)

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesContext/MPGAIncomeExpensesContext.tsx
```diff
@@ -0,0 +1,266 @@
...
+  // If year to date filter is selected, get first month index in the future to gray out future months in the table
+  const firstFutureMonthIndex = isYearToDate ? now.month : undefined;
+
+  const isFutureMonth = useCallback(
+    (index: number) =>
+      firstFutureMonthIndex !== undefined && index >= firstFutureMonthIndex,
+    [firstFutureMonthIndex],
+  );
+
+  const { startDate, endDate } = useMemo(() => {
```
kegrimes, line 94: **[Suggestion] Redundant date-window derivation (latent, not a live bug).** `ReportProvider` re-derives `startDate`/`endDate` here from `effectiveYear` and ignores the `filters.startDate`/`filters.endDate` that `SettingsDialog.getFiltersWithCalculatedDates` already computed — two derivations of the same window. `ReportContext` clamps year-end to `now` while the dialog returns an unclamped Dec 31; adversarial verify confirmed this can't diverge today (the MPGA dropdown offers only completed past years + YTD, so the clamp always agrees). Also note `useGetLastTwelveMonths` reads its own `DateTime.now()` separate from the `now` pinned at line 72 — harmless but can drift across a month boundary. Consider a single source of truth for the window.

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesContext/MPGAIncomeExpensesContext.tsx
```diff
@@ -0,0 +1,266 @@
...
+        endDate.month,
+        endDate.year,
+        locale,
+        true,
+        true,
+      ),
+    });
+  }, [selectedYear, isYearToDate, startDate, endDate, locale, t]);
+
+  const { data: reportData, loading } = useMpgaTransactionsQuery({
```
kegrimes, line 134: **[Suggestion] Provider concentrates many concerns.** `ReportProvider` fetches (`useMpgaTransactionsQuery`), deep-transforms the response, filters (`useFilteredFunds`), zero-fills future months, computes totals, derives the date range, and formats the `subtitle`. Consolidating to kill prop-drilling is the right call, but the fetch+transform+filter block would be cleaner extracted as a `useMpgaReportData(filters)` hook, leaving the provider to assemble the context value. Improves testability (also relevant to the missing graying-render tests).

src/components/Reports/MPGAIncomeExpensesReport/Charts/MonthlySummaryChart/MonthlySummaryChart.tsx
```diff
@@ -41,17 +36,20 @@ const chartColors = [
...
+  const {
+    allData: data,
+    dataLoading,
+    monthLabels: months,
+    currency,
+  } = useReport();
+  const incomeData = data.income ?? [];
+  const expenseData = data.expenses ?? [];
 
   const monthlyTotals = useMemo(() => {
```
kegrimes, line 50: **[Suggestion / cross-cutting] Zero-filled future months render as real `$0` net in the chart.** For YTD, `months` is the full 12 `monthLabels` and income/expense are zero-filled through December, so this map produces `net = 0` for every future month and the chart draws real `$0` bars/labels (`currencyFormat(0)` = `"$0"`, no empty guard). The tables gray future columns and render `-`, but the chart got no equivalent "future" treatment — the fix-one-fix-all gap. Slice the data to `firstFutureMonthIndex` in YTD, or gray/suppress future net labels to match the tables.

src/components/Reports/MPGAIncomeExpensesReport/BreakdownModal/BreakdownModal.tsx
```diff
@@ -76,11 +76,21 @@ export const BreakdownModal: React.FC<BreakdownModalProps> = ({
       open={open}
       onClose={onClose}
     >
-      <TableContainer>
+      <TableContainer
+        sx={{
+          '--breakdown-header-height': theme.spacing(6),
```
kegrimes: **[Suggestion] Dead CSS custom property.** `--breakdown-header-height` is defined here but never consumed via `var(--breakdown-header-height)` anywhere in the repo (the sticky header uses `top: 0`). Remove it, or wire it into the sticky offset it was meant to drive.

kegrimes (commented):

**PR #1929 — Add year filter to MPGA** · Mode: standard · 6 specialized agents + dependency analysis + 2 gap-review agents + 3 adversarial verifiers.

A well-executed refactor. It deletes `TotalsContext` and a large amount of prop-drilling, consolidates fetch/derive/totals into one `ReportContext`, and the new year/YTD date-window + zero-fill logic is **verified correct** (no off-by-one in `firstFutureMonthIndex`, totals not inflated by future zeros, date windows serialize correctly, no broken imports). **No blockers.** All findings below are improvements.

### Risk Assessment
`10/10 · CRITICAL` — driven by breadth (17 medium-risk feature files) + change volume + the financial-reporting domain, not by any inherently dangerous infra file. Recommended reviewer level: **SENIOR** (financial display logic).

### Dependency Impact
`ReportContext.tsx` has 15 dependents, all updated in this PR (self-contained). Every breaking edit — deleted `TotalsContext`, `useGetLastTwelveMonths(locale, year?)` signature change, `hideDateRange`→`isMpgaReport` rename, un-exported `getFiltersWithCalculatedDates` — has its call sites fixed in-PR. External consumer `StaffExpenseReport` (not in PR) verified safe.

---

### Findings not anchored to a diff line (coverage gaps, related files, pre-existing)

**Medium (5.0–6.9)**

**🟠 [6.5] Positional month↔label alignment is fragile — `src/components/Reports/MPGAIncomeExpensesReport/Helper/filterFunds.ts:92,135,187`** *(Data Integrity, PLAUSIBLE; related file, not in PR)*
Row `monthly[]` is built by array **index** from `breakdownByMonth`; `monthLabels` and the zero-fill are indexed independently — nothing joins on `month.month`. Now that arbitrary years / partial YTD windows are queryable, a sparse or reordered `breakdownByMonth` from the server would silently shift every column (April's amount under March's header; a real month grayed as "future"). **Fix:** key months by `month.month`, or assert the one-ordered-entry-per-requested-month server contract.

**🟠 [6.0] Deleted negative-net calc test not migrated — `Charts/MonthlySummaryChart` test (diff 374-407)** *(Testing)*
The removed test was the only assertion covering a negative monthly net (expenses > income). Re-add an equivalent net-calc assertion driven through the context mock.

**🟡 [5.0] Headline "graying" feature has no rendering test — `Tables/{TableCard,PrintTables,TotalRow}`** *(Testing, CONFIRMED; downgraded from 7.5 by adversarial verify)*
The derivation is unit-tested in `ReportContext.test.tsx`, but no component test activates YTD, so the `future-month` classes / gradient underline / colSpan-split paths never execute under test. `MPGAIncomeExpensesReportTestWrapper` can't seed filters — closing this needs an `initialFilters` seam on `ReportProvider`/the wrapper. Real regression risk, not a live bug.

**Suggestions (<5.0)**

- **[4.0] Empty-state copy hardcodes "in the last 12 months"** — `Tables/PrintTables.tsx:259-260`, `DisplayModes/ScreenOnlyReport.tsx:42,57` *(UX, in-scope but pre-existing lines)*. Now contradicts the dynamic `subtitle` when a specific year / YTD is selected and yields no data. Thread `subtitle` (already on `useReport()`) into the empty-state copy.
- **[3.0] Average column mixes two divisor conventions in YTD** — server `averagePerMonth` (plain/subcategory rows) vs client `total / monthly.length` (split/combined rows). Verified the client divisor uses the elapsed-month array N, so they agree **if** the server also divides by N; normalize to one convention. No demonstrable miscalc today.

**Pre-existing (informational, not counted toward verdict)**
- **Hardcoded `currency = 'USD'`** also in `BreakdownModal.tsx:41` (not migrated to `useReport()` like its siblings). No `currencyCode` is read from the data → wrong symbol for non-USD accounts. Worth a follow-up ticket given the report's purpose.
- `SettingsDialog` `<Dialog>` missing `aria-labelledby` → `DialogTitle`; empty-state `colSpan={15}` magic number in `PrintTables`; Cancel button `sx={{ color: 'black' }}` (hardcoded, use a theme token).

---

### Agent Summary
| Agent | Blockers | Medium | Suggestions | Confidence |
|---|:--:|:--:|:--:|:--:|
| Security | *(skipped — no auth/api/env/apollo-link files)* | | | |
| Architecture | 0 | 0 | 5 | Med-High |
| Data Integrity | 0 | 1 | 2 | High |
| Testing | 0 | 2 | 3 | High |
| UX | 0 | 1 | 4 | High |
| Standards | 0 | 1 | 0 | High |
| Financial Reporting | 0 | 0 | 2 | Med-High |

**Financial Checklist:** Money arithmetic safe ✅ · Currency mixing prevented ✅ (single server-aggregated figure) · Rounding at display boundary ✅ · Null/undefined handled ✅ · Luxon (not `new Date()`) ✅ · Server aggregations preferred ✅ *(caveat: Average mixes server+client — see above)*

> Inline comments below carry a `<!-- severity:X.X -->` tag. Reply `/dismiss: <reason>` on any finding with severity < 7 you disagree with, then re-run the review.

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesContext/MPGAIncomeExpensesContext.tsx
kegrimes: /dismiss this report is available to US staff only

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesContext/MPGAIncomeExpensesContext.tsx
kegrimes: /dismiss this is not worth doing in this PR

src/components/Reports/MPGAIncomeExpensesReport/Charts/MonthlySummaryChart/MonthlySummaryChart.tsx
kegrimes: /dismiss the design was already approved. this could be a future task if needed

src/components/Reports/MPGAIncomeExpensesReport/Tables/PrintTables.tsx
kegrimes: /dismiss

src/components/Reports/MPGAIncomeExpensesReport/Tables/TableCard.tsx
kegrimes: /dismiss I think the different shades are differing enough and the screen readers would be accurate in saying future months are 0

kegrimes:
@zweatshirt I realized Will was out! So sorry for the large PR🫣, but most of it is removing prop drilling. I noticed it was getting bad when I was adding the filters. Can you just make sure nothing is broken with the filters and the code (unrelated to context) looks decent?

src/components/Reports/MPGAIncomeExpensesReport/Charts/MonthlySummaryChart/MonthlySummaryChart.tsx
```diff
@@ -35,23 +30,23 @@ interface MonthlyTotal {
...
 
-  const { dataLoading } = useTotals();
+  const {
+    allData: data,
+    dataLoading,
+    monthLabels: months,
+    currency,
+  } = useReport();
+  const incomeData = data.income ?? [];
+  const expenseData = data.expenses ?? [];
```
zweatshirt: Optional: We can use de-structuring here

src/components/Reports/MPGAIncomeExpensesReport/DisplayModes/PrintOnlyReport.test.tsx
```diff
@@ -54,29 +32,19 @@ describe('PrintOnlyReport', () => {
...
-              <PrintOnlyReport
-                data={emptyData}
-                last12Months={months}
-                currency={currency}
-              />
-            </TotalsProvider>
-          </GqlMockedProvider>
-        </LocalizationProvider>
-      </ThemeProvider>,
+  it('displays the tables that should be showing', async () => {
```
zweatshirt, line 40: This test checks that there are 2 tables but only seem to add an assertion for one table by text. What is the second table that is supposed to display? Can we add an assertion for it in this test?

src/components/Reports/MPGAIncomeExpensesReport/ReportContext/ReportContext.tsx
```diff
@@ -0,0 +1,278 @@
...
+      transactionYears,
+      subtitle,
+      incomeTotal,
+      expensesTotal,
+      ministryTotal,
+      healthcareTotal,
+      assessmentTotal,
+      benefitsTotal,
+      salaryTotal,
+      otherTotal,
```
zweatshirt: I would group these, something like this:
```
 totals: {
  income, expenses, ministry, healthcare,
  assessment, benefits, salary, other,
}
```

src/components/Reports/MPGAIncomeExpensesReport/Tables/PrintTables.tsx
```diff
@@ -63,34 +75,59 @@ export const PrintTables: React.FC<PrintTablesProps> = ({
...
-                    colSpan={count}
-                    sx={{
-                      borderBottom: `2px solid ${borderColor}`,
-                      borderRight: `20px solid transparent`,
-                    }}
+                const monthOffset = monthCount
+                  .slice(0, index)
+                  .reduce((sum, group) => sum + group.count, 0);
+
+                const futureStart =
```
zweatshirt: What do you think of changing this to something like `pastCount` instead of `futureStart`?

src/components/Reports/MPGAIncomeExpensesReport/Tables/PrintTables.tsx
```diff
@@ -63,34 +75,59 @@ export const PrintTables: React.FC<PrintTablesProps> = ({
...
+                  firstFutureMonthIndex !== undefined
+                    ? Math.max(
```
zweatshirt: Can `futureStart` ever go negative? Wondering if this `Math.max` is needed or not

src/components/Reports/MPGAIncomeExpensesReport/Tables/TableCard.tsx
```diff
@@ -237,14 +268,21 @@ export const TableCard: React.FC<TableCardProps> = ({
...
             columns={columns}
             columnGroupingModel={columnGroupingModel}
+            sx={{
+              '& .future-month-header .MuiDataGrid-columnHeaderTitle': {
+                color: theme.palette.text.disabled,
+              },
+              '& .future-month': {
+                backgroundColor: theme.palette.action.hover,
+              },
+            }}
```
zweatshirt, line 283: Optional:

```suggestion
            sx={{
              '.future-month-header .MuiDataGrid-columnHeaderTitle': {
                color: theme.palette.text.disabled,
              },
              '.future-month': {
                backgroundColor: theme.palette.action.hover,
              },
            }}
```

src/components/Reports/MPGAIncomeExpensesReport/Tables/TableCard.tsx
```diff
@@ -223,7 +254,7 @@ export const TableCard: React.FC<TableCardProps> = ({
         ),
       },
     ];
-  }, [monthCount, getBorderColor, t]);
+  }, [monthCount, getBorderColor, firstFutureMonthIndex, t]);
```
zweatshirt, line 255: Probably won't ever trigger but `theme` could go in the dependency array

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesReport.tsx
```diff
@@ -51,32 +44,37 @@ export const MPGAIncomeExpensesReport: React.FC<
...
+  const isFilterActive = useMemo(
+    () =>
+      Boolean(
+        filters &&
+          (filters.selectedDateRange === DateRange.YearToDate ||
+            (filters.selectedYear !== null &&
+              filters.selectedYear !== undefined)),
+      ),
+    [filters],
+  );
```
zweatshirt: We can probably keep this un-memoized

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesReport.tsx
```diff
@@ -212,46 +160,30 @@ export const MPGAIncomeExpensesReport: React.FC<
...
-            if (newFilters) {
-              onFiltersChange(newFilters);
-            }
+            const hasActiveFilter = Boolean(
+              newFilters &&
+                (newFilters.categories ||
+                  newFilters.selectedDateRange === DateRange.YearToDate ||
+                  (newFilters.selectedYear !== null &&
+                    newFilters.selectedYear !== undefined)),
+            );
```
zweatshirt: This and `isFilterActive` are nearly identical, what do you think about creating a shared variable for both conditions?

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesContext/MPGAIncomeExpensesContext.tsx
```diff
@@ -0,0 +1,278 @@
...
+                transactedAt: transaction.transactedAt,
+                description: transaction.description ?? '',
+                amount: transaction.amount,
+              })),
+            })),
+          })),
+        })),
+      })),
+    [reportData],
+  );
```
zweatshirt, line 172: Wondering if it is worth adding a comment for this function explaining why we transform the data here

zweatshirt (approved):
Tried to comment on whatever I was able to find. The UI looks amazing and seems to behave as expected so I am going to approve. Just had some suggestions.

src/components/Reports/MPGAIncomeExpensesReport/Tables/PrintTables.tsx
kegrimes: You are right, thanks!

src/components/Reports/MPGAIncomeExpensesReport/Tables/TableCard.tsx
kegrimes: I don't think we will need to put this in the dependency array since it is an import not defined inside a component.

zweatshirt:
@kegrimes I believe the GitGuardian flag is a false positive from one of @canac's PRs last week, so no worries with that

PR #1930, opened 2026-07-20
MPDX-9818 - Add DA5 and A33 warnings/errors to goal settings

zweatshirt:
## Description

We want to show a warning to the admin when the salary the admin has entered is over the salary cap.
Backend PR: https://github.com/CruGlobal/mpdx_api/pull/3469

## Testing

- Go to the GoalSettingsForm
- Enter a large amount for 'Annual Requested Salary'
- Check that the warning displays that the amount exceeds the salary cap.

## Checklist:

### DIFF
--- a/src/components/HrTools/NsGoalCalculator/GoalSettings/Fields/useGoalSettingsField.test.tsx
+++ b/src/components/HrTools/NsGoalCalculator/GoalSettings/Fields/useGoalSettingsField.test.tsx
@@ -1,21 +1,73 @@
 import React from 'react';
 import { ThemeProvider } from '@mui/material';
 import { render } from '@testing-library/react';
-import { Form, Formik } from 'formik';
+import { Form, Formik, FormikErrors, FormikTouched } from 'formik';
+import { GqlMockedProvider } from '__tests__/util/graphqlMocking';
+import { NewStaffGoalCalculationSalaryOverCapEnum } from 'src/graphql/types.generated';
 import theme from 'src/theme';
+import { defaultGoalCalculation } from '../../NsGoalCalculatorTestWrapper';
+import { GoalSettingsPreviewProvider } from '../GoalSettingsPreviewContext';
+import { calculationToFormValues } from '../goalSettingsApiMapping';
+import { GoalSettingsFormValues } from '../goalSettingsFormValues';
+import { NewStaffGoalCalculation } from '../useNewStaffGoalCalculation';
 import { GoalSettingsTextField } from './GoalSettingsTextField';
 import { GoalSettingsFieldBaseProps } from './useGoalSettingsField';
 
+interface TestComponentProps extends GoalSettingsFieldBaseProps {
+  initialErrors?: FormikErrors<GoalSettingsFormValues>;
+  initialTouched?: FormikTouched<GoalSettingsFormValues>;
+  savedSalaryOverCap?: boolean;
+  allowSalaryOverCap?: NewStaffGoalCalculationSalaryOverCapEnum;
+  secaExempt?: GoalSettingsFormValues['secaExempt'];
+  spouseSecaExempt?: GoalSettingsFormValues['spouseSecaExempt'];
+}
+
 // Exercises the hook through a real consumer (GoalSettingsTextField).
-const TestComponent: React.FC<GoalSettingsFieldBaseProps> = (props) => (
-  <ThemeProvider theme={theme}>
-    <Formik initialValues={{ [props.name]: '' }} onSubmit={jest.fn()}>
-      <Form>
-        <GoalSettingsTextField {...props} />
-      </Form>
-    </Formik>
-  </ThemeProvider>
-);
+const TestComponent: React.FC<TestComponentProps> = ({
+  initialErrors,
+  initialTouched,
+  savedSalaryOverCap = false,
+  allowSalaryOverCap = NewStaffGoalCalculationSalaryOverCapEnum.No,
+  secaExempt = 'false',
+  spouseSecaExempt = 'false',
+  ...fieldProps
+}) => {
+  const calculation: NewStaffGoalCalculation = {
+    ...defaultGoalCalculation,
+    calculations: {
+      ...defaultGoalCalculation.calculations,
+      salaryOverCap: savedSalaryOverCap,
+      debtOverCap: false,
+    },
+  };
+
+  return (
+    <ThemeProvider theme={theme}>
+      <GqlMockedProvider>
+        <Formik
+          initialValues={{
+            ...calculationToFormValues(defaultGoalCalculation),
+            allowSalaryOverCap,
+            secaExempt,
+            spouseSecaExempt,
+          }}
+          initialErrors={initialErrors}
+          initialTouched={initialTouched}
+          onSubmit={jest.fn()}
+        >
+          <GoalSettingsPreviewProvider
+            accountListId="account-list-1"
+            calculation={calculation}
+          >
+            <Form>
+              <GoalSettingsTextField {...fieldProps} />
+            </Form>
+          </GoalSettingsPreviewProvider>
+        </Formik>
+      </GqlMockedProvider>
+    </ThemeProvider>
+  );
+};
 
 describe('useGoalSettingsField', () => {
   it('uses the label as the accessible name for a shared field', () => {
@@ -65,18 +117,12 @@ describe('useGoalSettingsField', () => {
 
   it('surfaces the Formik error as helperText once the field is touched', () => {
     const { getByRole, getByText } = render(
-      <ThemeProvider theme={theme}>
-        <Formik
-          initialValues={{ age: '' }}
-          initialErrors={{ age: 'Age is required' }}
-          initialTouched={{ age: true }}
-          onSubmit={jest.fn()}
-        >
-          <Form>
-            <GoalSettingsTextField name="age" label="Age" />
-          </Form>
-        </Formik>
-      </ThemeProvider>,
+      <TestComponent
+        name="age"
+        label="Age"
+        initialErrors={{ age: 'Age is required' }}
+        initialTouched={{ age: true }}
+      />,
     );
 
     expect(getByRole('textbox', { name: 'Age' })).toBeInvalid();
@@ -85,20 +131,114 @@ describe('useGoalSettingsField', () => {
 
   it('hides the error until the field is touched', () => {
     const { getByRole, queryByText } = render(
-      <ThemeProvider theme={theme}>
-        <Formik
-          initialValues={{ age: '' }}
-          initialErrors={{ age: 'Age is required' }}
-          onSubmit={jest.fn()}
-        >
-          <Form>
-            <GoalSettingsTextField name="age" label="Age" />
-          </Form>
-        </Formik>
-      </ThemeProvider>,
+      <TestComponent
+        name="age"
+        label="Age"
+        initialErrors={{ age: 'Age is required' }}
+      />,
     );
 
     expect(getByRole('textbox', { name: 'Age' })).toBeValid();
     expect(queryByText('Age is required')).not.toBeInTheDocument();
+  });
+
+  it('outlines a contributing field in the warning color when allowed', () => {
+    const { container } = render(
+      <TestComponent
+        name="annualRequestedSalary"
+        label="Salary"
+        savedSalaryOverCap
+        allowSalaryOverCap={NewStaffGoalCalculationSalaryOverCapEnum.YesAny}
+      />,
+    );
+
+    expect(
+      container.querySelector('.MuiOutlinedInput-notchedOutline'),
+    ).toHaveStyle({ borderColor: theme.palette.warning.main });
+  });
+
+  it('outlines a contributing field in the error color when not allowed', () => {
+    const { container } = render(
+      <TestComponent
+        name="annualRequestedSalary"
+        label="Salary"
+        savedSalaryOverCap
+      />,
+    );
+
+    expect(
+      container.querySelector('.MuiOutlinedInput-notchedOutline'),
+    ).toHaveStyle({ borderColor: theme.palette.error.main });
+  });
+
+  it('leaves fields that do not feed the warning alone', () => {
+    const { container } = render(
+      <TestComponent name="tenure" label="Tenure" savedSalaryOverCap />,
+    );
+
+    expect(
+      container.querySelector('.MuiOutlinedInput-notchedOutline'),
+    ).not.toHaveStyle({ borderColor: theme.palette.error.main });
+  });
+
+  it('does not outline anything while the cap is not exceeded', () => {
+    const { container } = render(
+      <TestComponent name="annualRequestedSalary" label="Salary" />,
+    );
+
+    expect(
+      container.querySelector('.MuiOutlinedInput-notchedOutline'),
+    ).not.toHaveStyle({ borderColor: theme.palette.error.main });
+  });
+
+  it('lets a validation error take over the field', () => {
+    const { getByRole, getByText } = render(
+      <TestComponent
+        name="annualRequestedSalary"
+        label="Salary"
+        savedSalaryOverCap
+        initialErrors={{ annualRequestedSalary: 'Salary is required' }}
+        initialTouched={{ annualRequestedSalary: true }}
+      />,
+    );
+
+    expect(getByRole('textbox', { name: 'Salary' })).toBeInvalid();
+    expect(getByText('Salary is required')).toBeInTheDocument();
+  });
+
+  it('outlines the exemption field the warning came from', () => {
+    const { container } = render(
+      <TestComponent name="secaExempt" label="SECA Exempt" secaExempt="true" />,
+    );
+
+    expect(
+      container.querySelector('.MuiOutlinedInput-notchedOutline'),
+    ).toHaveStyle({ borderColor: theme.palette.warning.main });
+  });
+
+  it('leaves the spouse exemption field alone when only staff opted out', () => {
+    const { container } = render(
+      <TestComponent
+        name="spouseSecaExempt"
+        label="SECA Exempt"
+        secaExempt="true"
+      />,
+    );
+
+    expect(
+      container.querySelector('.MuiOutlinedInput-notchedOutline'),
+    ).not.toHaveStyle({ borderColor: theme.palette.warning.main });
+  });
+
+  it('does not mark a highlighted field invalid to assistive tech', () => {
+    const { getByRole } = render(
+      <TestComponent
+        name="annualRequestedSalary"
+        label="Salary"
+        savedSalaryOverCap
+      />,
+    );
+
+    expect(getByRole('textbox', { name: 'Salary' })).toBeValid();
   });
 });
--- a/src/components/HrTools/NsGoalCalculator/GoalSettings/Fields/useGoalSettingsField.ts
+++ b/src/components/HrTools/NsGoalCalculator/GoalSettings/Fields/useGoalSettingsField.ts
@@ -1,6 +1,8 @@
 import { TextFieldProps } from '@mui/material';
 import { useField } from 'formik';
 import { NewStaffGoalCalculationAttributesInput } from 'src/graphql/types.generated';
+import { useGoalSettingsPreview } from '../GoalSettingsPreviewContext';
+import { WarningSeverity } from '../goalSettingsWarnings';
 
 export type GoalSettingsFieldBaseProps = Omit<TextFieldProps, 'name'> & {
   /** Formik field key. Required here even though MUI's `name` is optional. */
@@ -16,6 +18,19 @@ export type GoalSettingsFieldBaseProps = Omit<TextFieldProps, 'name'> & {
    */
   showLabel?: boolean;
 };
+
+const outlineSx = (severity: WarningSeverity) => ({
+  '.MuiOutlinedInput-notchedOutline': {
+    borderColor: `${severity}.main`,
+  },
+  // Needs the ampersand: :hover applies to the root, not a descendant.
+  '&:hover .MuiOutlinedInput-notchedOutline': {
+    borderColor: `${severity}.dark`,
+  },
+  '.Mui-focused .MuiOutlinedInput-notchedOutline': {
+    borderColor: `${severity}.main`,
+  },
+});
 
 /**
  * Shared wiring for every Goal Settings field: Formik binding, consistent MUI
@@ -38,6 +53,7 @@ export const useGoalSettingsField = ({
   ...props
 }: GoalSettingsFieldBaseProps): TextFieldProps => {
   const [field, meta] = useField(name);
+  const preview = useGoalSettingsPreview();
 
   const accessibleName =
     typeof label === 'string'
@@ -47,6 +63,8 @@ export const useGoalSettingsField = ({
       : undefined;
 
   const showError = Boolean(meta.touched && meta.error);
+  // A validation error is actionable, so it wins over the advisory outline.
+  const severity = showError ? undefined : preview?.fieldSeverity(name);
 
   return {
     id: name,
@@ -58,6 +76,10 @@ export const useGoalSettingsField = ({
     ...field,
     error: showError,
     helperText: showError ? meta.error : undefined,
+    sx: {
+      ...(severity ? outlineSx(severity) : {}),
+      ...props.sx,
+    },
     inputProps: {
       ...(!showLabel && accessibleName ? { 'aria-label': accessibleName } : {}),
       ...inputProps,
--- a/src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsForm.test.tsx
+++ b/src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsForm.test.tsx
@@ -339,6 +339,71 @@ describe('GoalSettingsForm', () => {
     );
   });
 
+  // Before the shared provider, each consumer ran its own preview request.
+  it('previews an edit exactly once with every consumer mounted', async () => {
+    const { findByRole } = render(<TestComponent />);
+
+    const salary = await findByRole('spinbutton', {
+      name: 'Annual Requested Salary — John',
+    });
+    userEvent.clear(salary);
+    userEvent.type(salary, '250000');
+
+    await waitFor(() =>
+      expect(mutationSpy).toHaveGraphqlOperation(
+        'PreviewNewStaffGoalCalculation',
+      ),
+    );
+
+    const previewCalls = mutationSpy.mock.calls.filter(
+      ([{ operation }]) =>
+        operation.operationName === 'PreviewNewStaffGoalCalculation',
+    );
+    expect(previewCalls).toHaveLength(1);
+  });
+
+  it('shows the salary over-cap warning', async () => {
+    const { findByText, findByRole } = render(
+      <TestComponent
+        goalCalculationMock={{
+          newStaffGoalCalculation: {
+            ...defaultMock.newStaffGoalCalculation,
+            calculations: {
+              ...defaultGoalCalculation.calculations,
+              salaryOverCap: true,
+            },
+          },
+        }}
+      />,
+    );
+
+    await findByRole('button', { name: 'Save & Share' });
+    expect(
+      await findByText('Total salary is over the standard cap'),
+    ).toBeInTheDocument();
+  });
+
+  it('shows the debt over-cap warning', async () => {
+    const { findByText, findByRole } = render(
+      <TestComponent
+        goalCalculationMock={{
+          newStaffGoalCalculation: {
+            ...defaultMock.newStaffGoalCalculation,
+            calculations: {
+              ...defaultGoalCalculation.calculations,
+              debtOverCap: true,
+            },
+          },
+        }}
+      />,
+    );
+
+    await findByRole('button', { name: 'Save & Share' });
+    expect(
+      await findByText('Annual debt is over the standard cap'),
+    ).toBeInTheDocument();
+  });
+
   it('clears spouse attributes when marital status changes from married to single', async () => {
     const { findByRole, getByRole } = render(<TestComponent />);
 
--- a/src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsForm.tsx
+++ b/src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsForm.tsx
@@ -11,7 +11,9 @@ import { Form, Formik } from 'formik';
 import { useTranslation } from 'react-i18next';
 import { NewStaffQuestionnaireMaritalStatusEnum } from 'src/graphql/types.generated';
 import { GoalSettingsHeader } from './GoalSettingsHeader';
+import { GoalSettingsPreviewProvider } from './GoalSettingsPreviewContext';
 import { GoalSettingsScrollContainer } from './GoalSettingsScrollContainer';
+import { GoalSettingsWarning } from './GoalSettingsWarning';
 import { ContactInformationSection } from './Sections/ContactInformationSection';
 import { ExemptionsSection } from './Sections/ExemptionsSection';
 import { FinancialInformationSection } from './Sections/FinancialInformationSection';
@@ -67,6 +69,8 @@ export const GoalSettingsForm: React.FC<GoalSettingsFormProps> = (props) => {
     goalCalculation: calculation,
     fallback,
     isScenario,
+    // `null` for scenario goals, which have no account list.
+    accountListId,
     save,
   } = useNewStaffGoalCalculation(props);
 
@@ -155,51 +159,60 @@ export const GoalSettingsForm: React.FC<GoalSettingsFormProps> = (props) => {
           };
 
           return (
-            <Form>
-              <GoalSettingsHeader
-                accountListId={
-                  // Scenario goals have no account list
-                  'accountListId' in props ? props.accountListId : null
-                }
-                calculationId={calculation.id}
-                primaryPerson={primaryPerson}
-                spousePerson={spousePerson}
-                mpdGoal={mpdGoal}
-                joinedStaffYear={calculation.joinedStaffYear}
-                isScenario={isScenario}
-                isComplete={isGoalSettingsComplete(values)}
-              />
+            <GoalSettingsPreviewProvider
+              accountListId={accountListId}
+              calculation={calculation}
+            >
+              <Form>
+                <GoalSettingsHeader
+                  primaryPerson={primaryPerson}
+                  spousePerson={spousePerson}
+                  mpdGoal={mpdGoal}
+                  joinedStaffYear={calculation.joinedStaffYear}
+                  isScenario={isScenario}
+                  isComplete={isGoalSettingsComplete(values)}
+                />
 
-              <Divider sx={{ mb: 3 }} />
+                <Divider sx={{ mb: 3 }} />
 
-              {isScenario && <ContactInformationSection {...sectionProps} />}
-              <PersonalInformationSection {...sectionProps} />
-              <FinancialInformationSection {...sectionProps} />
-              <HealthcareInformationSection {...sectionProps} />
-              <MinistryInformationSection {...sectionProps} />
-              <NsoInformationSection {...sectionProps} />
-              <ExemptionsSection {...sectionProps} />
+                {isScenario && <ContactInformationSection {...sectionProps} />}
+                <PersonalInformationSection {...sectionProps} />
+                <FinancialInformationSection {...sectionProps} />
+                <HealthcareInformationSection {...sectionProps} />
+                <MinistryInformationSection {...sectionProps} />
+                <NsoInformationSection {...sectionProps} />
+                <ExemptionsSection {...sectionProps} />
 
-              <StickyActionBar>
-                <Stack direction="row" spacing={2} justifyContent="flex-end">
-                  <Button color="inherit" onClick={() => resetForm()}>
-                    {t('Cancel')}
-                  </Button>
-                  <Button
-                    type="submit"
-                    variant="contained"
-                    disabled={!isValid || isSubmitting}
-                    startIcon={
-                      isSubmitting ? (
-                        <CircularProgress color="inherit" size={20} />
-                      ) : undefined
-                    }
+                <StickyActionBar>
+                  <Stack
+                    direction="row"
+                    spacing={2}
+                    alignItems="center"
+                    flexWrap="wrap"
+                    useFlexGap
                   >
-                    {t('Save & Share')}
-                  </Button>
-                </Stack>
-              </StickyActionBar>
-            </Form>
+                    <GoalSettingsWarning />
+                    <Stack direction="row" spacing={2} sx={{ ml: 'auto' }}>
+                      <Button color="inherit" onClick={() => resetForm()}>
+                        {t('Cancel')}
+                      </Button>
+                      <Button
+                        type="submit"
+                        variant="contained"
+                        disabled={!isValid || isSubmitting}
+                        startIcon={
+                          isSubmitting ? (
+                            <CircularProgress color="inherit" size={20} />
+                          ) : undefined
+                        }
+                      >
+                        {t('Save & Share')}
+                      </Button>
+                    </Stack>
+                  </Stack>
+                </StickyActionBar>
+              </Form>
+            </GoalSettingsPreviewProvider>
           );
         }}
       </Formik>
--- a/src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsPreviewContext.tsx
+++ b/src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsPreviewContext.tsx
@@ -1,0 +1,90 @@
+import React, { createContext, useContext, useMemo } from 'react';
+import { useFormikContext } from 'formik';
+import { useTranslation } from 'react-i18next';
+import { GoalSettingsFormValues } from './goalSettingsFormValues';
+import {
+  GoalSettingsWarningItem,
+  WarningSeverity,
+  buildGoalSettingsWarnings,
+  getFieldSeverity,
+} from './goalSettingsWarnings';
+import { useMpdGoalPreview } from './useMpdGoalPreview';
+import { NewStaffGoalCalculation } from './useNewStaffGoalCalculation';
+
+interface GoalSettingsPreviewValue {
+  calculating: boolean;
+  failed: boolean;
+  previewGoal: number | null;
+  warnings: GoalSettingsWarningItem[];
+  fieldSeverity: (name: string) => WarningSeverity | undefined;
+}
+
+const GoalSettingsPreviewContext =
+  createContext<GoalSettingsPreviewValue | null>(null);
+
+interface GoalSettingsPreviewProviderProps {
+  /** `null` for a scenario goal, which has no account list. */
+  accountListId: string | null;
+  calculation: NewStaffGoalCalculation;
+  children?: React.ReactNode;
+}
+
+/**
+ * Owns the Goal Settings preview for the whole form. Every consumer — the goal
+ * figure, the warnings, the field outlines — reads from here, so an edit costs
+ * one request no matter how many are mounted, and they can't disagree mid-flight.
+ *
+ * Must render inside the `<Formik>`, since the preview keys off live values.
+ */
+export const GoalSettingsPreviewProvider: React.FC<
+  GoalSettingsPreviewProviderProps
+> = ({ accountListId, calculation, children }) => {
+  const { t } = useTranslation();
+  const { values } = useFormikContext<GoalSettingsFormValues>();
+
+  const {
+    id: calculationId,
+    calculations: {
+      salaryOverCap: savedSalaryOverCap,
+      debtOverCap: savedDebtOverCap,
+    },
+  } = calculation;
+
+  const {
+    calculating,
+    failed,
+    previewGoal,
+    previewSalaryOverCap,
+    previewDebtOverCap,
+  } = useMpdGoalPreview({ accountListId, calculationId });
+
+  const salaryOverCap = previewSalaryOverCap ?? savedSalaryOverCap;
+  const debtOverCap = previewDebtOverCap ?? savedDebtOverCap;
+
+  const value = useMemo<GoalSettingsPreviewValue>(() => {
+    const warnings = buildGoalSettingsWarnings({
+      values,
+      salaryOverCap,
+      debtOverCap,
+      t,
+    });
+
+    return {
+      calculating,
+      failed,
+      previewGoal,
+      warnings,
+      fieldSeverity: (name) => getFieldSeverity(warnings, name),
+    };
+  }, [values, salaryOverCap, debtOverCap, t, calculating, failed, previewGoal]);
+
+  return (
+    <GoalSettingsPreviewContext.Provider value={value}>
+      {children}
+    </GoalSettingsPreviewContext.Provider>
+  );
+};
+
+/** `null` outside a provider, so standalone fields just skip highlighting. */
+export const useGoalSettingsPreview = (): GoalSettingsPreviewValue | null =>
+  useContext(GoalSettingsPreviewContext);
--- a/src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsWarning.test.tsx
+++ b/src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsWarning.test.tsx
@@ -1,0 +1,435 @@
+import React from 'react';
+import { render, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { Field, Form, Formik } from 'formik';
+import { GqlMockedProvider } from '__tests__/util/graphqlMocking';
+import {
+  MpdGoalBenefitsConstantPlanEnum,
+  NewStaffGoalCalculationSalaryOverCapEnum,
+  NewStaffQuestionnaireMaritalStatusEnum,
+} from 'src/graphql/types.generated';
+import i18n from 'src/lib/i18n';
+import { defaultGoalCalculation } from '../NsGoalCalculatorTestWrapper';
+import { GoalSettingsPreviewProvider } from './GoalSettingsPreviewContext';
+import { GoalSettingsWarning } from './GoalSettingsWarning';
+import { PreviewNewStaffGoalCalculationMutation } from './NewStaffGoalCalculation.generated';
+import { calculationToFormValues } from './goalSettingsApiMapping';
+import { GoalSettingsFormValues } from './goalSettingsFormValues';
+import { getGoalSettingsSchema } from './goalSettingsSchema';
+import { NewStaffGoalCalculation } from './useNewStaffGoalCalculation';
+
+const mutationSpy = jest.fn();
+
+const { Married, Single } = NewStaffQuestionnaireMaritalStatusEnum;
+const { Base, Select } = MpdGoalBenefitsConstantPlanEnum;
+const { No, UpToBcc } = NewStaffGoalCalculationSalaryOverCapEnum;
+
+interface TestComponentProps {
+  savedSalaryOverCap?: boolean;
+  savedDebtOverCap?: boolean;
+  previewSalaryOverCap?: boolean;
+  previewDebtOverCap?: boolean;
+  allowSalaryOverCap?: GoalSettingsFormValues['allowSalaryOverCap'];
+  allowDebtOverCap?: GoalSettingsFormValues['allowDebtOverCap'];
+  maritalStatus?: GoalSettingsFormValues['maritalStatus'];
+  healthcareExempt?: GoalSettingsFormValues['healthcareExempt'];
+  spouseHealthcareExempt?: GoalSettingsFormValues['spouseHealthcareExempt'];
+  secaExempt?: GoalSettingsFormValues['secaExempt'];
+  spouseSecaExempt?: GoalSettingsFormValues['spouseSecaExempt'];
+  benefitsPlan?: GoalSettingsFormValues['benefitsPlan'];
+  validationSchema?: ReturnType<typeof getGoalSettingsSchema>;
+}
+
+const TestComponent: React.FC<TestComponentProps> = ({
+  savedSalaryOverCap = false,
+  savedDebtOverCap = false,
+  previewSalaryOverCap = false,
+  previewDebtOverCap = false,
+  allowSalaryOverCap = No,
+  allowDebtOverCap = 'false',
+  maritalStatus = Single,
+  healthcareExempt = 'false',
+  spouseHealthcareExempt = 'false',
+  secaExempt = 'false',
+  spouseSecaExempt = 'false',
+  benefitsPlan = Base,
+  validationSchema,
+}) => {
+  const calculation: NewStaffGoalCalculation = {
+    ...defaultGoalCalculation,
+    calculations: {
+      ...defaultGoalCalculation.calculations,
+      salaryOverCap: savedSalaryOverCap,
+      debtOverCap: savedDebtOverCap,
+    },
+  };
+
+  return (
+    <GqlMockedProvider<{
+      PreviewNewStaffGoalCalculation: PreviewNewStaffGoalCalculationMutation;
+    }>
+      mocks={{
+        PreviewNewStaffGoalCalculation: {
+          previewNewStaffGoalCalculation: {
+            newStaffGoalCalculation: {
+              id: defaultGoalCalculation.id,
+              calculations: {
+                monthlyGoal: 5200,
+                salaryOverCap: previewSalaryOverCap,
+                debtOverCap: previewDebtOverCap,
+              },
+            },
+          },
+        },
+      }}
+      onCall={mutationSpy}
+    >
+      <Formik
+        initialValues={{
+          ...calculationToFormValues(defaultGoalCalculation),
+          allowSalaryOverCap,
+          allowDebtOverCap,
+          maritalStatus,
+          healthcareExempt,
+          spouseHealthcareExempt,
+          secaExempt,
+          spouseSecaExempt,
+          benefitsPlan,
+        }}
+        validationSchema={validationSchema}
+        onSubmit={jest.fn()}
+      >
+        {({ isValid }) => (
+          <GoalSettingsPreviewProvider
+            accountListId="account-list-1"
+            calculation={calculation}
+          >
+            <Form>
+              <Field
+                name="annualRequestedSalary"
+                type="number"
+                aria-label="Salary"
+              />
+              <span data-testid="is-valid">{String(isValid)}</span>
+              <GoalSettingsWarning />
+            </Form>
+          </GoalSettingsPreviewProvider>
+        )}
+      </Formik>
+    </GqlMockedProvider>
+  );
+};
+
+describe('GoalSettingsWarning', () => {
+  beforeEach(() => {
+    jest.useFakeTimers();
+  });
+
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
+  it('warns when the saved calculation is over the salary cap', () => {
+    const { getByText } = render(<TestComponent savedSalaryOverCap />);
+
+    expect(
+      getByText('Total salary is over the standard cap'),
+    ).toBeInTheDocument();
+  });
+
+  it('clears a salary warning an edit resolves, and keeps it clear while a later edit previews', async () => {
+    const { getByRole, getByText, queryByText } = render(
+      <TestComponent savedSalaryOverCap />,
+    );
+
+    expect(
+      getByText('Total salary is over the standard cap'),
+    ).toBeInTheDocument();
+
+    const salary = getByRole('spinbutton', { name: 'Salary' });
+    userEvent.clear(salary);
+    userEvent.type(salary, '40000');
+
+    await waitFor(() =>
+      expect(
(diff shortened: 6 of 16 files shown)


src/components/HrTools/NsGoalCalculator/GoalSettings/useMpdGoalPreview.ts
```diff
@@ -131,6 +148,7 @@ export const useMpdGoalPreview = ({
       ? 0
       : Math.round((previewGoal - savedMonthlyGoal) * 100) / 100;
   const changed = Math.abs(diff) >= CENT_EPSILON;
+  const salaryOverCap = settledPreview?.salaryOverCap ?? savedSalaryOverCap;
```
zweatshirt: [Medium] **The warning flickers and re-announces on every edit.**

`settledPreview` is invalidated by *object identity* (`preview?.attributes === previewable`), and `attributes` is memoized on **all** form values. So any keystroke in any of the seven sections drops `settledPreview` to `null` for 500ms + RTT, and this line falls back to `savedSalaryOverCap`.

The inverse case is the damaging one: saved record is over cap, the admin lowers the salary and watches the warning clear — then **every subsequent edit anywhere in the form makes the resolved warning reappear**. Because reappearance is a fresh DOM addition inside `aria-live`, a screen reader re-speaks the warning once per edit burst for the rest of the session, about a problem already fixed.

Current tests can't catch this — they assert only initial and final state, never a second edit after a settle.

```diff
-  const salaryOverCap = settledPreview?.salaryOverCap ?? savedSalaryOverCap;
+  // Hold the last settled warning while a new preview is in flight so it
+  // doesn't flicker (and re-announce) on every edit.
+  const salaryOverCap =
+    settledPreview?.salaryOverCap ?? preview?.salaryOverCap ?? savedSalaryOverCap;
```

Flagged independently by the UX, Data Integrity, and Financial Reporting agents.

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsForm.tsx
```diff
@@ -179,22 +179,34 @@ export const GoalSettingsForm: React.FC<GoalSettingsFormProps> = (props) => {
...
-                    {t('Save & Share')}
-                  </Button>
+                <Stack
+                  direction="row"
+                  spacing={2}
+                  alignItems="center"
+                  flexWrap="wrap"
+                  useFlexGap
+                >
+                  <GoalSettingsWarning
```
zweatshirt: [Medium] **This integration is completely unasserted — proven by mutation testing.**

Deleting this entire `<GoalSettingsWarning>` block and running the tree gives **34 suites, 213 tests, all passing**. The PR's user-facing deliverable can be removed with a fully green suite.

`GoalSettingsForm.test.tsx` is untouched by this PR. Note the warning text is actually present in the DOM of all 30 of its tests (`defaultGoalCalculation.calculations.salaryOverCap` seeds to `true`) — rendered and never looked at.

```tsx
it('shows the over-cap warning beside the save actions', async () => {
    <TestComponent
      goalCalculationMock={{
        newStaffGoalCalculation: {
          ...defaultMock.newStaffGoalCalculation,
          calculations: { ...defaultGoalCalculation.calculations, salaryOverCap: true },
        },
      }}
    />,
  );

  expect(
    await findByText('Requested salary is over the standard cap'),
  ).toBeInTheDocument();
});
```

src/components/HrTools/NsGoalCalculator/GoalSettings/useNewStaffGoalCalculation.tsx
```diff
@@ -82,6 +84,7 @@ export const useNewStaffGoalCalculation = (
     error,
     fallback,
     isScenario,
+    accountListId: isScenario ? null : source.accountListId,
```
zweatshirt, line 87: [Medium] **This data-scoping value is untested in both directions.**

Two mutations of this line each survived the full suite (213/213 passing):

| Mutation | Result |
|---|---|
| `accountListId: null` (real goals lose their account list) | 213 passed |
| `isScenario ? 'leaked-account-list' : ...` (scenario fabricates one) | 213 passed |

This value now feeds both `GoalSettingsHeader` and the `previewNewStaffGoalCalculation` mutation input, so the leak direction matters. `MpdGoalPreview.test.tsx` covers null handling but passes `accountListId={null}` as a prop directly — it never exercises this hook-to-component plumbing.

```tsx
// useNewStaffGoalCalculation.test.tsx
expect(result.current.accountListId).toBe('account-list-1'); // accountList mode
expect(result.current.accountListId).toBeNull();             // scenario mode
```

The refactor itself is a genuine improvement — it removes a duplicated `'accountListId' in props` narrowing. It just needs coverage.

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsWarning.tsx
```diff
@@ -0,0 +1,43 @@
...
+    savedMonthlyGoal: calculation.calculations.monthlyGoal,
+    savedSalaryOverCap: calculation.calculations.salaryOverCap,
+  });
+
+  return (
+    <Stack direction="row" spacing={1} alignItems="center" aria-live="polite">
+      {salaryOverCap && (
+        <>
+          <WarningAmber fontSize="small" sx={{ color: 'warning.main' }} />
+          <Typography variant="body2" sx={{ color: 'warning.main' }}>
```
zweatshirt: [Medium] **WCAG AA contrast failure: 3.11:1, needs 4.5:1.**

`src/theme.ts` never defines `palette.warning` — line 35's `warning: cruColors.vermilion` sits inside `statusColors` and is surfaced only as the custom `statusWarning` entry (line 262). So `warning.main` resolves to MUI's default `#ed6c02`, whose own source comments it as *"closest to orange[800] that pass 3:1"* — designed for the non-text threshold. `background.paper` is an unoverridden `#fff`.

`variant="body2"` is 14px normal weight, so 4.5:1 applies. **Computed 3.11:1 — fails.** The repo's own `statusWarning` (`#D34400`) gives 4.57:1.

The icon on line 35 is fine — it clears the 3:1 non-text bar and is correctly decorative (`aria-hidden`, no `titleAccess`).

**Please fix in `src/theme.ts`, not here.** HrTools uses `warning.main` in 7 of 7 pre-existing cases; changing only this file makes it the sole diverging one. See the review body for the two-line palette fix, which repairs all 9 call sites (`CardTitle.tsx:24` has the same latent defect with `success.main`). Pre-existing and tree-wide, so not a blocker on this PR.

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsWarning.tsx
```diff
@@ -0,0 +1,43 @@
...
+ * The non-blocking admin warning from NSGC A33, rendered beside the Goal
+ * Settings save actions so an admin sees it at the point of Save & Share
+ * regardless of how far the long form has scrolled.
+ */
+export const GoalSettingsWarning: React.FC<GoalSettingsWarningProps> = ({
+  accountListId,
+  calculation,
+}) => {
+  const { t } = useTranslation();
+  const { salaryOverCap } = useMpdGoalPreview({
```
zweatshirt: [Medium] **Second `useMpdGoalPreview` instance — every edit runs the goal calculation twice server-side.**

`useMpdGoalPreview` is a stateful orchestrator, not a selector: each call site owns a `useState`, a `useDebouncedValue` timer, and a `useEffect` firing `previewNewStaffGoalCalculation`. It is now mounted twice under the same Formik context — here, and via `GoalSettingsHeader` -> `MpdGoalPreview`.

Mitigating (verified): `src/lib/apollo/link.ts` routes these through `BatchHttpLink` (batchInterval 20ms, batchDebounce), and both fire in the same commit, so they coalesce into **one HTTP request carrying two operations**. Shared HTTP fate also makes header/warning divergence unlikely. The real cost is doubled server-side goal computation per edit, plus the state-ownership smell — which is why this is 6.5 rather than the 7.5 initially assessed.

Fix: hoist to a single instance behind a small context provider inside `<Form>`, consumed by both. `UseMpdGoalPreviewResult` is already exported (though currently unused), which suggests this was considered.

Worth confirming with the API team that `previewNewStaffGoalCalculation` is genuinely side-effect-free — it is a GraphQL *mutation*, and if it writes anything the doubling becomes a correctness issue rather than a performance one.

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsWarning.tsx
```diff
@@ -0,0 +1,43 @@
...
+    savedSalaryOverCap: calculation.calculations.salaryOverCap,
+  });
+
+  return (
+    <Stack direction="row" spacing={1} alignItems="center" aria-live="polite">
+      {salaryOverCap && (
+        <>
+          <WarningAmber fontSize="small" sx={{ color: 'warning.main' }} />
+          <Typography variant="body2" sx={{ color: 'warning.main' }}>
+            {t('Requested salary is over the standard cap')}
```
zweatshirt: [Medium] **Copy: "Requested salary" names a combined household figure, and doesn't convey that the warning survives a granted exception.**

Server-side: `salary_over_cap? = combined_uncapped_total_salary > base_salary_cap`, where the combined figure is `staff + spouse` and the cap is keyed on marital status (85k single / 135k married, higher in NYC).

The form directly above renders **two** fields both labelled "Annual Requested Salary" (`FinancialInformationSection.tsx:41-51`, disambiguated only by person name). John at 70k + Jane at 70k trips the 135k married cap while neither field looks unusual — the admin hunts for a problem in the wrong field.

Second issue: the API deliberately warns **even when an over-cap exception is granted** (spec: `'still warns when an over-cap exception is granted'`). The `Allow Salary Over Cap` select is in `ExemptionsSection.tsx:71-75`, immediately above this bar — so an admin can set "Yes, any amount" and watch an unclearable warning persist, with nothing explaining why.

To be explicit: **do not gate the warning on `allowSalaryOverCap`** — that would contradict the API's deliberate design. Fix the copy instead:

```tsx
{t('Total requested salary is over the standard cap')}
```

plus a tooltip noting it compares against the pre-exception cap and stays visible when an exception is applied.

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsWarning.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+const mutationSpy = jest.fn();
+
+interface TestComponentProps {
+  /** The warning flag on the saved record. */
+  savedSalaryOverCap?: boolean;
+  /** The warning flag the preview request comes back with. */
+  previewSalaryOverCap?: boolean;
+  /** Replaces the preview response, to make the request fail. */
+  mocks?: ApolloErgonoMockMap;
```
zweatshirt: [Medium] **Dead test prop — the failure path it was built for is never exercised.**

This prop is documented *"Replaces the preview response, to make the request fail"*, but **no test passes it**. All four call sites use only the boolean props. So the `mocks ??` fallback on line 32 is unreachable and the `ApolloErgonoMockMap` import exists only to type a prop nothing uses.

The path it would have covered is the `.catch()` in `useMpdGoalPreview.ts`, which sets `salaryOverCap: null` so the fallback yields the *saved* flag rather than clearing the warning. That behaviour is currently unasserted — if someone changed the catch to set `false`, a network blip would silently erase a live warning and every test would still pass.

Severity is moderated to 5.0 because the warning is advisory and self-corrects on save, so the blast radius is a missed hint rather than bad data.

**Either add the test or delete the prop.** Leaving documented-but-unused scaffolding implies the failure path is covered when it isn't — five independent agents read it that way.

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsWarning.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+              salaryOverCap: savedSalaryOverCap,
+            },
+          }}
+        />
+      </Form>
+    </Formik>
+  </GqlMockedProvider>
+);
+
+describe('GoalSettingsWarning', () => {
```
zweatshirt, line 123: [Medium] **Real timers against a 500ms debounce — ~15x slower than the sibling, with CI flake risk.**

This file never calls `jest.useFakeTimers()`. `useMpdGoalPreview` debounces via `useDebouncedValue(previewable, PREVIEW_DEBOUNCE_MS /* 500 */)`, so every preview test burns 500ms of real wall clock. Measured:

```
GoalSettingsWarning.test.tsx (real timers)
  raises a warning that an unsaved edit introduces     587 ms
  clears a saved warning that an unsaved edit resolves 565 ms

MpdGoalPreview.test.tsx (fake timers, same hook, same debounce)
  shows a spinner while the preview is loading          16 ms
  falls back to the saved goal when the preview fails   55 ms
```

The repo sets no `asyncUtilTimeout`, so RTL's `waitFor` budget is the 1000ms default. A fixed 500ms of that is spent before the request is even issued, leaving ~400ms for Apollo + React under parallel jest workers — the classic profile for a test that is green locally and intermittently red on a loaded runner.

```tsx
describe('GoalSettingsWarning', () => {
  beforeEach(() => { jest.useFakeTimers(); });
  afterEach(() => { jest.useRealTimers(); });
```

Matches the convention already set by `MpdGoalPreview.test.tsx:94-100`.

src/components/HrTools/NsGoalCalculator/GoalSettings/useMpdGoalPreview.ts
```diff
@@ -37,13 +44,20 @@ interface UseMpdGoalPreviewResult {
...
+  salaryOverCap: boolean;
 }
 
 /**
- * Orchestrates the goal preview for {@link MpdGoalPreview}: watches the Goal
- * Settings form and, whenever the unsaved, valid form values change, debounces
- * a call to the API to recompute the goal (`previewNewStaffGoalCalculation`),
- * exposing the new total plus the signed difference from the saved goal.
+ * Orchestrates the goal preview for {@link MpdGoalPreview} and
+ * `GoalSettingsWarnings`: watches the Goal Settings form and, whenever the
```
zweatshirt: [Medium] **Stale doc reference — `GoalSettingsWarnings` (plural) doesn't exist.**

The component shipped in this PR is `GoalSettingsWarning` (singular); `grep -rn "GoalSettingsWarnings" src/` returns only this docblock. Line 55 already uses `{@link MpdGoalPreview}`, so the bare backtick form is inconsistent too.

```diff
- * Orchestrates the goal preview for {@link MpdGoalPreview} and
- * `GoalSettingsWarnings`: watches the Goal Settings form and, whenever the
+ * Orchestrates the goal preview for {@link MpdGoalPreview} and
+ * {@link GoalSettingsWarning}: watches the Goal Settings form and, whenever the
```

src/components/HrTools/NsGoalCalculator/GoalSettings/useMpdGoalPreview.ts
```diff
@@ -17,16 +17,23 @@ interface PreviewState {
...
   savedMonthlyGoal: number;
+  /**
+   * The warning flag on the saved record, shown until a preview settles.
+   * Optional for callers that only render the goal figure.
+   */
+  savedSalaryOverCap?: boolean;
 }
 
-interface UseMpdGoalPreviewResult {
+export interface UseMpdGoalPreviewResult {
```
zweatshirt: [Suggestion] **`export` added with no consumer.**

`grep -rn "UseMpdGoalPreviewResult" src pages` finds zero references outside this file. Either it's vestigial from an abandoned context approach — in which case revert to a local `interface` — or it's the right export for the shared-provider refactor suggested on `GoalSettingsWarning.tsx:24`, in which case land the consumer alongside it.

src/components/HrTools/NsGoalCalculator/GoalSettings/useMpdGoalPreview.ts
```diff
@@ -52,6 +66,7 @@ export const useMpdGoalPreview = ({
   accountListId,
   calculationId,
   savedMonthlyGoal,
+  savedSalaryOverCap = false,
```
zweatshirt: [Suggestion] **Optional arg defaulting to `false` fails open — currently harmless.**

A future consumer that reads `salaryOverCap` without passing `savedSalaryOverCap` gets a confident "no warning" rather than a type error. Downgraded from 4.5 to 2.0 on verification: the only caller that omits it (`MpdGoalPreview.tsx:31-33`) destructures `{ calculating, displayGoal, diff, changed, failed }` and never reads `salaryOverCap`, so no live defect exists.

If the shared-provider refactor lands, the single call site always supplies the value and this disappears on its own.

zweatshirt (commented):
## Multi-Agent Code Review — PR #1930

**Verdict: APPROVED WITH SUGGESTIONS** — no blockers. Highest finding is 6.5/10.

**Risk: 7/10 (HIGH)** — `src/components/**/*.graphql` (+2), 4x feature components (+4), 146 non-test lines (+1). Senior review recommended. Codegen verified clean (`yarn gql` succeeds, `yarn lint:ts` exits 0, 213 tests pass).

Two rounds: 6 specialized agents each. Round 2 was re-run with round-1 findings adjudicated against verified facts, which recalibrated three findings downward and surfaced two better ones.

### Key recalibrations from round 1

| Finding | Was | Now | Why |
|---|---|---|---|
| Failed preview silently suppresses warning | 7.5 | 3.5 | Self-corrects on save (`UpdateNewStaffGoalCalculation` returns the fragment incl. `salaryOverCap`; `enableReinitialize` set; Save never gated on it) |
| Duplicate preview mutations | 7.5 | 6.5 | `BatchHttpLink` (batchInterval 20ms) coalesces both into one HTTP request; cost is doubled server compute, not round-trips |
| `savedSalaryOverCap = false` fails open | 4.5 | 2.0 | The only caller omitting it (`MpdGoalPreview`) never reads `salaryOverCap` |

### Verified correct (no action)

- The always-mounted empty `aria-live` region is proper live-region technique — mutation-tested: deleting `aria-live` and making the region conditional both fail the suite.
- Cache normalization correct (`NewStaffGoalCalculationCalculations` has no `id`; non-normalizable value type). `fetchPolicy: 'no-cache'` correct (`NewStaffGoalCalculationPreview` is a distinct typename).
- Destructuring-with-defaults change is behaviourally identical (`monthlyGoal: Float!`, `salaryOverCap: Boolean!`).
- Scenario-goal path (`accountListId: null`) is valid per schema and pre-existing via `MpdGoalPreview`.
- Save-while-preview-in-flight race traced and safe.
- Cap semantics: a non-joining **senior-staff spouse's** salary IS included against the married cap (`applicable? = !spouse? || !single?`) — coherent, no bug. SOSA correctly treated as single.
- Mock fixtures deterministic (`seed = 'seed'`, probed across processes).
- `salaryOverCap: false` added to `MpdGoalPreview.test.tsx` is inert — removing it still passes 15/15.
- Missing `public/locales/en` key follows the repo's batch-extraction convention. Not a violation.

### Findings outside this PR's diff (informational, not blockers)

**1. The contrast fix belongs in `src/theme.ts`, not in this component.** `src/theme.ts` never defines `palette.warning`, so `warning.main` resolves to MUI's default `#ed6c02` (3.11:1 on `#fff`). But HrTools uses `warning.main` in **7 of 7** pre-existing cases — changing only this file would make it the sole diverging file. Systemic fix:
```ts
warning: { main: statusColors.warning },  // #D34400 -> 4.57:1
success: { main: statusColors.success },  // CardTitle.tsx:24 has the same latent defect
```
Pre-existing HrTools-wide. Recommend a standalone ticket rather than blocking this PR.

**2. Review-process gap.** `.claude/rules/code-review.md` lists `src/components/User/impersonate*`, which matches nothing — impersonation lives in `Settings/Admin/ImpersonateUser/`, `Settings/Organization/ImpersonateUser/`, and `ProfileMenu/`. **The Security agent's impersonation trigger never fires.** Also stale: `src/components/Reports/SalaryCalculator/**` and `pages/api/Schema/uploads/**` don't exist.

**3. Doc drift.** `CLAUDE.md` and the rules file say "Material UI v5"; `yarn.lock` resolves `@mui/material` **7.3.11**.

**4. Questions for the API team (mpdx_api #3469):**
- `salary_over_cap?` calls `base_salary_cap` unconditionally, whereas `salary_cap` only reaches it when `allow_salary_over_cap == 'no'`. It resolves via `find_by!`, which raises. Are the four `board_approved_salary_calc` misc constants seeded for every historical `calculations_year`? A raise would null the whole `calculations` block via non-null propagation, taking `monthlyGoal` down with it.
- `salary_over_cap?` uses strict `>` — confirm equal-to-cap is intended as compliant.

### Recommended order of work

1. The flicker fix (`useMpdGoalPreview.ts:151`) — one line, real user-facing defect
2. Two test assertions (warning renders in the form; `accountListId` in both directions) — closes a mutation-proven hole
3. Resolve the dead `mocks` prop either way
4. Copy revision
5. Defer: theme palette, doc drift, rules globs

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsWarning.tsx
zweatshirt: /dismiss Out of scope but I am thinking we should make tickets for this

src/components/HrTools/NsGoalCalculator/GoalSettings/useMpdGoalPreview.ts
```diff
@@ -131,6 +148,8 @@ export const useMpdGoalPreview = ({
       ? 0
       : Math.round((previewGoal - savedMonthlyGoal) * 100) / 100;
   const changed = Math.abs(diff) >= CENT_EPSILON;
+  const salaryOverCap =
+    (dirty ? preview?.salaryOverCap : null) ?? savedSalaryOverCap;
```
zweatshirt: [Medium] **This guard diverges from the rest of the hook.**

Flagged independently by four agents (Architecture 6.0, Data Integrity 5.0, UX 4.5, Financial 3.5).

`previewable` is `dirty && isValid` (line 88). Every other derived value in this hook gates on it; this line gates on `dirty` alone. They differ on exactly one state — **dirty and invalid** — and the comment already on lines 86-87 documents the invariant that breaks:

> `// The attributes to preview, or null while the form is clean or invalid (which also clears any prior preview).`

Two reachable consequences, both reproduced by agents against this commit:

**1. An invalid form freezes a stale verdict.** Previewing stops entirely when `isValid` goes false, so the verdict can never be re-checked — but it keeps displaying. Meanwhile `displayGoal` reverts to `savedMonthlyGoal`, so the header and the warning describe different form states indefinitely. `MpdGoalPreview.test.tsx` already pins the opposite behavior for the goal figure (`drops the preview when a previewed edit is then made invalid`).

**2. An undone edit's verdict resurrects.** `preview` is never cleared, so it survives dirty → pristine → dirty. Type an over-cap salary, undo it, then edit any other field, and the warning from the reverted edit reappears for the debounce + round-trip.

**The one-line swap to `previewable !== null` fixes only #1.** Comparing against the pre-PR behavior settles which fix is right — pre-PR, `settledPreview` required *both* `previewable !== null` and an attributes-identity match, so it returned the saved value in every state below except the one the flicker fix deliberately targets:

| State | Pre-PR | `previewable` swap | Clearing effect |
|---|---|---|---|
| Pristine | saved | saved | saved |
| Invalid | saved | saved | saved |
| New edit in flight | saved | stale *(intended)* | stale *(intended)* |
| Undo then re-edit | saved | **stale** | saved |

So bounding `preview`'s lifetime is the minimal deviation from pre-existing behavior:

```ts
// after: const debounced = useDebouncedValue(previewable, PREVIEW_DEBOUNCE_MS);

// Drop a previewed answer once it can no longer be re-checked. Going pristine
// or invalid stops previewing, so keeping one would let the warning assert a
// verdict for values the form no longer holds.
useEffect(() => {
  if (previewable === null) {
    setPreview(null);
  }
}, [previewable]);
```

Keep the `dirty ?` guard — it suppresses a one-render flash before the effect commits. This provably cannot affect the goal figure: the effect only fires when `previewable === null`, and in that state `settledPreview` is already `null` regardless of `preview`.

Not a blocker — the warning is advisory, never gates Save, and in case #1 Save is disabled anyway.

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsWarning.tsx
```diff
@@ -0,0 +1,43 @@
...
+    savedSalaryOverCap: calculation.calculations.salaryOverCap,
+  });
+
+  return (
+    <Stack direction="row" spacing={1} alignItems="center" aria-live="polite">
+      {salaryOverCap && (
+        <>
+          <WarningAmber fontSize="small" sx={{ color: 'warning.main' }} />
+          <Typography variant="body2" sx={{ color: 'warning.main' }}>
+            {t('Total requested salary is over the standard cap')}
```
zweatshirt: [Medium] **"requested salary" is inaccurate when the salary field is empty.**

Reported from live use: the warning appears even with Annual Requested Salary blank. Traced to the API rather than to stale data — `salary_calculation.rb:73-81`:

```ruby
requested_annual_salary = calculator_attribute(:requested_annual_salary)&.to_d
requested_annual_salary || ((geographic_multiplier * base) + tenure_adjustment + debt_salary)
```

An empty field maps to `null` (`toNumberOrNull('')` in `goalSettingsApiMapping.ts:141`), and the server falls back to a **computed** salary — geographic multiplier x base + tenure adjustment + debt salary — then compares that against the cap.

So the flag is right (the goal calculation genuinely uses that figure) but the sentence claims a requested salary that doesn't exist. An admin looking at an empty field and a warning about their "requested salary" has no way to reconcile the two.

Minimal fix, no API dependency:

```diff
-{t('Total requested salary is over the standard cap')}
+{t('Total salary is over the standard cap')}
```

Accurate whether the figure was requested or computed, and it keeps the combined-household sense that "Total" carries.

The better long-term fix is to name the amounts — "Total salary of $148,000 is over the standard cap of $135,000" — which also makes the empty-field case self-explanatory. That needs two new fields on the API; see the review body.

Note this is the third revision of this string, and it churns 13 assertions across three test files, so it's worth deciding once.

src/components/HrTools/NsGoalCalculator/GoalSettings/useMpdGoalPreview.ts
```diff
@@ -17,13 +17,20 @@ interface PreviewState {
...
 interface UseMpdGoalPreviewArgs {
   /** Account list the goal belongs to, or `null` for a scenario goal. */
   accountListId: string | null;
   calculationId: string;
   savedMonthlyGoal: number;
+  /**
+   * The warning flag on the saved record, shown until a preview settles.
+   * Optional for callers that only render the goal figure.
+   */
+  savedSalaryOverCap?: boolean;
```
zweatshirt: [Medium] **Optional input, non-optional output — fails open.**

`savedSalaryOverCap?: boolean` defaults to `false` (line 69), but `UseMpdGoalPreviewResult.salaryOverCap` (line 51) is a required `boolean` documented as an authoritative business fact ("the uncapped salary exceeds the base salary cap").

So any caller that reads `salaryOverCap` without passing `savedSalaryOverCap` gets a confident `false` — "no warning" — for a record that is genuinely over cap, with a clean typecheck and no signal that the value is conditional on an optional input. `false` is exactly the plausible-looking wrong answer for a compliance-adjacent flag.

No live defect today: the only caller that omits it (`MpdGoalPreview.tsx:32-34`) destructures `{ calculating, displayGoal, diff, changed, failed }` and never reads it. This is contract hygiene, not a bug.

The field is now on the `NewStaffGoalCalculationFields` fragment and is `Boolean!` in the schema, so it's always available at the call site — making the arg required costs one prop:

```diff
-savedSalaryOverCap?: boolean;   // ... savedSalaryOverCap = false
+savedSalaryOverCap: boolean;    // no default; every caller supplies it
```

src/components/HrTools/NsGoalCalculator/GoalSettings/useMpdGoalPreview.ts
```diff
@@ -94,19 +109,21 @@ export const useMpdGoalPreview = ({
...
       .catch(() => {
         // Record the failed attributes so the spinner stops and we fall back to
-        // the saved goal.
+        // the saved goal and warning.
         if (active) {
-          setPreview({ attributes: debounced, monthlyGoal: null });
+          setPreview({
+            attributes: debounced,
+            monthlyGoal: null,
+            salaryOverCap: null,
```
zweatshirt, line 125: [Suggestion] **A failed preview silently erases a warning the user already raised — and nothing tests it.**

This `.catch` sets `salaryOverCap: null`, which line 152's `??` coalesces to `savedSalaryOverCap`. So an admin who raises the salary over cap, sees the warning, then makes any further edit whose preview request fails, watches the warning **disappear** — and can reasonably read that as "resolved". There is no snackbar (`context: { suppressErrors: true }`, line 108) and `GoalSettingsWarning` never consumes the hook's `failed` flag, so unlike `MpdGoalPreview` — which renders "Preview unavailable" for the same failure — the warning bar shows nothing at all.

This also contradicts the flicker fix's own principle: every other path holds the last known answer rather than reverting to saved.

**Untested.** Changing this `null` to `false` left **165 of 165 tests passing**. It became unreachable from the suite when the `mocks` prop was removed from `GoalSettingsWarning.test.tsx` — that prop was the only handle for forcing a failure.

Two defensible resolutions:

```ts
// (a) hold the last known verdict rather than dropping it
.catch(() => {
  if (active) {
    setPreview((prev) => ({
      attributes: debounced,
      monthlyGoal: null,
      salaryOverCap: prev?.salaryOverCap ?? null,
    }));
  }
});
```

or **(b)** keep reverting to saved, but have `GoalSettingsWarning` render the `failed` state the way `MpdGoalPreview` does, so the disappearance is explained rather than silent.

Either way, add a test — re-introducing the `mocks` override (the pattern still present in `MpdGoalPreview.test.tsx`) is enough.

Severity is bounded because the warning is advisory, never gates Save, and self-corrects on the next successful preview or on save.

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsForm.test.tsx
```diff
@@ -334,6 +334,26 @@ describe('GoalSettingsForm', () => {
...
+              ...defaultGoalCalculation.calculations,
+              salaryOverCap: true,
+            },
+          },
+        }}
+      />,
+    );
+
+    expect(
+      await findByText('Total requested salary is over the standard cap'),
```
zweatshirt: [Suggestion] **Load-sensitive assertion.**

`await findByText(...)` with RTL's default 1s budget is the first and only assertion here, and it gates on the full `NewStaffGoalCalculation` query resolving *plus* the entire heavy form rendering.

It failed twice during review under parallel load (once as `1 failed, 58 passed` across four suites at 25.9s wall time) but passes reliably in isolation and did not reproduce across four `--maxWorkers=4` stress runs. So: genuine CI-load intermittency rather than a deterministic break.

Anchor on something cheap first so the 1s budget covers only the warning:

```tsx
await findByRole('button', { name: 'Save & Share' });
expect(
  await findByText('Total requested salary is over the standard cap'),
).toBeInTheDocument();
```

or pass `{ timeout: 3000 }`.

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsWarning.tsx
```diff
@@ -0,0 +1,43 @@
...
+import WarningAmber from '@mui/icons-material/WarningAmber';
+import { Stack, Typography } from '@mui/material';
+import { useTranslation } from 'react-i18next';
+import { useMpdGoalPreview } from './useMpdGoalPreview';
+import { NewStaffGoalCalculation } from './useNewStaffGoalCalculation';
+
+interface GoalSettingsWarningProps {
+  /** Account list the goal belongs to, or `null` for a scenario goal. */
+  accountListId: string | null;
+  calculation: NewStaffGoalCalculation;
```
zweatshirt: [Suggestion] **Prop shape diverges from its sibling.**

This component takes the whole `calculation: NewStaffGoalCalculation` and reads exactly three fields from it (`.id`, `.calculations.monthlyGoal`, `.calculations.salaryOverCap`). Its direct sibling and co-consumer of the same hook, `MpdGoalPreview.tsx:7-16`, takes flat scalars (`calculationId`, `savedMonthlyGoal`).

Two components rendered from the same parent, driven by the same hook, with opposite prop conventions. The object form also couples a purely presentational warning to the full generated fragment type — which is why both test files have to spread a whole `defaultGoalCalculation` just to flip one boolean.

Matching the sibling (`calculationId`, `savedMonthlyGoal`, `savedSalaryOverCap`) would also resolve the optional-arg finding above, since `GoalSettingsForm` has all three in scope.

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsWarning.test.tsx
```diff
@@ -0,0 +1,191 @@
...
+  savedSalaryOverCap?: boolean;
+  /** The warning flag the preview request comes back with. */
+  previewSalaryOverCap?: boolean;
+}
+
+const TestComponent: React.FC<TestComponentProps> = ({
+  savedSalaryOverCap = false,
+  previewSalaryOverCap = false,
+}) => (
+  <GqlMockedProvider<{
```
zweatshirt: [Suggestion] **No `ThemeProvider`, unlike every sibling.**

11 test files in this directory wrap in `<ThemeProvider theme={theme}>`, including the closest analogue `MpdGoalPreview.test.tsx:56-58`.

`GoalSettingsWarning.tsx:35-36` styles on `warning.main`, so in tests it resolves against MUI's **default** palette rather than `src/theme`. The suite therefore cannot catch a token that exists in the default theme but not the project's — which is precisely the situation here (`palette.warning` is undefined in `src/theme.ts`, see the review body).

Low blast radius, but it's the reason the contrast issue was invisible to CI.

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsWarning.test.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
+  it('warns about an over-cap salary on the saved calculation', () => {
+    const { getByText } = render(<TestComponent savedSalaryOverCap />);
+
+    expect(
+      getByText('Total requested salary is over the standard cap'),
```
zweatshirt: [Suggestion] **Repeated literal and inlined helper.**

The warning string appears **14 times** across this file, and `userEvent.clear(...)` + `userEvent.type(...)` is inlined 8 times where the sibling `MpdGoalPreview.test.tsx:91` already defines an `editField(field, value)` helper.

Maintenance only — no correctness risk — but a copy change currently means 14 edits here plus one in `GoalSettingsForm.test.tsx`, which is a real cost given the string has now been revised twice.

```tsx
const overCapWarning = 'Total requested salary is over the standard cap';
const editField = (field: HTMLElement, value: string) => {
  userEvent.clear(field);
  userEvent.type(field, value);
};
```

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsForm.test.tsx
```diff
@@ -334,6 +334,26 @@ describe('GoalSettingsForm', () => {
     );
   });
 
+  it('shows the over-cap warning beside the save actions', async () => {
```
zweatshirt: [Suggestion] **Test name overclaims what is asserted.**

`shows the over-cap warning beside the save actions` asserts only `findByText(...)` — presence anywhere in the document. The placement claim ("beside the save actions") is the component's stated reason for existing per its docblock, but moving `<GoalSettingsWarning>` out of `<StickyActionBar>` would not fail this test.

The test does its primary job — deleting the component from the form fails it — so the cheapest fix is to rename to `shows the over-cap warning`, rather than add a brittle DOM-containment assertion.

zweatshirt (commented):
## Multi-Agent Code Review — PR #1930 (Round 3)

**Verdict: APPROVED WITH SUGGESTIONS** — no blockers. Highest finding 6.0/10.

**Risk: 7/10 (HIGH)** — `src/components/**/*.graphql` (+2), 4x feature components (+4), ~145 non-test lines (+1). Senior review recommended.

6 agents (Architecture, Testing, Standards, Data Integrity, UX, Financial Reporting). Third round on this branch; the six fix-up commits since the last review closed every prior finding, so this pass focuses on the new code.

### Confirmed fixed since the last review

Verified by mutation testing — each item below now fails a targeted test when reverted:

| Fix | Guarded by |
|---|---|
| Warning flicker on every edit | `does not re-raise a resolved warning…` |
| `accountListId` untested in both directions | both `useNewStaffGoalCalculation` tests |
| Warning-in-form integration unasserted | `shows the over-cap warning beside the save actions` |
| Real timers against a 500ms debounce | fake timers added; probed with a 5,000,000ms debounce |
| Dead `mocks` test prop | removed |
| Stale plural JSDoc, unused `export` | fixed |

**Mutation audit: 9 of 10 mutations caught.** The one gap is the `.catch` fallback — see the inline comment on `useMpdGoalPreview.ts`.

### Root cause found for "the warning shows when Annual Requested Salary is empty"

This was reported during review and is **not** stale persistence. `app/services/new_staff_goals/salary_calculation.rb:73-81`:

```ruby
requested_annual_salary = calculator_attribute(:requested_annual_salary)&.to_d
requested_annual_salary || ((geographic_multiplier * base) + tenure_adjustment + debt_salary)
```

An empty field maps to `null` on the client (`toNumberOrNull('')`), and the API then **computes a default salary** and compares that against the cap. The flag is correct — the goal calculation really does use that computed figure. The copy is what's wrong: it says "requested salary" when nothing was requested. See the inline comment on `GoalSettingsWarning.tsx`.

### Findings on related files (not in this PR)

**1. Contrast fix belongs in `src/theme.ts`, not in the component.** `palette.warning` is never defined, so `warning.main` resolves to MUI's default `#ed6c02` — **3.11:1** on `#fff`, failing the 4.5:1 bar for 14px `body2`. The repo's own `statusWarning` (`#D34400`) is 4.57:1. But HrTools uses `warning.main` in **7 of 7** pre-existing cases, so changing only this file would make it the sole outlier. Systemic fix:
```ts
warning: { main: statusColors.warning },  // #D34400 -> 4.57:1
success: { main: statusColors.success },  // CardTitle.tsx:24 has the same latent defect
```
Pre-existing and tree-wide — should not block this PR. Standalone ticket recommended.

**2. Review-process gap.** `.claude/rules/code-review.md` lists `src/components/User/impersonate*`, which matches nothing — impersonation lives under `Settings/Admin/ImpersonateUser/`, `Settings/Organization/ImpersonateUser/`, and `ProfileMenu/`. **The Security agent's impersonation trigger never fires.** Also stale: `src/components/Reports/SalaryCalculator/**` and `pages/api/Schema/uploads/**`.

**3. Doc drift.** `CLAUDE.md` and the rules file say "Material UI v5"; `yarn.lock` resolves `@mui/material` **7.3.11**.

### Questions for the API team (mpdx_api #3469)

1. **Expose the amounts behind the warning.** To say "Total salary of $X is over the standard cap of $Y" the client needs `combined_uncapped_total_salary` and `base_salary_cap` on `NewStaffGoalCalculationCalculations`. Neither is exposed today, and the three salary fields that are (`salary`, `salarySubtotal`, `totalSalary`) are all **monthly** and **post-capping** — `base_salary` scales down when the combined total exceeds the cap, so `totalSalary` is by construction never over the cap. Rendering any of them beside "is over the cap" would show a number that isn't. This is the blocker on improving the copy further.
2. **Seeding.** `salary_over_cap?` calls `base_salary_cap` unconditionally, whereas `salary_cap` only reaches it when `allow_salary_over_cap == 'no'`. It resolves via `MpdGoalMiscConstant.find_by!`, which raises, and a raise would null the whole `calculations` block via non-null propagation — taking `monthlyGoal` with it. The in-repo seeder is all-or-nothing per year (so the risk is pre-existing, not new), but production seeding isn't in the repo. Is the prod path also all-or-nothing per year, and is every `calculations_year` in use backfilled?
3. **Seeded cap values differ from the spec.** `MpdGoalMiscConstantsSeeder` has 80k/125k/90k/145k; `goal_calculator_spec.rb` stubs 85k. Which figures are current?

canac:
@zweatshirt Are you also going to add warnings that the user is healthcare exempt, SECA exempt, or over the debt cap?

(Sorry I missed the debt cap before with the API PR. I haven't dug in, but that check may need to happen on the server.)

zweatshirt:
@canac I can add that to this PR. My brain assumed we weren't adding them due to what we concluded with the backend PR, but it makes sense we want them still.

canac:
Yesterday, I was just thinking about the `NSGC!DA5` checks, but it slipped my mind that we also need to implement `NSGC!A33`: `IF(CV5<>"No","SECA Opted Out("&CV5&") - ","")&IF(CU5<>"No","Healthcare Exempt("&CU5&") - ","")&IF(CW5<>"No","Allow OverSalCap("&CW5&") - ","")&IF(CX5<>"No","Allow OverDebtCap("&CX5&") - ","")`. Sorry about that!

zweatshirt:
@canac The backend PR added a boolean for healthcare exemption but you mentioned we don't need to check for it and so I removed it, are we still wanting to implement it?

canac:
The sheet just surfaces a non-blocking warning when someone has an opt out or exemption. I think we can just read the booleans from the goal calculation and show a message if any of them are true. Like it says something like "SECA Opted Out (Both)", for example. I don't think that needs to be a validation in the backend calculations. What do you think?

zweatshirt:
@canac That was what I was intending with the boolean in the GraphQL for healthcare exemption. Is the goal to do as much of the logic for the warnings in the the frontend then? My idea was that we could send booleans to the frontend on which warnings to display and have the frontend act on those (or in the case of SECA we could maybe do an enum)

I guess the frontend has enough information for most of these apart from the salary cap and debt cap to do it in the frontend instead, though

zweatshirt:
@canac This is basically ready for review. I've done multiple review agent iterations, it is considered critical. Sorry, it turned out to be a much bigger PR than I expected. Feel free to share constructive criticism.

zweatshirt:
@canac Did you want me to redo this PR? I noticed it hadn't been reviewed in a while. It ended up being a big PR, and there are always areas to improve, so I am happy to rework it if needed.

canac:
Oh sorry, for some reason I thought you were going to make more changes. I can look over it tomorrow if it's ready for review!

src/components/HrTools/NsGoalCalculator/GoalSettings/Fields/useGoalSettingsField.test.tsx
```diff
@@ -1,19 +1,61 @@
...
+          ...calculationToFormValues(defaultGoalCalculation),
+          allowSalaryOverCap,
+          secaExempt,
+          spouseSecaExempt,
+        }}
+        initialErrors={initialErrors}
+        initialTouched={initialTouched}
+        onSubmit={jest.fn()}
+      >
+        <GoalSettingsPreviewProvider
```
zweatshirt: [Suggestion] Every test in this file now renders inside `GoalSettingsPreviewProvider`, so the `useGoalSettingsPreview() === null` (no-provider) branch is no longer exercised *here*. It is still covered by the untouched field-component tests (`GoalSettingsTextField/Select/NumberField/YesNoField.test.tsx`), so this is a locality nit, not a coverage hole — optional: re-add one no-provider case asserting no outline is applied.

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsWarning.test.tsx
```diff
@@ -0,0 +1,550 @@
...
+        maritalStatus={Single}
+        secaExempt="true"
+        spouseSecaExempt="true"
+      />,
+    );
+
+    expect(getByText('Staff opted out of SECA')).toBeInTheDocument();
+    expect(queryByText('Both opted out of SECA')).not.toBeInTheDocument();
+  });
+  it('announces warnings politely, from a region that exists while empty', () => {
```
zweatshirt, line 424: [Medium] Untested error path. A **failed** preview leaves `previewSalaryOverCap`/`previewDebtOverCap` null, so the warning should revert to the *saved* over-cap flags. The goal-figure failure fallback is tested (`MpdGoalPreview.test.tsx`), but the warning-side fallback is not. Add a test with a rejecting `PreviewNewStaffGoalCalculation` mock + `savedSalaryOverCap` asserting the saved warning stays shown after an edit fails.

src/components/HrTools/NsGoalCalculator/GoalSettings/goalSettingsWarnings.ts
```diff
@@ -0,0 +1,179 @@
...
+
+export const getFieldSeverity = (
+  warnings: GoalSettingsWarningItem[],
+  name: string,
+): WarningSeverity | undefined => {
+  const severities = warnings
+    .filter(({ fields }) => fields.includes(name))
+    .map(({ severity }) => severity);
+
+  return severities.find((severity) => severity === 'error') ?? severities[0];
```
zweatshirt, line 143: [Suggestion] `severities.find((s) => s === 'error') ?? severities[0]` — the error-precedence branch is currently **unreachable**: `SALARY_CAP_FIELDS`, `SECA_FIELDS`, and `HEALTHCARE_FIELDS` are disjoint and the other warnings carry `fields: []`, so no field ever appears in two warnings and `severities.length` is always 0 or 1. Either add a direct unit test with a field in both an error and a warning to lock the precedence in, or note in-code that it is defensive.

src/components/HrTools/NsGoalCalculator/GoalSettings/goalSettingsWarnings.ts
```diff
@@ -0,0 +1,179 @@
...
+      severity: 'warning',
+      message: healthcareParty ? healthcareMessages[healthcareParty] : '',
+      fields: healthcareParty ? HEALTHCARE_FIELDS[healthcareParty] : [],
+    },
+  ];
+
+  return candidates.filter(({ active }) => active);
+};
+
+export const getFieldSeverity = (
```
zweatshirt, line 135: [Suggestion] This pure rule engine (`buildGoalSettingsWarnings` / `getFieldSeverity`) is only covered indirectly through three component test files. A colocated `goalSettingsWarnings.test.ts` exercising the DA5/A33 truth table directly would be cheaper, document the rules in one place, and satisfy the repo's 'colocated test per new lib function' checklist item. Non-blocking — behavior is well covered indirectly today.

src/components/HrTools/NsGoalCalculator/GoalSettings/Fields/useGoalSettingsField.ts
```diff
@@ -17,6 +19,19 @@ export type GoalSettingsFieldBaseProps = Omit<TextFieldProps, 'name'> & {
   showLabel?: boolean;
 };
 
+const outlineSx = (severity: WarningSeverity) => ({
```
zweatshirt, line 22: [Suggestion] The contributing-field outline is **color-only**. For salary/debt the warning text names the concept, but fields like `geographicLocation` and `contribution403bPercentage` (in `SALARY_CAP_FIELDS`) get no textual/icon/ARIA pointer — the tinted border is their sole cue. `aria-invalid` is correctly avoided (these are advisory, not validation errors), so this is an acceptable tradeoff, but a known a11y residual worth a follow-up if those fields need a non-color indicator.

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsWarning.tsx
```diff
@@ -0,0 +1,50 @@
...
+      {SEVERITIES.map((severity) => {
+        const grouped = warnings.filter(
+          (warning) => warning.severity === severity,
+        );
+
+        return grouped.length ? (
+          <Alert
+            key={severity}
+            severity={severity}
+            role="presentation"
```
zweatshirt, line 37: [Suggestion] The error-vs-warning distinction is conveyed only visually (MUI's severity icon/color). The same text 'Total salary is over the standard cap' lands in the error or warning group depending on whether an exception is approved, but a screen-reader/colorblind user gets no cue to the allowed-vs-not distinction. The primary 'over cap' fact is fully in text, so low-priority polish.

src/components/HrTools/NsGoalCalculator/GoalSettings/useMpdGoalPreview.ts
```diff
@@ -125,12 +143,14 @@ export const useMpdGoalPreview = ({
...
-  const diff =
-    previewGoal === null
-      ? 0
-      : Math.round((previewGoal - savedMonthlyGoal) * 100) / 100;
-  const changed = Math.abs(diff) >= CENT_EPSILON;
+  const previewSalaryOverCap = dirty ? (preview?.salaryOverCap ?? null) : null;
+  const previewDebtOverCap = dirty ? (preview?.debtOverCap ?? null) : null;
 
-  return { calculating, displayGoal, diff, changed, failed };
+  return {
```
zweatshirt, line 149: [Suggestion] The hook now returns `previewSalaryOverCap`/`previewDebtOverCap` (driving the A33/DA5 warnings), not just the MPD goal figure, and is consumed only by `GoalSettingsPreviewProvider`. The name `useMpdGoalPreview` reads narrower than its actual role — consider a rename (e.g. `useGoalPreview` / `useGoalSettingsPreviewData`). Purely cosmetic.

src/components/HrTools/NsGoalCalculator/GoalSettings/goalSettingsWarnings.ts
```diff
@@ -0,0 +1,179 @@
...
+  // Discrepancy with the sheet: DA5's plan list has no "Exempt" option, so it
+  // never defined this case. Following it literally makes Exempt a violating
+  // plan. Revisit with whoever owns the plan list if that reading is wrong.
+  const exemptPlanMismatch =
+    everyoneExempt &&
+    values.benefitsPlan !== '' &&
+    values.benefitsPlan !== MpdGoalBenefitsConstantPlanEnum.Select;
+
+  // Only reachable via `everyoneExempt`, which is either both spouses or a lone applicant
+  const staffName = values.firstName || t('Staff');
```
zweatshirt: [Suggestion] The `t('Staff')` / `t('Spouse')` name fallbacks (when `firstName`/`spouseFirstName` are empty) are never exercised — every exempt-plan test supplies real names. A one-line test with an empty name would cover them.

src/components/HrTools/NsGoalCalculator/GoalSettings/useMpdGoalPreview.ts
```diff
@@ -125,12 +143,14 @@ export const useMpdGoalPreview = ({
   const failed = settledPreview !== null && settledPreview.monthlyGoal === null;
 
   const previewGoal = settledPreview?.monthlyGoal ?? null;
-  const displayGoal = previewGoal ?? savedMonthlyGoal;
-  const diff =
-    previewGoal === null
-      ? 0
-      : Math.round((previewGoal - savedMonthlyGoal) * 100) / 100;
-  const changed = Math.abs(diff) >= CENT_EPSILON;
+  const previewSalaryOverCap = dirty ? (preview?.salaryOverCap ?? null) : null;
```
zweatshirt, line 146: [Dismissed by author: intentional, documented behavior] [Suggestion] `previewSalaryOverCap`/`previewDebtOverCap` are held from the last *settled* preview across an in-flight/invalid edit (gated only on `dirty`), while `previewGoal` uses `settledPreview`. So during an in-flight or invalid edit the header can show the saved goal/spinner while the cap warning/outline persists — a transient 'divergence.' Deliberate anti-flicker / anti-re-announce behavior; the 'can't disagree at rest' invariant holds when a concrete goal is shown, and Save is gated on `isValid`. Raised by Financial + Data Integrity as intentional.

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsPreviewContext.tsx
```diff
@@ -0,0 +1,89 @@
...
+    failed,
+    previewGoal,
+    previewSalaryOverCap,
+    previewDebtOverCap,
+  } = useMpdGoalPreview({ accountListId, calculationId });
+
+  const salaryOverCap = previewSalaryOverCap ?? savedSalaryOverCap;
+  const debtOverCap = previewDebtOverCap ?? savedDebtOverCap;
+
+  const value = useMemo<GoalSettingsPreviewValue>(() => {
```
zweatshirt, line 64: [Dismissed by author: intentional, documented behavior] [Suggestion] The `useMemo` depends on `values`, so `buildGoalSettingsWarnings` (and its eager `t()` message tables) rebuilds on every keystroke. The work is trivially cheap (a few `t()` lookups + array ops), so fine as-is; noting only that the message tables are built even when no exemption is active.

zweatshirt (commented):
## 🤖 Multi-Agent Code Review — ✅ Approved with Suggestions

**6 specialized reviewers** (Architecture, Testing, Standards, Data Integrity, UX, Financial-domain) + dependency analysis. All returned **High confidence with zero critical / high / important findings**. What remains is a set of optional low-severity test-coverage and a11y-polish suggestions.

> **Scope note:** this review deliberately **excludes local commit `e67821d21`** (review-process / governance files) per the author's request. The reviewed tree is the remote PR head `bb84087ac` — the 16 goal-settings feature files — so it matches this PR's diff exactly.

### Risk: 🟢 LOW
| Axis | Score | Why |
|---|---|---|
| Blast Radius | 1 | Contained to `GoalSettings/`; one additive `.graphql` fragment (0 external dependents on every changed export) |
| Complexity | 2 | New warning rule-engine + shared preview state held across in-flight edits |
| Sensitivity | 1 | Displays financial figures; `.graphql` floor. No auth / PII / money-write |
| Recoverability | 0 | CI-gated frontend, fully revertible, no persisted side effects |

**effort 3 · danger 1** → auto-approve eligible.

### Dependency impact
Blast radius genuinely contained. `useMpdGoalPreview`, `MpdGoalPreview`, `GoalSettingsHeader`, `useGoalSettingsField`, and all new exports have **0 external dependents**; every in-directory call site was updated. The only cross-boundary change — adding `salaryOverCap` / `debtOverCap` to `NewStaffGoalCalculationFields` — is purely additive.

### Highlights
- ✅ **Data integrity sound:** previewing (a `mutation`) cannot drift the saved display — protected three ways (`fetchPolicy: 'no-cache'`, distinct `NewStaffGoalCalculationPreview` typename, non-normalized value-object `calculations`).
- ✅ **N+1 removed:** the shared provider collapses per-consumer preview requests into one (verified by "previews an edit exactly once with every consumer mounted").
- ✅ **A11y choices are deliberate & correct:** `aria-live="polite"` + `role="presentation"` on the grouped alerts, and no `aria-invalid` on advisory outlines.
- ✅ **Money math safe:** the single arithmetic line is the established cents-rounding idiom; no `-$0.00` path.

### Findings
8 open suggestions (all severity ≤ 5.0) posted inline. Nothing blocks merge.

### Dismissed by author (acknowledged — not counted toward the verdict)
- **useMpdGoalPreview.ts** (3.5) — held over-cap flag / transient header-vs-warning divergence during in-flight edits — *intentional, documented behavior*
- **GoalSettingsPreviewContext.tsx** (3.0) — warnings rebuild each keystroke — *intentional, documented behavior*

Also: the prior sev-6.5 `/dismiss`ed finding ("every edit runs the goal calculation twice") is **resolved by this PR's** shared-provider refactor — no new finding matches it.

<sub>Security agent skipped (no api/auth/apollo-config/env/workflow/dependency changes in scope). Generated by a multi-agent review harness — findings are advisory.</sub>

zweatshirt:
@canac I can finish up the code given the agent review (sorry, I know it's produced a lot of comments, I wanted to test the new review scoring here). If you are up for it, I was mostly hoping for a human look at the UI.

src/components/HrTools/NsGoalCalculator/GoalSettings/goalSettingsWarnings.ts
```diff
@@ -0,0 +1,179 @@
...
+  };
+
+  // DA5 fires on exempt="Both", or exempt="One" with a Single or SOSA status —
+  // both of which mean everyone on the plan is exempt.
+  const everyoneExempt =
+    healthcareParty === 'both' || (!hasSpouse && healthcareParty === 'staff');
+  // Discrepancy with the sheet: DA5's plan list has no "Exempt" option, so it
+  // never defined this case. Following it literally makes Exempt a violating
+  // plan. Revisit with whoever owns the plan list if that reading is wrong.
+  const exemptPlanMismatch =
```
canac: We don't need this check anymore on the frontend either.

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsForm.tsx
```diff
@@ -155,51 +159,62 @@ export const GoalSettingsForm: React.FC<GoalSettingsFormProps> = (props) => {
...
-                    startIcon={
-                      isSubmitting ? (
-                        <CircularProgress color="inherit" size={20} />
-                      ) : undefined
-                    }
+            <GoalSettingsPreviewProvider
+              accountListId={accountListId}
+              calculationId={calculation.id}
+              savedSalaryOverCap={calculation.calculations.salaryOverCap}
+              savedDebtOverCap={calculation.calculations.debtOverCap}
```
canac: What if we passed the full calculation to the provider and then it can read whichever fields it needs?

```suggestion
              calculation={calculation}
```

src/components/HrTools/NsGoalCalculator/GoalSettings/useMpdGoalPreview.ts
```diff
@@ -125,12 +143,14 @@ export const useMpdGoalPreview = ({
...
+  const previewDebtOverCap = dirty ? (preview?.debtOverCap ?? null) : null;
```
canac, line 147: If we're adding more fields to the preview, maybe we should add the calculated 403b amounts too?

src/components/HrTools/NsGoalCalculator/GoalSettings/goalSettingsWarnings.ts
```diff
@@ -0,0 +1,179 @@
...
+  fields: readonly string[];
+}
+
+const SALARY_CAP_FIELDS: readonly (keyof GoalSettingsFormValues)[] = [
+  'annualRequestedSalary',
+  'spouseRequestedAnnualSalary',
+  'contribution403bPercentage',
+  'spouseContribution403bPercentage',
+  // Sets the cost-of-living multiplier, and so the cap itself.
+  'geographicLocation',
```
canac: We can leave this and see what stakeholders think if you want, but I'm thinking that we shouldn't mark this as invalid since it can't be changed (as long as it's correct). Age, tenure, and marital status also affect the salary cap, but we don't mark those fields as invalid.

canac (approved):
I don't really know what stakeholders want for validation, but this seems like a great start! Hopefully we can get more feedback at some point.

Highlighting the fields that contribute to the goal is a really cool idea!

src/components/HrTools/NsGoalCalculator/GoalSettings/goalSettingsWarnings.ts
zweatshirt: I was implementing this in the backend previously. Are you saying we should drop it completely? I can do that.

canac: Sorry if it was confusing. I think I was telling you in the backend that it wasn't necessary there. It won't be necessary in the frontend at all. The old Google Sheet calculations would be incorrect if you didn't choose Select. But in my Ruby calculations, you can just choose Exempt and everything works.

src/components/HrTools/NsGoalCalculator/GoalSettings/goalSettingsWarnings.ts
canac: Rereading my comment, I don't think I explained it well. I just mean that the location affects the cap but isn't really invalid (unless they typed in the wrong location). Whereas if they said 403b percentage of 50% or requested a salary of $250K, that might be invalid and should be corrected.

zweatshirt: Ah yeah, thanks for the clarity on that, I definitely get what you mean. Maybe it is best to only mark the Annual Requested Salary field as invalid, excluding the fields that that the Annual Requested Salary may depend on to avoid added complexity.

canac: I'm not opposed to keeping the 403b fields marked since they're under the person's control and can push them over their cap. But I don't feel strongly either way.

zweatshirt: I kept the 403b fields markable as invalid and removed geographicLocation from this array

src/components/HrTools/NsGoalCalculator/GoalSettings/useMpdGoalPreview.ts
zweatshirt: https://jira.cru.org/browse/MPDX-9904

PR #1932, opened 2026-07-21
[MPDX-9866] - Disable all inputs of read-only goal calculations

wjames111:
## Description

- Depends on CruGlobal/mpdx_api#3472, which adds the `GoalCalculation.readOnly` field and rejects all mutations (including deletion) against read-only goals. Read-only goals are created automatically when a new staff member reports (MPDX-9865).
- Fetch the new `readOnly` field in the goal calculator and goals list queries, and expose `isReadOnly` through `GoalCalculatorContext`.
- Disable every input of a read-only goal while still showing the inputs and calculated totals: all autosave fields (via the central `useGoalAutoSave` hook), the SECA status select, the geographic location autocomplete, budget grid cell editing, Add Line Item and delete line-item actions, the Lump Sum/Line Item mode toggle, and the lump sum total field.
- Guard the two effects that fire mutations automatically (incompatible benefits plan clearing and default line-item repopulation) so they don't run against read-only goals.
- Hide the Delete button and show a "Read-Only" chip on read-only goal cards in the goals list (`GoalCard.onDelete` is now optional).
- Related Jira: [MPDX-9866](https://jira.cru.org/browse/MPDX-9866), [MPDX-9865](https://jira.cru.org/browse/MPDX-9865)

Note: until the API PR reaches staging, `yarn gql` on this branch needs a staging schema snapshot with `readOnly` grafted on (`API_URL=<snapshot file> yarn gql`).

## Testing

- Mark a goal calculation as read-only (until MPDX-9865's conversion is live, set it directly: `UPDATE goal_calculations SET read_only = true WHERE id = '...'` on staging, or mock `readOnly: true` locally)
- Go to HR Tools → Goal Calculator
- Check that the read-only goal's card shows a "Read-Only" chip and no Delete button, while other goals still have one
- Open the read-only goal with View
- Check that every field in Calculator Settings (goal name, personal, financial) is disabled but still shows its saved value
- Check that in Household/Ministry Expenses the Lump Sum/Line Item toggle and Add Line Item are disabled, rows can't be edited by double-clicking, no delete icons appear on hover, and category totals still display
- Check that the Summary Report still renders the calculated totals
- Open a normal (non-read-only) goal and check everything is still editable

## Checklist:

### DIFF
--- a/src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/Autosave/AutosaveTextField.test.tsx
+++ b/src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/Autosave/AutosaveTextField.test.tsx
@@ -16,12 +16,14 @@ const mutationSpy = jest.fn();
 
 interface TestComponentProps {
   schema?: yup.Schema;
+  readOnly?: boolean;
 }
 
 const TestComponent: React.FC<TestComponentProps> = ({
   schema = defaultSchema,
+  readOnly,
 }) => (
-  <GoalCalculatorTestWrapper onCall={mutationSpy}>
+  <GoalCalculatorTestWrapper onCall={mutationSpy} readOnly={readOnly}>
     <AutosaveTextField
       label="MHA Amount"
       fieldName="mhaAmount"
@@ -145,6 +147,14 @@ describe('AutosaveTextField', () => {
     expect(input).toHaveAccessibleDescription('');
   });
 
+  it('disables the input when the goal is read-only but still shows its value', async () => {
+    const { getByRole } = render(<TestComponent readOnly />);
+
+    const input = getByRole('textbox', { name: 'MHA Amount' });
+    await waitFor(() => expect(input).toHaveValue('1000'));
+    expect(input).toBeDisabled();
+  });
+
   it('shows validation error for invalid type', async () => {
     const { getByRole } = render(<TestComponent />);
 
@@ -195,6 +205,26 @@ describe('AutosaveTextField', () => {
       await waitFor(() =>
         expect(mutationSpy).not.toHaveGraphqlOperation('UpdateGoalCalculation'),
       );
+    });
+
+    it('disables the select when the goal is read-only but still shows its value', async () => {
+      const { getByRole } = render(
+        <GoalCalculatorTestWrapper onCall={mutationSpy} readOnly>
+          <AutosaveTextField
+            label="MHA Amount"
+            fieldName="mhaAmount"
+            schema={defaultSchema}
+            select
+          >
+            <MenuItem value={1000}>1000</MenuItem>
+            <MenuItem value={2000}>2000</MenuItem>
+          </AutosaveTextField>
+        </GoalCalculatorTestWrapper>,
+      );
+
+      const input = getByRole('combobox', { name: 'MHA Amount' });
+      await waitFor(() => expect(input).toHaveTextContent('1000'));
+      expect(input).toHaveAttribute('aria-disabled', 'true');
     });
   });
 });
--- a/src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/Autosave/useGoalAutosave.ts
+++ b/src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/Autosave/useGoalAutosave.ts
@@ -17,13 +17,14 @@ export const useGoalAutoSave = ({
   const saveField = useSaveField();
   const {
     goalCalculationResult: { data },
+    isReadOnly,
   } = useGoalCalculator();
 
   return useAutoSave({
     value: data?.goalCalculation[fieldName],
     saveValue: (value) => saveField({ [fieldName]: value }),
     fieldName,
-    disabled: !data,
+    disabled: !data || isReadOnly,
     ...options,
   });
 };
--- a/src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/InformationCategory/InformationCategoryForm/InformationCategoryFinancialForm.tsx
+++ b/src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/InformationCategory/InformationCategoryForm/InformationCategoryFinancialForm.tsx
@@ -30,6 +30,7 @@ export const InformationCategoryFinancialForm: React.FC<
   const {
     setRightPanelContent,
     goalCalculationResult: { data },
+    isReadOnly,
   } = useGoalCalculator();
 
   const saveField = useSaveField();
@@ -89,6 +90,7 @@ export const InformationCategoryFinancialForm: React.FC<
             fullWidth
             size="small"
             select
+            disabled={!data || isReadOnly}
             label={
               isSpouse
                 ? t('Spouse SECA (Social Security) Status')
--- a/src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/InformationCategory/InformationCategoryForm/InformationCategoryPersonalForm.tsx
+++ b/src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/InformationCategory/InformationCategoryForm/InformationCategoryPersonalForm.tsx
@@ -42,6 +42,7 @@ export const InformationCategoryPersonalForm: React.FC<
   const { t } = useTranslation();
   const {
     goalCalculationResult: { data },
+    isReadOnly,
     setRightPanelContent,
   } = useGoalCalculator();
   const { goalGeographicConstantMap, goalBenefitsPlans } =
@@ -77,6 +78,11 @@ export const InformationCategoryPersonalForm: React.FC<
   }, [goalBenefitsPlans, familySize]);
 
   useEffect(() => {
+    // Read-only goals reject mutations, so don't try to fix an incompatible plan
+    if (isReadOnly) {
+      return;
+    }
+
     // Clear benefits plan if it's not compatible with selected family size
     if (familySize && benefitsPlan) {
       const isPlanValid = planOptions.some(([plan]) => plan === benefitsPlan);
@@ -84,7 +90,7 @@ export const InformationCategoryPersonalForm: React.FC<
         saveField({ benefitsPlan: null });
       }
     }
-  }, [familySize, benefitsPlan, planOptions]);
+  }, [isReadOnly, familySize, benefitsPlan, planOptions]);
 
   return (
     <>
@@ -134,7 +140,7 @@ export const InformationCategoryPersonalForm: React.FC<
               onChange={(_, newValue) =>
                 saveField({ geographicLocation: newValue })
               }
-              disabled={!data}
+              disabled={!data || isReadOnly}
               size="small"
               renderInput={(params) => (
                 <TextField
--- a/src/components/HrTools/GoalCalculator/GoalCalculatorTestWrapper.tsx
+++ b/src/components/HrTools/GoalCalculator/GoalCalculatorTestWrapper.tsx
@@ -34,6 +34,7 @@ export const goalCalculationMock = gqlMock<
     goalCalculation: {
       id: 'goal-calculation-1',
       name: 'Initial Goal Name',
+      readOnly: false,
       firstName: 'John',
       spouseFirstName: 'Jane',
       lastName: 'Doe',
@@ -270,19 +271,38 @@ export const constantsMock = gqlMock<
   },
 }).constant;
 
-interface GoalCalculatorTestWrapperProps {
+interface MockedGoalCalculatorTestWrapperProps {
+  noMocks?: false;
   onCall?: MockLinkCallHandler;
-  noMocks?: boolean;
+  readOnly?: boolean;
   /** Override the mocked goal calculation (defaults to `goalCalculationMock`). */
   goalCalculation?: GoalCalculationQuery['goalCalculation'];
   children?: React.ReactNode;
 }
+
+interface NoMocksGoalCalculatorTestWrapperProps {
+  /**
+   * Skip the `GqlMockedProvider` entirely (the test supplies its own Apollo
+   * provider). `onCall`, `readOnly`, and `goalCalculation` only configure the
+   * mocked provider, so they are disallowed here — they would silently no-op.
+   */
+  noMocks: true;
+  onCall?: never;
+  readOnly?: never;
+  goalCalculation?: never;
+  children?: React.ReactNode;
+}
+
+type GoalCalculatorTestWrapperProps =
+  | MockedGoalCalculatorTestWrapperProps
+  | NoMocksGoalCalculatorTestWrapperProps;
 
 export const GoalCalculatorTestWrapper: React.FC<
   GoalCalculatorTestWrapperProps
 > = ({
   onCall,
   noMocks = false,
+  readOnly = false,
   goalCalculation = goalCalculationMock,
   children,
 }) => {
@@ -307,7 +327,9 @@ export const GoalCalculatorTestWrapper: React.FC<
             }>
               mocks={{
                 GoalCalculation: {
-                  goalCalculation,
+                  goalCalculation: readOnly
+                    ? { ...goalCalculation, readOnly: true }
+                    : goalCalculation,
                 },
                 GoalCalculatorConstants: {
                   constant: constantsMock,
--- a/src/components/HrTools/GoalCalculator/GoalCard/MpdGoalCard.test.tsx
+++ b/src/components/HrTools/GoalCalculator/GoalCard/MpdGoalCard.test.tsx
@@ -14,7 +14,11 @@ import { MpdGoalCard } from './MpdGoalCard';
 
 const mutationSpy = jest.fn();
 
-const TestComponent: React.FC = () => (
+interface TestComponentProps {
+  readOnly?: boolean;
+}
+
+const TestComponent: React.FC<TestComponentProps> = ({ readOnly = false }) => (
   <TestRouter>
     <ThemeProvider theme={theme}>
       <GqlMockedProvider<{
@@ -27,7 +31,9 @@ const TestComponent: React.FC = () => (
           },
         }}
       >
-        <MpdGoalCard goal={{ ...goalCalculationMock, id: 'goal-1' }} />
+        <MpdGoalCard
+          goal={{ ...goalCalculationMock, id: 'goal-1', readOnly }}
+        />
       </GqlMockedProvider>
     </ThemeProvider>
   </TestRouter>
@@ -49,12 +55,13 @@ describe('MpdGoalCard', () => {
   });
 
   it('builds the View link with the goal calculator path', () => {
-    const { getByRole } = render(<TestComponent />);
+    const { getByRole, queryByText } = render(<TestComponent />);
 
     expect(getByRole('link', { name: 'View' })).toHaveAttribute(
       'href',
       '/accountLists/account-list-1/hrTools/goalCalculator/goal-1',
     );
+    expect(queryByText('Read-Only')).not.toBeInTheDocument();
   });
 
   it('calculates and displays the goal total', async () => {
@@ -69,5 +76,17 @@ describe('MpdGoalCard', () => {
     expect(
       getByTestId('goal-amount-value').querySelector('.MuiSkeleton-root'),
     ).toBeInTheDocument();
+  });
+
+  describe('read-only goal', () => {
+    it('hides the Delete button and shows the Read-Only badge', () => {
+      const { getByText, getByRole, queryByRole } = render(
+        <TestComponent readOnly />,
+      );
+
+      expect(queryByRole('button', { name: 'Delete' })).not.toBeInTheDocument();
+      expect(getByText('Read-Only')).toBeInTheDocument();
+      expect(getByRole('link', { name: 'View' })).toBeInTheDocument();
+    });
   });
 });
--- a/src/components/HrTools/GoalCalculator/GoalCard/MpdGoalCard.tsx
+++ b/src/components/HrTools/GoalCalculator/GoalCard/MpdGoalCard.tsx
@@ -1,4 +1,6 @@
 import React, { useMemo } from 'react';
+import { Chip } from '@mui/material';
+import { useTranslation } from 'react-i18next';
 import { GoalCard } from 'src/components/Reports/Shared/GoalCard/GoalCard';
 import { useAccountListId } from 'src/hooks/useAccountListId';
 import { useGoalCalculatorConstants } from 'src/hooks/useGoalCalculatorConstants';
@@ -13,6 +15,7 @@ export interface MpdGoalCardProps {
 }
 
 export const MpdGoalCard: React.FC<MpdGoalCardProps> = ({ goal }) => {
+  const { t } = useTranslation();
   const accountListId = useAccountListId();
   const constants = useGoalCalculatorConstants();
   const [deleteGoalCalculation] = useDeleteGoalCalculationMutation();
@@ -43,7 +46,13 @@ export const MpdGoalCard: React.FC<MpdGoalCardProps> = ({ goal }) => {
       loading={constants.loading}
       updatedAt={goal.updatedAt}
       viewHref={`/accountLists/${accountListId}/hrTools/goalCalculator/${goal.id}`}
-      onDelete={handleDelete}
+      // Read-only goals reject deletion, so don't offer it
+      onDelete={goal.readOnly ? undefined : handleDelete}
+      badge={
+        goal.readOnly ? (
+          <Chip label={t('Read-Only')} size="small" variant="outlined" />
+        ) : undefined
+      }
     />
   );
 };
--- a/src/components/HrTools/GoalCalculator/Shared/GoalCalculation.graphql
+++ b/src/components/HrTools/GoalCalculator/Shared/GoalCalculation.graphql
@@ -1,5 +1,6 @@
 fragment GoalCalculationSettings on GoalCalculation {
   name
+  readOnly
   firstName
   spouseFirstName
   lastName
--- a/src/components/HrTools/GoalCalculator/Shared/GoalCalculatorContext.tsx
+++ b/src/components/HrTools/GoalCalculator/Shared/GoalCalculatorContext.tsx
@@ -48,10 +48,18 @@ export type GoalCalculatorType = {
   goalCalculationResult: ReturnType<typeof useGoalCalculationQuery>;
   goalTotals: GoalTotals;
 
+  /** Whether the goal is read-only and all inputs should be disabled */
+  isReadOnly: boolean;
+
   /** Whether any mutations are currently in progress */
   isMutating: boolean;
-  /** Call with the mutation promise to track the start and end of mutations */
-  trackMutation: <T>(mutation: Promise<T>) => Promise<T>;
+  /**
+   * Call with a thunk that starts the mutation to track the start and end of
+   * mutations. As a defense-in-depth guard, the thunk is not invoked and the
+   * mutation does not fire when the goal is read-only; in that case the
+   * returned promise resolves to `undefined`.
+   */
+  trackMutation: <T>(mutate: () => Promise<T>) => Promise<T | undefined>;
   percentComplete: number;
 
   isMarried: boolean;
@@ -89,6 +97,9 @@ export const GoalCalculatorProvider: React.FC<Props> = ({ children }) => {
       id: goalCalculationId,
     },
   });
+
+  const isReadOnly =
+    goalCalculationResult.data?.goalCalculation?.readOnly ?? false;
 
   const role = goalCalculationResult.data?.goalCalculation?.role ?? null;
   const familySize =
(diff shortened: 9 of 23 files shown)


src/components/HrTools/GoalCalculator/Shared/GoalCalculatorContext.tsx
```diff
@@ -90,6 +93,9 @@ export const GoalCalculatorProvider: React.FC<Props> = ({ children }) => {
     },
   });
 
+  const isReadOnly =
```
wjames111, line 101: [Important] **`HouseholdExpensesHeader` is a missed mutation surface — the one remaining gap in "disable all inputs".** (Consensus of 4 agents; anchored here because the gap is in the rollout of this `isReadOnly` flag.) `HouseholdExpensesHeader.tsx` destructures `trackMutation` from this context but never `isReadOnly`; its five controls (Override paycheck amount, edit pencil, Save, Use paycheck amount, Use categories total) all reach `setDirectInput` → `useUpdateHouseholdDirectInputMutation` with an `optimisticResponse`. On a read-only goal the optimistic `directInput` write cascades through `calculateTotals` (monthly budget → net salary → taxes → gross → admin/attrition → `overallTotal`), so the **entire displayed goal** changes across MpdGoalTable/MpdGoalHeaderCards/PresentingYourGoal until the server rejects and it snaps back with an error toast. Debate verified: no cache/persistence corruption (optimistic layer rolls back atomically; server enforcement holds) — this is a ticket-contract violation and a jarring UX, not data damage.

**Fix:** consume `isReadOnly` in `HouseholdExpensesHeader` and hide the `CardActions` block and the edit `IconButton` (the hide-what-can-never-work pattern this PR already uses for delete affordances).

src/components/HrTools/GoalCalculator/GoalCard/MpdGoalCard.tsx
```diff
@@ -43,7 +46,13 @@ export const MpdGoalCard: React.FC<MpdGoalCardProps> = ({ goal }) => {
       loading={constants.loading}
       updatedAt={goal.updatedAt}
       viewHref={`/accountLists/${accountListId}/hrTools/goalCalculator/${goal.id}`}
-      onDelete={handleDelete}
+      // Read-only goals reject deletion, so don't offer it
+      onDelete={goal.readOnly ? undefined : handleDelete}
+      badge={
+        goal.readOnly ? (
+          <Chip label={t('Read-Only')} size="small" variant="outlined" />
```
wjames111, line 53: [Medium] **No in-calculator read-only explanation.** This chip on the list card is the only place the read-only state is ever explained; inside the calculator every input is simply grey — a user following a direct URL sees pages of dead inputs that read as a loading failure or permissions bug (and screen-reader users get "unavailable" with no reason). Recommend an `<Alert severity="info">{t('This goal is finalized and read-only')}</Alert>` in the calculator step layout when `isReadOnly` — it also carries the announce-reason that disabled inputs cannot.

src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/InformationCategory/InformationCategoryForm/InformationCategoryPersonalForm.tsx
```diff
@@ -77,6 +78,11 @@ export const InformationCategoryPersonalForm: React.FC<
   }, [goalBenefitsPlans, familySize]);
 
   useEffect(() => {
+    // Read-only goals reject mutations, so don't try to fix an incompatible plan
+    if (isReadOnly) {
```
wjames111, line 82: [Medium] **The `isReadOnly` guard branch of this effect is untested.** This is the one place in the PR where a mutation fires with zero user interaction (auto-clearing an incompatible benefits plan), so a regressed guard means an error snackbar the moment a read-only goal opens. The positive path is covered (`InformationCategory.test.tsx` "clears invalid benefits plan"); the new early-return isn't. A `GoalCalculatorTestWrapper readOnly` + `not.toHaveGraphqlOperation('UpdateGoalCalculation')` test locks both branches. Note: `InformationCategory.test.tsx` uses `noMocks` with its own provider — the wrapper's `readOnly` prop silently no-ops with `noMocks` (see the test-wrapper comment).

src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/InformationCategory/InformationCategoryForm/InformationCategoryPersonalForm.tsx
```diff
@@ -77,6 +78,11 @@ export const InformationCategoryPersonalForm: React.FC<
...
+      return;
+    }
```
wjames111, line 84: [Medium] **`isReadOnly` is read inside this effect but missing from its dependency array** (`[familySize, benefitsPlan, planOptions]`, line 93) — inconsistent with this same PR's parallel change in `GoalCalculatorGrid.tsx`, which did add `isReadOnly` to its deps. Flagged independently by 5 agents. Behaviorally harmless today (`readOnly` arrives with the same query payload), but it's a stale-closure landmine and a documented-convention violation. One-line fix: `}, [isReadOnly, familySize, benefitsPlan, planOptions]);`

src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/InformationCategory/InformationCategoryForm/InformationCategoryFinancialForm.tsx
```diff
@@ -89,6 +90,7 @@ export const InformationCategoryFinancialForm: React.FC<
             fullWidth
             size="small"
             select
+            disabled={!data || isReadOnly}
```
wjames111, line 93: [Medium] **The two hand-wired disabled inputs are untested: this SECA select and the geographic `Autocomplete`** (`InformationCategoryPersonalForm.tsx:143`). Every `AutosaveTextField` gets read-only disabling for free via the tested `useGoalAutoSave` path; these two bypass it and each carries a one-off `disabled` expression — exactly the surface that silently regresses in a refactor. Two `toBeDisabled()` assertions in a read-only context cover them.

src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/Autosave/useGoalAutosave.ts
```diff
@@ -17,13 +17,14 @@ export const useGoalAutoSave = ({
   const saveField = useSaveField();
   const {
     goalCalculationResult: { data },
+    isReadOnly,
```
wjames111, line 20: [Medium] **Defense-in-depth: guard the real mutation funnel, not only each input.** (Architecture finding, revised in debate.) Read-only enforcement is distributed across ~10 UI sites; the near-complete funnel is `GoalCalculatorContext.trackMutation` (all in-calculator mutation sites flow through it except the grid's defaults-effect call). Recommended shape: make the context's `trackMutation` accept a thunk (`() => Promise<T>`) and short-circuit when `isReadOnly` — keeping the shared `useTrackMutation` hook generic (5 other HrTools contexts use it) — and also wrap `GoalCalculatorGrid.tsx:247`'s effect mutation, which currently evades both the funnel and the effect's own `isMutating` gate. One context-level test then proves no mutation fires on a read-only goal across every funneled path; this design would have caught the HouseholdExpensesHeader gap by construction.

src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/Autosave/useGoalAutosave.ts
```diff
@@ -17,13 +17,14 @@ export const useGoalAutoSave = ({
...
   } = useGoalCalculator();
 
   return useAutoSave({
     value: data?.goalCalculation[fieldName],
     saveValue: (value) => saveField({ [fieldName]: value }),
     fieldName,
-    disabled: !data,
+    disabled: !data || isReadOnly,
```
wjames111, line 27: [Suggestion] **`disabled` vs `readOnly` semantics (advisory).** Disabled MUI inputs leave the tab order, announce no reason, render values at ~38% opacity, and are un-copyable — on a screen whose entire purpose is *reading* frozen financial values. `InputProps={{ readOnly: true }}` would keep amounts focusable/legible/copyable. Debate verdict: NOT a drop-in — MUI `readOnly` doesn't cleanly cover Select/Autocomplete/DataGrid, and the autosave blur path must be guarded first (see the funnel suggestion) or readOnly fields could fire saves. Treat as a possible follow-up paired with the funnel guard, not a change for this PR.

src/components/HrTools/GoalCalculator/SharedComponents/GoalCalculatorGrid/GoalCalculatorGrid.test.tsx
```diff
@@ -356,4 +356,59 @@ describe('GoalCalculatorGrid', () => {
       await findByText('Only the portion not reimbursed as ministry expense.'),
     ).toBeInTheDocument();
   });
+
+  describe('read-only goal', () => {
+    it('disables the mode toggle and add buttons and hides delete actions', async () => {
```
wjames111: [Suggestion] **The read-only grid tests assert UI state but never assert "no mutation fired".** None of the four tests in this describe pass `onCall`, so they'd stay green if `processRowUpdate` or the defaults effect fired mutations anyway. A mount-time `mutationSpy` + `await waitFor(() => expect(mutationSpy).not.toHaveGraphqlOperation('UpdateSubBudgetCategory'))` here is the highest-value single test addition in the PR — it also covers the untested `isReadOnly` guard on the defaults-repopulation effect for free. Six sibling HrTools test files already use the `not.toHaveGraphqlOperation` pattern.

wjames111 (commented):
## Multi-Agent Code Review — Verdict: APPROVED WITH SUGGESTIONS

6 specialized agents (Architecture, Testing, Standards, Data Integrity, UX, Financial Reporting) + 2 fresh gap reviewers + dependency-impact analysis, followed by a full cross-examination debate and rebuttal round.

**Risk: LOW** (Blast Radius 2, Complexity 1, Sensitivity 1, Recoverability 0 — effort 3, danger 1) · **Blockers: 0** · 1 Important, 6 Medium, 13 Suggestions

### The one thing worth fixing before merge

**`HouseholdExpensesHeader` still fires mutations on read-only goals** — the single remaining ungated mutation surface after this PR's sweep (every other path is guarded: autosave fields, SECA select, autocomplete, grid editing/add/delete, mode toggle, card delete, both auto-firing effects). Its five controls mutate `directInput` with an optimistic response, so a read-only goal's **entire displayed total** flashes to a different value before the server rejection snaps it back. Details in the inline comment on `GoalCalculatorContext.tsx`.

### Verified sound — checked and fine

- **No breaking changes.** `GoalCalculatorContext` has 25 dependents but the change is purely additive; `GoalCard.onDelete` widening verified against both consumers (PdsGoalCard unchanged).
- **Cache normalization is correct.** Both fragments select `readOnly` on the same normalized entity; all four operations that spread them agree; optimistic-response shapes include the new field; `cache.evict` delete path intact.
- **The two auto-firing effect guards are correctly ordered** — `isReadOnly` derives synchronously from the same query payload, so no initial-load race can bypass them.
- **Test infrastructure is sound**: pinning `readOnly: false` in the shared mock is *required* (an unpinned ergonomock value would randomly flip other suites into read-only mode); the wrapper prop composes correctly with the `goalCalculation` override.
- **No financial arithmetic changed** — the `$1,450` total assertion was verified against the mock fixtures; totals still derive from server data for read-only goals.
- **Debate-verified**: the optimistic flash cannot corrupt or persist (Apollo optimistic layers roll back atomically and are excluded from cache persistence), and server-side enforcement (CruGlobal/mpdx_api#3472 model guards) backstops every client gap.

### Product confirmation worth a comment (no change requested)

**"Apply Goal to MPDX" stays enabled for read-only goals.** It writes account preferences (outside the goal's read-only enforcement boundary), and applying a finalized goal's total is arguably the core remaining use of a read-only goal — the panel's conclusion is this is correct behavior, but it deserves a one-line confirmation from the ticket owner. (Side note: the sibling "Save Goal Without Applying" button's label is misleading — it saves nothing on any goal.)

### Pre-existing Issues (Informational — not counted in the verdict)

- **[Pre-existing] 8.0 — Initial-load `defaultTypeChanged` false-positive silently rewrites saved line-item amounts on *editable* Office/married goals.** `getDefaultType(null, false)` seeds the ref with `SingleField` during loading; when the query resolves, `defaultTypeChanged` latches and the grid effect rewrites default-seeded rows to new-staff defaults via unawaited mutations — silently moving the computed support goal. This PR's `isReadOnly` guard incidentally fixes it for read-only goals only. Verified end-to-end by two agents; deserves its own ticket.
- **[Pre-existing] 8.0 — Benefits-plan clearing effect can null a saved plan when constants resolve after goal data** (`planOptions` is empty while `GoalCalculatorConstants` is in flight). Same shape: this PR fixes the read-only symptom; the editable-goal race remains. Deserves its own ticket.

### Debate highlights

- F1 (HouseholdExpensesHeader) settled at **7.0** after a full round trip: Architecture argued down to 6.5 on server-enforcement grounds, then rejoined 7.0 after Financial traced the optimistic write's cascade through `calculateTotals` into every displayed total.
- The `disabled`→`readOnly` a11y suggestion was **conceded down to 4.5** by its own author after verifying MUI `readOnly` is not a drop-in for Select/Autocomplete/DataGrid and the autosave blur path must be guarded first.
- The "centralize in `useSaveField`" suggestion was **revised**: `useSaveField` funnels none of the escaped mutations; the correct choke point is a thunk-accepting `trackMutation` wrapper in `GoalCalculatorContext` (details inline).
- The deployment-ordering note was **raised from 3.0 to 5.0** by three agents: codegen validates against staging only, so merge ordering with the API's production deploy is the actual gate.

Severity < 7 comments you disagree with can be dismissed by replying `/dismiss: <reason>`.

src/components/HrTools/GoalCalculator/SharedComponents/GoalCalculatorGrid/GoalCalculatorGrid.tsx
```diff
@@ -497,8 +501,8 @@ export const GoalCalculatorGrid: React.FC<GoalCalculatorGridProps> = ({
       headerName: '',
       width: 60,
       getActions: (params) => {
-        // Don't show delete action for total row
-        if (params.id === 'total') {
+        // Don't show delete action for total row, or at all when the goal is read-only
+        if (params.id === 'total' || isReadOnly) {
```
wjames111: [Suggestion] When read-only, `getActions` returns `[]` for every row but the actions column still renders as a permanent empty 60px column. Consider excluding the column from `columns` entirely when `isReadOnly`.

src/components/HrTools/GoalCalculator/SharedComponents/GoalCalculatorGrid/GoalCalculatorGrid.tsx
```diff
@@ -276,6 +279,7 @@ export const GoalCalculatorGrid: React.FC<GoalCalculatorGridProps> = ({
     saveValue: updateDirectInput,
     fieldName: 'amount',
     schema: directInputSchema,
+    disabled: isReadOnly,
```
wjames111, line 285: [Suggestion] `useAutoSave`'s `onChange`/`onBlur` closures never check `disabled` — suppression relies entirely on the DOM attribute (a disabled input can't focus/blur). Three stacked guards make this unreachable today (debate consensus: severity split 3.5/2.0), but an `if (disabled) return;` at the top of both handlers is cheap hardening if the readOnly follow-up ever lands.

src/components/HrTools/GoalCalculator/SharedComponents/GoalCalculatorGrid/GoalCalculatorGrid.tsx
```diff
@@ -538,6 +542,7 @@ export const GoalCalculatorGrid: React.FC<GoalCalculatorGridProps> = ({
               size="small"
               onClick={() => handleDirectInputToggle(true)}
               startIcon={<FunctionsIcon />}
+              disabled={isReadOnly}
```
wjames111, line 553: [Suggestion] With both mode buttons disabled, the contained-vs-outlined distinction degrades toward two grey buttons and the active mode isn't conveyed non-visually (no `aria-pressed`). The mode stays legible from structure (lump-sum renders a single Total field; line-item renders the grid), so this is a polish item only.

src/components/Reports/Shared/GoalCard/GoalCard.tsx
```diff
@@ -81,7 +82,7 @@ export const GoalCard: React.FC<GoalCardProps> = ({
 
   const displayName = name ?? t('Unnamed Goal');
 
-  const handleConfirmDelete = async () => onDelete();
+  const handleConfirmDelete = async () => onDelete?.();
```
wjames111: [Suggestion] The `Confirmation` stays mounted with `mutation={handleConfirmDelete}` even when `onDelete` is undefined — if `readOnly` flips true under an already-open dialog (list refetch mid-confirmation), confirming resolves successfully and deletes nothing. Cheap hardening: only render the `Confirmation` when `onDelete` is defined.

src/components/Reports/Shared/GoalCard/GoalCard.tsx
```diff
@@ -158,11 +159,16 @@ export const GoalCard: React.FC<GoalCardProps> = ({
...
-          </Button>
+          {onDelete ? (
+            <Button onClick={() => setDeleting(true)}>
+              <Typography variant="body2" fontWeight="bold" color="error">
+                {t('Delete')}
+              </Typography>
+            </Button>
+          ) : (
+            // Keep the View button on the right when there is no delete button
+            <Box />
```
wjames111: [Suggestion] The empty `<Box />` spacer works (and is commented), but `marginLeft: 'auto'` on the View button — or a conditional `justifyContent: 'flex-end'` — would encode the layout intent without a phantom element.

src/components/HrTools/GoalCalculator/Shared/GoalCalculatorContext.tsx
```diff
@@ -90,6 +93,9 @@ export const GoalCalculatorProvider: React.FC<Props> = ({ children }) => {
...
+    goalCalculationResult.data?.goalCalculation?.readOnly ?? false;
```
wjames111, line 102: [Suggestion] The `?? false` branch is only exercised incidentally — the test wrapper pins `readOnly: false` in the mock, so the field-absent path is never hit. A one-line assertion in `GoalCalculatorContext.test.tsx` exposing `isReadOnly` would pin the derivation.

src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/Autosave/AutosaveTextField.test.tsx
```diff
@@ -145,6 +145,22 @@ describe('AutosaveTextField', () => {
     expect(input).toHaveAccessibleDescription('');
   });
 
+  it('disables the input when the goal is read-only but still shows its value', async () => {
```
wjames111, line 150: [Suggestion] The select-variant `AutosaveTextField` (Role/Family Size/Benefits Plan — the `saveOnChange` path) relies on the same `disabled` prop but is a distinct MUI rendering path; one `combobox` + `toBeDisabled()` assertion in the existing select describe would close it.

src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/Autosave/AutosaveTextField.test.tsx
```diff
@@ -145,6 +145,22 @@ describe('AutosaveTextField', () => {
...
+    const { getByRole } = render(
+      <GoalCalculatorTestWrapper onCall={mutationSpy} readOnly>
```
wjames111: [Suggestion] This test inlines the wrapper + component instead of extending the file's existing `TestComponent` (which already parameterizes `schema`). Adding `readOnly?: boolean` to `TestComponentProps` keeps one copy of the setup — the same move `MpdGoalCard.test.tsx` made in this PR.

src/components/HrTools/GoalCalculator/Shared/GoalCalculation.graphql
```diff
@@ -1,5 +1,6 @@
 fragment GoalCalculationSettings on GoalCalculation {
   name
+  readOnly
```
wjames111, line 3: [Medium] **Deployment gate (raised from 3.0 to 5.0 by three agents in debate).** Both fragments select `readOnly` unconditionally, and codegen validates against **staging** (`codegen.ts` defaults to `api.stage.mpdx.org`), which runs ahead of the API's master branch. If this frontend reaches production before the mpdx_api MPDX-9866 deploy, an unknown-field validation error fails the *entire* `GoalCalculations`, `GoalCalculation`, and settings queries — goals list and calculator hard-down; the `?? false` fallback never executes on a failed document. No code change needed: **verify CruGlobal/mpdx_api#3472 is deployed to production before merging**, and note the ordering in the PR description.

src/components/HrTools/GoalCalculator/GoalCard/MpdGoalCard.test.tsx
```diff
@@ -70,4 +76,16 @@ describe('MpdGoalCard', () => {
       getByTestId('goal-amount-value').querySelector('.MuiSkeleton-root'),
     ).toBeInTheDocument();
   });
+
+  describe('read-only goal', () => {
```
wjames111, line 81: [Suggestion] No negative assertion that the Read-Only chip is absent for a normal goal — if the ternary regressed to always-badge, nothing fails. Add `expect(queryByText('Read-Only')).not.toBeInTheDocument();` to one of the default-mode tests.

src/components/HrTools/GoalCalculator/GoalCalculatorTestWrapper.tsx
```diff
@@ -273,6 +274,7 @@ export const constantsMock = gqlMock<
 interface GoalCalculatorTestWrapperProps {
   onCall?: MockLinkCallHandler;
   noMocks?: boolean;
+  readOnly?: boolean;
```
wjames111, line 277: [Suggestion] `readOnly` silently no-ops when combined with `noMocks` (it only applies inside the mocked branch) — and the two test files that the Important/Medium fixes must touch (`HouseholdExpensesHeader.test.tsx`, `InformationCategory.test.tsx`) both use `noMocks`, so a naive `noMocks readOnly` there would green-light a broken guard. Minimum: a doc comment on the prop; better: a union type making them mutually exclusive.

src/components/HrTools/GoalCalculator/Shared/GoalCalculatorContext.tsx
wjames111: 🤖 Fixed in 274ae377c: Hid the monthly-budget edit controls (edit pencil + CardActions) on read-only goals, with regression tests

src/components/HrTools/GoalCalculator/GoalCard/MpdGoalCard.tsx
wjames111: 🤖 Fixed in fbdba220c: Added a read-only info banner to GoalCalculatorLayout so every calculator step explains the disabled state

src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/InformationCategory/InformationCategoryForm/InformationCategoryPersonalForm.tsx
wjames111: 🤖 Fixed in a8204bb6a: Added a test proving the benefits-plan clearing effect does not fire on read-only goals (verified it fails with the guard removed)

src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/InformationCategory/InformationCategoryForm/InformationCategoryPersonalForm.tsx
wjames111: 🤖 Fixed in b0e383ab3: Added isReadOnly to the benefits-plan effect dependency array

src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/InformationCategory/InformationCategoryForm/InformationCategoryFinancialForm.tsx
wjames111: 🤖 Fixed in cfa366fce: Added tests that the SECA select and geographic autocomplete are disabled on read-only goals

src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/Autosave/useGoalAutosave.ts
wjames111: 🤖 Fixed in c99965b9f: GoalCalculatorContext.trackMutation is now a thunk-accepting funnel that no-ops when read-only; all six call sites migrated and the grid effect mutation is funneled too (fixing its isMutating blind spot)

src/components/HrTools/GoalCalculator/SharedComponents/GoalCalculatorGrid/GoalCalculatorGrid.test.tsx
wjames111: 🤖 Fixed in ee21b7955: Read-only grid tests now assert no budget mutations fire, at mount and after attempted interactions

src/components/HrTools/GoalCalculator/SharedComponents/GoalCalculatorGrid/GoalCalculatorGrid.tsx
wjames111: 🤖 Fixed in 2f94be050: The actions column is omitted entirely for read-only goals instead of rendering empty

src/components/HrTools/GoalCalculator/SharedComponents/GoalCalculatorGrid/GoalCalculatorGrid.tsx
wjames111: 🤖 Fixed in 46c8c1ccc: useAutoSave's onChange/onBlur now short-circuit when disabled, with a test

src/components/HrTools/GoalCalculator/SharedComponents/GoalCalculatorGrid/GoalCalculatorGrid.tsx
wjames111: 🤖 Fixed in 519b19518: Added aria-pressed to the Lump Sum/Line Item buttons so the active mode is conveyed non-visually

src/components/Reports/Shared/GoalCard/GoalCard.tsx
wjames111: 🤖 Fixed in 3d3e9a00c: The delete Confirmation only mounts when onDelete is provided, with tests covering the mid-dialog flip

src/components/Reports/Shared/GoalCard/GoalCard.tsx
wjames111: 🤖 Fixed in 8ebf14cba: Replaced the empty Box spacer with marginLeft auto on the View button

src/components/HrTools/GoalCalculator/Shared/GoalCalculatorContext.tsx
wjames111: 🤖 Fixed in dded1506f: Added context tests pinning the isReadOnly ?? false derivation (loading and read-only cases)

src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/Autosave/AutosaveTextField.test.tsx
wjames111: 🤖 Fixed in 1e04e5cbd: Covered the select-variant read-only disabled state (asserting aria-disabled per MUI Select rendering)

src/components/HrTools/GoalCalculator/CalculatorSettings/Categories/Autosave/AutosaveTextField.test.tsx
wjames111: 🤖 Fixed in 07366f68e: Routed the read-only test through the shared TestComponent

src/components/HrTools/GoalCalculator/GoalCard/MpdGoalCard.test.tsx
wjames111: 🤖 Fixed in 187f985d0: Added the negative Read-Only chip assertion for normal goals

src/components/HrTools/GoalCalculator/GoalCalculatorTestWrapper.tsx
wjames111: 🤖 Fixed in 066a784a9: Wrapper props are now a discriminated union - noMocks with readOnly is a compile error

canac (approved):
I skimmed, but high-level this looks good!

PR #1933, opened 2026-07-21
No-Jira - Add validation for 403(b) contribution percentage to cap at 99%

zweatshirt:
## Description

This stops a runtime error from occurring when a user enters a 403b contribution that is equal to 100. The backend has validation to stop a divide by zero error.

## Testing

- Go to Step 3 of the NSO MPD Questionnaire
- Using the test account, go to the question 'What is Min's 403(b) contribution percentage? *' and try to enter 100 or greater.

## Checklist:

### DIFF
--- a/src/components/HrTools/NsoMpdQuestionnaire/FinancialInformation/VariantQuestions.test.tsx
+++ b/src/components/HrTools/NsoMpdQuestionnaire/FinancialInformation/VariantQuestions.test.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { render, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import { MockLinkCallHandler } from 'graphql-ergonomock/dist/apollo/MockLink';
 import { NewStaffQuestionnaireVariantEnum } from 'src/graphql/types.generated';
 import { NsoMpdQuestionnaireTestWrapper } from '../NsoMpdQuestionnaireTestWrapper';
@@ -112,6 +113,49 @@ describe('VariantQuestions', () => {
           name: 'How many total dependents would you like to cover by your Cru healthcare?',
         }),
       ).not.toBeInTheDocument();
+    });
+
+    it('rejects a 403(b) percentage of 100', async () => {
+      const { findByRole, findByText } = render(
+        <TestComponent
+          onCall={mutationSpy}
+          newStaffQuestionnaire={{
+            variant: NewStaffQuestionnaireVariantEnum.SpouseSeniorStaff,
+          }}
+        />,
+      );
+
+      const field = await findByRole('spinbutton', {
+        name: "What is Jane's 403(b) contribution percentage?",
+      });
+      userEvent.clear(field);
+      userEvent.type(field, '100');
+      userEvent.tab();
+
+      expect(
+        await findByText(
+          '403(b) contribution percentage must be less than 100%',
+        ),
+      ).toBeInTheDocument();
+      expect(mutationSpy).not.toHaveGraphqlOperation(
+        'UpdateNewStaffQuestionnaire',
+      );
+
+      userEvent.clear(field);
+      userEvent.type(field, '99.99');
+      userEvent.tab();
+
+      await waitFor(() =>
+        expect(mutationSpy).toHaveGraphqlOperation(
+          'UpdateNewStaffQuestionnaire',
+          {
+            input: {
+              accountListId: 'account-list-1',
+              attributes: { spouseContribution403bPercentage: 99.99 },
+            },
+          },
+        ),
+      );
     });
 
     it("interpolates the spouse's first name into the questions", async () => {
--- a/src/components/HrTools/NsoMpdQuestionnaire/FinancialInformation/VariantQuestions.tsx
+++ b/src/components/HrTools/NsoMpdQuestionnaire/FinancialInformation/VariantQuestions.tsx
@@ -22,7 +22,9 @@ export const getVariantQuestionsSchema = (t: TFunction) =>
     spouseContribution403bPercentage: percentage(
       t('403(b) contribution percentage'),
       t,
-    ).required(t('Please enter a percentage, or 0 if you contribute none.')),
+    )
+      .lessThan(100, t('403(b) contribution percentage must be less than 100%'))
+      .required(t('Please enter a percentage, or 0 if you contribute none.')),
     spouseMhaAmount: getAmountSchema(t),
     staffConferenceTransfer: getAmountSchema(t),
     accountTransfers: getAmountSchema(t),
@@ -75,7 +77,9 @@ export const VariantQuestions: React.FC = () => {
                 spouseName,
               },
             )}
-            helperText={t('Enter a percentage between 0 and 100.')}
+            helperText={t(
+              'Enter a percentage greater than or equal to 0 and less than 100.',
+            )}
             endAdornment={<PercentageAdornment />}
           />
           <NumberQuestion


src/components/HrTools/NsoMpdQuestionnaire/FinancialInformation/VariantQuestions.tsx
```diff
@@ -22,7 +22,9 @@ export const getVariantQuestionsSchema = (t: TFunction) =>
     spouseContribution403bPercentage: percentage(
       t('403(b) contribution percentage'),
       t,
-    ).required(t('Please enter a percentage, or 0 if you contribute none.')),
+    )
+      .lessThan(100, t('403(b) contribution percentage must be less than 100%'))
```
zweatshirt, line 26: [Suggestion] **Consider a brief comment recording why the ceiling is 100.**

The constraint is not derivable from anything in this repo — `schema.graphql:8515` declares this field as `PositiveFloat` with no upper bound, no directive, and no description of the rule. The knowledge that the API rejects `100` lives only in the PR description and this test's name. A one-liner would preserve it at the call site:

```tsx
// The API rejects a contribution of 100%, so block it client-side rather than surfacing a runtime error.
```

Minor, non-blocking — the test name carries most of this already.

---

**Implementation note (no action needed):** unlike `.max()`, `.lessThan()` is a *different* Yup test from the `.max(100)` already applied inside `percentage()` (`src/lib/yupHelpers.ts:88`), so it composes with it rather than replacing it. `≤100 AND <100` reduces to `<100`, and entering `100` correctly surfaces this message rather than the helper's "must not exceed 100%" — verified by the rendered output in the test run. Behavior is right; just worth knowing it's composition, not override, if anyone edits `percentage()` later.

src/components/HrTools/NsoMpdQuestionnaire/FinancialInformation/VariantQuestions.test.tsx
```diff
@@ -114,6 +115,49 @@ describe('VariantQuestions', () => {
       ).not.toBeInTheDocument();
     });
 
+    it('rejects a 403(b) percentage of 100', async () => {
```
zweatshirt, line 118: [Suggestion] **`0` — the value the helper text points users toward — has no coverage.**

This test pins the upper boundary well (mutation-tested: removing `.lessThan()` makes it fail). The lower boundary is untested anywhere: `percentage` has no tests in `src/lib/yupHelpers.test.ts`, and none of `'Please enter a percentage, or 0'`, `'must be positive'`, or `'must be a number'` appear in any test under `NsoMpdQuestionnaire/`.

`0` is the highest-value addition because `prepareDataForValidation` (`useAutosave.ts:36`) converts `''` to `undefined` — so the boundary between "user entered zero" and "user left it blank" is exactly where a regression would hide, and the required-message explicitly instructs users to enter `0`.

```tsx
it.each([
  ['0', 0],
  ['99.99', 99.99],
])('saves the boundary value %s', async (typed, saved) => {
    <TestComponent
      onCall={mutationSpy}
      newStaffQuestionnaire={{
        variant: NewStaffQuestionnaireVariantEnum.SpouseSeniorStaff,
      }}
    />,
  );

  const field = await findByRole('spinbutton', {
    name: "What is Jane's 403(b) contribution percentage?",
  });
  userEvent.clear(field);
  userEvent.type(field, typed);
  userEvent.tab();

  await waitFor(() =>
    expect(mutationSpy).toHaveGraphqlOperation('UpdateNewStaffQuestionnaire', {
      input: {
        accountListId: 'account-list-1',
        attributes: { spouseContribution403bPercentage: saved },
      },
    }),
  );
});
```

If you add this, the accept-half of the current test becomes redundant and can be dropped, leaving it focused purely on the rejection path.

zweatshirt (commented):
## Multi-Agent Code Review

**Verdict: CLEAN** — no blockers, no findings at or above severity 5.0.

**Risk: 1/10 (LOW)** · 2 files · +49/-2 (7 source lines) · Required reviewer level: ANY

### What this PR does

The NSO Questionnaire's spouse 403(b) contribution percentage accepted `100`, which the API rejects — producing an unhandled runtime error. Client-side Yup validation now enforces `< 100`, so the user gets an inline field error instead.

### Agents run

|---|---|---|---|---|---|
| Architecture | 0 | 0 | 0 | 2 | High |
| Testing & Quality | 0 | 0 | 0 | 3 | High |
| UX & Accessibility | 0 | 0 | 0 | 3 | High |
| Financial Reporting (domain) | 0 | 0 | 1 → resolved | 2 | High |
| **Total** | **0** | **0** | **0** | **12** | |

Plus a dependency-impact pass and an adjudication round on two contested findings.

### Verification performed

- `yarn test VariantQuestions.test.tsx` — 6/6 pass
- `yarn lint:ts` — clean · `eslint` on both changed files — clean
- **Mutation-tested three times.** Removing `.max()` (earlier revision), changing the typed value while leaving the assertion, and removing `.lessThan()` each produced the expected failure. The test genuinely guards the rule rather than passing incidentally.
- Save path confirmed synchronous (`useAutosave.ts:100-105`, no debounce, `saveOnChange` absent), so the negative assertion is not racy. `clearMocks: true` (`jest.config.js:25`) prevents spy bleed.

### Resolved during review

The original revision capped at `.max(99)`, then `.max(99.99)`. The Financial Reporting agent flagged (severity 6.0) that an arbitrary numeric ceiling is stricter than the stated rule and wrongly rejects values the API accepts. An adjudication pass confirmed decimals are genuinely reachable — `NumberQuestion.tsx` renders `type="number"` with no `step`, and saving is blur-triggered rather than a native submit — and that decimal tolerance is deliberate house convention (`percentage()` has no `.integer()`, while a separate `getWholeNumberSchema` does and is used for count fields in this same schema).

The final `.lessThan(100)` states the actual known constraint, matches the helper text exactly, and leaves no boundary gap. **This concern is resolved in the current head.**

These surfaced during review on files related to the change but not modified here, so they cannot be posted as line comments. They are informational — they do not count toward the verdict.

**[Suggestion] `src/components/HrTools/NsGoalCalculator/GoalSettings/goalSettingsSchema.ts:40-44` — sibling 403(b) field still capped at 100**
- Severity: 4.5/10 · Flagged by: Financial Reporting + Dependency Analysis
- `contribution403bPercentage` and `spouseContribution403bPercentage` validate via `optionalPercentage` → `percentage()` → `.max(100)`, mapped at `goalSettingsApiMapping.ts:148-149`.
- Adjudication confirmed these are genuinely **distinct** GraphQL entities — `NewStaffGoalCalculationAttributesInput` (schema.graphql:7582) vs `NewStaffQuestionnaireAttributesInput` (schema.graphql:8515) on separate mutations. However, both are `PositiveFloat`, both carry byte-identical descriptions (`"""Spouse's 403(b) contribution percentage"""`, schema.graphql:7580 and 8513), and neither carries a constraint directive. **The frontend contains no information distinguishing the two rules.**
- Recommended action: one question for the API team — *do `NewStaffQuestionnaire` and `NewStaffGoalCalculation` share a validator on this field, and is the rule `less_than: 100` or `less_than_or_equal_to: 99`?* If shared, GoalSettings has the identical unfixed crash path. Follow-up ticket, not this PR.

**[Suggestion] `src/components/HrTools/NsoMpdQuestionnaire/Shared/NumberQuestion.tsx:53` — native input has no `max`**
- Severity: 4.0/10 · Flagged by: UX, Architecture, Testing, Financial (4 agents independently)
- `htmlInput: { min: 0, inputMode: 'numeric', ...aria }` sets no upper bound, so the number spinner steps past the ceiling the helper text promises. Yup still blocks the save, so this is a papercut rather than a correctness bug.
- Fixing it means threading a new optional prop through the shared `NumberQuestion`, which is broader than this change warrants. Noted so it is a decision rather than an oversight.

**[Informational] `src/components/HrTools/GoalCalculator/.../calculateTotals.ts:72` — genuine divide-by-zero on a different 403(b) entry point**
- Severity: ~8.0/10 if confirmed · Flagged by: Financial Reporting · **Out of scope for this PR**
- `salaryPreIra / (1 - rothContributionPercentage) - salaryPreIra`. The feeding fields (`InformationCategory.tsx:76-91`) validate via `percentage()` at max 100. A blended value of exactly 1.0 yields `Infinity`, which propagates into `grossMonthlySalary`, `grossAnnualSalary`, `overallSubtotal`, and `overallTotal` (`calculateTotals.ts:75-95`) with no guard — rendering `Infinity` as a dollar figure.
- This is the strongest argument that `< 100` is a real domain constraint rather than an API quirk, and that it may belong in a shared helper. Worth its own ticket.

### Notes

- **i18n:** both new strings are static literals inside `t()` — no dynamic keys, no interpolation. `yarn extract` will pick them up. This feature's strings are unextracted repo-wide and land in periodic batch commits, so no locale update was expected here.
- **Standards checklist:** all applicable items PASS; GraphQL/Apollo items N/A (no `.graphql` changes).
- **Blast radius: minimal.** `getVariantQuestionsSchema` has zero consumers outside its own module; `VariantQuestions` has one. Step completion and the submit gate run on value *presence* (`stepCompletion.ts`), not this schema, so a stored `100` on an existing questionnaire cannot silently change completion state or block submission.

PR #1938, opened 2026-07-23
[no-Jira] Force reauthentication for old sessions to fix ministry partner reminders

canac:
## Description

Force reauthentication for any user who logged in before https://github.com/CruGlobal/cru-terraform/pull/11223. Reauthentication will ensure that `mpdx_api` has their primary designation account and update their information in the database so that partner reminders works.

## Testing

*It's hard to test the reauth flow unless you have an old token*

- Login
- Check that you aren't logged out

## Checklist:

### DIFF
--- a/src/components/RouterGuard/RouterGuard.test.tsx
+++ b/src/components/RouterGuard/RouterGuard.test.tsx
@@ -76,5 +76,54 @@ describe('RouterGuard', () => {
 
       await waitFor(() => expect(signIn).toHaveBeenCalledWith('okta'));
     });
+
+    // Remove alongside the reauth-cutoff effect in RouterGuard after 2026-08-10.
+    it('should reauthenticate when the API token predates the reauth cutoff', async () => {
+      (useSession as jest.MockedFn<typeof useSession>).mockReturnValue({
+        // Future session expiry so the token-expiry effect does not fire; the
+        // API token exp is 1000000000 / 2001-09-09, before the 2026-08-10 cutoff.
+        data: {
+          ...session,
+          expires: '2099-01-01T00:00:00Z',
+          user: {
+            ...session.user,
+            apiToken:
+              'eyJhbGciOiJIUzI1NiJ9.eyJleHAiOjEwMDAwMDAwMDB9.h2Qk01iTa6nH_U-OpImeSecS7owIx6YMqeh6yfWO7Xg',
+          },
+        },
+        status: 'authenticated',
+        update: () => Promise.resolve(null),
+      });
+
+      render(<TestComponent pathname="/authRoute">Authed route</TestComponent>);
+
+      await waitFor(() => expect(signIn).toHaveBeenCalledWith('okta'));
+    });
+
+    it('should not reauthenticate when the API token is after the cutoff', async () => {
+      (useSession as jest.MockedFn<typeof useSession>).mockReturnValue({
+        // API token exp is 2000000000 / 2033-05-18, after the 2026-08-10 cutoff.
+        data: {
+          ...session,
+          expires: '2099-01-01T00:00:00Z',
+          user: {
+            ...session.user,
+            apiToken:
+              'eyJhbGciOiJIUzI1NiJ9.eyJleHAiOjIwMDAwMDAwMDB9.XIy4tZ8x7c86GgrOHqwZwR_i28BWxknyjPpHpklBw4U',
+          },
+        },
+        status: 'authenticated',
+        update: () => Promise.resolve(null),
+      });
+
+      const { getByText } = render(
+        <TestComponent pathname="/authRoute">Authed route</TestComponent>,
+      );
+
+      await waitFor(() =>
+        expect(getByText('Authed route')).toBeInTheDocument(),
+      );
+      expect(signIn).not.toHaveBeenCalled();
+    });
   });
 });
--- a/src/components/RouterGuard/RouterGuard.tsx
+++ b/src/components/RouterGuard/RouterGuard.tsx
@@ -27,6 +27,30 @@ export const RouterGuard: React.FC<Props> = ({ children = null }) => {
     }
   }, [session]);
 
+  // cru-terraform#11223 (merged 2026-07-10T18:42:16Z) added Okta profile
+  // claims. Sessions that authenticated before the merge must reauthenticate to
+  // receive them and fix partner reminders. Remove this effect after 2026-08-10,
+  // once every pre-merge token has expired on its own and this check can no longer match.
+  useEffect(() => {
+    if (session.status !== 'authenticated') {
+      return;
+    }
+    const reauthCutoff = DateTime.fromISO('2026-08-10T00:00:00Z').toSeconds();
+    let apiTokenExp: number | null = null;
+    try {
+      const payload = session.data.user.apiToken.split('.')[1];
+      apiTokenExp = JSON.parse(
+        atob(payload.replace(/-/g, '+').replace(/_/g, '/')),
+      ).exp;
+    } catch {
+      // Ignore malformed JWTs
+    }
+
+    if (typeof apiTokenExp === 'number' && apiTokenExp < reauthCutoff) {
+      signIn('okta');
+    }
+  }, [session]);
+
   return session.status === 'authenticated' ? (
     children
   ) : (


src/components/RouterGuard/RouterGuard.tsx
```diff
@@ -27,6 +27,28 @@ export const RouterGuard: React.FC<Props> = ({ children = null }) => {
...
+      return;
+    }
+    const reauthCutoff = DateTime.fromISO('2026-08-10T00:00:00Z').toSeconds();
+    let apiTokenExp: number | null = null;
+    try {
+      const payload = session.data.user.apiToken.split('.')[1];
+      apiTokenExp = JSON.parse(
+        atob(payload.replace(/-/g, '+').replace(/_/g, '/')),
+      ).exp;
+    } catch {}
```
canac: **[Medium] Empty `catch {}` swallows the JWT-decode failure.** Violates the CLAUDE.md Code-Quality standard "No empty `catch {}` blocks that swallow errors silently" — this is the only empty catch in the entire `src`/`pages` tree (ESLint won't flag it; `no-empty` isn't enabled). The swallow is intentional (a malformed token should skip reauth, not crash), so satisfy the standard with a one-line comment rather than changing behavior:

```diff
-    } catch {}
+    } catch {
+      // Malformed / non-JWT apiToken — leave apiTokenExp null and skip reauth.
+    }
```

Flagged by all 4 agents.

src/components/RouterGuard/RouterGuard.tsx
```diff
@@ -27,6 +27,28 @@ export const RouterGuard: React.FC<Props> = ({ children = null }) => {
     }
   }, [session]);
 
+  // cru-terraform#11223 (merged 2026-07-10T18:42:16Z) added Okta profile
+  // claims. Sessions that authenticated before the merge must reauthenticate to
+  // receive them and fix partner reminders. Remove this effect after 2026-08-10,
+  // once every pre-merge token has expired on its own and this check can no longer match.
+  useEffect(() => {
```
canac, line 34: **[Suggestion] Reauth-loop safety rests on an externally-owned, undocumented assumption.** This effect has no loop guard: if `signIn('okta')` mints a fresh token whose `exp` is still `< cutoff`, the branch re-fires on the next render → an app-wide silent Okta redirect loop (this component gates every authenticated route).

A cross-examination round established the risk is **low, not a blocker**: the apiToken is minted once at sign-in and never refreshed, so any live session already has `exp > now`; a loop is therefore only possible for token lifetime `L ∈ (13d, 18d)`, self-clears within days as `now` advances, and vanishes entirely at the 2026-08-10 cutoff. But `L` is set by api.mpdx.org and is not verifiable in-repo (the token carries only `exp`, no `iat`).

Because this gates the whole app, cheap insurance is worth it:
- **Minimum:** add one line to the removal comment — loop-freedom depends on the token lifetime exceeding `cutoff − now`.
- **Recommended (~3 lines):** a one-shot `sessionStorage` guard set before `signIn('okta')`, skipped if already set — degrades any wrong assumption from "app-wide lockout" to "one extra reauth attempt."

src/components/RouterGuard/RouterGuard.test.tsx
```diff
@@ -76,5 +76,54 @@ describe('RouterGuard', () => {
...
+          },
+        },
+        status: 'authenticated',
+        update: () => Promise.resolve(null),
+      });
+
+      render(<TestComponent pathname="/authRoute">Authed route</TestComponent>);
+
+      await waitFor(() => expect(signIn).toHaveBeenCalledWith('okta'));
+    });
```
canac, line 101: **[Medium] No explicit test for the malformed / missing-token branch.** The empty-`catch` path (decode failure → no reauth) is executed implicitly but never asserted. The repo's Testing rules call for error-path + edge-case coverage. Suggested test:

```tsx
it('should not reauthenticate when the API token is malformed', async () => {
  (useSession as jest.MockedFn<typeof useSession>).mockReturnValue({
    data: { ...session, expires: '2099-01-01T00:00:00Z',
      user: { ...session.user, apiToken: 'not-a-jwt' } },
    status: 'authenticated', update: () => Promise.resolve(null),
  });
    <TestComponent pathname="/authRoute">Authed route</TestComponent>,
  );
  expect(await findByText('Authed route')).toBeInTheDocument();
  expect(signIn).not.toHaveBeenCalled();
});
```

src/components/RouterGuard/RouterGuard.test.tsx
```diff
@@ -76,5 +76,54 @@ describe('RouterGuard', () => {
...
+
+    it('should not reauthenticate when the API token is after the cutoff', async () => {
```
canac, line 103: **[Suggestion] This negative test doesn't isolate the branch it claims to.** A decode *failure* would also make this test pass (both produce "signIn not called"), so on its own it doesn't prove the base64url decode actually works — only the positive test above (`exp:1000000000`) does. Fine in combination, but using an `exp` a few seconds either side of the 2026-08-10 cutoff (instead of a value 7 years away) would pin the comparison boundary directly. Minor.

Also: `await waitFor(() => expect(getByText(...)))` at :123-125 → prefer `expect(await findByText(...))` per the repo rule (nit — matches existing local style).

canac (commented):
## 🤖 Multi-Agent Review — APPROVED WITH SUGGESTIONS

**Verdict:** ✅ APPROVED_WITH_SUGGESTIONS · **Risk:** LOW (1/10) · **Blockers:** 0

4 specialized agents (Security, Architecture, Testing, Standards) + 1 focused debate round. Test suite passes 5/5.

### Design is sound
The core change checks out: decoding the JWT without signature verification is fine here (it only reads `exp` to decide whether to redirect — real authz is enforced server-side by `isJwtExpired` in the NextAuth session callback, and the session is signed). No new token exposure (`apiToken` is already client-side today), no open-redirect, Luxon used, named export, colocated test, and a clean scheduled-removal comment referencing `cru-terraform#11223` with a hard 2026-08-10 removal date.

⚠️ Note: file-pattern risk is LOW, but `RouterGuard` gates **every authenticated route** (`pages/_app.page.tsx`), so a regression here has app-wide blast radius.

### Findings

**🟡 Medium (6.0) — empty `catch {}` (`RouterGuard.tsx:45`)** — violates the CLAUDE.md "no empty catch" standard; the only one in `src`/`pages`. The swallow is intentional, so the fix is a one-line comment. See inline.

**🔵 Suggestion (4.5, debated down from 7.0) — reauth-loop assumption (`RouterGuard.tsx:34`)** — the effect has no loop guard; loop-freedom depends on the api.mpdx.org token lifetime exceeding `cutoff − now`. Debate established a loop is only possible for token lifetime `L ∈ (13d, 18d)`, self-clears within days, and vanishes at the cutoff — **not a blocker**. Recommend documenting the assumption (min) or a ~3-line `sessionStorage` one-shot guard (hardening). See inline.

**🔵 Suggestions — testing:** no explicit malformed-token test (the empty-catch branch is executed but never asserted); the negative test doesn't isolate its branch on its own. See inline (skeleton provided).

**🔵 Minor:** two `[session]` effects can both call `signIn('okta')` in one commit (harmless); `waitFor(()=>getByText)` → `findByText` nit at `:123` (matches existing local style); `JSON.parse(...).exp` is implicit `any` narrowed by a `typeof` guard (cosmetic).

### ℹ️ Pre-existing (informational, not a blocker)
`signIn('okta')` is hardcoded (`:48`, and the pre-existing `:26`). In an `API_OAUTH` deployment there is no `okta` provider (`pages/login.page.tsx` derives the id dynamically), so reauth would target a non-configured provider. The new effect faithfully copies the sibling effect's behavior and the whole change is Okta-claim-specific, so it's consistent — flagged for awareness.

src/components/RouterGuard/RouterGuard.tsx
canac: /dismiss This is temporary code and we trust that the tokens always last 30 days.

src/components/RouterGuard/RouterGuard.test.tsx
canac: /dismiss This is temporary code

src/components/RouterGuard/RouterGuard.test.tsx
canac: /dismiss This is temporary code

PR #1940, opened 2026-07-23
[no-Jira] Skip reauth for impersonators

canac:
## Description

- The forced-reauthentication effect added in #1938 (fix MPD partner reminders for pre-cutoff sessions) also caught impersonated sessions and bounced impersonators to Okta.
- Impersonation tokens are short-lived and can't be reauthenticated the same way, so this skips the reauth-cutoff check when `session.data.user.impersonating` is true.
- Adds a `RouterGuard` test covering an impersonated pre-cutoff token that should not trigger reauth.

No Jira ticket. Related to #1938 (the reauth-cutoff effect this guards against); both should be removed after 2026-08-10.

## Testing

- Impersonate another user from the admin/organization impersonation UI.
- Confirm you land on an authenticated route (e.g. the dashboard) and are **not** redirected to the Okta sign-in.
- As a non-impersonated user with a pre-cutoff token, confirm the existing forced-reauth behavior still fires.

## Checklist:

### DIFF
--- a/src/components/RouterGuard/RouterGuard.test.tsx
+++ b/src/components/RouterGuard/RouterGuard.test.tsx
@@ -100,6 +100,34 @@ describe('RouterGuard', () => {
       await waitFor(() => expect(signIn).toHaveBeenCalledWith('okta'));
     });
 
+    // Remove alongside the reauth-cutoff effect in RouterGuard after 2026-08-10.
+    it('should not reauthenticate an impersonated pre-cutoff token', async () => {
+      (useSession as jest.MockedFn<typeof useSession>).mockReturnValue({
+        data: {
+          ...session,
+          expires: '2099-01-01T00:00:00Z',
+          user: {
+            ...session.user,
+            impersonating: true,
+            apiToken:
+              'eyJhbGciOiJIUzI1NiJ9.eyJleHAiOjEwMDAwMDAwMDB9.h2Qk01iTa6nH_U-OpImeSecS7owIx6YMqeh6yfWO7Xg',
+          },
+        },
+        status: 'authenticated',
+        update: () => Promise.resolve(null),
+      });
+
+      const { getByText } = render(
+        <TestComponent pathname="/authRoute">Authed route</TestComponent>,
+      );
+
+      await waitFor(() =>
+        expect(getByText('Authed route')).toBeInTheDocument(),
+      );
+      expect(signIn).not.toHaveBeenCalled();
+    });
+
+    // Remove alongside the reauth-cutoff effect in RouterGuard after 2026-08-10.
     it('should not reauthenticate when the API token is after the cutoff', async () => {
       (useSession as jest.MockedFn<typeof useSession>).mockReturnValue({
         // API token exp is 2000000000 / 2033-05-18, after the 2026-08-10 cutoff.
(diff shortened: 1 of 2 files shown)


src/components/RouterGuard/RouterGuard.tsx
```diff
@@ -35,6 +35,10 @@ export const RouterGuard: React.FC<Props> = ({ children = null }) => {
     if (session.status !== 'authenticated') {
       return;
     }
+    // Ignore short-lived impersonation tokens
```
canac, line 38: [Suggestion] This branch is temporary code but the comment only documents *what* it does, not *that* it's throwaway. It's enclosed by the effect already tagged "Remove this effect after 2026-08-10", so it gets deleted automatically — no orphaning risk. Optional: append "(temporary — goes away with this effect)" for symmetry with the removal-tagged test comments. Non-blocking.

src/components/RouterGuard/RouterGuard.test.tsx
```diff
@@ -100,6 +100,34 @@ describe('RouterGuard', () => {
...
+        },
+        status: 'authenticated',
+        update: () => Promise.resolve(null),
+      });
+
+      const { getByText } = render(
+        <TestComponent pathname="/authRoute">Authed route</TestComponent>,
+      );
+
+      await waitFor(() =>
```
canac, line 124: [Suggestion] Repo testing convention prefers `await findByText('Authed route')` over `await waitFor(() => expect(getByText(...)).toBeInTheDocument())`. However every existing test in this file uses the same `waitFor(getByText)` form, so matching the file is the right call here — changing only the new test would make the file inconsistent. Worth migrating the whole file to `findBy*` during the 2026-08-10 cleanup. Non-blocking, informational.

canac (commented):

**Risk: 1/10 (LOW)** · 5 agents (Security, Architecture, Testing, Standards, UX) + dependency analysis · standard mode

No blockers, no important issues. A 4-line guard that skips the temporary pre-cutoff forced-reauth for impersonation sessions, with a well-designed test (uses a genuinely pre-cutoff token `exp=2001-09-09` and a matching `impersonating: false` control test that *does* reauth — so the guard is proven to be the differentiator). The change **removes** a latent bug (impersonators bounced to Okta) rather than adding risk.

### Security (High confidence — clean)
- `session.data.user.impersonating` is **server-set** (`pages/api/auth/helpers.ts:34` → `!!impersonateJWT`) and persisted in the encrypted NextAuth JWT (keyed by `JWT_SECRET`) — **not client-spoofable**.
- The bypass can't keep a stale session alive: the session-expiry effect (`RouterGuard.tsx:22-28`) still fires for impersonators, and server-side `isJwtExpired(apiToken)` (`[...nextauth].page.ts:266`) rejects expired tokens regardless.

### Suggestions (severity < 5 — informational, no `/dismiss` needed)
See inline comments. All are low-value / consistent-with-file-convention.

---

### 📌 Related observation (pre-existing, not in this diff)
Raised independently by 4 agents: the **first** `useEffect` (permanent, `expires`-based reauth at `RouterGuard.tsx:22-28`) has **no** impersonation guard — an impersonation session whose `session.data.expires` has passed would still hit `signIn('okta')` and drop impersonation. Out of this PR's scope (the new test sidesteps it with `expires: '2099-…'`), and likely fine since impersonation tokens are short-lived — but worth confirming the asymmetry is intentional. This does not affect the verdict.

PR #1949, opened 2026-07-28
MPDX-9811 Split NSO and IBS training costs into separate sections

EJTison:
## Description

The **Edit Training Costs** modal in the MPD Goal Calculator admin captured NSO and IBS as a single combined **"NSO & IBS Cost"** section. That made it impossible to price a cohort for someone who attends NSO but *not* IBS — and it left two backend columns with no way to be filled.

This splits that one section into two:

- **NSO Cost** — Individual (1 in room), Individual (2 in room), Couple, Family *(unchanged, just no longer carrying IBS)*
- **IBS Cost** — Single, Couple *(new)*

The modal goes from **11 to 13 fields**, matching the design and the API.

### Why this lands separately from MPDX-9701

The modal itself shipped in #1894 (MPDX-9701). On the same day it merged, Caleb left [Figma comment #127](https://www.figma.com/design/O3zztrqYUv0lxHhYw3Fhw4/New-Staff-Goal-Calculation?node-id=789-32532#1838857237) on the modal's design node:

> "I think we need some extra fields to be able to calculate how much it costs for someone to attend NSO but not IBS. I think 'NSO & IBS Cost' should be 'NSO Cost' and we need a new IBS row with single and couple inputs."

The design was updated to match, but the modal wasn't. This is that follow-up.

### Field naming

The keys now line up one-for-one with `NewStaffCohort::COST_FIELDS` in `mpdx_api` (added in CruGlobal/mpdx_api#3425), so `ibs_single_cost` and `ibs_couple_cost` have somewhere to come from once the cohort API exists:

| Backend column | Frontend key |
| --- | --- |
| `nso_individual_1_in_room_cost` | `nsoIndividual1InRoom` |
| `nso_individual_2_in_room_cost` | `nsoIndividual2InRoom` |
| `nso_couple_cost` | `nsoCouple` |
| `nso_family_cost` | `nsoFamily` |
| `ibs_single_cost` | `ibsSingle` ✨ |
| `ibs_couple_cost` | `ibsCouple` ✨ |
| `refresh_retreat_{single,couple}_cost` | `refreshRetreat{Single,Couple}` |
| `faith_and_finance_{single,couple}_cost` | `faithAndFinance{Single,Couple}` |
| `cru_conference_{single,couple,family}_cost` | `cruConference{Single,Couple,Family}` |

<details>
<summary>Why the keys don't carry the columns' <code>_cost</code> suffix</summary>

[...]

PR #1950, opened 2026-07-28
Ignore the root tasks/ scratch directory

EJTison:
## Description

Adds `/tasks/` to `.gitignore`.

Development process files — todo lists, plans, specs — get written to a root-level `tasks/` directory while working, and they shouldn't ride along into feature PRs. This came up while building #1949: `tasks/todo.md` showed as untracked and had to be avoided by staging each file by hand, which is easy to get wrong.

### Why `/tasks/` and not `tasks/`

The leading slash anchors the pattern to the repo root. Without it, gitignore matches a directory of that name at **any** depth, which would also cover two unrelated source directories:

- `pages/accountLists/[accountListId]/tasks/`
- `src/lib/tasks/`

Files already committed there would keep tracking either way, but any **new** file added to them would be silently ignored — a confusing failure to run into later. Anchoring avoids that entirely.

## Testing

- `git check-ignore -v tasks/todo.md` → matches `.gitignore:60:/tasks/`
- `git check-ignore` on a hypothetical new file in `pages/accountLists/[accountListId]/tasks/` and in `src/lib/tasks/` → not ignored, both still track
- `git status` reports no change to any currently-tracked file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,10 @@ node_modules
 .lighthouseci
 lighthouse-results.md
 
+# Local scratch/process notes (todo lists, plans, specs) written while working.
+# Root-level only, so the tasks/ source directories under pages/ and src/ still track.
+/tasks/
+
 # AI Code Review - Temporary Files
 # These are working files that get regenerated each review
 .ai-review.json


zweatshirt:
Try running `/quality:agent-review` to get this auto approved.

If you don't have the skill yet, you can get it from here: https://github.com/CruGlobal/claude-code-plugins

We use it to run agent reviews and auto approve small PRs like this.

EJTison:
@zweatshirt cool thanks for the tip!  this one liner seems too simple to be worth burning those tokens but I'll give it a shot despite claude being partially down at the moment haha 😅

zweatshirt:
Another heads up is it often isn't required to use the staging label, we tend to lean toward using preview environments via the `Preview Environment` label. But it's ultimately up to you of course. Any PR merged into main gets auto-merged into staging (assuming there are no conflicts)

EJTison (commented):
## Multi-Agent Review — CLEAN

**Verdict: CLEAN** · Risk: **LOW (0/10)** · Blockers: **0** · Mode: `quick` (3 agents: Testing, Standards, Architecture)

No findings at any severity. Debate round skipped — nothing to cross-examine.

### Risk assessment

`.gitignore` matches no Critical/High/Medium pattern (universal or repo-defined); 4 added lines (<50); single domain; no dependency-manifest, schema, or CI/CD-workflow change. Score 0/10.

### What was actually verified

The one real risk in this change is pattern scope — a bare `tasks/` would match a directory of that name at *any* depth, silently swallowing two live source directories. The diff uses root-anchored `/tasks/`, and all three agents confirmed empirically (not by reading the diff) that this is correct:

| Check | Result |
| --- | --- |
| `tasks/todo.md` ignored | ignored via `.gitignore:60:/tasks/` |
| `pages/accountLists/[accountListId]/tasks/` | not ignored — still tracks |
| `src/lib/tasks/` | not ignored — still tracks |
| **New** file added under either source dir | shows as untracked/addable, not silently hidden |
| Files currently tracked under a root `tasks/` | none — nothing becomes hidden |
| Nested `pages/api/Schema/Tasks/` (case variant) | not ignored |
| Plain file (not dir) named `tasks` at root | not ignored — trailing slash restricts to directories |
| Symlink named `tasks` at root | not ignored |
| Tooling deps on a root `tasks/` path | none in `package.json`, `.github/workflows/**`, `tsconfig`, `jest`, `next`, or `codegen` config |

The inline comment's claim that the nested source directories still track was checked against `git ls-files` and is accurate.

### Convention compliance

Leading-slash root anchoring is already used 5+ times in this file (`/coverage`, `/.next`, `/src/graphql`, `/public/sw*`, the `/tmp/...` block), and leading **and** trailing slash together appears in 3 existing entries (`/tmp/automated_fixes/`, `/tmp/dependency_analysis/`, `/tmp/changed_file_contents/`). The two-line comment style matches existing precedent at the `# AI Code Review - Temporary Files` and `.claude/review-history` blocks. Comment lines measure 79 and 86 chars, within the 120-char guidance.

### Technical debt

Net **positive**. Removes a live footgun: `CLAUDE.md`'s Standard Workflow directs contributors to write plans to `tasks/todo.md`, which was previously untracked-but-not-ignored and therefore capturable by `git add -A`. A per-developer `.git/info/exclude` would only protect one machine; since the convention is repo-wide, the shared `.gitignore` is the correct home.

### Review summary table

| Testing | 0 | 0 | 0 | 0 | High |
| Architecture | 0 | 0 | 0 | 0 | High |
| **Total** | **0** | **0** | **0** | **0** | High |

zweatshirt:
Thanks for running the agent review ;) I am actively developing it so I was curious to see how it would work on a small PR

PR #1952, opened 2026-07-30
MPDX-9887 Fix broken filtering for UTC timestamps

kegrimes:
## Description

Transaction `transactedAt` timestamps are sent in UTC. The frontend was converting them into the user's local timezone, so for users behind UTC a date like July 1st became June 30th, which would exclude the transaction from the July filter and from the report totals, even though the server's balances still included it.

Instead we take the transaction's UTC calendar date, so it reads as July 1st for everyone.

Jira ticket: [MPDX-9887](https://jira.cru.org/browse/MPDX-9887)

## Testing

Impersonate: courtney.campbell@cru.org

Staff expenses test:
- Check that the transaction `-$203.20` exists in the expenses table
- Verify the start and end date calculations add up correctly

MPGA test:
- Click on "Other" breakdown in Income table
- Check transaction `$203.20` says "July 1, 2026" as the date

## Checklist:

### DIFF
--- a/src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesContext/MPGAIncomeExpensesContext.tsx
+++ b/src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesContext/MPGAIncomeExpensesContext.tsx
@@ -6,6 +6,7 @@ import { useFilteredFunds } from 'src/hooks/useFilteredFunds';
 import { useGetLastTwelveMonths } from 'src/hooks/useGetLastTwelveMonths';
 import { useLocale } from 'src/hooks/useLocale';
 import { monthYearFormat } from 'src/lib/intlFormat';
+import { transformTransactionDate } from '../../Shared/Helpers/transformTransactionDate';
 import { Filters } from '../../Shared/SettingsDialog/SettingsDialog';
 import { DateRange } from '../../StaffExpenseReport/Helpers/StaffReportEnum';
 import { FundTypes, Funds } from '../Helper/MPGAReportEnum';
@@ -160,7 +161,9 @@ export const MPGAIncomeExpensesReportProvider: React.FC<Props> = ({
             breakdownByMonth: subcategory.breakdownByMonth.map((month) => ({
               ...month,
               transactions: (month.transactions ?? []).map((transaction) => ({
-                transactedAt: transaction.transactedAt,
+                transactedAt: transformTransactionDate(
+                  transaction.transactedAt,
+                ),
                 description: transaction.description ?? '',
                 amount: transaction.amount,
               })),
--- a/src/components/Reports/Shared/Helpers/transformTransactionDate.ts
+++ b/src/components/Reports/Shared/Helpers/transformTransactionDate.ts
@@ -1,0 +1,13 @@
+import { DateTime } from 'luxon';
+
+/**
+ * The calendar day a transaction happened on, as YYYY-MM-DD.
+ *
+ * The API returns transactedAt as a UTC timestamp. Reading it in
+ * the local zone shifts it into the neighboring day for timezones
+ * behind UTC, which displayed transactions on the wrong day and
+ * dropped ones on the 1st out of the month's filter entirely.
+ *
+ */
+export const transformTransactionDate = (transactedAt: string): string =>
+  DateTime.fromISO(transactedAt, { zone: 'utc' }).toISODate() ?? transactedAt;
--- a/src/components/Reports/StaffExpenseReport/Helpers/filterTransactions.test.tsx
+++ b/src/components/Reports/StaffExpenseReport/Helpers/filterTransactions.test.tsx
@@ -1,4 +1,4 @@
-import { DateTime } from 'luxon';
+import { DateTime, Settings } from 'luxon';
 import {
   Fund,
   StaffExpenseCategoryEnum,
@@ -39,19 +39,19 @@ const mockFund: Fund = {
                 {
                   id: 'transaction-1',
                   amount: 100,
-                  transactedAt: '2025-01-15',
+                  transactedAt: '2025-01-15T00:00:00Z',
                   description: 'January Additional Salary',
                 },
                 {
                   id: 'transaction-2',
                   amount: -100,
-                  transactedAt: '2025-01-20',
+                  transactedAt: '2025-01-20T00:00:00Z',
                   description: 'Star Wars Costume',
                 },
                 {
                   id: 'transaction-3',
                   amount: -1000,
-                  transactedAt: '2025-01-24',
+                  transactedAt: '2025-01-24T00:00:00Z',
                   description: 'January Additional Salary',
                 },
               ],
@@ -79,7 +79,7 @@ const mockFund: Fund = {
                 {
                   id: 'transaction-4',
                   amount: -50,
-                  transactedAt: '2025-01-10',
+                  transactedAt: '2025-01-10T00:00:00Z',
                   description: 'Health Welfare Payment',
                 },
               ],
@@ -301,7 +301,7 @@ describe('filterTransactions', () => {
     expect(grouped.groupedTransactions).toHaveLength(2);
   });
 
-  it('uses earliest transaction date for grouped transaction', () => {
+  it('uses the earliest transaction date, transformed from UTC, for grouped transaction', () => {
     const result = filterTransactions({
       ...baseParams,
       filters: {
@@ -311,7 +311,7 @@ describe('filterTransactions', () => {
     });
 
     const grouped = result[0] as GroupedTransaction;
-    // Earliest transaction is transaction-3 at 2025-01-20
+    // Earliest transaction is transaction-2 at 2025-01-20
     expect(grouped.transactedAt).toBe('2025-01-20');
   });
 
@@ -382,5 +382,86 @@ describe('getAvailableCategories', () => {
     const result = getAvailableCategories([], null, targetTime);
 
     expect(result).toEqual([]);
+  });
+});
+
+describe('UTC timestamps in a behind-UTC timezone', () => {
+  const originalZone = Settings.defaultZone;
+
+  beforeEach(() => {
+    Settings.defaultZone = 'America/Denver';
+  });
+
+  afterEach(() => {
+    Settings.defaultZone = originalZone;
+  });
+
+  const firstOfMonthFund: Fund = {
+    ...mockFund,
+    categories: [
+      {
+        category: StaffExpenseCategoryEnum.Benefits,
+        total: -50,
+        averagePerMonth: -50,
+        subcategories: [
+          {
+            subCategory: StaffExpensesSubCategoryEnum.HealthWelfare,
+            total: -50,
+            averagePerMonth: -50,
+            breakdownByMonth: [
+              {
+                month: '2025-02-01',
+                total: -50,
+                transactions: [
+                  {
+                    id: 'transaction-first',
+                    amount: -50,
+                    transactedAt: '2025-02-01T00:00:00Z',
+                    description: 'First of the month',
+                  },
+                ],
+              },
+            ],
+          },
+        ],
+        breakdownByMonth: [{ month: '2025-02-01', total: -50 }],
+      },
+    ],
+  };
+
+  it('keeps a first-of-month transaction inside the default month filter', () => {
+    const result = filterTransactions({
+      ...baseParams,
+      fund: firstOfMonthFund,
+      targetTime: DateTime.fromISO('2025-02-15'),
+      filters: null,
+    });
+
+    expect(result).toHaveLength(1);
+    expect(result[0].transactedAt).toBe('2025-02-01');
+  });
+
+  it('keeps a first-of-month transaction inside a custom date range', () => {
+    const result = filterTransactions({
+      ...baseParams,
+      fund: firstOfMonthFund,
+      filters: {
+        ...baseFilters,
+        startDate: DateTime.fromISO('2025-02-01'),
+        endDate: DateTime.fromISO('2025-02-28'),
+      },
+    });
+
+    expect(result).toHaveLength(1);
+  });
+
+  it('getAvailableCategories includes a category whose only transaction is on the 1st', () => {
+    expect(
+      getAvailableCategories(
+        [firstOfMonthFund],
+        null,
+        DateTime.fromISO('2025-02-15'),
+      ),
+    ).toEqual([StaffExpenseCategoryEnum.Benefits]);
   });
 });
--- a/src/components/Reports/StaffExpenseReport/Helpers/filterTransactions.tsx
+++ b/src/components/Reports/StaffExpenseReport/Helpers/filterTransactions.tsx
@@ -10,6 +10,7 @@ import {
   getLocalizedCategory,
   getLocalizedSubCategory,
 } from '../../Shared/Helpers/transformStaffExpenseEnums';
+import { transformTransactionDate } from '../../Shared/Helpers/transformTransactionDate';
 import { ReportType } from './StaffReportEnum';
 
 export interface Transaction {
@@ -121,7 +122,9 @@ export const filterTransactions = ({
             breakdown.transactions
               ?.filter((transaction) => {
                 const isInRange = isInDateRange(
-                  DateTime.fromISO(transaction.transactedAt),
+                  DateTime.fromISO(
+                    transformTransactionDate(transaction.transactedAt),
+                  ),
                 );
                 const matchesType =
                   tableType === ReportType.Income
@@ -145,6 +148,9 @@ export const filterTransactions = ({
 
                 return {
                   ...transaction,
+                  transactedAt: transformTransactionDate(
+                    transaction.transactedAt,
+                  ),
                   fundType: fund.fundType,
                   category: category.category,
                   subcategory: subcategory.subCategory,
@@ -220,7 +226,11 @@ export const getAvailableCategories = (
         subcategory.breakdownByMonth?.forEach((breakdown) => {
           const hasTransactionsInRange = breakdown.transactions?.some(
             (transaction) =>
-              isInDateRange(DateTime.fromISO(transaction.transactedAt)),
+              isInDateRange(
+                DateTime.fromISO(
+                  transformTransactionDate(transaction.transactedAt),
+                ),
+              ),
           );
           if (hasTransactionsInRange) {
             categoriesWithTransactions.add(category.category);


src/components/Reports/Shared/Helpers/transformTransactionDate.ts
```diff
@@ -0,0 +1,13 @@
...
+ * The calendar day a transaction happened on, as YYYY-MM-DD.
+ *
+ * The API returns transactedAt as a UTC timestamp. Reading it in
+ * the local zone shifts it into the neighboring day for timezones
+ * behind UTC, which displayed transactions on the wrong day and
+ * dropped ones on the 1st out of the month's filter entirely.
+ *
+ */
+export const transformTransaction = (transactedAt: string): string =>
+  DateTime.fromISO(transactedAt, { zone: 'utc' }).toISODate() ?? transactedAt;
```
kegrimes, line 13: [Important] **New lib function ships without a colocated test.**

`.claude/rules/code-review.md` lists this as mandatory: "Every new component, hook, and lib function has a colocated `*.test.{ts,tsx}`". This helper is now shared across two report features, and the `?? transactedAt` fallback branch is unreachable from `filterTransactions.test.tsx`, so it has no coverage at all.

Flagged by Testing (7.0), Standards (5.5), Architecture (4.0), Financial Reporting (4.0).

Honest context: the only sibling in this directory (`transformStaffExpenseEnums.ts`) also has no test, and 12 of 15 `Reports/**/Helper*` files lack one — so this is a pre-existing repo gap the PR doesn't widen. It's still the cheapest place to pin the behavior that this whole PR exists to produce.

Ready to apply — create `src/components/Reports/Shared/Helpers/transformTransactionDate.test.ts` (verified: 15/15 pass on the fixed code, fails loudly on an identity helper):

```ts
import { Settings } from 'luxon';
import { transformTransaction } from './transformTransactionDate';

describe('transformTransaction', () => {
  const originalZone = Settings.defaultZone;

  afterEach(() => {
    Settings.defaultZone = originalZone;
  });

  it('returns the UTC calendar day for a UTC timestamp', () => {
    expect(transformTransaction('2025-01-15T00:00:00Z')).toBe('2025-01-15');
  });

  describe.each(['America/Denver', 'Pacific/Honolulu', 'America/New_York'])(
    'in %s (behind UTC)',
    (zone) => {
      beforeEach(() => {
        Settings.defaultZone = zone;
      });

      it('does not shift a midnight-UTC timestamp back a day', () => {
        expect(transformTransaction('2025-01-15T00:00:00Z')).toBe('2025-01-15');
      });

      it('keeps a transaction on the 1st of the month on the 1st', () => {
        expect(transformTransaction('2025-02-01T00:00:00Z')).toBe('2025-02-01');
      });

      it('keeps a transaction on the last day of the month in that month', () => {
        expect(transformTransaction('2025-01-31T23:59:59Z')).toBe('2025-01-31');
      });
    },
  );

  it('does not shift forward a day in a zone ahead of UTC', () => {
    Settings.defaultZone = 'Australia/Sydney';
    expect(transformTransaction('2025-01-31T23:00:00Z')).toBe('2025-01-31');
  });

  it('is idempotent for an already date-only string', () => {
    const once = transformTransaction('2025-01-15');
    expect(once).toBe('2025-01-15');
    expect(transformTransaction(once)).toBe('2025-01-15');
  });

  it('falls back to the original value for an unparseable date', () => {
    expect(transformTransaction('not-a-date')).toBe('not-a-date');
  });

  it('falls back to the original value for an empty string', () => {
    expect(transformTransaction('')).toBe('');
  });
});
```

src/components/Reports/StaffExpenseReport/Helpers/filterTransactions.test.tsx
```diff
@@ -39,19 +39,19 @@ const mockFund: Fund = {
                 {
                   id: 'transaction-1',
                   amount: 100,
-                  transactedAt: '2025-01-15',
+                  transactedAt: '2025-01-15T00:00:00Z',
```
kegrimes, line 42: [Medium] **This fixture rewrite is a no-op — the bug being fixed is still unguarded.**

`__tests__/util/globalSetup.ts:2` pins `process.env.TZ = 'UTC'` for the whole suite (wired at `jest.config.js:19`), and it overwrites an explicit developer `TZ` override too — verified empirically in both worker and `--runInBand` modes. Under UTC, `DateTime.fromISO('2025-01-15T00:00:00Z')` and `DateTime.fromISO('2025-01-15')` are the same instant, so `'2025-01-15'` → `'2025-01-15T00:00:00Z'` changes nothing for the date-range predicate.

Measured, not assumed — replacing the helper with `(x) => x` and re-running:

| Scope | Tests | Failures with the fix removed |
|---|---|---|
| this file | 19 | **1** (line 315, output string format only) |
| full report surface | 256 | **1** (the same assertion) |

So `filterTransactions.tsx:126` and `:229` — the two call sites that implement the reported bug — have zero protection. All four fixtures are also mid-month (10th, 15th, 20th, 24th), never on the 1st or last, which is exactly the case the helper's own JSDoc says was broken.

(The rewrite isn't entirely wasted: it's what gives line 315 its teeth. With pre-PR date-only fixtures *and* an identity helper, all 19 still pass.)

Ready to apply — change line 1 to `import { DateTime, Settings } from 'luxon';` and append (verified: 22/22 pass on the fixed code, all 3 fail on an identity helper):

```tsx
describe('UTC timestamp handling in a behind-UTC timezone', () => {
  const originalZone = Settings.defaultZone;

  beforeEach(() => {
    Settings.defaultZone = 'America/Denver';
  });

  afterEach(() => {
    Settings.defaultZone = originalZone;
  });

  // Midnight UTC on the 1st is 17:00 the previous day in Denver, so reading
  // transactedAt in the local zone used to push it out of the month's filter.
  const firstOfMonthFund: Fund = {
    ...mockFund,
    categories: [
      {
        category: StaffExpenseCategoryEnum.Benefits,
        total: -50,
        averagePerMonth: -50,
        subcategories: [
          {
            subCategory: StaffExpensesSubCategoryEnum.HealthWelfare,
            total: -50,
            averagePerMonth: -50,
            breakdownByMonth: [
              {
                month: '2025-02-01',
                total: -50,
                transactions: [
                  {
                    id: 'transaction-first',
                    amount: -50,
                    transactedAt: '2025-02-01T00:00:00Z',
                    description: 'First of the month',
                  },
                ],
              },
            ],
          },
        ],
        breakdownByMonth: [{ month: '2025-02-01', total: -50 }],
      },
    ],
  };

  it('keeps a first-of-month transaction inside the default month filter', () => {
    const result = filterTransactions({
      ...baseParams,
      fund: firstOfMonthFund,
      targetTime: DateTime.fromISO('2025-02-15'),
      filters: null,
    });

    expect(result).toHaveLength(1);
    expect(result[0].transactedAt).toBe('2025-02-01');
  });

  it('keeps a first-of-month transaction inside a custom date range', () => {
    const result = filterTransactions({
      ...baseParams,
      fund: firstOfMonthFund,
      filters: {
        ...baseFilters,
        startDate: DateTime.fromISO('2025-02-01'),
        endDate: DateTime.fromISO('2025-02-28'),
      },
    });

    expect(result).toHaveLength(1);
  });

  it('getAvailableCategories includes a category whose only transaction is on the 1st', () => {
    expect(
      getAvailableCategories(
        [firstOfMonthFund],
        null,
        DateTime.fromISO('2025-02-15'),
      ),
    ).toEqual([StaffExpenseCategoryEnum.Benefits]);
  });
});
```

kegrimes (commented):
## Multi-Agent Code Review — APPROVED WITH SUGGESTIONS

**Mode:** standard · **Agents:** 5 specialized reviewers + dependency analysis + 2 adversarial verifiers
**Risk:** LOW (3/10) · **Blockers:** 0

### The fix is correct — verified by execution, not inference

An adversarial verifier tried hard to break this and could not. It ran a brute-force sweep against a UTC-calendar-day ground truth using the **real** exported `createDateRangeFilter`, `filterTransactions`, and `getAvailableCategories`: 12 timezones × 5 window types × hourly timestamps from `2025-01-28T00Z` to `2025-03-04T00Z` (~50,000 evaluations), including half-hour offsets (`Asia/Kolkata`, `Pacific/Chatham`, `Pacific/Marquesas`) and 30-minute-DST zones (`Australia/Lord_Howe`).

```
BEFORE (pre-PR) mismatches vs UTC-day truth: 889
AFTER  (post-PR) mismatches vs UTC-day truth:   0
```

Zero cases where the post-PR behavior is wrong. A separate DST probe across 9 DST-transition zones × months 2015–2030 (1,920 month boundaries) found no instant where `DateTime.fromISO('YYYY-MM-01')` and `targetTime.startOf('month')` diverge — so the local-midnight round-trip is safe even across spring-forward.

The zone-consistency question was the main thing three agents independently checked: every date-window boundary is constructed in the local zone (`targetTime = DateTime.now().startOf('month')` at `StaffExpenseReport.tsx:85`; presets from `DateTime.now()` at `SettingsDialog.tsx:107-134`; DatePicker values via `DateTime.fromFormat(raw, 'D')`), and after the transform the transaction date is also local midnight. Both sides of every comparison are local wall-clock. Confirmed.

The PR also fixes an end-boundary bug nobody called out: pre-PR, a custom midnight-to-midnight range dropped 23 of 24 hours of the final day **even in UTC**, and 34–51 hours in ahead-of-UTC zones. Post-PR: 0.

### Fix-one-fix-all: complete

Three agents independently swept every `transactedAt` consumer in `src/` and `pages/`. **No unfixed call site remains.** All Staff Expense renderers (`StaffReportTable.tsx:97`, `PrintTables.tsx:57`, `CategoryBreakdownDialog.tsx:38,67`, `downloadReport.tsx:20,35`) consume `filterTransactions` output; the MPGA path (`filterFunds.ts:144` → `BreakdownAccordion.tsx:124`) consumes the context's transformed value. `SettingsDialog.tsx:204` is covered via `getAvailableCategories`.

Adjacent date fields were checked and cleared: `donationDate` and `breakdownByMonth.month` are `ISO8601Date` (date-only, no time component), so `FinancialAccountsReport` / `DonationsReport` do not have this bug. `transactedAt` is the only report-facing `ISO8601DateTime!`. `HrTools/SavingsFundTransfer` reads it raw but is already correct by a different mechanism (`{ setZone: true }` + Luxon `.toFormat`, never `Intl`-based `dateFormat`).

Also verified: `transformTransaction` is idempotent on bare `YYYY-MM-DD`, so the many existing fixtures using date-only strings keep working; and the lexicographic `earliestDate` comparison at `filterTransactions.tsx:82-85` is now safe because all inputs are uniform.

### The one real gap: nothing guards the regression

`__tests__/util/globalSetup.ts:2` pins `process.env.TZ = 'UTC'` for the entire suite (wired at `jest.config.js:19`), and it **overwrites** an explicit developer `TZ` override — verified empirically in both worker and `--runInBand` modes. Under UTC, the rewritten fixtures are the same instant as the originals.

Measured by replacing the helper with an identity function and re-running:

| Scope | Tests | Failures with the fix removed |
|---|---|---|
| `filterTransactions.test.tsx` | 19 | **1** (asserts output string format only) |
| Full report surface (StaffExpense + MPGA + Shared + useFilteredFunds) | 256 | **1** (the same assertion) |

The two call sites that implement the reported bug — the filter predicate at `filterTransactions.tsx:126` and `getAvailableCategories` at `:229` — have **zero** regression protection across 256 tests. A future refactor could reintroduce this exact bug with a fully green suite. See the inline comments for ready-to-apply tests.

Note: `Settings.defaultZone` appears **0 times** in tracked code today, so a zone-pinned test would be the first in this repo. Worth adding to the testing conventions in `.claude/rules/code-review.md`.

### Debate outcomes — findings that did NOT survive verification

Four findings were raised and then refuted or downgraded with concrete counter-evidence. Recording them so they aren't re-raised:

- **"Test-coverage gap is severity 8.5 / blocking"** → downgraded to **6.5**. The production fix is verifiably correct and there is no user-facing defect; blocking a correct bugfix over a missing test would leave the real bug in production.
- **"`SettingsDialog.test.tsx:48,62,76` fixtures using `DateTime.now().toISO()` introduce a month-boundary flake"** → **refuted**. `globalSetup` unconditionally forces `TZ=UTC` and overrides even an explicit developer override, so the shift is unreachable in CI and locally.
- **"End-boundary now depends on exact local-midnight equality; a nonzero `endDate` time would make final-day transactions vanish"** → **refuted**, mechanism is backwards. `transactionDate` is local midnight, so `midnight <= endDate` holds for *any* nonzero end time — it can only ever admit more, never drop. Empirically: `endDate 12:00 → BEFORE=true, AFTER=true`.
- **"Ahead-of-UTC users get worse output if the API ever returns a real time-of-day"** → **not a finding**. The direction is real but it's a *correction*: showing the UTC day makes the row date agree with the server-bucketed column it sits under. Every `transactedAt` fixture in the repo is bare-date or exactly UTC midnight.

### Informational — pre-existing, not introduced here

- **`BalanceCard` reconciliation.** `BalanceCard.tsx:111-141` renders server `startBalance`/`endBalance` alongside client-computed `transfersIn`/`transfersOut` (`StaffExpenseReport.tsx:219-236`) in an arithmetic-looking stack. This PR makes reconciliation **strictly better** for behind-UTC users (the Feb-1-UTC row is now counted in `transfersIn`, matching the server's Feb `endBalance`). A larger pre-existing break is untouched: `getMonthRange.ts:8-19` rounds custom ranges to whole months, so a Feb 5–20 selection gets whole-February server balances against a Feb 5–20 client sum.
- **The corrected comment at `filterTransactions.test.tsx:314`** is right and worth keeping. Verified against the fixtures: `baseParams` uses `ReportType.Expense`, so `transaction-1` (income) is filtered out, leaving `transaction-2` (2025-01-20) as earliest. The original comment named the wrong id while quoting the right date.

> ⚠️ **1 Important finding (severity 7.0-7.9)** — strongly recommended, does not block merge, and cannot be dismissed via `/dismiss`. Address it in this PR or open a follow-up ticket.

### Review Summary Table

|-------|----------|------|-----------|--------|-------------|------------|
| Architecture | 0 | 0 | 0 | 0 | 5 | High |
| Testing | 0 | 0 | 1 | 2 | 2 | High |
| Standards | 0 | 0 | 0 | 1 | 2 | High |
| UX | 0 | 0 | 0 | 0 | 1 | High |
| Financial Reporting | 0 | 0 | 0 | 1 | 4 | High |
| **Total (post-debate, deduped)** | **0** | **0** | **1** | **2** | **4** | |

**Standards checklist:** all mandatory items PASS except the colocated-test item. `yarn eslint`, `yarn tsc --noEmit`, and `yarn prettier --check` all run clean on the four changed files; 36 suites / 256 tests pass under `TZ=UTC`, `America/Denver`, `Asia/Tokyo`, and `Pacific/Kiritimati`. No `any`, no `new Date()`, no debug output, no breaking changes, named exports only.

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesContext/MPGAIncomeExpensesContext.tsx
```diff
@@ -160,7 +161,7 @@ export const MPGAIncomeExpensesReportProvider: React.FC<Props> = ({
             breakdownByMonth: subcategory.breakdownByMonth.map((month) => ({
               ...month,
               transactions: (month.transactions ?? []).map((transaction) => ({
-                transactedAt: transaction.transactedAt,
+                transactedAt: transformTransaction(transaction.transactedAt),
```
kegrimes: [Medium] **This transform has no test coverage.**

`MPGAIncomeExpensesContext.test.tsx` exists (231 lines) but never references `transactedAt` — its assertions cover derived totals, the subtitle string, `MPGATransactions` query variables, and the transaction-years filter.

The transformed value flows onward to `Helper/filterFunds.ts:144` (`date: transaction.transactedAt`) and into `BreakdownAccordion.tsx:124`, where `dateFormat(DateTime.fromISO(transaction.date), locale)` renders it. A wrong value here surfaces as a wrong displayed date with nothing catching it.

Suggested: extend the existing `TestComponent` consumer to render a breakdown row's `date`, mock `MPGATransactions` with `transactedAt: '2025-02-01T00:00:00Z'` under `Settings.defaultZone = 'America/Denver'`, and assert the rendered date is `2025-02-01`.

src/components/Reports/Shared/Helpers/transformTransactionDate.ts
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * The calendar day a transaction happened on, as YYYY-MM-DD.
+ *
+ * The API returns transactedAt as a UTC timestamp. Reading it in
+ * the local zone shifts it into the neighboring day for timezones
+ * behind UTC, which displayed transactions on the wrong day and
+ * dropped ones on the 1st out of the month's filter entirely.
+ *
+ */
+export const transformTransaction = (transactedAt: string): string =>
```
kegrimes: [Suggestion] **`transformTransaction` names a transaction but operates on a date string.**

It takes and returns a date; it never sees a transaction. At the call sites this reads oddly: `transactedAt: transformTransaction(transaction.transactedAt)` (`filterTransactions.tsx:151`). The file is already named `transformTransactionDate.ts`, so the export is the odd one out.

Flagged independently by all 5 agents (3.0–4.0).

To be clear about what is *not* being flagged: "helper filename must match its primary export" is **not** a convention here — `deserializeJsonApi.ts` → `fetchAllData`, `getErrorFromCatch.ts` → `getErrorMessage`, `removeObjectNulls.ts` → `filterNullValues`, `getMonthRange.ts` → `getStaffExpenseMonthRange`, `transformStaffExpenseEnums.ts` → `getLocalizedCategory`. This is purely about the name's semantic accuracy.

```ts
// before
export const transformTransaction = (transactedAt: string): string =>
// after
export const transformTransactionDate = (transactedAt: string): string =>
```

Call sites to update: `filterTransactions.tsx:13,126,151,229` and `MPGAIncomeExpensesContext.tsx:9,164`.

src/components/Reports/StaffExpenseReport/Helpers/filterTransactions.tsx
```diff
@@ -145,6 +148,7 @@ export const filterTransactions = ({
 
                 return {
                   ...transaction,
+                  transactedAt: transformTransaction(transaction.transactedAt),
```
kegrimes: [Suggestion] **Each surviving transaction is transformed twice, and the string round-trips through a `DateTime`.**

`DateTime.fromISO(transformTransaction(x))` at line 126 is parse → format → parse, and this line transforms the same value again. Not a real perf concern at report-sized arrays, but the duplication is avoidable: map first, then filter on the already-normalized field.

```tsx
// after
breakdown.transactions
  ?.map((transaction) => {
    /* localizedCategory / localizedSubCategory / displayCategory unchanged */
    return {
      ...transaction,
      transactedAt: transformTransaction(transaction.transactedAt),
      fundType: fund.fundType,
      category: category.category,
      subcategory: subcategory.subCategory,
      displayCategory,
    };
  })
  .filter(
    (transaction) =>
      isInDateRange(DateTime.fromISO(transaction.transactedAt)) &&
      (tableType === ReportType.Income
        ? transaction.amount > 0
        : transaction.amount < 0),
  ) ?? []
```

src/components/Reports/StaffExpenseReport/Helpers/filterTransactions.tsx
```diff
@@ -121,7 +122,9 @@ export const filterTransactions = ({
             breakdown.transactions
               ?.filter((transaction) => {
                 const isInRange = isInDateRange(
-                  DateTime.fromISO(transaction.transactedAt),
+                  DateTime.fromISO(
```
kegrimes, line 125: [Suggestion] **Consider normalizing once at the ingest boundary, for symmetry with MPGA.**

MPGA normalizes at the query→domain boundary (`MPGAIncomeExpensesContext.tsx:164`), so every downstream consumer is correct by construction. StaffExpenseReport takes the opposite approach: `StaffExpenseReport.tsx:119-125` builds `allFunds` straight from raw `data.reportsStaffExpenses.funds`, and correctness lives in three inline calls here (`:126`, `:151`, `:229`).

**This is future fragility, not a present defect.** An adversarial verifier enumerated every `allFunds` consumer and confirmed all are date-safe today: `defaultFundType` (`:128-130`, reads `fundType`), `selectedFund` (`:142-144`, reads balances and `fundType`), `overallBalance` (`:256-257`, reads `endBalance`), `<BalanceCardList funds={allFunds}>` (`:346` — zero `transactedAt` hits in `BalanceCard/`), and the `transactions` memo (`:162-185`, which goes through `filterTransactions`). An initial 6.0 rating was revised down to 3.0 on that evidence.

The risk is that the invariant is unwritten: `Transaction.transactedAt` is `YYYY-MM-DD` after `filterTransactions` but a UTC timestamp before it, with no type-level distinction. The next component that reads `allFunds` directly silently reintroduces this bug — as a *dropped transaction*, the failure mode `.claude/rules/code-review.md` calls out as silently misleading staff.

Minimum viable version: a comment on `Transaction.transactedAt` (`filterTransactions.tsx:19`) stating it is always normalized, and one on `allFunds` stating it is not.

src/components/Reports/Shared/Helpers/transformTransactionDate.ts
```diff
@@ -0,0 +1,13 @@
+import { DateTime } from 'luxon';
+
+/**
+ * The calendar day a transaction happened on, as YYYY-MM-DD.
+ *
+ * The API returns transactedAt as a UTC timestamp. Reading it in
+ * the local zone shifts it into the neighboring day for timezones
+ * behind UTC, which displayed transactions on the wrong day and
+ * dropped ones on the 1st out of the month's filter entirely.
+ *
```
kegrimes, line 10: [Suggestion] **Two nits on this JSDoc block.**

**1. Document the UTC-day assumption.** `src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.tsx:116-118` solves the same problem for the same field with `DateTime.fromISO(x, { setZone: true })` — two idioms for one API contract. Both are correct today (every observed value is UTC midnight) and this PR's choice is the right one *here*, since normalizing to local midnight is what makes the value comparable against the local-zone filter boundaries. But they diverge the moment the API emits a non-UTC offset: `'2025-01-15T23:00:00-05:00'` → `zone:'utc'` gives `2025-01-16`, `setZone:true` gives `2025-01-15`. One line stating "assumes the API's canonical calendar day is the UTC day" would stop the two from drifting.

**2. Stray blank line** — line 10 is an empty ` *` immediately before the closing `*/`. Confirmed *not* a Prettier violation (`prettier --check` passes; Prettier doesn't reflow block-comment interiors), purely cosmetic.

Also noted and deliberately not flagged: the `?? transactedAt` fallback forwards unparseable input, which downstream either silently drops the transaction (`filterTransactions` → Invalid DateTime → all comparisons false) or throws `RangeError: Invalid time value` from `Intl.DateTimeFormat` (MPGA → `BreakdownAccordion.tsx:124`). Both were reproduced by execution — but both are **byte-for-byte identical to pre-PR behavior**, and the schema types `transactedAt` as non-null `ISO8601DateTime!` with no non-conforming fixture anywhere in the repo. Not a regression.

src/components/Reports/Shared/Helpers/transformTransactionDate.ts
kegrimes: /dismiss coverage is okay without this

src/components/Reports/MPGAIncomeExpensesReport/MPGAIncomeExpensesContext/MPGAIncomeExpensesContext.tsx
kegrimes: /dismiss added tests for transformation

PR #1955, opened 2026-07-31
MPDX-9826 Hide Re-Entry and Return Travel

kegrimes:
## Description

On Staff expenses report, hide Re-Entry and Return Travel fund types unless a user is an international staff or staff stint. These new US staff groups have already been added on the backend.

Extra:
- Reorder fund types on staff expenses AND savings fund transfer: Primary, Conference Savings, Savings, ...
- Add Re-entry and Return Travel funds to savings fund transfer

Jira ticket: [MPDX-9826](https://jira.cru.org/browse/MPDX-9826)

## Testing

International Staff: ✅
- Impersonate: barrett.harkins@cru.org
- Check that all 5 fund types appear
- Verify the order is correct: Primary, Conference, Savings, Return Travel, Re-Entry
- Go to `/hrTools/staffSavingFund/transfers`
- Check that all 5 fund types appear
- Verify the order is correct: Primary, Conference, Savings, Return Travel, Re-Entry

Senior Staff: ✅
- Check that only 3 fund types appear
- Verify the order is correct: Primary, Conference, Savings
- Go to `/hrTools/staffSavingFund/transfers`
- Check that only 3 fund types appear
- Verify the order is correct: Primary, Conference, Savings

## Checklist:

### DIFF
--- a/src/components/HrTools/SavingsFundTransfer/BalanceCard/BalanceCard.test.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/BalanceCard/BalanceCard.test.tsx
@@ -92,12 +92,12 @@ describe('BalanceCard', () => {
       expect(getByTestId('SavingsIcon')).toBeInTheDocument();
     });
 
-    it('should display group icon', () => {
+    it('should display conference savings icon', () => {
       const { getByTestId } = render(
         <Components
           fund={{
             ...defaultFund,
-            fundType: 'Conference Savings',
+            fundType: FundTypeEnum.ConferenceSavings,
           }}
         />,
       );
@@ -116,6 +116,45 @@ describe('BalanceCard', () => {
       );
 
       expect(getByTestId('WalletIcon')).toBeInTheDocument();
+    });
+
+    it('should display return travel icon', () => {
+      const { getByTestId } = render(
+        <Components
+          fund={{
+            ...defaultFund,
+            fundType: FundTypeEnum.ReturnTravel,
+          }}
+        />,
+      );
+
+      expect(getByTestId('FlightIcon')).toBeInTheDocument();
+    });
+
+    it('should display re-entry icon', () => {
+      const { getByTestId } = render(
+        <Components
+          fund={{
+            ...defaultFund,
+            fundType: FundTypeEnum.ReEntry,
+          }}
+        />,
+      );
+
+      expect(getByTestId('HomeIcon')).toBeInTheDocument();
+    });
+
+    it('should display the fallback icon for an unknown fund type', () => {
+      const { getByTestId } = render(
+        <Components
+          fund={{
+            ...defaultFund,
+            fundType: 'Unknown Fund',
+          }}
+        />,
+      );
+
+      expect(getByTestId('GroupsIcon')).toBeInTheDocument();
     });
   });
 
--- a/src/components/HrTools/SavingsFundTransfer/BalanceCard/BalanceCard.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/BalanceCard/BalanceCard.tsx
@@ -1,13 +1,17 @@
 import React from 'react';
-import { Diversity1, Outbox, Savings, Wallet } from '@mui/icons-material';
+import { Outbox } from '@mui/icons-material';
 import { Box, Button, Card, Typography } from '@mui/material';
 import { useTranslation } from 'react-i18next';
 import { SimpleScreenOnly } from 'src/components/Reports/styledComponents';
 import { useLocale } from 'src/hooks/useLocale';
 import { currencyFormat } from 'src/lib/intlFormat';
 import theme from 'src/theme';
+import {
+  getIconColorForFundType,
+  getIconForFundType,
+} from '../../../Reports/StaffExpenseReport/Helpers/fundTypeHelpers';
 import { FundFieldsFragment } from '../ReportsSavingsFund.generated';
-import { FundTypeEnum, TransferModalData } from '../mockData';
+import { TransferModalData } from '../mockData';
 
 export interface BalanceCardProps {
   fund: FundFieldsFragment;
@@ -24,18 +28,8 @@ export const BalanceCard: React.FC<BalanceCardProps> = ({
   const locale = useLocale();
 
   const title = t('{{ name }} Account Balance', { name: fund.fundType });
-  const Icon =
-    fund.fundType === FundTypeEnum.Primary
-      ? Wallet
-      : fund.fundType === FundTypeEnum.Savings
-        ? Savings
-        : Diversity1;
-  const iconBgColor =
-    fund.fundType === FundTypeEnum.Primary
-      ? theme.palette.chartOrange.main
-      : fund.fundType === FundTypeEnum.Savings
-        ? theme.palette.chartBlueDark.main
-        : theme.palette.chartBlue.main;
+  const Icon = getIconForFundType(fund.fundType);
+  const iconBgColor = getIconColorForFundType(fund.fundType, theme);
 
   const handleTransferFrom = () => {
     handleOpenTransferModal({
@@ -60,7 +54,7 @@ export const BalanceCard: React.FC<BalanceCardProps> = ({
         flexDirection: 'column',
       }}
     >
-      <Box display={'flex'} flexDirection="row" alignItems="center" gap={1}>
+      <Box display={'flex'} flexDirection="row" alignItems="flex-start" gap={1}>
         <Box
           sx={{
             backgroundColor: iconBgColor || 'primary.main',
@@ -88,8 +82,6 @@ export const BalanceCard: React.FC<BalanceCardProps> = ({
               fontWeight: 500,
               fontSize: '13pt',
               minHeight: '3em',
-              display: 'flex',
-              alignItems: 'center',
             }}
           >
             {title}
--- a/src/components/HrTools/SavingsFundTransfer/Helper/TransferIcons.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/Helper/TransferIcons.tsx
@@ -1,47 +1,53 @@
-import { Groups, Savings, Wallet } from '@mui/icons-material';
+import {
+  getIconColorForFundType,
+  getIconForFundType,
+} from 'src/components/Reports/StaffExpenseReport/Helpers/fundTypeHelpers';
+import theme from 'src/theme';
+import { FundTypeEnum } from '../mockData';
 
-export const PrimaryAccount = (
-  <Wallet
-    titleAccess="Primary Account"
-    sx={{
-      backgroundColor: '#F08020',
-      color: 'primary.contrastText',
-      borderRadius: 1,
-      p: 0.25,
-      display: 'flex',
-      alignItems: 'center',
-      justifyContent: 'center',
-      mr: 1,
-    }}
-  />
-);
-export const SavingsAccount = (
-  <Savings
-    titleAccess="Savings Account"
-    sx={{
-      backgroundColor: '#007890',
-      color: 'primary.contrastText',
-      borderRadius: 1,
-      p: 0.25,
-      display: 'flex',
-      alignItems: 'center',
-      justifyContent: 'center',
-      mr: 1,
-    }}
-  />
-);
-export const ConferenceSavingsAccount = (
-  <Groups
-    titleAccess="Staff Conference Savings Account"
-    sx={{
-      backgroundColor: '#00C0D8',
-      color: 'primary.contrastText',
-      borderRadius: 1,
-      p: 0.25,
-      display: 'flex',
-      alignItems: 'center',
-      justifyContent: 'center',
-      mr: 1,
-    }}
-  />
-);
+const createFundIcon = (fundType: FundTypeEnum, titleAccess: string) => {
+  const Icon = getIconForFundType(fundType);
+
+  return (
+    <Icon
+      titleAccess={titleAccess}
+      sx={{
+        backgroundColor: getIconColorForFundType(fundType, theme),
+        color: 'primary.contrastText',
+        borderRadius: 1,
+        p: 0.25,
+        display: 'flex',
+        alignItems: 'center',
+        justifyContent: 'center',
+        mr: 1,
+      }}
+    />
+  );
+};
+
+const fundIcons: Record<FundTypeEnum, JSX.Element> = {
+  [FundTypeEnum.Primary]: createFundIcon(
+    FundTypeEnum.Primary,
+    'Primary Account',
+  ),
+  [FundTypeEnum.Savings]: createFundIcon(
+    FundTypeEnum.Savings,
+    'Savings Account',
+  ),
+  [FundTypeEnum.ConferenceSavings]: createFundIcon(
+    FundTypeEnum.ConferenceSavings,
+    'Staff Conference Savings Account',
+  ),
+  [FundTypeEnum.ReturnTravel]: createFundIcon(
+    FundTypeEnum.ReturnTravel,
+    'Return Travel Account',
+  ),
+  [FundTypeEnum.ReEntry]: createFundIcon(
+    FundTypeEnum.ReEntry,
+    'Re-Entry Account',
+  ),
+};
+
+export const getFundIcon = (
+  fundType: string | undefined,
+): JSX.Element | undefined => fundIcons[fundType as FundTypeEnum];
--- a/src/components/HrTools/SavingsFundTransfer/Helper/getFundLabel.ts
+++ b/src/components/HrTools/SavingsFundTransfer/Helper/getFundLabel.ts
@@ -11,5 +11,11 @@ export const getFundLabel = (fund: string | undefined, t: TFunction) => {
   if (fund === FundTypeEnum.ConferenceSavings) {
     return t('Staff Conference Savings Account');
   }
+  if (fund === FundTypeEnum.ReturnTravel) {
+    return t('Return Travel Account');
+  }
+  if (fund === FundTypeEnum.ReEntry) {
+    return t('Re-Entry Account');
+  }
   return t('N/A');
 };
--- a/src/components/HrTools/SavingsFundTransfer/Table/Row/createTableRow.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/Table/Row/createTableRow.tsx
@@ -13,15 +13,15 @@ import {
 } from '@mui/material';
 import { TFunction } from 'react-i18next';
 import { currencyFormat } from 'src/lib/intlFormat';
+import { getFundIcon } from '../../Helper/TransferIcons';
 import {
-  FundTypeEnum,
   ScheduleEnum,
   StatusEnum,
   TableTypeEnum,
   Transfers,
 } from '../../mockData';
 import { RenderCell } from '../TransfersTable';
-import { chipStyle, iconMap } from './createTableRowHelper';
+import { chipStyle } from './createTableRowHelper';
 
 type Options = {
   type: TableTypeEnum;
@@ -57,23 +57,8 @@ export function populateTransferRows(options: Options) {
   }, []);
 
   const transfers: RenderCell = ({ row }) => {
-    const fromIconName =
-      row.transferFrom?.toLowerCase() ===
-      FundTypeEnum.ConferenceSavings.toLowerCase()
-        ? 'conference'
-        : row.transferFrom
-          ? row.transferFrom.toLowerCase()
-          : 'primary';
-    const toIconName =
-      row.transferTo?.toLowerCase() ===
-      FundTypeEnum.ConferenceSavings.toLowerCase()
-        ? 'conference'
-        : row.transferTo
-          ? row.transferTo.toLowerCase()
-          : 'savings';
-
-    const fromIcon = iconMap[fromIconName];
-    const toIcon = iconMap[toIconName];
+    const fromIcon = getFundIcon(row.transferFrom);
+    const toIcon = getFundIcon(row.transferTo);
 
     if (fromIcon && toIcon) {
       return (
--- a/src/components/HrTools/SavingsFundTransfer/TransferModal/Helper/FundInfoDisplay.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/TransferModal/Helper/FundInfoDisplay.tsx
@@ -1,13 +1,8 @@
 import { Box, Typography } from '@mui/material';
 import { useLocale } from 'src/hooks/useLocale';
 import { currencyFormat } from 'src/lib/intlFormat';
-import {
-  ConferenceSavingsAccount,
-  PrimaryAccount,
-  SavingsAccount,
-} from '../../Helper/TransferIcons';
+import { getFundIcon } from '../../Helper/TransferIcons';
 import { FundFieldsFragment } from '../../ReportsSavingsFund.generated';
-import { FundTypeEnum } from '../../mockData';
 
 type Fund = Pick<FundFieldsFragment, 'fundType' | 'endBalance'>;
 
@@ -24,11 +19,7 @@ export const FundInfoDisplay: React.FC<FundInfoDisplayProps> = ({ fund }) => {
 
   return (
     <Box sx={{ display: 'flex', alignItems: 'center' }}>
-      {fund.fundType === FundTypeEnum.Primary
-        ? PrimaryAccount
-        : fund.fundType === FundTypeEnum.Savings
-          ? SavingsAccount
-          : ConferenceSavingsAccount}
+      {getFundIcon(fund.fundType)}
       <Box
         sx={{
           overflow: 'hidden',
--- a/src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.test.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.test.tsx
@@ -9,6 +9,8 @@ import { SnackbarProvider } from 'notistack';
 import TestRouter from '__tests__/util/TestRouter';
 import { GqlMockedProvider } from '__tests__/util/graphqlMocking';
 import { StaffAccountQuery } from 'src/components/Shared/StaffAccount/StaffAccount.generated';
+import { GetUserQuery } from 'src/components/User/GetUser.generated';
+import { UsStaffGroupEnum } from 'src/graphql/types.generated';
 import theme from 'src/theme';
 import { StaffSavingFundContext } from '../../StaffSavingFund/StaffSavingFundContext';
 import {
@@ -207,8 +209,10 @@ jest.mock('notistack', () => ({
 
 const Components = ({
   title = 'Staff Savings Fund Transfers',
+  usStaffGroup = UsStaffGroupEnum.SeniorStaff,
 }: {
   title?: string;
+  usStaffGroup?: UsStaffGroupEnum;
 }) => (
   <SnackbarProvider>
     <ThemeProvider theme={theme}>
@@ -218,8 +222,9 @@ const Components = ({
             StaffAccount: StaffAccountQuery;
             ReportsSavingsFundTransfer: ReportsSavingsFundTransferQuery;
             FundBalances: FundBalancesQuery;
+            GetUser: GetUserQuery;
           }>
-            mocks={mock}
+            mocks={{ ...mock, GetUser: { user: { usStaffGroup } } }}
             onCall={mutationSpy}
           >
             <MockStaffSavingFundProvider>
@@ -268,7 +273,7 @@ describe('TransfersPage', () => {
     expect(getByText('$25,000.00')).toBeInTheDocument();
   });
 
-  it('should sort fund cards in ascending order by id', async () => {
+  it('should sort fund cards into fund type display order', async () => {
     const { findAllByText } = render(<Components />);
 
     const fundCards = await findAllByText(/Account Balance/i);
@@ -294,6 +299,44 @@ describe('TransfersPage', () => {
     expect(tables.length).toBe(2);
 
     expect(within(tables[0]).getAllByRole('columnheader')).toHaveLength(8);
+  });
+
+  it.each([
+    UsStaffGroupEnum.SeniorInternationalStaff,
+    UsStaffGroupEnum.NewInternationalStaff,
+    UsStaffGroupEnum.SeniorStaffStint,
+    UsStaffGroupEnum.NewStaffStint,
+  ])('requests re-entry and return travel for %s', async (usStaffGroup) => {
+    render(<Components usStaffGroup={usStaffGroup} />);
+
+    await waitFor(() =>
+      expect(mutationSpy).toHaveGraphqlOperation('FundBalances', {
+        fundTypes: [
+          'Primary',
+          'Savings',
+          'Staff Conference Savings',
+          'Return Travel',
+          'Re-Entry',
+        ],
+      }),
+    );
+  });
+
+  it('never requests re-entry and return travel for everyone else', async () => {
+    const { findByText } = render(<Components />);
+
+    await waitFor(() => expect(mutationSpy).toHaveGraphqlOperation('GetUser'));
+    expect(await findByText('Primary Account Balance')).toBeInTheDocument();
+
+    expect(mutationSpy).not.toHaveGraphqlOperation('FundBalances', {
+      fundTypes: [
+        'Primary',
+        'Savings',
+        'Staff Conference Savings',
+        'Return Travel',
+        'Re-Entry',
+      ],
+    });
   });
 
   it('should display stopped recurring transfers in history with a stopped status and no actions', async () => {
--- a/src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.tsx
@@ -24,7 +24,10 @@ import {
   MultiPageHeader,
 } from 'src/components/Shared/MultiPageLayout/MultiPageHeader';
 import { useStaffAccountQuery } from 'src/components/Shared/StaffAccount/StaffAccount.generated';
+import { useGetUserQuery } from 'src/components/User/GetUser.generated';
+import { UsStaffGroupEnum } from 'src/graphql/types.generated';
 import theme from 'src/theme';
+import { compareFundTypes } from '../../../Reports/StaffExpenseReport/Helpers/fundTypeHelpers';
 import {
   StaffSavingFundContext,
   StaffSavingFundType,
@@ -41,7 +44,6 @@ import { PrintTable } from '../Table/PrintTable';
 import { TransfersTable } from '../Table/TransfersTable';
 import { DynamicTransferModal } from '../TransferModal/DynamicTransferModal';
 import {
-  FundTypeEnum,
   ScheduleEnum,
   StatusEnum,
   TableTypeEnum,
@@ -70,7 +72,7 @@ interface TransfersPageProps {
 
 const StyledCardsBox = styled(Box)({
   flex: 1,
-  minWidth: 250,
+  minWidth: 215,
   display: 'flex',
   gap: theme.spacing(4),
 });
@@ -82,6 +84,24 @@ export const TransfersPage: React.FC<TransfersPageProps> = ({ title }) => {
     StaffSavingFundContext,
   ) as StaffSavingFundType;
 
+  const { data: userData, loading: userLoading } = useGetUserQuery();
+  const usStaffGroup = userData?.user.usStaffGroup;
+  const hideFunds =
+    usStaffGroup !== UsStaffGroupEnum.SeniorInternationalStaff &&
+    usStaffGroup !== UsStaffGroupEnum.NewInternationalStaff &&
+    usStaffGroup !== UsStaffGroupEnum.SeniorStaffStint &&
+    usStaffGroup !== UsStaffGroupEnum.NewStaffStint;
+
+  const fundTypes = hideFunds
+    ? ['Primary', 'Savings', 'Staff Conference Savings']
+    : [
+        'Primary',
+        'Savings',
+        'Staff Conference Savings',
+        'Return Travel',
+        'Re-Entry',
+      ];
+
   const { data: staffAccountData, error: staffAccountError } =
     useStaffAccountQuery();
 
@@ -89,18 +109,15 @@ export const TransfersPage: React.FC<TransfersPageProps> = ({ title }) => {
     useReportsSavingsFundTransferQuery();
   const { data: fundsData, error: fundsError } = useFundBalancesQuery({
     variables: {
-      fundTypes: [
-        FundTypeEnum.Primary,
-        FundTypeEnum.Savings,
-        FundTypeEnum.ConferenceSavings,
-      ],
+      fundTypes,
     },
+    skip: userLoading,
   });
 
   const funds = useMemo(
     () =>
       (fundsData?.reportsStaffExpenses?.funds ?? []).toSorted((a, b) =>
-        a.id.localeCompare(b.id),
+        compareFundTypes(a.fundType, b.fundType),
       ),
     [fundsData],
   );
@@ -312,11 +329,12 @@ export const TransfersPage: React.FC<TransfersPageProps> = ({ title }) => {
                 </StyledCardsBox>
               ) : (
                 funds.map((fund) => (
-                  <BalanceCard
-                    fund={fund}
-                    key={fund.id}
-                    handleOpenTransferModal={handleOpenTransferModal}
-                  />
+                  <StyledCardsBox key={fund.id}>
+                    <BalanceCard
+                      fund={fund}
+                      handleOpenTransferModal={handleOpenTransferModal}
+                    />
+                  </StyledCardsBox>
                 ))
               )}
             </Box>
--- a/src/components/Reports/StaffExpenseReport/Helpers/fundTypeHelpers.ts
+++ b/src/components/Reports/StaffExpenseReport/Helpers/fundTypeHelpers.ts
@@ -37,6 +37,18 @@ const fundTypeConfig: Record<string, FundTypeConfig> = {
   },
 };
 
+const fundTypeOrder: Record<string, number> = {
+  Primary: 0,
+  'Staff Conference Savings': 1,
+  Savings: 2,
+  'Return Travel': 3,
+  'Re-Entry': 4,
+};
+
+export const compareFundTypes = (a: string, b: string): number => {
+  return fundTypeOrder[a] - fundTypeOrder[b] || a.localeCompare(b);
+};
+
 const defaultFundTypeConfig: FundTypeConfig = {
   icon: Groups,
   getColor: (theme) => theme.palette.chartBlue.main,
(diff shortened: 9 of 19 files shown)


src/components/Reports/StaffExpenseReport/StaffExpenseReport.tsx
```diff
@@ -84,22 +86,34 @@ export const StaffExpenseReport: React.FC<StaffExpenseReportProps> = ({
...
     return Boolean(
       filters &&
         (filters.startDate || filters.endDate || filters.selectedDateRange),
     );
   }, [filters]);
 
-  const { data, loading } = useReportsStaffExpensesQuery({
-    variables: {
-      fundTypes: [
+  const fundTypes = hideFunds
```
kegrimes, line 104: **[Critical] 🔴 BLOCKER — the headline `overallBalance` silently loses these funds' dollars**

Flagged by Financial Reporting, **confirmed empirically by Testing**, supported by UX + Standards.

Narrowing `fundTypes` here narrows `allFunds` (`:133-139`), which is reduced into `overallBalance` at `:269-272` and rendered by `AccountInfoBox.tsx:57-63` as a **bare, unlabeled currency figure** beside the staff name — no "3 of 5 funds" qualifier. There is no server-provided grand total; this client `reduce` is the only one.

**Measured** (same account, same balances, three probes):

| Funds returned | Requested `fundTypes` | Rendered total |
|---|---|---|
| all 5 (P 2000, S 1000, SCS 500, RT 750, RE 1250) | 3 types | **$5,500.00** |
| only 3 (production reality) | 3 types | **$3,500.00** |

**$2,000 silently removed.** This is a strict regression — `git log -S"Return Travel"` shows commit `82699382e` deliberately expanded `fundTypes` from 3 to 5 **for every user**; this branch removes them for anyone outside the four-group allowlist.

**Architectural note:** grepping every `usStaffGroup ===`/`!==` comparison outside tests turns up exactly eight — all eight added by this PR. This is the **first place in the codebase where a user's staff group determines GraphQL query variables**. Every prior visibility decision was made at the display boundary over a full data set.

**Fix — fetch everything, filter at the display boundary, total over the truth:**

```tsx
const ALL_FUND_TYPES = [
  'Primary', 'Savings', 'Staff Conference Savings', 'Return Travel', 'Re-Entry',
];

const { data, loading } = useReportsStaffExpensesQuery({
  variables: { fundTypes: ALL_FUND_TYPES, ...getStaffExpenseMonthRange(filters, time) },
});

// keep overallBalance reducing over ALL funds;
// filter only the CARD LIST by hideFunds
const visibleFunds = useMemo(
  () => allFunds.filter((f) =>
    !hideFunds || !['Return Travel', 'Re-Entry'].includes(f.fundType) || (f.endBalance ?? 0) !== 0),
  [allFunds, hideFunds],
);
```

This also structurally dissolves the loading-race concern, since `fundTypes` stops depending on `usStaffGroup` entirely.

**⚠️ Open question:** can a plain `SeniorStaff` / `NewStaff` user hold a *nonzero* Return Travel or Re-Entry balance? Nothing in this repo answers it. If yes → real money is hidden. If always $0 → cosmetic, and this stops being a blocker. Please confirm with the API team.

src/components/Reports/StaffExpenseReport/StaffExpenseReport.test.tsx
```diff
@@ -280,6 +290,52 @@ describe('StaffExpenseReport', () => {
...
+          'Savings',
+          'Staff Conference Savings',
+          'Return Travel',
+          'Re-Entry',
+        ],
+      }),
+    );
+  });
+
+  it('does not request re-entry and return travel funds for everyone else', async () => {
```
kegrimes: **[High] 🟠 BLOCKER — this test passes for every user group, including the one it excludes**

Proven by mutation: rendering `<TestComponent usStaffGroup={UsStaffGroupEnum.SeniorInternationalStaff} />` — the exact user this assertion is meant to exclude — leaves the test **green**.

**Root cause:** `useGetUserQuery()` resolves asynchronously, so on first render `usStaffGroup === undefined` → `hideFunds === true` → the query **always** fires once with the 3-type list, then refires with 5. `toHaveGraphqlOperation` matches with `operations.some(...)` (`__tests__/extensions/toHaveGraphqlOperation.ts:21`), so that first call satisfies this assertion for everyone.

A second mutation confirms the coverage gap is wider: **deleting `NewInternationalStaff` and `NewStaffStint`** from the predicate at `StaffExpenseReport.tsx:91-95` leaves all 12 tests passing. Two of the four groups the feature exists for are entirely unconstrained.

**Fix — assert absence directly.** Order-independent, count-independent, and it survives a future `skip: userLoading` fix:

```ts
it('never requests re-entry and return travel funds for everyone else', async () => {
  render(<TestComponent />);

  await waitFor(() =>
    expect(mutationSpy).toHaveGraphqlOperation('ReportsStaffExpenses', {
      fundTypes: ['Primary', 'Savings', 'Staff Conference Savings'],
    }),
  );
  expect(mutationSpy).not.toHaveGraphqlOperation('ReportsStaffExpenses', {
    fundTypes: [
      'Primary', 'Savings', 'Staff Conference Savings', 'Return Travel', 'Re-Entry',
    ],
  });
});
```

Plus `it.each` over all four gated groups for the positive case.

**Do not** reach for `calls.length === 1` instead — it discriminates today, but silently re-vacuates the moment the loading race is fixed, with no test failure to warn anyone.

src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.test.tsx
```diff
@@ -296,6 +301,50 @@ describe('TransfersPage', () => {
...
+          'Savings',
+          'Staff Conference Savings',
+          'Return Travel',
+          'Re-Entry',
+        ],
+      }),
+    );
+  });
+
+  it('does not request re-entry and return travel for everyone else', async () => {
```
kegrimes: **[High] 🟠 BLOCKER — same vacuous assertion as `StaffExpenseReport.test.tsx:329`**

Rendering `<Components usStaffGroup={UsStaffGroupEnum.SeniorInternationalStaff} />` leaves this test green. The async `useGetUserQuery` makes the first render fire a 3-type `FundBalances` call for **every** user group, and `toHaveGraphqlOperation` matches with `.some(...)`.

Also only 2 of the 4 gated groups (`SeniorInternationalStaff`, `SeniorStaffStint`) are covered here — `NewInternationalStaff` and `NewStaffStint` are unconstrained.

**Fix:** same shape as the StaffExpenseReport comment — add a `.not.toHaveGraphqlOperation(...)` assertion for the 5-type variant, and convert the positive tests to `it.each` over all four groups.

src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.tsx
```diff
@@ -82,25 +84,39 @@ export const TransfersPage: React.FC<TransfersPageProps> = ({ title }) => {
...
 
+  const { data: userData } = useGetUserQuery();
+  const usStaffGroup = userData?.user.usStaffGroup;
+  const hideFunds =
+    usStaffGroup !== UsStaffGroupEnum.SeniorInternationalStaff &&
+    usStaffGroup !== UsStaffGroupEnum.NewInternationalStaff &&
+    usStaffGroup !== UsStaffGroupEnum.SeniorStaffStint &&
+    usStaffGroup !== UsStaffGroupEnum.NewStaffStint;
+
+  const fundTypes = hideFunds
```
kegrimes, line 95: **[Medium] This predicate + both arrays are duplicated byte-for-byte in `StaffExpenseReport.tsx:91-112`**

And a *third* encoding of the same four groups is `BASE_GROUPS` in `useIneligibleByGroup.ts:5-10` — added by this same PR. Adding a fifth international/stint group, or a sixth fund type, requires three synchronized edits; the duplicated tests mean a drift would produce two green suites asserting different things.

Extracting one hook fixes the duplication, the loading race, and (with the card-filter change) the `overallBalance` blocker in a single move:

```ts
// src/hooks/useFundTypes.ts
const EXTRA_FUND_GROUPS = new Set([
  UsStaffGroupEnum.SeniorInternationalStaff,
  UsStaffGroupEnum.NewInternationalStaff,
  UsStaffGroupEnum.SeniorStaffStint,
  UsStaffGroupEnum.NewStaffStint,
]);

export const useFundTypes = () => {
  const { data, loading } = useGetUserQuery();
  const group = data?.user.usStaffGroup;
  return { loading, showRestrictedFunds: !!group && EXTRA_FUND_GROUPS.has(group) };
};
```

Both call sites then reduce to one line.

src/hooks/useIneligibleByGroup.ts
```diff
@@ -2,6 +2,20 @@ import { useMemo } from 'react';
 import { useGetUserQuery } from 'src/components/User/GetUser.generated';
 import { UsStaffGroupEnum } from 'src/graphql/types.generated';
 
+const BASE_GROUPS: Partial<Record<UsStaffGroupEnum, UsStaffGroupEnum>> = {
```
kegrimes, line 7: **[Medium] Undocumented access-widening riding along on a fund-visibility ticket**

Mapping these four groups to base groups changes **every** eligibility flag, not just fund visibility. Blast radius: 2 direct consumers (`UserTypeAccess.tsx:47`, `useHrToolsNavItems.ts:22`), both consuming all six flags, fanning out to ~18 guarded pages. Newly reachable for these groups: Salary Calculation Form, MPD Goal Calculator + Admin Table, MHA Calculation Tool, Additional Salary Request, NS Goal Calculator, NSO MPD Questionnaire.

**Debate revised this down from 7.8 → 5.5.** The evidence says it's a regression fix, not scope creep: the predecessor hook `useUsStaffGroups` gated on HCM booleans (`asrEligibility`, `salaryRequestEligible`, …); main-branch refactor `5ec62c491` replaced that with enum allowlists covering only 4 of 10 `UsStaffGroupEnum` members, dropping international/stint **by omission**. This PR restores exactly 4 of the 6 dropped groups (leaving out `Intern` and `PartTimeFieldStaff` — semantically coherent) with four explicit per-group tests. `HrTools/CLAUDE.md` also states these guards are "UX gating, **not** the security boundary."

**What still needs doing:**

1. **MHA and ASR have independent server-side eligibility flags** (`hcmData.mhaEit.mhaEligibility`, `user.asrEit.asrEligibility`) — those paths are protected. **SalaryCalculator has none**; this hook is its entire client gate. Please confirm with HR that `SeniorStaffStint` / `SeniorInternationalStaff` are intended to submit salary calculations.
2. Add a ticket reference or a comment explaining the base-group mapping — nothing in the code or commit history states the intent.
3. Update the eligibility table in `src/components/HrTools/CLAUDE.md` (rows for SalaryCalculator, GoalCalculator, AdditionalSalaryRequest, MinisterHousingAllowance, NsGoalCalculator).

src/components/HrTools/SavingsFundTransfer/Helper/TransferIcons.tsx
```diff
@@ -1,47 +1,53 @@
...
+    FundTypeEnum.Savings,
+    'Savings Account',
+  ),
+  [FundTypeEnum.ConferenceSavings]: createFundIcon(
+    FundTypeEnum.ConferenceSavings,
+    'Staff Conference Savings Account',
+  ),
+  [FundTypeEnum.ReturnTravel]: createFundIcon(
+    FundTypeEnum.ReturnTravel,
+    'Return Travel Account',
```
kegrimes, line 43: **[Medium] Hard-coded English `titleAccess` — this is the screen-reader accessible name**

MUI's `titleAccess` renders an SVG `<title>` and sets `role="img"`, so these five strings are what a screen reader announces.

Independently measured across the repo: **63 of 67** `titleAccess` sites localize via `t()`. The single raw-literal precedent (`createTableRow.tsx:67`, `titleAccess="Arrow"`) is itself a bug — a decorative separator should be `aria-hidden`, not announced.

The sharper problem is **divergence within this PR**: `getFundLabel.ts:14-19` localizes these exact strings — including `t('Return Travel Account')` and `t('Re-Entry Account')`, added in this same commit. `TransfersTable.tsx:200-203` builds the DataGrid cell's value from the localized `getFundLabel`, while `createTableRow.tsx:60-61` renders that same cell's AT content from `getFundIcon` in English only. Same cell, two languages.

Root cause: `fundIcons` is a module-scope `Record`, so `t` is unreachable. The fix is not separable per-string — make the lookup lazy and reuse `getFundLabel` as the single source of truth:

```tsx
import { TFunction } from 'i18next';
import { getFundLabel } from './getFundLabel';

export const getFundIcon = (
  fundType: string | undefined,
  t: TFunction,
): JSX.Element | undefined =>
  fundType && Object.values<string>(FundTypeEnum).includes(fundType)
    ? createFundIcon(fundType as FundTypeEnum, getFundLabel(fundType, t))
    : undefined;
```

`createTableRow.tsx` already has `t` in scope via `Options`; `FundInfoDisplay.tsx` needs `const { t } = useTranslation();`.

src/components/HrTools/SavingsFundTransfer/Helper/TransferIcons.tsx
```diff
@@ -1,47 +1,53 @@
...
+  ),
+  [FundTypeEnum.ReEntry]: createFundIcon(
+    FundTypeEnum.ReEntry,
+    'Re-Entry Account',
+  ),
+};
+
+export const getFundIcon = (
```
kegrimes, line 51: **[Medium] New exported API with no colocated test**

`TransferIcons.tsx` is now the **only** module in `SavingsFundTransfer/Helper/` without a test — `filterTransfers.test.ts`, `getFundLabel.test.ts`, and `getStatus.test.ts` all exist — and it's the one file in that directory that gained public API in this PR.

A test here would have caught three other findings at once: the untranslated `titleAccess`, the `undefined` return path that `FundInfoDisplay.tsx:22` renders bare, and the case-sensitivity change in `createTableRow.tsx:60-61`.

```tsx
// Helper/TransferIcons.test.tsx
describe('getFundIcon', () => {
  it.each([
    [FundTypeEnum.Primary, 'Primary Account'],
    [FundTypeEnum.ReturnTravel, 'Return Travel Account'],
    [FundTypeEnum.ReEntry, 'Re-Entry Account'],
  ])('renders an accessible icon for %s', (fundType, name) => {
    const { getByRole } = render(<>{getFundIcon(fundType)}</>);
    expect(getByRole('img', { name })).toBeInTheDocument();
  });

  it('returns undefined for an unknown fund type', () => {
    expect(getFundIcon('staffAccount')).toBeUndefined();
  });

  it('is case sensitive', () => {
    expect(getFundIcon('primary')).toBeUndefined();
  });
});
```

src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.tsx
```diff
@@ -312,11 +328,12 @@ export const TransfersPage: React.FC<TransfersPageProps> = ({ title }) => {
                 </StyledCardsBox>
               ) : (
                 funds.map((fund) => (
-                  <BalanceCard
-                    fund={fund}
-                    key={fund.id}
-                    handleOpenTransferModal={handleOpenTransferModal}
-                  />
+                  <StyledCardsBox key={fund.id}>
```
kegrimes, line 332: **[Medium] Skeleton and loaded branches now differ in count, width, and gutter — guaranteed layout shift**

| | skeleton branch (`:324-328`) | loaded branch (here) |
|---|---|---|
| item count | always 3 | 3 **or 5** |
| per-item `minWidth` | 250 (`CardSkeleton.tsx:38`) | 215 (`StyledCardsBox`, `:75`) |
| inter-item gap | 32px (`StyledCardsBox` `gap: theme.spacing(4)`) | 16px (parent `<Box gap={2}>`) |

The three skeletons share **one** `StyledCardsBox` so its 32px gap applies between them; each card now gets its **own** wrapper, so that gap has no siblings and the effective spacing falls to the parent's 16px. Combined with the loading race, a cold-session international staffer sees three distinct layouts in one page load: 3 skeletons → 3 cards → 5 cards.

**Fix** — drive the skeleton count off the same source of truth and match the width:

```tsx
{!fundsData && !fundsError
  ? fundTypes.map((fundType) => (
      <StyledCardsBox key={fundType}><CardSkeleton /></StyledCardsBox>
    ))
  : funds.map((fund) => (
      <StyledCardsBox key={fund.id}>
        <BalanceCard fund={fund} handleOpenTransferModal={handleOpenTransferModal} />
      </StyledCardsBox>
    ))}
```

and drop the now-inert `minWidth: 250` from `CardSkeleton.tsx:38`.

**Separately (severity 6.0): no error or empty state.** The guard at `:323` is `!fundsData && !fundsError` — when `fundsError` is truthy the skeleton is suppressed and `funds` is `[]`, so this `.map()` renders **nothing**. A blank region between the account box and the tables, indistinguishable from "you have no funds." The sibling tables at `:346-352` and `:360-366` do supply `emptyPlaceholder`, and the repo has four `<Alert severity="error">` precedents.

kegrimes (commented):
## 🤖 Multi-Agent Code Review — BLOCKERS FOUND

**5 specialized agents + dependency analysis + gap review + 4 cross-examination rounds.**

### Verdict

**2 blockers** before merge:

1. **`overallBalance` silently under-reports** — a `SeniorStaff` user holding Return Travel / Re-Entry funds sees **$3,500.00 instead of $5,500.00** in the Staff Expense Report headline. Measured empirically.
2. **The four-group `usStaffGroup` predicate has no effective test coverage** — the two "does not request" tests pass for *every* user group, and two of the four gated groups can be deleted from the source with the suite still green.

### Risk

**10/10 (CRITICAL)** — but driven by file count (11 Medium-risk feature files) rather than any high-risk surface. No auth, API, Apollo config, schema, workflow, or dependency files touched. Reviewer level: **MID-LEVEL/SENIOR**.

### Dependency impact

- **Breaking changes: none.** `PrimaryAccount`, `SavingsAccount`, `ConferenceSavingsAccount`, `iconMap` — 0 remaining references anywhere.
- **CRITICAL fan-out:** `src/hooks/useIneligibleByGroup.ts` — 2 direct consumers, ~18 transitive guarded pages.

---

### ❓ One open question that decides blocker #1

`reportsStaffExpenses` is a primary-API root field with no local resolver, so nothing in this repo establishes whether **a plain SeniorStaff / NewStaff user can hold a *nonzero* Return Travel or Re-Entry balance.**

- If **yes** → real money is silently subtracted from the headline total. Blocker stands at 9.0.
- If **always $0** for those groups → the under-report is cosmetic and this drops to ~3.0, no longer blocking.

The realistic scenario is the group transition: a Re-Entry fund is at its maximum exactly when HCM flips `SeniorInternationalStaff → SeniorStaff`. **Worth a one-line answer from the API team before merging.**

---

### ✅ Findings the debate rounds *cleared*

Three findings looked serious in the first pass and did not survive cross-examination. Recording the corrections so they aren't re-litigated:

| Finding | Pre-debate | Post-debate | What refuted it |
|---|---|---|---|
| `useGetUserQuery` missing `loading`/`error` | 8.5 | **4.5** | Both pages mount inside `<UserTypeAccess>`, which blocks children on `userLoading` and returns `<LimitedAccess userGroupError />` on error (`staffExpense/index.page.tsx:36`, `StaffSavingFundLayout.tsx:42`, `UserTypeAccess.tsx:71-77`). Two agents ran guarded probes: **one** fund query, correct variables, first try. The double-fire everyone observed is an artifact of tests rendering the bare component — which is also the root cause of blocker #2. |
| `useIneligibleByGroup` access-widening | 7.8 | **5.5** | Git evidence: the predecessor hook `useUsStaffGroups` gated on HCM booleans; main-branch refactor `5ec62c491` replaced it with enum allowlists covering only 4 of 10 groups, dropping international/stint **by omission**. This PR restores them with four explicit per-group tests. `HrTools/CLAUDE.md` self-declares its eligibility table non-authoritative and states the guards are "UX gating, **not** the security boundary." |
| `compareFundTypes` intransitivity | 7.0 | **5.5** | Real and reproduced by three agents, but `fundTypes` is a server-side allowlist and both call sites pass literals that are a subset of `fundTypeOrder`'s keys — **reachability is zero today**. Latent maintenance hazard, one-line fix. |

---

### Credit where due

- Three hardcoded hex colors (`#F08020`, `#007890`, `#00C0D8`) replaced with `theme.palette.*` tokens — a genuine design-system win.
- Nested-ternary icon selection and the `iconMap` string-munging layer both deleted; fund icon/color logic now has one implementation instead of two.
- `fundTypeHelpers.test.ts` is a new test file for a previously untested module.
- **Clean:** no `any`, `@ts-ignore`, `!` assertions, `console.*`, `debugger`, bare TODOs, `new Date()`, `fireEvent`, commented-out code, or empty catch blocks. `yarn eslint` and `yarn tsc --noEmit` pass. All 6 affected suites pass (68 tests).

---

### Findings on related files (not in this PR)

**[Pre-existing, informational — do not block on this]** `src/components/HrTools/SavingsFundTransfer/Table/Row/createTableRow.tsx:199-213, 234-242, 263-271`

`<IconButton><Edit onClick={…} /></IconButton>` puts the handler on the child SVG. Mouse clicks work; **keyboard Enter/Space fires on the `<button>` and the handler never runs** — keyboard-only users cannot Edit, Cancel Transfer, or Stop Transfer. Sibling buttons at `:217-223` and `:246-252` do it correctly. Confirmed pre-existing via `git blame` (`69ac02febb`, not on this branch). Severity 8.0. **Worth a follow-up ticket.**

---

### Summary

| Agent | Critical | High | Medium | Suggestions | Confidence |
|-------|----------|------|--------|-------------|------------|
| Architecture | 0 | 0 | 5 | 4 | High |
| Testing | 1 | 1 | 4 | 6 | High |
| Standards | 0 | 0 | 5 | 2 | High |
| UX | 0 | 0 | 4 | 5 | High |
| Financial Reporting | 1 | 0 | 2 | 3 | High |
| Gap Review | 0 | 0 | 0 | 2 | High |
| **Total (deduped)** | **1** | **1** | **12** | **8** | |

To dismiss a finding below severity 7, reply `/dismiss: <reason>` on its comment.

src/components/Reports/StaffExpenseReport/Helpers/fundTypeHelpers.ts
```diff
@@ -37,6 +37,18 @@ const fundTypeConfig: Record<string, FundTypeConfig> = {
...
+const fundTypeOrder: Record<string, number> = {
+  Primary: 0,
+  'Staff Conference Savings': 1,
+  Savings: 2,
+  'Return Travel': 3,
+  'Re-Entry': 4,
+};
+
+export const compareFundTypes = (a: string, b: string): number => {
+  return fundTypeOrder[a] - fundTypeOrder[b] || a.localeCompare(b);
```
kegrimes: **[Medium] Non-transitive comparator — one unknown fund type reorders the *known* ones**

`fundTypeOrder[a] - fundTypeOrder[b]` is `NaN` for an unmapped key. `NaN` is falsy so it *does* fall through to `localeCompare` — no `NaN` escapes. The defect is mixing two orderings: `fundTypeOrder` puts `'Staff Conference Savings'` (1) **before** `'Savings'` (2), while alphabetically `'Savings' < 'Staff Conference Savings'`.

Reproduced by three agents independently:

```
compare('Special Fund', 'Savings')                 =  1
compare('Special Fund', 'Staff Conference Savings') = -1
compare('Savings',      'Staff Conference Savings') =  1   ← cycle

['Savings','Special Fund','SCS'].toSorted(c)  → ['Savings','Special Fund','SCS']   ✗ wrong
['Special Fund','Savings','SCS'].toSorted(c)  → ['SCS','Savings','Special Fund']   ✓ right
```

Same set, two orders depending on input permutation — `Array.prototype.sort` is UB with an inconsistent comparator.

**Reachability is zero today** (debate revised 7.0 → 5.5): `fundTypes` is a server-side allowlist and both call sites pass literals that are a subset of `fundTypeOrder`'s keys. But the same module ships `defaultFundTypeConfig` precisely because unknown fund types are anticipated for *display*, and the next person to add a sixth type to the request array without touching `fundTypeOrder` gets silent mis-ordering rather than a type error.

```ts
export const compareFundTypes = (a: string, b: string): number => {
  const orderA = fundTypeOrder[a] ?? Number.MAX_SAFE_INTEGER;
  const orderB = fundTypeOrder[b] ?? Number.MAX_SAFE_INTEGER;
  return orderA - orderB || a.localeCompare(b);
};
```

The new `fundTypeHelpers.test.ts` covers only the all-known case — worth adding a mixed-input case.

src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.tsx
```diff
@@ -12,6 +12,7 @@ import {
 import { styled } from '@mui/material/styles';
 import { DateTime } from 'luxon';
 import { useTranslation } from 'react-i18next';
+import { UsStaffGroupEnum } from 'pages/api/graphql-rest.page.generated';
```
kegrimes: **[Medium] `UsStaffGroupEnum` imported from the server-side generated module**

Measured: **17 of 19** non-generated importers use `src/graphql/types.generated`; the only 2 exceptions are this file and its test. `StaffExpenseReport.tsx:23` in this same PR does it the conventional way.

`pages/api/graphql-rest.page.generated.ts` imports `{ Context } from './graphql-rest.page'` — the server lambda entry point — so this pulls a runtime enum value across the client/lambda boundary. Both enums were verified byte-identical (same members, same string values), so there is **no live bug**; the concern is layering and drift.

```ts
// before
import { UsStaffGroupEnum } from 'pages/api/graphql-rest.page.generated';
// after
import { UsStaffGroupEnum } from 'src/graphql/types.generated';
```

Same change in `TransfersPage.test.tsx:11`. Note import-order lint: `src/graphql/...` sorts after `src/components/User/GetUser.generated` and before `src/theme`.

src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.tsx
```diff
@@ -82,25 +84,39 @@ export const TransfersPage: React.FC<TransfersPageProps> = ({ title }) => {
     StaffSavingFundContext,
   ) as StaffSavingFundType;
 
+  const { data: userData } = useGetUserQuery();
```
kegrimes: **[Suggestion] `loading` / `error` dropped — safe here only because of an ancestor guard**

On first render `usStaffGroup` is `undefined`, so `hideFunds` is `true` for **everyone**, and the funds query fires with the truncated list before refiring with the full one.

**The debate cleared the production impact** (8.5 → 4.5): both pages mount inside `<UserTypeAccess>` (`StaffSavingFundLayout.tsx:42`, `staffExpense/index.page.tsx:36`), which blocks children on `userLoading` (`UserTypeAccess.tsx:77`) and returns `<LimitedAccess userGroupError />` on error (`:71`). Two agents ran guarded probes and observed exactly **one** fund query with correct variables in both branches. The double-fire is a test artifact — which is precisely why the negative tests are vacuous.

What remains is an **undeclared dependency on that ancestor guard**. Nothing in this file says "something above me guarantees GetUser has resolved," and any future page that renders `TransfersPage` without the wrapper reintroduces every symptom. Note `useIneligibleByGroup.ts:27` — edited by this very PR — exposes `userLoading`/`userError` for exactly this reason.

```tsx
const { data: userData, loading: userLoading } = useGetUserQuery();
...
const { data: fundsData, error: fundsError } = useFundBalancesQuery({
  variables: { fundTypes },
  // Don't fire with a truncated fund list before the staff group is known.
  skip: userLoading,
});
```

src/hooks/useIneligibleByGroup.ts
```diff
@@ -12,8 +26,8 @@ const isInEligibleGroup = (
 export function useIneligibleByGroup() {
   const { data, loading: userLoading, error: userError } = useGetUserQuery();
 
-  const usStaffGroup = data?.user.usStaffGroup;
-  const spouseUsStaffGroup = data?.user.spouseUsStaffGroup;
+  const usStaffGroup = getBaseGroup(data?.user.usStaffGroup);
+  const spouseUsStaffGroup = getBaseGroup(data?.user.spouseUsStaffGroup);
```
kegrimes, line 32: **[Medium] Spouse base-group mapping is untested**

All four new tests (`useIneligibleByGroup.test.tsx:178-235`) pass `spouseUsStaffGroup: null`, and the pre-existing spouse block uses plain `SeniorStaff`. Removing `getBaseGroup` from **this line only** would fail nothing.

That matters because the spouse path grants real access: a user with no `usStaffGroup` whose spouse is `SeniorInternationalStaff` now becomes ASR- and MHA-eligible.

```tsx
it('maps a senior international staff spouse to senior staff', async () => {
  const { result } = renderUseIneligibleByGroup({
    usStaffGroup: null,
    spouseUsStaffGroup: UsStaffGroupEnum.SeniorInternationalStaff,
  });

  await waitFor(() => expect(result.current.userLoading).toBe(false));
  expect(result.current).toMatchObject({
    inAsrIneligibleGroup: false,
    inMhaIneligibleGroup: false,
    inSalaryCalcIneligibleGroup: true,
  });
});
```

src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.test.tsx
```diff
@@ -296,6 +301,50 @@ describe('TransfersPage', () => {
     expect(within(tables[0]).getAllByRole('columnheader')).toHaveLength(8);
   });
 
+  it('requests re-entry and return travel for international staff', async () => {
```
kegrimes: **[Medium] No page-level mock contains a Return Travel, Re-Entry, or Staff Conference Savings fund**

The `FundBalances` mock (`:134-151`) and the `ReportsStaffExpenses` mock (`StaffExpenseReport.test.tsx:69-73`) contain only Primary and Savings. Three consequences fall out of that one fact:

1. **The PR's actual feature is never rendered.** Nothing proves a Return Travel or Re-Entry fund produces a balance card, a Flight/Home icon, or a transfer-modal option. `BalanceCard.test.tsx` covers the icons in isolation; no integration path exists.
2. **The sort-key change is untested.** Primary(id 1) + Savings(id 2) sort identically under the old `a.id.localeCompare(b.id)` and the new `compareFundTypes` — the test would pass if the change were reverted.
3. **`:276` is still titled `'should sort fund cards in ascending order by id'`** — sorting is no longer by id.

Adding the three missing funds to one mock closes all three, and finally covers the discriminating case: **Staff Conference Savings must render before Savings** — neither alphabetical nor id order.

src/components/HrTools/SavingsFundTransfer/Helper/getFundLabel.ts
```diff
@@ -11,5 +11,11 @@ export const getFundLabel = (fund: string | undefined, t: TFunction) => {
   if (fund === FundTypeEnum.ConferenceSavings) {
     return t('Staff Conference Savings Account');
   }
+  if (fund === FundTypeEnum.ReturnTravel) {
```
kegrimes, line 14: **[Suggestion] These two new branches are untested — cheapest gap in the PR**

`getFundLabel.test.ts` already exists (Primary / Savings / ConferenceSavings / unknown / undefined) and wasn't extended. Deleting **both** new `if` blocks leaves all five existing cases green.

These are also the only place the new localized labels are asserted at all, so a typo (`'Re Entry Account'`) would ship silently into the transfer table and modal.

```ts
it('returns the Return Travel Account label', () => {
  expect(getFundLabel(FundTypeEnum.ReturnTravel, i18n.t)).toBe('Return Travel Account');
});

it('returns the Re-Entry Account label', () => {
  expect(getFundLabel(FundTypeEnum.ReEntry, i18n.t)).toBe('Re-Entry Account');
});
```

src/components/HrTools/SavingsFundTransfer/BalanceCard/BalanceCard.tsx
```diff
@@ -60,7 +54,7 @@ export const BalanceCard: React.FC<BalanceCardProps> = ({
         flexDirection: 'column',
       }}
     >
-      <Box display={'flex'} flexDirection="row" alignItems="center" gap={1}>
+      <Box display={'flex'} flexDirection="row" alignItems="flex-start" gap={1}>
```
kegrimes, line 57: **[Suggestion] `flex-start` makes the common case *worse* than the `center` it replaced**

Derived from the actual box model: the icon box (`:58-75`) is `p: 1` around a 24px `SvgIcon` → **40px** tall (corroborated by `CardSkeleton.tsx:23-24`, which hardcodes 40×40 for the same box). The title box (`:76-91`) sets `minHeight: '3em'` at 13pt → ~**52px**, with `display: flex; alignItems: center` so the text self-centers inside it.

- Old `center`: icon center 26px, text center 26px — **aligned exactly**.
- New `flex-start`: icon center 20px, text center 26px — **6px mismatch**.

At ~26px/line, `minHeight: 3em` is exactly two lines, so `flex-start` only helps for **3+ line** titles — and hurts the 1-2 line case, which dominates at these card widths.

If you were fixing a specific long-title case, the better lever is removing the inner centering rather than changing the outer alignment. If not, revert to `alignItems="center"`. (Flagged at medium confidence — reasoned from the box model; jsdom does not lay out, so this deserves one look in a browser at `md`/`lg` with an international-staff account.)

src/components/HrTools/SavingsFundTransfer/TransferModal/Helper/FundInfoDisplay.tsx
```diff
@@ -24,11 +19,7 @@ export const FundInfoDisplay: React.FC<FundInfoDisplayProps> = ({ fund }) => {
 
   return (
     <Box sx={{ display: 'flex', alignItems: 'center' }}>
-      {fund.fundType === FundTypeEnum.Primary
-        ? PrimaryAccount
-        : fund.fundType === FundTypeEnum.Savings
-          ? SavingsAccount
-          : ConferenceSavingsAccount}
+      {getFundIcon(fund.fundType)}
```
kegrimes, line 22: **[Medium] `getFundIcon` can return `undefined`; this call site renders it bare**

The removed ternary always produced an icon (falling through to `ConferenceSavingsAccount`); this renders nothing for an unrecognized `fundType`. `createTableRow.tsx:63` guards with `if (fromIcon && toIcon)` — this one doesn't.

**Two mitigating facts, which is why this is 5.0 and not higher:** every value the `FundBalances` query can currently return is in `FundTypeEnum`, so the branch is practically unreachable; and the old fallback actively **mislabeled** unknown funds as "Staff Conference Savings Account" to screen readers. Degraded-but-honest beats confident-and-wrong.

The durable point is the inconsistency: `getIconForFundType` (in `fundTypeHelpers`) falls back to `Groups`, while `getFundIcon` returns `undefined` — two APIs over the same data with opposite unknown-type policies. Pick one and document it. Making the membership check explicit also removes the unchecked `as` cast:

```ts
export const getFundIcon = (fundType: string | undefined): JSX.Element | undefined =>
  fundType && fundType in fundIcons ? fundIcons[fundType as FundTypeEnum] : undefined;
```

src/components/Reports/StaffExpenseReport/Helpers/fundTypeHelpers.ts
```diff
@@ -14,7 +14,7 @@ interface FundTypeConfig {
   getColor: (theme: Theme) => string;
 }
 
-const fundTypeConfig: Record<string, FundTypeConfig> = {
+export const fundTypeConfig: Record<string, FundTypeConfig> = {
```
kegrimes: **[Suggestion] `fundTypeConfig` was exported but has zero importers**

Grepped all of `src/` and `pages/` — the only references are `:58` and `:64`, inside this same file. `TransferIcons.tsx` reaches for `getIconForFundType` / `getIconColorForFundType` instead.

Either revert to a private `const`, or put it to use as the membership check in the `getFundIcon` fix (see the `FundInfoDisplay.tsx` comment) — which would give it a real consumer and remove the unchecked cast at the same time.

src/components/HrTools/SavingsFundTransfer/BalanceCard/BalanceCard.test.tsx
```diff
@@ -97,7 +97,7 @@ describe('BalanceCard', () => {
         <Components
           fund={{
             ...defaultFund,
-            fundType: 'Conference Savings',
+            fundType: FundTypeEnum.ConferenceSavings,
```
kegrimes, line 100: **[Suggestion] This fixture change leaves the fallback-icon branch covered nowhere**

`'Conference Savings'` was not a key in `fundTypeConfig` — it fell through to `defaultFundTypeConfig` (`fundTypeHelpers.ts:52-55`). Switching to `FundTypeEnum.ConferenceSavings` (`'Staff Conference Savings'`) makes this a known-key test, and grepping the whole test tree confirms **nothing else anywhere asserts the `Groups` fallback**.

Given that this PR's premise is fund types being conditionally present — and that the API returns `fundType` as a plain `String`, not an enum — the unknown-type path is the one most likely to regress unnoticed.

Also: the test name at `:95` (`'should display group icon'`) described the old fallback behavior and no longer matches its body.

```tsx
it('should display the fallback icon for an unknown fund type', () => {
  const { getByTestId } = render(
    <Components fund={{ ...defaultFund, fundType: 'Unknown Fund' }} />,
  );

  expect(getByTestId('GroupsIcon')).toBeInTheDocument();
});
```

src/components/HrTools/SavingsFundTransfer/Table/Row/createTableRow.tsx
```diff
@@ -57,23 +57,8 @@ export function populateTransferRows(options: Options) {
...
-      row.transferTo?.toLowerCase() ===
-      FundTypeEnum.ConferenceSavings.toLowerCase()
-        ? 'conference'
-        : row.transferTo
-          ? row.transferTo.toLowerCase()
-          : 'savings';
-
-    const fromIcon = iconMap[fromIconName];
-    const toIcon = iconMap[toIconName];
+    const fromIcon = getFundIcon(row.transferFrom);
```
kegrimes, line 60: **[Suggestion] Icon lookup is now exact-case — behavior change, entirely uncovered**

The old `iconMap` path lowercased both sides and defaulted a null `transferFrom` → `'primary'` / `transferTo` → `'savings'`. This matches `'Primary'` exactly. The source is `tx.transfer.sourceFundTypeName` (`TransfersPage.tsx:192, 228`) — a free-form API string. If upstream casing ever shifts, both icons come back `undefined`, the `if (fromIcon && toIcon)` guard fails, and the Transfers column renders the literal `N/A`; the DataGrid `valueGetter` degrades to `'N/A to N/A'`.

**Both reviewing agents revised this down** — the countervailing argument is strong: `getFundLabel.ts` already compares case-sensitively against the same values, so the removed normalization protected only one of two consumers of the same string; and the old null-fallbacks rendered a **confidently wrong** icon rather than admitting ignorance. Net a11y improvement.

What's left is that the change is untested. `TransfersTable.test.tsx:36-47` uses exact-cased values so it passes either way. One case with `transferFrom: 'savings'` asserting the `N/A` fallback would at least document the contract.

src/components/Reports/StaffExpenseReport/StaffExpenseReport.tsx
kegrimes: /dismiss no we only want to show the funds we query. If we have 3 funds showing for a senior staff, then the total should add up to all 3 funds NOT 5

src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.tsx
kegrimes: /dismiss This is okay for now. If introduced a third time we will update

src/hooks/useIneligibleByGroup.ts
kegrimes: /dismiss Senior staff stint and senior international staff are senior staff and should be able to have the eligibility permissions a senior staff has.

src/components/HrTools/SavingsFundTransfer/Helper/TransferIcons.tsx
kegrimes: /dismiss these reports are US staff only.

src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.tsx
kegrimes: /dismiss this is okay for now.

src/components/Reports/StaffExpenseReport/Helpers/fundTypeHelpers.ts
kegrimes: /dismiss there will never be any unknown funds. I will remove the default case though

src/components/HrTools/SavingsFundTransfer/TransferModal/Helper/FundInfoDisplay.tsx
kegrimes: /dismiss

src/components/HrTools/SavingsFundTransfer/Table/Row/createTableRow.tsx
kegrimes: /dismiss

kegrimes:
@zweatshirt I've been fighting with Claude for a couple hours now, so could you look this over for me? I think everything should be goo, and your reviews are pretty thorough so I appreciate anything you can give!

zweatshirt:
@kegrimes What issue are you fighting with? Is it related to the agent-review? If so, please let me know, as I am the one to fix it

zweatshirt:
If you think there are any suggestions the review agent made where it shouldn't block you, feel free to tag me in those too and I'll try to tweak the agent

zweatshirt (approved):
This looks and works great! Tried to find something to nitpick. Unsure why Claude is giving you a hard time.

kegrimes:
@zweatshirt Thank you!

I think it was mostly concerned about the total at the top adding up to all fund accounts even if Re-entry and Return Travel were hidden. That was wrong behavior and it kept mentioning that. Also, the other blocker it found was on a test which I don't think should be blocking at all even though the test wasn't testing correctly. I could just read that as a suggestion and fix it! 

I think I was just also annoyed at how long it was taking. I ran two reviews and I think one took like 30-45 min. It kept writing tests and then removing them which I think is beneficial but I felt like there was a lot it was doing for such a simple fix.

zweatshirt:
> @zweatshirt Thank you!
> 
> I think it was mostly concerned about the total at the top adding up to all fund accounts even if Re-entry and Return Travel were hidden. That was wrong behavior and it kept mentioning that. Also, the other blocker it found was on a test which I don't think should be blocking at all even though the test wasn't testing correctly. I could just read that as a suggestion and fix it!
> 
> I think I was just also annoyed at how long it was taking. I ran two reviews and I think one took like 30-45 min. It kept writing tests and then removing them which I think is beneficial but I felt like there was a lot it was doing for such a simple fix.

Yeah I don't know why it writes scratch tests. That's not something that exists in the prompt for the agent-review. Claude just decided to start doing that one day (even before I made changes to the skill). I'll try to tweak what I can. I think an issue we will run into in the future is that as the model changes, the skill itself will need to be changed to match the model

PR #1956, opened 2026-07-31
[No-Jira - Do not merge] Test new `/quality:agent-review`  option to dismiss high severity issues

zweatshirt:
## Description

This PR emulates a change that the `quality:agent-review` bot flags as blocking, to test if the developer can optionally override the blocking status vs explicit dismissal.
- `quality:agent-review` PR: https://github.com/CruGlobal/claude-code-plugins/pull/30

## Testing

- Read the Agent review comments

## Checklist:

- [x] I have given my PR a title with the format "MPDX-(JIRA#) (summary sentence max 80 chars)"
- [ ] I have applied the appropriate labels (_Add the label "Preview" to automatically create a preview environment_)
- [ ] I have run the Claude Code `/quality:agent-review` command locally and fixed any relevant suggestions
- [ ] I have requested a review from another person on the project
- [ ] I have tested my changes in preview or in staging
- [ ] I have cleaned up my commit history

PR #1958, opened 2026-07-31
[No-Jira - Do not merge] Test /agent-review dismiss model — fresh slate #2

zweatshirt:
Fresh-slate test PR for the reworked `/agent-review` single-`/dismiss` model (one dismissed counter, any-severity dismissal, ⚠ was a blocker flag). Same top commit as #1956 (`14bb103ef`), on a new branch so the review starts with no prior findings.

Do not merge — testing only.

### DIFF
--- a/src/components/Reports/FourteenMonthReports/useFourteenMonthReport.ts
+++ b/src/components/Reports/FourteenMonthReports/useFourteenMonthReport.ts
@@ -113,7 +113,7 @@ export const useFourteenMonthReport = (
                 months: contactDonationInfo.months.map((month, index) => {
                   const salaryCurrencyTotal = month.donations.reduce(
                     (convertedTotal, donation) =>
-                      convertedTotal + Number(donation.converted_amount),
+                      convertedTotal + Number(donation.amount),
                     0,
                   );
                   return {


src/components/Reports/FourteenMonthReports/useFourteenMonthReport.ts
```diff
@@ -113,7 +113,7 @@ export const useFourteenMonthReport = (
                 months: contactDonationInfo.months.map((month, index) => {
                   const salaryCurrencyTotal = month.donations.reduce(
                     (convertedTotal, donation) =>
-                      convertedTotal + Number(donation.converted_amount),
+                      convertedTotal + Number(donation.amount),
```
zweatshirt, line 116: [High] **Currency-basis regression: the salary report now sums donations in the donor's native currency.**

`converted_amount` is the salary-denominated value; `amount` is the donor's native one. Four independent proofs:

1. The deleted REST datahandler (`git show 60946c688^:pages/api/Schema/reports/fourteenMonth/datahandler.ts`) contains this reduce **character-for-character** using `converted_amount`, and types each donation with all four of `amount`/`currency`/`converted_amount`/`converted_currency` — `converted_amount` is **required**, refuting any "the field may be absent" reading.
2. The identical `year`/`year_converted` convention at lines 76-80 still selects the converted variant under the same `isSalaryType` flag. **One flag cannot mean "converted" on line 78 and "unconverted" on line 122.** This argument needs no knowledge of the upstream API.
3. `git log -L` confirms `4e1d87c9a` (2021) *introduced* this reduce so the salary report would differ from the donor report; `6eb86ad4c` (2024) depended on that branch. This collapses it, restoring the pre-2021 behavior `4e1d87c9a` existed to fix.
4. Reductio: post-change every production-visible salary cell is native, so "Contributions by Salary Currency" and "Contributions by Partner Currency" would render identical data.

**Impact.** Bounded to accounts with foreign-currency donors — where `amount === converted_amount` this is a literal no-op. For those accounts it is **silent**: `numberFormat` uses `style: 'decimal'` (`src/lib/intlFormat.ts:14`), so no cell prints a currency symbol, and neither `tsc` nor any test distinguishes the two versions. The value feeds the CSV "Missing In Hand Monthly Equivalent" column, so a staffer could conclude they are funded when they are not.

A dedicated adversarial agent was tasked with refuting this and could not (`refuted: false`).

**Required fix:**
```suggestion
                      convertedTotal + Number(donation.converted_amount),
```

If the motivating complaint was that the salary row's columns don't reconcile, the currency-consistent fix is to obtain converted values for `total` / `complete_months_total` / `average` / `minimum` (API work) — not to un-convert the month cells. See the pre-existing issues section of the review body.

_Flagged by: Financial Reporting (9.5→8.5), Architecture (7.5→8.0, challenged to 8.5), UX (6.5), Testing (support), Dependency analysis._

zweatshirt, line 116: [Medium] **Nothing in this repo can adjudicate this line — it is executed but never asserted.**

`currencyGroups` is a `JSON!` scalar hand-cast at line 68, and both `amount` and `converted_amount` are declared `string`, so `tsc` accepts either. The line is *executed* by 8 test suites (96.55% line coverage on this hook), but **no test asserts a per-contact month cell**:

- `Layout/Table/Table.test.tsx` is the only salary fixture with non-empty donations, and its assertions cover `totalGivenByContact` (= `completeMonthsTotal`), `monthlyTotals` (= the `totals` prop), `averageTotal`, `minimumTotal`, `overallTotal` — all server-supplied and untouched by this line.
- `FourteenMonthReport.test.tsx` runs the real hook but its salary fixture has `donations: []` on all 12 month objects, so the reduce returns `0` under either field name. Its numeric month assertions sit in the Donor-type block, where this branch is unreachable.
- `useCsvData.test.ts` and `helpers.test.ts` accept pre-built fixtures and never invoke the hook.

One `expect` on a per-contact month cell would have decided this entire review.

The coverage gap itself is **pre-existing** — this hook has never had a colocated test — which is why this is scored 5.5 rather than higher. But this specific line was just changed, converting latent debt into active risk, and tests are the *only* possible guard here since the compiler provably cannot help.

**Suggested:** add `useFourteenMonthReport.test.tsx` (or extend `Table.test.tsx`) with a salary fixture where `amount !== converted_amount`, asserting on `contacts[0].months[n].total`.

_Flagged by: Testing (7.0→5.5 after debate), Standards, Architecture, Financial._

zweatshirt, line 116: [Medium] **No nullish guard on the newly-read field.**

`Number(donation.amount)` yields `NaN` for an absent value, and one `NaN` poisons the whole month's reduce. The two surfaces fail differently:

- **Table:** renders a convincing `0`, not a visible error — `src/lib/intlFormat.ts:17` coerces non-finite input via `Number.isFinite(value) ? value : 0`. Silent wrong data in a donations report is the worse failure mode.
- **CSV:** unguarded. `useCsvData.ts:119` pushes raw `Math.round(month.total)` into the array, so a literal `NaN` reaches the file — and it propagates to the totals row, the grand total (`:151`), `inHandMonthlyEquivalent` (`:83`) and the three support-gap columns derived from it. The sanitizing reduces at `:138`/`:142`/`:146` guard with `typeof row[n] === 'number'`, and **`typeof NaN === 'number'`**, so they do not stop it.

Scored 5.0 rather than higher because both fields are non-optional in the real payload type, making `NaN` reachable mainly from hand-written fixtures — and the hazard is **symmetric and pre-existing** (the `converted_amount` path had the same exposure). Contested during debate: Financial 3.0, Testing 4.5, Architecture and UX 6.0.

**Suggested:** `Number(donation.converted_amount ?? 0)` when applying the blocker fix above. If the report should distinguish "no donation" from "amount unknown", surface that rather than defaulting to `0`.

_Flagged by: Financial, UX, Testing, Architecture._

src/components/Reports/FourteenMonthReports/useFourteenMonthReport.ts
```diff
@@ -113,7 +113,7 @@ export const useFourteenMonthReport = (
                 months: contactDonationInfo.months.map((month, index) => {
                   const salaryCurrencyTotal = month.donations.reduce(
                     (convertedTotal, donation) =>
```
zweatshirt, line 115: [Medium] **Three identifiers now misdescribe the value they hold.**

The accumulator is `convertedTotal` (this line), the binding is `salaryCurrencyTotal` (lines 114, 124), and the field is declared on the shared `Contact` type at `Layout/Table/TableHead/TableHead.tsx:29`. All three assert "this is the salary-converted amount," while the reduce now sums the donor's native `amount`.

Misleading names on money variables are how this class of bug survives review — and because the field escapes onto a type consumed across the `Layout/` tree, the wrong mental model propagates beyond this file. Nothing in the type system catches it (`currencyGroups` is `JSON!`, hand-cast at line 68).

**This finding is contingent** — it evaporates entirely if the blocker above is fixed, since reverting restores name/value agreement with no rename needed. It only stands as an independent issue if the change is kept deliberately, in which case both identifiers need renaming and the now-redundant reduce should yield to `Number(month.total)`.

_Flagged by: Architecture (6.5→5.5), Standards (5.5), Financial (4.0→3.5)._

zweatshirt (commented):
## Multi-Agent Code Review — Verdict: **BLOCKERS FOUND**

5 specialized agents + dependency analysis + 4 debate rounds (including a dedicated adversarial refuter). Mode: `standard`.

**1 high-priority blocker.** Note the deliberate split: the risk gate is **LOW** (read-only display path, one route, one-token revert, nothing persisted) while the finding is a blocker. That is the model working as intended — `danger` measures damage surface and recoverability; correctness bugs are caught by per-finding severity, not by the gate.

### Risk Assessment

| Axis | Score | Reason |
|---|---|---|
| Blast Radius | 1 | 3 direct dependents (1 production), confirmed by dependency analysis |
| Complexity | 1 | one-token field swap, no new branching |
| Sensitivity | 1 | read-only donation report display path |
| Recoverability | 0 | plain revert restores prior state, no side effects |

**effort 2/6 · danger 1/6 → LOW** · Suggested reviewer level: ANY

### The blocker in one line

**[8.5]** `useFourteenMonthReport.ts:116` — the salary-currency report now sums each donation's **native** currency amount instead of the salary-converted one. Four independent proofs, all verified against code:

1. The deleted REST datahandler (`git show 60946c688^:pages/api/Schema/reports/fourteenMonth/datahandler.ts`) contains this reduce **character-for-character** using `converted_amount`, and types each donation with all four of `amount`/`currency`/`converted_amount`/`converted_currency` — with `converted_amount` **required**, refuting any "the field may be absent" reading.
2. The identical `year`/`year_converted` convention at lines 76-80 still selects the converted variant under the same `isSalaryType` flag. **One flag cannot mean "converted" on line 78 and "unconverted" on line 122.** This argument needs no knowledge of the upstream API.
3. `git log -L` confirms commit `4e1d87c9a` (2021) *introduced* this reduce specifically so the salary report would differ from the donor report; `6eb86ad4c` (2024) depended on that branch. This PR collapses it, restoring the pre-2021 behavior `4e1d87c9a` existed to fix.
4. Reductio: post-change every production-visible salary cell is native, making "Contributions by Salary Currency" and "Contributions by Partner Currency" render identical data.

A dedicated adversarial agent was tasked with refuting this finding and **could not** (`refuted: false`).

### Dependency Impact

**No breaking changes** — signature and return shape byte-identical. 3 direct dependents (1 production). Affects only the `/reports/salaryCurrency` route.

Downstream surfaces reached by the changed value: per-contact month cells (`Layout/Table/Table.tsx:215`), the totals row via `calculateTotals`, column sort order, print output, and the CSV month columns, grand total, and **"In Hand / Missing In Hand Monthly Equivalent"** support-gap columns (`useCsvData.ts:83`, `:106-117`).

---

These were identified during review on files or lines outside this PR's diff hunk, so they cannot be posted as line comments. They are informational — they do **not** count toward the verdict or the blocker count, and cannot be dismissed via `/dismiss`.

**[Suggestion] `src/components/Reports/FourteenMonthReports/useFourteenMonthReport.ts:22`** — `converted_amount` becomes the only unreferenced field on the `CurrencyGroup` interface. Its real value is as *evidence*: a diff that orphans the last `_converted` field on the route whose entire purpose is currency conversion is the smell that identifies the bug.
- Severity: 4.0/10
- Flagged by: Architecture, Standards, Testing
- Recommended Action: resolves on revert; otherwise delete the field or comment why it is intentionally unread.

**[Suggestion] `src/components/Reports/FourteenMonthReports/useFourteenMonthReport.ts:124`** — the exposed `salaryCurrencyTotal` field has had no production reader since `6eb86ad4c` (Sept 2024). Only a type declaration (`Layout/Table/TableHead/TableHead.tsx:29`) and ~30 test fixtures reference it.
- Severity: 3.5/10
- Flagged by: Architecture, Dependency analysis, Testing
- Recommended Action: dead-field cleanup, ideally in a separate PR.

**[Suggestion] `src/components/Reports/FourteenMonthReports/Layout/Table/Table.test.tsx:129`** — the test passes `totals={currencyGroup.totals.months}` (server-supplied) while production passes `calculateTotals(currencyGroup.contacts)` (`FourteenMonthReport.tsx:75`). The test exercises a wiring production never uses, which is *why* the salary totals row is structurally immune to this bug class. Pre-existing.
- Severity: 4.5/10
- Flagged by: Testing, Architecture
- Recommended Action: align the test's `totals` prop with production wiring.

**[Suggestion] `src/components/Reports/FourteenMonthReports/Layout/Table/Table.test.tsx:40` and `:53-66`** — the fixture supplies `converted_amount` (no `amount`) on one contact and `amount` (no `converted_amount`) on the other. One contact renders a broken cell under the old code and the *other* under the new code, unasserted either way. Pre-existing.
- Severity: 4.0/10
- Flagged by: Testing, Standards, Architecture, Financial
- Recommended Action: set both fields, with **different** values, on every donation in the fixture.

---

These predate this PR, do not count toward the verdict, and are best addressed separately.

- **[~6]** The salary report's Avg / Min / "12 Month Total" columns have **never** had converted variants — the API supplies no `_converted` sibling for `total` / `complete_months_total` / `average` / `minimum` (lines 107-112). For foreign-currency groups the row has never reconciled. This is the real API-side gap, and it is the strongest argument in the change's favor: it makes the row internally consistent — by making every number consistently wrong for the report's stated purpose. **If the motivating complaint was that salary rows don't add up, the currency-consistent fix is to obtain converted values for those four aggregates, not to un-convert the month cells.**
- **[~4]** `fourteenMonthReport.salaryCurrency` (line 71) is computed but never consumed; `FourteenMonthReport.tsx:122` passes the currency-group key to the table header instead.
- **[~3]** The entire `currencyGroup.totals` object (lines 75-95) has no production reader.

---

### Convention Drift (Advisory)

Non-scored, non-blocking, cannot be dismissed. Flags that a convention doc may need updating to match this PR.

- **`.claude/rules/code-review.md`** → Domain Agents → Financial Reporting Agent → Focus areas → "Currency mixing". The bullet conflates two failure modes: *mechanical* (don't add EUR to USD) and *semantic* (the aggregate must be in the report's declared reporting currency). This diff **passes** the first — the reduce is scoped inside a single `currencyGroups` bucket (line 72) — and **fails** the second.
  - Suggested update: split the bullet in two so the distinction is checkable. This does **not** exonerate this diff.

---

### Unresolved Debates

1. **Severity 7.0 vs 8.5.** The adversarial refuter argued 7.0 (read-only path, already-inconsistent surface); the Financial and Architecture agents independently landed on 8.5 after being pushed to recalibrate downward. Spread of 1.5 is below the escalation threshold. Resolved at **8.5**.
2. **The upstream payload is unverifiable from this repo.** `currencyGroups` is an opaque `JSON!` scalar from the primary API, hand-cast at line 68. If the salary endpoint carries *only* `converted_amount`, every salary month cell silently renders `0` and the CSV fills with `NaN` — a worse outcome than the currency regression. Either way the change is wrong; only the failure mode differs. **Recommendation: attach a real API payload to the PR** so this can be settled from evidence rather than inference.
3. **Is the cross-column inconsistency pre-existing or introduced?** Adjudicated as **pre-existing** — the change flips *which* columns are wrong, not whether they disagree.

---

### Review Summary Table

| Financial Reporting | 0 | 1 | 0 | 2 | High |
| Architecture | 0 | 1 | 0 | 4 | High |
| Testing & Quality | 0 | 0 | 0 | 4 | High |
| Standards | 0 | 0 | 0 | 3 | High |
| UX | 0 | 0 | 0 | 3 | High/Med |
| Adversarial refuter | — | confirmed | — | — | High |
| **Total (deduped)** | **0** | **1** | **0** | **7** | |

### Resolving this review

1. Fix the severity ≥ 8 blocker (revert line 116 to `Number(donation.converted_amount)`).
2. For severity < 7 findings you disagree with, reply `/dismiss: <reason>` on the review comment.
3. Push your fixes and re-run `/agent-review`.

Severity 7.0-7.9 findings cannot be dismissed and do not block merge — there are none in this review. Suggestions (< 5.0) are informational and require no action.

src/components/Reports/FourteenMonthReports/useFourteenMonthReport.ts
zweatshirt: /dismiss This will eventually be the normal behavior in this file.

PR #1960, opened 2026-07-31
MPDX-9901, MPDX-9902 Remove person number from UI

kegrimes:
## Description

Some UAT tickets:
- Update user type modal to correctly display "Cru Hybrid" correctly (was displaying "Unknown")
- **Remove all person number and staff account ids from UI**
- Default note in Savings Fund transfer to "{last_name} transfer to ____ account {date}"
- Remove NameDisplay from any forms that were just using the user's name on the card

Jira tickets: [MPDX-9901](https://jira.cru.org/browse/MPDX-9901) and [MPDX-9902](https://jira.cru.org/browse/MPDX-9902)

## Testing

Cru Hybrid test:
- Impersonate: jerry.varghese@cru.org
- Go to dashboard (wait for modal to appear)
- Check that modal has "Cru Hybrid Staff" with new helper text present

Person number/staff account id test:
- ASR ✅
- MHA ✅
- Ministry Partner Reminders ✅
- MPD Supervisor --> Kept all staff account id and person number references
- NS Goal Calculator ✅
- NSO Questionnaire ✅
- Salary Calculator ✅
- Savings Fund Transfer ✅
- MPGA ✅
- Staff Expenses ✅

Default note test:
- Impersonate: timothy.ethington@cru.org
- Go to `/HrTools/staffSavingFund/transfers`
- Create a One Time transfer and select a destination account
- Verify the default note appears
- Delete the note and type another one to ensure overriding works

## Checklist:

### DIFF
--- a/src/components/HrTools/AdditionalSalaryRequest/Shared/useFormUserInfo.ts
+++ b/src/components/HrTools/AdditionalSalaryRequest/Shared/useFormUserInfo.ts
@@ -4,17 +4,12 @@ export const useFormUserInfo = () => {
   const { requestData, user } = useAdditionalSalaryRequest();
   const { staffAccountBalance } =
     requestData?.latestAdditionalSalaryRequest?.calculations || {};
-  const {
-    emailAddress: email,
-    preferredName: name,
-    personNumber: accountNumber,
-  } = user?.staffInfo ?? {};
+  const { emailAddress: email, preferredName: name } = user?.staffInfo ?? {};
   const primaryAccountBalance = staffAccountBalance ?? 0;
 
   return {
     email,
     name,
-    accountNumber,
     primaryAccountBalance,
   };
 };
--- a/src/components/HrTools/MinisterHousingAllowance/MinisterHousingAllowance.tsx
+++ b/src/components/HrTools/MinisterHousingAllowance/MinisterHousingAllowance.tsx
@@ -11,7 +11,6 @@ import { getMhiEligibility } from 'src/components/Reports/Shared/HousingAllowanc
 import { MhaStatusEnum } from 'src/graphql/types.generated';
 import { useAccountListId } from 'src/hooks/useAccountListId';
 import theme from 'src/theme';
-import { NameDisplay } from '../Shared/CalculationReports/NameDisplay/NameDisplay';
 import { PanelLayout } from '../Shared/CalculationReports/PanelLayout/PanelLayout';
 import { PanelTypeEnum } from '../Shared/CalculationReports/Shared/sharedTypes';
 import { EligibleDisplay } from './MainPages/EligibleDisplay';
@@ -50,13 +49,6 @@ export const MinisterHousingAllowanceReport = () => {
     userHcmData,
     spouseHcmData,
   } = useMinisterHousingAllowance();
-
-  const lastName = userHcmData?.staffInfo.lastName ?? '';
-  const spouseLastName = spouseHcmData?.staffInfo.lastName ?? '';
-
-  const names = isMarried
-    ? `${preferredName} ${lastName} and ${spousePreferredName} ${spouseLastName}`
-    : `${preferredName} ${lastName}`;
 
   const [createMHA] = useCreateHousingAllowanceRequestMutation();
   const [creatingRequest, setCreatingRequest] = useState(false);
@@ -154,8 +146,6 @@ export const MinisterHousingAllowanceReport = () => {
                   <Typography variant="h5">{t('Your MHA')}</Typography>
                 </Box>
 
-                <NameDisplay names={names} />
-
                 {hasNoRequests ? (
                   <>
                     <NoRequestsDisplay />
@@ -188,12 +178,14 @@ export const MinisterHousingAllowanceReport = () => {
                   </>
                 ) : (
                   eitherPersonEligible && (
-                    <EligibleDisplay
-                      isPending={isCurrentRequestPending}
-                      isEditable={
-                        currentRequest.status === MhaStatusEnum.ActionRequired
-                      }
-                    />
+                    <Box mb={2}>
+                      <EligibleDisplay
+                        isPending={isCurrentRequestPending}
+                        isEditable={
+                          currentRequest.status === MhaStatusEnum.ActionRequired
+                        }
+                      />
+                    </Box>
                   )
                 )}
 
--- a/src/components/HrTools/MinisterHousingAllowance/MinisterHousingAllowance.tsx
+++ b/src/components/HrTools/MinisterHousingAllowance/MinisterHousingAllowance.tsx
@@ -51,17 +51,12 @@ export const MinisterHousingAllowanceReport = () => {
     spouseHcmData,
   } = useMinisterHousingAllowance();
 
-  const personNumber = userHcmData?.staffInfo.personNumber ?? '';
-  const spousePersonNumber = spouseHcmData?.staffInfo.personNumber ?? '';
   const lastName = userHcmData?.staffInfo.lastName ?? '';
   const spouseLastName = spouseHcmData?.staffInfo.lastName ?? '';
 
   const names = isMarried
     ? `${preferredName} ${lastName} and ${spousePreferredName} ${spouseLastName}`
     : `${preferredName} ${lastName}`;
-  const personNumbers = isMarried
-    ? `${personNumber} and ${spousePersonNumber}`
-    : personNumber;
 
   const [createMHA] = useCreateHousingAllowanceRequestMutation();
   const [creatingRequest, setCreatingRequest] = useState(false);
@@ -159,11 +154,7 @@ export const MinisterHousingAllowanceReport = () => {
                   <Typography variant="h5">{t('Your MHA')}</Typography>
                 </Box>
 
-                <NameDisplay
-                  names={names}
-                  personNumbers={personNumbers}
-                  personNumberCount={isMarried ? 2 : 1}
-                />
+                <NameDisplay names={names} />
 
                 {hasNoRequests ? (
                   <>
--- a/src/components/HrTools/MinisterHousingAllowance/Steps/StepThree/CalcComponents/PersonInfo.tsx
+++ b/src/components/HrTools/MinisterHousingAllowance/Steps/StepThree/CalcComponents/PersonInfo.tsx
@@ -32,8 +32,8 @@ export const PersonInfo: React.FC = () => {
           }}
         >
           <Typography>
-            {t('Staff Account Number: {{personNumber}}', {
-              personNumber: person?.personNumber,
+            {t('Email: {{emailAddress}}', {
+              emailAddress: person?.emailAddress,
             })}
           </Typography>
         </Grid>
@@ -60,8 +60,8 @@ export const PersonInfo: React.FC = () => {
           }}
         >
           <Typography>
-            {t('Email: {{emailAddress}}', {
-              emailAddress: person?.emailAddress,
+            {t('Phone number: {{primaryPhoneNumber}}', {
+              primaryPhoneNumber: person?.primaryPhoneNumber,
             })}
           </Typography>
         </Grid>
@@ -75,18 +75,6 @@ export const PersonInfo: React.FC = () => {
         >
           <Typography>
             {person?.city}, {person?.state} {person?.zipCode}
-          </Typography>
-        </Grid>
-        <Grid
-          size={{
-            xs: 12,
-            sm: 6,
-          }}
-        >
-          <Typography>
-            {t('Phone number: {{primaryPhoneNumber}}', {
-              primaryPhoneNumber: person?.primaryPhoneNumber,
-            })}
           </Typography>
         </Grid>
       </Grid>
--- a/src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsHeader.tsx
+++ b/src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsHeader.tsx
@@ -43,15 +43,10 @@ interface PersonInfoCardProps {
 }
 
 const PersonInfoCard: React.FC<PersonInfoCardProps> = ({ person }) => {
-  const { t } = useTranslation();
-
   return (
     <Stack spacing={0.5}>
       <Typography variant="subtitle1" fontWeight="medium">
         {person.firstName} {person.lastName}
-      </Typography>
-      <Typography variant="body2" color="text.secondary">
-        {t('Person Number: {{number}}', { number: person.personNumber })}
       </Typography>
       <Stack spacing={0.75} sx={{ mt: 0.5 }}>
         <ContactLine icon={<MailOutline />}>{person.emailAddress}</ContactLine>
--- a/src/components/HrTools/NsoMpdQuestionnaire/PersonalInformation/StaffInformation.tsx
+++ b/src/components/HrTools/NsoMpdQuestionnaire/PersonalInformation/StaffInformation.tsx
@@ -84,9 +84,6 @@ export const StaffInformation: React.FC = () => {
           name,
           // Backend doesn't currently support accessing the spouse's avatar
           avatarSrc: viewingSpouse ? null : userData?.user.avatar,
-          staffAccountId: viewingSpouse
-            ? questionnaire?.spousePersonNumber
-            : questionnaire?.personNumber,
         }}
         toggle={
           hasSpouse
--- a/src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
@@ -24,6 +24,7 @@ import {
 import { TransferModal } from './TransferModal';
 
 const accountListId = 'abc';
+const lastName = 'Doe';
 const router = {
   query: { accountListId },
   isReady: true,
@@ -89,6 +90,7 @@ const Components = ({
                   transfer,
                 }}
                 funds={fundsMock}
+                lastName={lastName}
                 handleClose={handleClose}
               />
             </StaffSavingFundProvider>
@@ -267,6 +269,7 @@ describe('TransferModal', () => {
       userEvent.clear(amountField);
       userEvent.type(amountField, '100');
 
+      userEvent.clear(noteField);
       userEvent.type(noteField, 'Test note');
 
       userEvent.click(getByRole('button', { name: /submit/i }));
@@ -295,6 +298,7 @@ describe('TransferModal', () => {
       userEvent.clear(amountField);
       userEvent.type(amountField, '100');
 
+      userEvent.clear(noteField);
       userEvent.type(noteField, 'Test note');
 
       userEvent.click(submitButton);
@@ -309,50 +313,136 @@ describe('TransferModal', () => {
       });
     });
 
-    it('should require a note for one-time transfers', async () => {
-      const { getByRole, findByText } = render(<Components />);
+    it('should leave the note empty until a to account is chosen', () => {
+      const { getByRole } = render(<Components />);
 
-      const toAccount = getByRole('combobox', { name: /to account/i });
-      const amountField = getByRole('spinbutton', { name: /amount/i });
-      const noteField = getByRole('textbox', { name: /note/i });
+      expect(getByRole('textbox', { name: /note/i })).toHaveValue('');
+    });
 
-      userEvent.click(toAccount);
+    it('should default the note once a to account is chosen', async () => {
+      const { getByRole } = render(<Components />);
+
+      userEvent.click(getByRole('combobox', { name: /to account/i }));
       userEvent.click(getByRole('option', { name: /staff savings/i }));
 
+      await waitFor(() =>
+        expect(getByRole('textbox', { name: /note/i })).toHaveValue(
+          'Doe transfer to Staff Savings 1/1/2020',
+        ),
+      );
+    });
+
+    it('should update the note when the transfer date changes', async () => {
+      const { getByRole, getByLabelText } = render(<Components />);
+
+      userEvent.click(getByRole('combobox', { name: /to account/i }));
+      userEvent.click(getByRole('option', { name: /staff savings/i }));
+
+      const transferDate = getByLabelText(/transfer date/i);
+      userEvent.clear(transferDate);
+      userEvent.type(transferDate, '02/15/2020');
+      userEvent.tab();
+
+      await waitFor(() =>
+        expect(getByRole('textbox', { name: /note/i })).toHaveValue(
+          'Doe transfer to Staff Savings 2/15/2020',
+        ),
+      );
+    });
+
+    it('should let the user override the default note', async () => {
+      const { getByRole } = render(<Components />);
+
+      userEvent.click(getByRole('combobox', { name: /to account/i }));
+      userEvent.click(getByRole('option', { name: /staff savings/i }));
+
+      const noteField = getByRole('textbox', { name: /note/i });
+      userEvent.clear(noteField);
+      userEvent.type(noteField, 'My own note');
+
+      await waitFor(() => expect(noteField).toHaveValue('My own note'));
+    });
+
+    it('should keep a custom note when the transfer date changes', async () => {
+      const { getByRole, getByLabelText } = render(<Components />);
+
+      userEvent.click(getByRole('combobox', { name: /to account/i }));
+      userEvent.click(getByRole('option', { name: /staff savings/i }));
+
+      const noteField = getByRole('textbox', { name: /note/i });
+      userEvent.clear(noteField);
+      userEvent.type(noteField, 'Money for the new roof');
+      await waitFor(() =>
+        expect(noteField).toHaveValue('Money for the new roof'),
+      );
+
+      const transferDate = getByLabelText(/transfer date/i);
+      userEvent.clear(transferDate);
+      userEvent.type(transferDate, '02/15/2020');
+      userEvent.tab();
+
+      await waitFor(() => expect(transferDate).toHaveValue('2/15/2020'));
+      expect(noteField).toHaveValue('Money for the new roof');
+    });
+
+    it('should enable submit when the amount is entered before the to account', async () => {
+      const { getByRole } = render(<Components />);
+
+      const amountField = getByRole('spinbutton', { name: /amount/i });
       userEvent.clear(amountField);
       userEvent.type(amountField, '100');
 
-      await waitFor(() =>
-        expect(getByRole('button', { name: /submit/i })).toBeDisabled(),
-      );
+      userEvent.click(getByRole('combobox', { name: /to account/i }));
+      userEvent.click(getByRole('option', { name: /staff savings/i }));
 
-      userEvent.click(noteField);
+      await waitFor(() =>
+        expect(getByRole('textbox', { name: /note/i })).toHaveValue(
+          'Doe transfer to Staff Savings 1/1/2020',
+        ),
+      );
+      await waitFor(() =>
+        expect(getByRole('button', { name: /submit/i })).toBeEnabled(),
+      );
+    });
+
+    it('should require a note for one-time transfers when cleared', async () => {
+      const { getByRole, findByText } = render(<Components />);
+
+      userEvent.click(getByRole('combobox', { name: /to account/i }));
+      userEvent.click(getByRole('option', { name: /staff savings/i }));
+
+      const amountField = getByRole('spinbutton', { name: /amount/i });
+      userEvent.clear(amountField);
+      userEvent.type(amountField, '100');
+
+      userEvent.clear(getByRole('textbox', { name: /note/i }));
       userEvent.tab();
 
       expect(await findByText('Note is required')).toBeInTheDocument();
+      await waitFor(() =>
+        expect(getByRole('button', { name: /submit/i })).toBeDisabled(),
+      );
     });
 
     it('should reject a whitespace-only note for one-time transfers', async () => {
       const { getByRole, findByText } = render(<Components />);
 
-      const toAccount = getByRole('combobox', { name: /to account/i });
-      const amountField = getByRole('spinbutton', { name: /amount/i });
-      const noteField = getByRole('textbox', { name: /note/i });
-
-      userEvent.click(toAccount);
+      userEvent.click(getByRole('combobox', { name: /to account/i }));
       userEvent.click(getByRole('option', { name: /staff savings/i }));
 
+      const amountField = getByRole('spinbutton', { name: /amount/i });
       userEvent.clear(amountField);
       userEvent.type(amountField, '100');
 
+      const noteField = getByRole('textbox', { name: /note/i });
+      userEvent.clear(noteField);
       userEvent.type(noteField, '   ');
       userEvent.tab();
 
+      expect(await findByText('Note is required')).toBeInTheDocument();
       await waitFor(() =>
         expect(getByRole('button', { name: /submit/i })).toBeDisabled(),
       );
-
-      expect(await findByText('Note is required')).toBeInTheDocument();
     });
   });
 
@@ -499,6 +589,7 @@ describe('TransferModal', () => {
       userEvent.clear(amountField);
       userEvent.type(amountField, '100');
 
+      userEvent.clear(noteField);
       userEvent.type(noteField, 'Test note');
 
       const transferDate = getByLabelText(/transfer date/i);
@@ -616,6 +707,7 @@ describe('TransferModal', () => {
       userEvent.clear(amountField);
       userEvent.type(amountField, '100');
 
+      userEvent.clear(noteField);
       userEvent.type(noteField, '  Test note  ');
 
       userEvent.click(getByRole('button', { name: /submit/i }));
@@ -657,6 +749,7 @@ describe('TransferModal', () => {
       expect(transferDate).toHaveValue('02/15/2020');
       userEvent.tab();
 
+      userEvent.clear(noteField);
       userEvent.type(noteField, 'Future transfer');
 
       userEvent.click(getByRole('button', { name: /submit/i }));
--- a/src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
@@ -19,7 +19,7 @@ import {
 import { Formik } from 'formik';
 import { DateTime } from 'luxon';
 import { useSnackbar } from 'notistack';
-import { useTranslation } from 'react-i18next';
+import { TFunction, useTranslation } from 'react-i18next';
 import * as yup from 'yup';
 import { CustomDateField } from 'src/components/Shared/DateTimePickers/CustomDateField';
 import {
@@ -29,7 +29,11 @@ import {
 import Modal from 'src/components/Shared/Modal/Modal';
 import { useLocale } from 'src/hooks/useLocale';
 import i18n from 'src/lib/i18n';
-import { currencyFormat, dateFormat } from 'src/lib/intlFormat';
+import {
+  currencyFormat,
+  dateFormat,
+  dateFormatShort,
+} from 'src/lib/intlFormat';
 import { FundFieldsFragment } from '../ReportsSavingsFund.generated';
 import {
   useCreateRecurringTransferMutation,
@@ -60,6 +64,28 @@ const getTomorrow = (): DateTime => {
 
 const minimumTransferDate = (schedule: ScheduleEnum): DateTime =>
   schedule === ScheduleEnum.OneTime ? getToday() : getTomorrow();
+
+const buildDefaultNote = (
+  lastName: string,
+  transferTo: string,
+  transferDate: DateTime | null,
+  t: TFunction,
+  locale: string,
+): string => {
+  if (!transferTo) {
+    return '';
+  }
+
+  const date = transferDate?.isValid
+    ? dateFormatShort(transferDate, locale)
+    : '';
+
+  return t('{{lastName}} transfer to {{transferTo}} {{date}}', {
+    lastName,
+    transferTo,
+    date,
+  }).trim();
+};
 
 const pastDateMessage = (schedule: ScheduleEnum): string =>
   schedule === ScheduleEnum.OneTime
@@ -139,12 +165,14 @@ const transferSchema = (locale: string) =>
 interface TransferModalProps {
   data: TransferModalData;
   funds: FundFieldsFragment[];
+  lastName: string;
   handleClose: () => void;
 }
 
 export const TransferModal: React.FC<TransferModalProps> = ({
   data,
   funds,
+  lastName,
   handleClose,
 }) => {
   const { t } = useTranslation();
@@ -265,6 +293,7 @@ export const TransferModal: React.FC<TransferModalProps> = ({
         onSubmit={handleSubmit}
       >
         {({
+          values,
           values: {
             transferFrom,
             transferTo,
@@ -283,6 +312,7 @@ export const TransferModal: React.FC<TransferModalProps> = ({
           setFieldValue,
           setFieldTouched,
           validateField,
+          validateForm,
           handleBlur,
         }) => {
           const locale = useLocale();
@@ -360,7 +390,24 @@ export const TransferModal: React.FC<TransferModalProps> = ({
                             labelId="transferTo"
                             name="transferTo"
                             value={transferTo}
-                            onChange={handleChange}
+                            onChange={(event) => {
+                              const destination = String(event.target.value);
+                              const nextNote = buildDefaultNote(
+                                lastName,
+                                destination,
+                                transferDate,
+                                t,
+                                locale,
+                              );
+
+                              setFieldValue('note', nextNote, false);
+                              setFieldValue('transferTo', destination, false);
+                              validateForm({
+                                ...values,
+                                transferTo: destination,
+                                note: nextNote,
+                              });
+                            }}
                             onBlur={handleBlur}
                             error={
                               touched.transferTo && Boolean(errors.transferTo)
@@ -459,8 +506,43 @@ export const TransferModal: React.FC<TransferModalProps> = ({
                         label={t('Transfer Date')}
                         value={transferDate}
                         onChange={(date) => {
-                          setFieldValue('transferDate', date);
+                          const dateChanged =
+                            date?.isValid &&
+                            date.toMillis() !== transferDate?.toMillis();
+
+                          // Changing the destination account always rebuilds
+                          // the note, but a date change only refreshes a note
+                          // the user hasn't overridden
+                          const noteIsDefault =
+                            !note.trim() ||
+                            note ===
+                              buildDefaultNote(
+                                lastName,
+                                transferTo,
+                                transferDate,
+                                t,
+                                locale,
+                              );
+
+                          const nextNote =
+                            dateChanged && noteIsDefault
+                              ? buildDefaultNote(
+                                  lastName,
+                                  transferTo,
+                                  date,
+                                  t,
+                                  locale,
+                                )
+                              : note;
+
+                          setFieldValue('note', nextNote, false);
+                          setFieldValue('transferDate', date, false);
                           setFieldTouched('transferDate', true, false);
+                          validateForm({
+                            ...values,
+                            transferDate: date,
+                            note: nextNote,
+                          });
                         }}
                         error={
                           touched.transferDate && Boolean(errors.transferDate)
--- a/src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.tsx
@@ -308,10 +308,7 @@ export const TransfersPage: React.FC<TransfersPageProps> = ({ title }) => {
             {!staffAccountData && !staffAccountError ? (
               <AccountInfoBoxSkeleton />
             ) : (
-              <AccountInfoBox
-                name={staffAccountData?.staffAccount?.name}
-                accountId={staffAccountData?.staffAccount?.id}
-              />
+              <AccountInfoBox name={staffAccountData?.staffAccount?.name} />
             )}
             <Box
               display="flex"
--- a/src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.tsx
+++ b/src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.tsx
@@ -384,6 +384,7 @@ export const TransfersPage: React.FC<TransfersPageProps> = ({ title }) => {
             handleClose={() => setModalData(null)}
             data={modalData}
             funds={funds}
+            lastName={userData?.user.lastName ?? ''}
           />
         )}
       </Box>
--- a/src/components/HrTools/Shared/AccountInfoBox/AccountInfoBox.test.tsx
+++ b/src/components/HrTools/Shared/AccountInfoBox/AccountInfoBox.test.tsx
@@ -5,51 +5,28 @@ import { AccountInfoBox } from './AccountInfoBox';
 describe('AccountInfoBox', () => {
   it('renders name, accountId, and overallBalance when provided', () => {
     const { getByTestId } = render(
-      <AccountInfoBox
-        name="Test Name"
-        accountId="12345"
-        overallBalance={1000}
-      />,
+      <AccountInfoBox name="Test Name" overallBalance={1000} />,
     );
     expect(getByTestId('account-info')).toBeInTheDocument();
     expect(getByTestId('name')).toBeInTheDocument();
-    expect(getByTestId('account-id')).toBeInTheDocument();
-    expect(getByTestId('InfoOutlinedIcon')).toBeInTheDocument();
     expect(getByTestId('overall-balance')).toBeInTheDocument();
   });
 
-  it('renders empty strings when name, accountId, and overallBalance are not provided', () => {
+  it('renders empty strings when name and overallBalance are not provided', () => {
     const { getByTestId, queryByTestId } = render(<AccountInfoBox />);
     expect(getByTestId('name').textContent).toBe('');
-    expect(queryByTestId('account-id')).not.toBeInTheDocument();
-    expect(queryByTestId('InfoOutlinedIcon')).not.toBeInTheDocument();
     expect(queryByTestId('overall-balance')).not.toBeInTheDocument();
   });
 
-  it('displays only name when accountId and overallBalance are not provided', () => {
-    const { getByTestId, queryByTestId } = render(
-      <AccountInfoBox name="Only Name" />,
-    );
+  it('displays only name when overallBalance is not provided', () => {
+    const { getByTestId } = render(<AccountInfoBox name="Only Name" />);
     expect(getByTestId('name').textContent).toBe('Only Name');
-    expect(queryByTestId('account-id')).not.toBeInTheDocument();
-    expect(queryByTestId('InfoOutlinedIcon')).not.toBeInTheDocument();
   });
 
-  it('renders only accountId when name and overallBalance are not provided', () => {
-    const { getByTestId } = render(<AccountInfoBox accountId="OnlyId" />);
-    expect(getByTestId('account-id').textContent).toBe('OnlyId');
-    expect(getByTestId('InfoOutlinedIcon')).toBeInTheDocument();
-    expect(getByTestId('name').textContent).toBe('');
-  });
-
-  it('renders only overallBalance when name and accountId are not provided', () => {
-    const { getByTestId, queryByTestId } = render(
-      <AccountInfoBox overallBalance={1000} />,
-    );
+  it('renders only overallBalance when name is not provided', () => {
+    const { getByTestId } = render(<AccountInfoBox overallBalance={1000} />);
     expect(getByTestId('overall-balance').textContent).toBe('$1,000.00');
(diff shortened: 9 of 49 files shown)


src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.tsx
```diff
@@ -85,6 +85,8 @@ export const TransfersPage: React.FC<TransfersPageProps> = ({ title }) => {
   const { data: staffAccountData, error: staffAccountError } =
     useStaffAccountQuery();
 
+  const lastName = staffAccountData?.staffAccount?.name?.split(' ')[1] ?? '';
```
kegrimes: **[NEEDS HUMAN REVIEW]** Surname derived by positional string split — contested 4.0 vs 9.0, needs your call.

```ts
const lastName = staffAccountData?.staffAccount?.name?.split(' ')[1] ?? '';
```

This takes the **second whitespace token** of a staff *account* name and treats it as a surname. An adversarial verifier tried to refute this and confirmed every factual element:

- `StaffAccount.name` has **no format contract** — `schema.graphql:13011-13015` is a bare `String!` with no doc comment, and it's documented as "Staff account associated with user's account" (an account label, not a person name)
- **No REST datahandler constructs it** — `staffAccount` is a native root field (`rootFields.generated.ts:62`); nothing under `pages/api/Schema/` produces it, so the hoped-for `"${firstName} ${lastName}"` constructor doesn't exist in this repo
- **Not a repo convention** — of 10 `split(' ')` uses in `src/`, the other 9 are all month/date parsing (`useMonthHeaders.ts:12`, `TableCard.tsx:143`, …). This is the only name-parse in the codebase
- **A correct source already exists in this tree** — `Shared/HcmData/Hcm.graphql:6` exposes `staffInfo.lastName`, used by `MinisterHousingAllowance.tsx:54` and `StaffInfoSummaryCard.tsx:25`
- **The value is persisted** — reaches the server as `description` on `CreateTransfer` (`TransferModal.tsx:244`)
- **Zero test coverage** — `TransferModal.test.tsx:27` injects `lastName='Doe'` as a prop, bypassing this line entirely. `TransfersPage.test.tsx` mocks `name: 'Test Account'` (→ `'Account'`) and never asserts the note

Behaviour today: `"Test Account"` → `"Account"` · `"Mary Ann Smith"` → `"Ann"` · `"Doe, John"` → `"John"` (the *first* name) · `"Cher"` → `""`

**What pulls severity down to 4.0:** the note is a visible, `required`, editable `TextField` the user passes through before Submit, and the value only reaches the wire on the one-time-transfer path. A wrong default the user can fix is a nuisance, not corruption.

**The question only you can answer: what is the real production shape of `staffAccount.name`?** If the upstream API concatenates first + last for single-holder accounts, this is cosmetic (~2.0) and just brittle. If accounts are ever joint or `"Last, First"`, it's a genuine data-quality defect (~5.5).

Preferred fix either way — read the structured field:
```ts
const { data: hcmData } = useHcmQuery();
const lastName = hcmData?.hcm?.[0]?.staffInfo?.lastName ?? '';
```
If HCM is out of scope here, at minimum take the *last* token rather than index 1, and add a `TransfersPage.test.tsx` case asserting the derived value.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
```diff
@@ -360,7 +385,19 @@ export const TransferModal: React.FC<TransferModalProps> = ({
                             labelId="transferTo"
                             name="transferTo"
                             value={transferTo}
-                            onChange={handleChange}
+                            onChange={(event) => {
+                              const destination = String(event.target.value);
+                              setValues((previous) => ({
+                                ...previous,
+                                transferTo: destination,
+                                note: buildDefaultNote(
```
kegrimes: **[Medium]** This handler (and the date handler at line ~502) unconditionally overwrites `note`, destroying text the user typed.

Both call `setValues((previous) => ({ ...previous, <field>, note: buildDefaultNote(...) }))` with no check for whether the user has already edited the note.

**A verifier reproduced this.** With a custom note in place, *merely focusing and tabbing out of the Transfer Date field* replaces it — no date change required — because `DesktopDateField.tsx:95-98` calls `onChange(parsedDate)` on every blur with no equality check:

```
NOTE AFTER BLUR => Doe transfer to Staff Savings 1/1/2020
DATE AFTER BLUR => 1/1/2020
```

**Scope is narrow** (verified, and narrower than the initial 7.5 rating assumed):
- ✅ **New + One-Time** — the only flow with consequence; note renders and is sent as `description`
- ➖ New + Monthly — clobber happens in Formik state but the Note field isn't rendered (line 565 gates on `OneTime`) and `createRecurringTransfer` sends no description
- ➖ Edit mode — the To Account select isn't rendered (`isNew` false), edit only opens for recurring rows, and `UpdateRecurringTransfer` accepts no `description`

Also: the overwrite is visible on screen before Submit, so this is an unexpected overwrite rather than silent data loss.

**The test suite gives false confidence.** `'should let the user override the default note'` (test:351-362) types a custom note and then never touches the date, so it passes trivially. The mutation test (test:688-728) sets the date *first* and the note *second* — encoding the safe ordering. No test performs note-then-date; all 110 tests pass and are structurally incapable of catching this.

**Fix** — only regenerate while the note is still the untouched default. Extract one helper so both handlers share it:

```ts
const nextNote = (previous, next, lastName, locale) => {
  const previousDefault = buildDefaultNote(
    lastName, previous.transferTo, previous.transferDate, locale,
  );
  return previous.note === '' || previous.note === previousDefault
    ? buildDefaultNote(lastName, next.transferTo, next.transferDate, locale)
    : previous.note;
};
```

This also fixes a second case: changing the date *before* picking a destination currently wipes the note to `''`, since `buildDefaultNote` returns `''` when `transferTo` is empty.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
```diff
@@ -61,6 +65,23 @@ const getTomorrow = (): DateTime => {
...
+): string => {
+  if (!transferTo) {
+    return '';
+  }
+
+  const date = transferDate?.isValid
+    ? dateFormatShort(transferDate, locale)
+    : '';
+
+  return `${lastName} transfer to ${transferTo} ${date}`.trim();
```
kegrimes: **[Medium]** Hard-coded English in a user-visible string, bypassing i18n.

```ts
return `${lastName} transfer to ${transferTo} ${date}`.trim();
```

This lands in a `required`, editable Note `TextField` the user reads (line 565-584), is persisted as `description`, and is rendered back into the transfers table (`TransfersPage.tsx:187`). It's both display text *and* data — so the "it's only a payload" framing doesn't hold.

It's also the **only outlier in its own file**: `pastDateMessage` two lines below at :87-88 uses `i18n.t()`, as do lines 92, 93, 97, 100, 120, 140, 153, 154, 157. 30+ non-test files use `i18n.t` for module-scope strings.

I had an agent specifically investigate whether this is *intentional* English for HR reconciliation. **That defense failed on evidence:**
- No documented `description` format anywhere (`createTransfer` is a native root field; nothing under `pages/api/Schema/` references it)
- Zero comments in `src/` or `pages/` justifying any intentionally-untranslated persisted string
- A **direct counter-precedent** exists — `Announcements.tsx:130` builds a composed default for a persisted, user-editable name field and *does* wrap it in `i18n.t`

Independently, the current state is **incoherent under either reading**: the date *is* locale-formatted via `dateFormatShort(transferDate, locale)`, while the surrounding words are not. A French-locale user gets `Doe transfer to Staff Savings 15/02/2020`. If it's UI text it should be fully localized; if it's a stable reconciliation record it should use a fixed ISO date, not one that varies by the user's saved locale preference.

**Fix** — single interpolated key (not concatenated `t()` fragments; word order varies by language):

```ts
return i18n
  .t('{{lastName}} transfer to {{transferTo}} {{date}}', { lastName, transferTo, date })
  .trim();
```

`i18n` is already imported at line 31. Existing tests keep passing — `t()` returns the interpolated source string in the test env.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
```diff
@@ -309,50 +311,94 @@ describe('TransferModal', () => {
       });
```
kegrimes, line 313: **[Medium]** Two pre-existing tests in this file now append to a pre-filled note (lines ~300 and ~548, outside the diff hunks so I can't anchor there directly).

Both select a To Account — which now pre-fills the note — then call `userEvent.type(noteField, 'Test note')` **without a preceding `userEvent.clear(noteField)`**. `userEvent.type` appends, so the submitted note is:

```
"Doe transfer to Staff Savings 1/1/2020Test note"
```

They still pass because they only assert the success snackbar, never the payload — so two tests whose names claim to cover form submission are now asserting nothing about it while feeding a malformed value.

You correctly added `userEvent.clear(noteField)` to the two mutation-asserting tests in this PR (the new ones in this hunk) — these two were just missed. Same one-line fix at both sites:

```tsx
userEvent.clear(noteField);
userEvent.type(noteField, 'Test note');
```

While you're here, one genuinely missing case: no test submits *without editing the note* and asserts `description: 'Doe transfer to Staff Savings 1/1/2020'` on the `CreateTransfer` mutation. That's the single most common production path for this feature — the whole point is that the user doesn't type a note.

src/components/Shared/ConfirmUserGroupModal/Helper/getUserType.ts
```diff
@@ -12,6 +12,8 @@ export const getUserType = (
       return t('Cru Global Staff');
     case UserTypeEnum.NonCru:
       return t('Non Cru User');
+    case UserTypeEnum.HybridStaff:
```
kegrimes, line 28: **[Suggestion]** This new case has no test, and `getUserType.test.ts` enumerates every *other* enum member (`UsStaff` :8, `GlobalStaff` :12, `NonCru` :16, `undefined` :20) — which makes the gap conspicuous and the fix nearly free:

```ts
it('returns correct user type for Hybrid Staff', () => {
  expect(getUserType(UserTypeEnum.HybridStaff, t)).toBe('Cru Hybrid Staff');
});
```

This was initially rated 9.0/BLOCKING; a dedicated adjudication brought it to 3.5, and I agree with that. Reasons: the cited checklist item says "every **new** lib function" and this is a modified pre-existing one; `HybridStaff` is already tested at all four *behavioral* access-gating sites (`UserTypeAccess.test.tsx:235-268`, `useNavPages.test.tsx:89`, `useReportNavItems.test.tsx:101`); the failure mode is a wrong label visible on first render with no data or access-control consequence; and the change **fixes a real bug** — before this, hybrid users hit `default: return t('Unknown')` and saw "The user group for your account is: Unknown" with no description at all.

Worth adding, not worth gating the merge on.

src/components/Shared/ConfirmUserGroupModal/ConfirmUserGroupModal.tsx
```diff
@@ -52,7 +52,11 @@ export const ConfirmUserGroupModal: React.FC<ConfirmUserGroupModalProps> = ({
         ? t(
             'Users in this group receive (mostly) non-US donations and are paid through our Global NetSuite system.',
           )
-        : undefined;
+        : userType === UserTypeEnum.HybridStaff
```
kegrimes: **[Suggestion]** This makes the description chain a three-deep nested ternary, and it's the second differently-shaped mapping of `UserTypeEnum` in the same folder — the sibling `Helper/getUserType.ts` maps the same enum with a `switch`.

So adding a user type now requires edits in two places with two different shapes, and neither has exhaustiveness checking (both have a `default` / trailing `: undefined` fallback), so the compiler won't catch a miss.

Suggest extracting a `getUserTypeDescription(userType, t)` switch into `Helper/` alongside `getUserType.ts`, so the label and description for a user type live in one file with one shape.

Verified this branch is genuinely reachable, not dead code falling through to "Unknown": `HybridStaff = 'HYBRID_STAFF'` exists in the generated enum, and `Dashboard.tsx:66` passes `data.user.userType` through unfiltered. The copy also matches the definition in `src/components/HrTools/CLAUDE.md`. ✅

Note the pre-existing `GlobalStaff` description branch (lines 51-54) has no test either — if you add a hybrid test, worth covering that one in the same commit rather than leaving the file half-enumerated.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
```diff
@@ -255,7 +278,8 @@ export const TransferModal: React.FC<TransferModalProps> = ({
           amount: data.transfer.amount ?? 0,
           schedule: data.transfer.schedule ?? ScheduleEnum.OneTime,
           status: data.transfer.status ?? '',
-          transferDate: data.transfer.transferDate ?? getToday(),
+          transferDate: (data.transfer.transferDate ??
```
kegrimes: **[Suggestion]** `as DateTime | null` widens an inferred type where fixing the declared type would be cleaner.

The cast exists so `setValues` can accept the `DateTime | null` that `CustomDateField.onChange` emits. But `TransferFormValues` already exists at lines 47-55 and already declares `transferDate: DateTime<boolean> | null` — it's just not applied to `<Formik>`, so the form's real value type is inferred and the interface is decorative. (`initialValues` also carries three fields the interface never mentions: `status`, `isEditing`, `originalStart`.)

Adding those three to `TransferFormValues` and passing it as the generic — `<Formik<TransferFormValues> initialValues={...}>` — removes the cast entirely and gets you compile-time checking on `setValues` / `setFieldValue`, which the new note logic leans on.

src/components/HrTools/Shared/CalculationReports/NameDisplay/NameDisplay.tsx
```diff
@@ -25,32 +22,20 @@ interface NameDisplayProps {
...
   amountTwo,
   spouseComponent,
 }) => {
-  const { t } = useTranslation();
   const locale = useLocale();
   const theme = useTheme();
   const currency = 'USD';
   const title = (
     <Box sx={{ display: 'flex', justifyContent: 'space-between' }}>
       <Box sx={{ display: 'flex', flexDirection: 'column' }}>
```
kegrimes: **[Suggestion]** Vestigial wrapper — this `<Box sx={{ display: 'flex', flexDirection: 'column' }}>` existed to stack the name above the person-number line, and now wraps a single child.

```tsx
<Box sx={{ display: 'flex', justifyContent: 'space-between' }}>
  <Typography variant="h6">{names}</Typography>
  {spouseComponent}
</Box>
```

No visual change (column flex with one child), just dead scaffolding that suggests a second row was intended.

Same shape in `StaffInfoCard.tsx:49-53` — the `<Stack>` now holds one `<Typography>`, and its `lineHeight={1.2}` existed only to tighten the two-line name + person-number block.

src/components/HrTools/AdditionalSalaryRequest/Shared/useFormUserInfo.ts
```diff
@@ -4,17 +4,12 @@ export const useFormUserInfo = () => {
   const { requestData, user } = useAdditionalSalaryRequest();
   const { staffAccountBalance } =
     requestData?.latestAdditionalSalaryRequest?.calculations || {};
-  const {
-    emailAddress: email,
-    preferredName: name,
-    personNumber: accountNumber,
-  } = user?.staffInfo ?? {};
+  const { emailAddress: email, preferredName: name } = user?.staffInfo ?? {};
```
kegrimes, line 7: **[Suggestion]** While you're cleaning this hook: `email` is dead too.

This line correctly dropped the unread `personNumber: accountNumber` alias, but kept the equally-unread `emailAddress: email`. All five consumers destructure only `name` and/or `primaryAccountBalance` — `AboutForm.tsx:18`, `NewForm.tsx:26`, `EditForm.tsx:27`, `ViewForm.tsx:27`, `useAdditionalSalaryRequestForm.ts:39`. The one place that needs the email reads it straight off the context instead (`useAdditionalSalaryRequestForm.ts:80,101`).

```ts
const { preferredName: name } = user?.staffInfo ?? {};
const primaryAccountBalance = staffAccountBalance ?? 0;
return { name, primaryAccountBalance };
```

(and drop the `result.current.email` assertions at `useFormUserInfo.test.ts:40,79`)

Minor related note: `personNumber: '00123456'` is still sitting in `AboutForm.test.tsx:43,53` and `FormVersions/testUtils.tsx:21` after the assertions reading it were deleted — read by nothing now.

kegrimes (commented):
## Multi-Agent Review — PR #1960

**Verdict: BLOCKERS_FOUND — 1 item needs human adjudication (0 severity-8+ defects)**

To be clear about what this means: no agent found a severity-8+ defect. The verdict is triggered by the severity-spread rule — on one finding, one agent rated 9.0/BLOCKING and a dedicated adversarial verifier rated 4.0 with evidence. Rather than average them away, it goes to you.

**The person-number removal itself — the PR's title change, 21 of 23 source files — came back clean.** All three substantive findings are in the *savings-fund default note* feature that rode along on this branch.

### Setup

6 specialized agents (Architecture, Testing, Standards, UX, Data Integrity, Financial Reporting) + dependency-impact sweep + 2 fresh-eyes gap agents + 3 adversarial verifiers that were told to *refute* the top findings. Repo conventions loaded from `CLAUDE.md` and `.claude/rules/code-review.md`.

### Risk

**10/10 CRITICAL** — but driven purely by file count (21 medium-risk files × 1pt, cross-cutting ×1.7). Actual change is +76/−136 non-test lines, mostly deletion. Read this as "wide blast radius," not "dangerous operation." Suggested reviewer level: mid-level/senior.

### Verification actually run

- `tsc --noEmit` — clean
- `yarn jest src/components/HrTools/SavingsFundTransfer` — **110/110 pass**
- ASR suites 56/56 pass; reports + landings 146 tests pass
- Zero surviving references to any removed prop across `src/` and `pages/`

### Dependency impact

No CRITICAL-impact files. Highest fan-out: `NameDisplay.tsx` (7 dependents), `useFormUserInfo.ts` (5) — every dependent updated in this PR. No breaking changes reach an unupdated call site.

### Findings on files not in this PR's diff

These can't be posted as line comments.

**[Medium] Orphaned person-number data paths — severity 5.0** — PII still fetched with no reader, which matters given this ticket's intent:
- `NsoMpdQuestionnaire/Shared/NewStaffQuestionnaire.graphql:8,12` — `personNumber` / `spousePersonNumber`, zero readers after `StaffInformation.tsx` was updated
- `MinisterHousingAllowance/MinisterHousingAllowance.graphql:27,41,60` — `personNumber`, zero readers
- `NsGoalCalculator/GoalSettings/goalSettingsFormValues.ts:29` — `GoalSettingsPerson.personNumber` is still a **required** type member, still populated at `GoalSettingsForm.tsx:98,107`, now write-only

⚠️ `Shared/HcmData/Hcm.graphql`'s `personNumber` is still **live** (`orientSalaryRequest.ts`, `useLandingData.ts:159`, `RequestedSalaryCard.tsx:46`) — do not remove it in the cleanup.

**[Suggestion] `getUserType.test.ts` — severity 3.5** — add `expect(getUserType(UserTypeEnum.HybridStaff, t)).toBe('Cru Hybrid Staff')`. Adjudicated down from an initial 9.0: the cited checklist item says "every **new** lib function" and this is a *modified pre-existing* one; `HybridStaff` is already tested at all four behavioral access-gating sites (`UserTypeAccess.test.tsx`, `useNavPages.test.tsx`, `useReportNavItems.test.tsx`); the sibling `GlobalStaff` description branch was already untested before this PR; and the change **fixes a real visible bug** (hybrid users previously saw "Unknown"). Nearly-free fix, not a gate.

**[Context] `src/components/Shared/DateTimePickers/DesktopDateField.tsx:95-98`** — `onBlur` calls `onChange(parsedDate)` unconditionally, with no equality check. Pre-existing and not changed here, but it's what makes the note-clobber finding below reachable by focus+tab alone.

### What came back clean

Skeleton/real parity verified at all 4 `AccountInfoBox` call sites · no table column or `colSpan` drift · no money arithmetic touched, no `new Date()` anywhere, no client-side aggregation changed · `id` retained in every GraphQL selection set (the `id` removals in the diff are test mocks only) · `UserTypeEnum` "fix one, fix all" **passes** — all 7 other branch sites already handle `HybridStaff` · no `any`, no `console.*`, no commented-out code, no orphaned imports · person-number sweep complete, the only survivors are the deliberately reverted `MpdSupervisorReport/`

### Summary

| | Critical | High | Needs human review | Medium | Suggestions |
|---|---|---|---|---|---|
| Deduped across 12 agents | 0 | 0 | 1 | 4 | 8 |

Severity < 5.0 findings are informational and don't require `/dismiss`. For severity 5.0–6.9 findings you disagree with, reply `/dismiss: <reason>` on the comment.

src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.tsx
kegrimes: /dismiss Savings fund transfer does not use HCM data at all and middle names do not appear

src/components/HrTools/AdditionalSalaryRequest/Shared/useFormUserInfo.ts
kegrimes: /dismiss I will keep this for now incase we want to make future changes!

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
```diff
@@ -459,6 +501,23 @@ export const TransferModal: React.FC<TransferModalProps> = ({
                         label={t('Transfer Date')}
                         value={transferDate}
                         onChange={(date) => {
+                          const dateChanged =
```
kegrimes, line 509: **[Important]** Changing the transfer date silently overwrites a note the user typed.

Flagged by all 5 agents and **reproduced with real probe tests by 4 of them**. `buildDefaultNote` is written back unconditionally whenever the date changes — there's no check for whether the note is still the auto-generated one.

The trigger is more ordinary than it looks: `DesktopDateField.tsx:95-98` fires `onChange` on **blur**, so simply tabbing out of the date field after typing a note replaces it. Same unguarded overwrite exists on the To Account handler at `:391-405`.

**Severity came *down* from 9.0 during debate, and that matters.** A round-1 probe claimed Edit mode destroys an already-*saved* description. Three agents traced the open-path and disproved it — `getStatus.ts:5-7` marks every one-time transfer `Complete` → `TransfersPage.tsx:188` denies it `edit-delete` → `createTableRow.tsx:211` → `TransfersTable.tsx:117`. Every reachable Edit modal is `Monthly`, where the note field isn't rendered (`:578`) and `UpdateRecurringTransfer` carries no `description`. **No persisted data is at risk** — this is unsaved input on a visible field.

Your test `should keep the note when tabbing through the transfer date` doesn't cover this: it clicks and tabs without changing the value, so `dateChanged` is never true. Deleting the whole `if (dateChanged)` block leaves it green.

**Recommended fix closes this *and* the stranded-Submit bug below.** Architecture applied and verified it — 34/34 existing tests still pass, `yarn lint:ts` clean:

```tsx
const noteEdited = useRef(false);   // outer component body

// To Account onChange
onChange={(event) => {
  const destination = String(event.target.value);
  setValues((prev) => ({ ...prev, transferTo: destination,
    note: noteEdited.current ? prev.note
      : buildDefaultNote(lastName, destination, transferDate, t, locale) }));
}}

// Transfer Date onChange
onChange={(date) => {
  const dateChanged = date?.isValid && date.toMillis() !== transferDate?.toMillis();
  setValues((prev) => ({ ...prev, transferDate: date,
    note: dateChanged && !noteEdited.current
      ? buildDefaultNote(lastName, transferTo, date, t, locale) : prev.note }));
  setFieldTouched('transferDate', true, false);
}}

// Note field onChange
onChange={(event) => { noteEdited.current = true; handleChange(event); }}
```

A boolean ref was chosen over comparing against the generated string: that comparison runs on **translated** output, so a locale switch or a missing translation would silently break it.

src/components/HrTools/SavingsFundTransfer/TransfersPage/TransfersPage.tsx
kegrimes: **[Important]** Surname derived by positional split — all 5 agents + dependency analysis.

**Your `/dismiss` reasoning was verified correct.** Architecture and Financial both independently confirmed HCM is unreachable here: `StaffSavingFundContext` carries only `isNavListOpen`/`onNavListToggle`, and this page's three queries are `StaffAccount`, `ReportsSavingsFundTransfer`, `FundBalances`. The dismissal couldn't apply (threshold is <7.0), but your objection did its job — it killed the wrong fix.

**Three competing fixes collapsed into one during debate.** Architecture withdrew its `deriveLastName` helper; Financial retracted `useHcmQuery` (it would have added a network round-trip for a ~45-field payload including `personNumber` — reintroducing exactly the PII this PR removes). Unanimous answer:

```ts
const { data: userData } = useGetUserQuery();
const lastName = userData?.user.lastName ?? '';
```

Verified: `GetUser.graphql:5` selects `lastName` as a discrete server field — **no string parsing at all**, so no heuristic can be wrong. `UserPreferenceProvider` already calls `useGetUserQuery()` at `_app.page.tsx:65` app-wide, making this a **free warm-cache read**. It also resolves at app boot, before the lazily-mounted modal exists, closing a load-order race where an unresolved query would bake an empty surname in permanently.

**Why this is worth fixing rather than living with:** Financial traced the consequence nobody had. The note becomes the transfer `description`, and `TransfersPage.tsx:187` reads it back into the history grid, the print view, and the **CSV export** (`downloadTable.tsx:52`). There's no in-app path to correct a submitted transfer's description. Today: `'Test Account'` → `'Account'`, `'Mary Jo Smith'` → `'Jo'`, `'Doe, John'` → `'John'`, single-token → `''`.

**One product question only you can answer:** `user.lastName` is the *session user*; `staffAccount.name` is the *account holder*. If those can diverge on a shared account, which one belongs in the description?

No test exercises this line today — `TransferModal.test.tsx` injects `lastName='Doe'` as a prop, and `TransfersPage.test.tsx` mocks `name: 'Test Account'` and asserts nothing about the note.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
```diff
@@ -360,7 +388,21 @@ export const TransferModal: React.FC<TransferModalProps> = ({
                             labelId="transferTo"
                             name="transferTo"
                             value={transferTo}
-                            onChange={handleChange}
+                            onChange={(event) => {
```
kegrimes, line 393: **[Medium]** Submit button strands with a contradictory error — **new regression, found during debate**.

Enter the **amount before** picking a To Account, then touch nothing. Submit stays disabled indefinitely on a fully valid form, showing *"Note is required"* underneath a Note field that visibly contains `Doe transfer to Staff Savings 1/1/2020`. It only recovers when you touch an unrelated field.

**Four independent probes reproduced this.** The decisive pair: with every value valid, Submit is disabled; then a blur that changes **zero values** flips it to enabled — proving the validation ran against a stale snapshot.

Root cause, confirmed against the Formik 2.2.9 source (`setFieldValue` → `validateFormWithHighPriority(setIn(state.values, field, value))`): `state.values` is the **render-closure** state. Two synchronous `setFieldValue` calls in one handler produce no intervening render, so the second call validates an object where `note` is still `''`, setting `errors.note` and `isValid = false`.

This is a regression from this PR — the diff replaced `onChange={handleChange}` with the paired `setFieldValue` calls. Architecture initially cleared it on reasoning, then ran the probe in rebuttal and **conceded in full**.

The same latent pattern exists on the date handler at `:509-521`.

**Fixed by the `setValues` half of the patch in the note-clobber comment above** — `setValues` validates the object it actually writes. A test that fails on head today:

```tsx
it('enables submit when the amount is entered before the destination', async () => {
  const { getByRole } = render(<Components />);
  const amountField = getByRole('spinbutton', { name: /amount/i });
  userEvent.clear(amountField);
  userEvent.type(amountField, '100');
  userEvent.click(getByRole('combobox', { name: /to account/i }));
  userEvent.click(getByRole('option', { name: /staff savings/i }));

  await waitFor(() =>
    expect(getByRole('button', { name: /submit/i })).toBeEnabled(),
  );
});
```

src/components/Shared/ConfirmUserGroupModal/Helper/getUserType.ts
```diff
@@ -7,12 +7,35 @@ export const getUserType = (
...
+        ),
+      };
     case UserTypeEnum.NonCru:
-      return t('Non Cru User');
+      return {
+        label: t('Non Cru User'),
+        sublabel: null,
+      };
+    case UserTypeEnum.HybridStaff:
+      return {
```
kegrimes, line 29: **[Medium]** The PR's headline fix has no test at the modal level.

The new `HybridStaff` case is well covered in `getUserType.test.ts`, and the helper is now enum-exhaustive — `UserTypeEnum` has exactly 4 members and none falls through to `Unknown` any more. Verified.

But the bug you fixed ("Cru Hybrid staff see *Unknown*") lived in the **rendered modal**. `ConfirmUserGroupModal.test.tsx` exists and passes, but covers only `UsStaff` and `NonCru` — it isn't in this PR's changed files even though `ConfirmUserGroupModal.tsx` is. So the `label`/`sublabel` destructure and the reworked `mb: sublabel ? 0 : 2` branch are untested at the render layer, and a future refactor could reintroduce a blank label while the helper test stays green.

That file's `TestComponent` already accepts a `userType` prop, so this is ~10 lines:

```tsx
it('renders the hybrid staff user group and description', async () => {
    <TestComponent userType={UserTypeEnum.HybridStaff} />,
  );

  expect(await findByText('Cru Hybrid Staff')).toBeInTheDocument();
  expect(
    await findByText(
      'Users in this group receive donations through our Global NetSuite system but also need access to US HR forms (Salary Calc, MHA Calc, etc).',
    ),
  ).toBeInTheDocument();
});
```

Spacing was verified correct across all five branches — no change needed there.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.test.tsx
```diff
@@ -309,50 +313,111 @@ describe('TransferModal', () => {
...
+
+      const noteField = getByRole('textbox', { name: /note/i });
+      userEvent.clear(noteField);
+      userEvent.type(noteField, 'My own note');
 
-      userEvent.click(noteField);
+      await waitFor(() => expect(noteField).toHaveValue('My own note'));
+    });
+
+    it('should keep the note when tabbing through the transfer date', async () => {
```
kegrimes: **[Medium]** `should keep the note when tabbing through the transfer date` asserts nothing.

This test clicks the date field and tabs, but never changes the value — so `CustomDateField`'s `onChange` never fires and the `dateChanged` guard is never evaluated. **Deleting the entire `if (dateChanged)` block leaves this test green.** It's a mutation-testing no-op.

That's especially unfortunate because the name describes exactly the broken behavior: per `DesktopDateField.tsx:95-98`, `onChange` fires on *blur*, so "tabbing through the transfer date" is precisely the gesture that triggers the clobber.

Replace it with a test that actually changes the date after the user types a note:

```tsx
it('should preserve a custom note when the transfer date changes', async () => {
  const { getByRole, getByLabelText } = render(<Components />);

  userEvent.click(getByRole('combobox', { name: /to account/i }));
  userEvent.click(getByRole('option', { name: /staff savings/i }));

  const noteField = getByRole('textbox', { name: /note/i });
  userEvent.clear(noteField);
  userEvent.type(noteField, 'My own note');
  await waitFor(() => expect(noteField).toHaveValue('My own note'));

  const transferDate = getByLabelText(/transfer date/i);
  userEvent.clear(transferDate);
  userEvent.type(transferDate, '02/15/2020');
  userEvent.tab();

  await waitFor(() => expect(transferDate).toHaveValue('02/15/2020'));
  expect(noteField).toHaveValue('My own note');
});
```

Keep `should update the note when the transfer date changes` — with the fix applied, an *untouched* note should still regenerate.

Also worth adding: the empty-`lastName` boundary case, which is the most likely production shape today.

One thing that checked out cleanly: your date assertions are time-safe — `__tests__/util/setup.ts:67-70` pins Luxon's `Settings.now` to 2020-01-01 and `globalSetup.ts:2` pins `TZ=UTC`.

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsHeader.tsx
```diff
@@ -43,16 +43,11 @@ interface PersonInfoCardProps {
 }
 
 const PersonInfoCard: React.FC<PersonInfoCardProps> = ({ person }) => {
-  const { t } = useTranslation();
-
   return (
```
kegrimes, line 46: **[Medium]** Removing this display orphaned the data behind it.

The deleted `t('Person Number: {{number}}', …)` line was the **only reader** of `person.personNumber`. It's now written but never read:

- `goalSettingsFormValues.ts:29` — the interface field
- `GoalSettingsForm.tsx:98,107` — the only writes
- `NewStaffGoalCalculation.graphql:55` — the selection

Neither `tsc` nor eslint catches this, because it's a required interface property rather than an unused local.

Standards traced `goalSettingsApiMapping.ts` to be sure deletion is safe: `personNumber` appears there only inside a JSDoc comment, and `GoalSettingsPerson` is a sibling type that never enters `formValuesToAttributes`. Safe to delete all four, then `yarn gql`.

Worth doing in this PR specifically because it's the PR's own goal: with the selection retained, the identifier still ships to the browser and persists into `localStorage` via `persistCache` (`src/lib/apollo/client.ts:21-26`), so it stays visible in devtools even though nothing renders it.

The same applies to `NewStaffQuestionnaire.graphql:8` (see the NSO comment). Note that `MinisterHousingAllowance.graphql` and `AdditionalSalaryRequest.graphql` were verified **already** orphaned before this PR — leave those for a separate cleanup.

src/components/HrTools/NsoMpdQuestionnaire/PersonalInformation/StaffInformation.tsx
```diff
@@ -84,9 +84,6 @@ export const StaffInformation: React.FC = () => {
           name,
           // Backend doesn't currently support accessing the spouse's avatar
```
kegrimes, line 85: **[Medium]** Same orphaning as NsGoalCalculator, verified at the base commit.

A gap-review agent confirmed via `git grep` against `80f6e52` that lines 88-89 of this file (`questionnaire?.spousePersonNumber` / `questionnaire?.personNumber`) were the **only** readers in the repo. With them removed, `NewStaffQuestionnaire.graphql:8` selects `personNumber` for nothing.

That fragment is reused by the `UpdateNewStaffQuestionnaire` and `CompleteNewStaffQuestionnaire` mutation responses too, so it ships on every questionnaire load *and* every save.

```graphql
fragment NewStaffInformationFields on NewStaffQuestionnaire {
  id
  address
  age
  firstName
  lastName
  maritalStatus
-  personNumber
  spouseAge
  spouseFirstName
  spouseJoining
-  spousePersonNumber
  spouseTenure
  tenure
  variant
}
```

Then `yarn gql`, and drop the now-dead mocks at `NsoMpdQuestionnaireTestWrapper.tsx:22,25`.

src/components/HrTools/MinisterHousingAllowance/MinisterHousingAllowance.tsx
```diff
@@ -197,12 +178,14 @@ export const MinisterHousingAllowanceReport = () => {
                   </>
                 ) : (
                   eitherPersonEligible && (
```
kegrimes, line 180: **[Suggestion]** `[Pre-existing — aggravated]` One reachable state now renders an empty panel.

When `!hasNoRequests && !eitherPersonEligible`, this `else` branch evaluates to `false`, and `showCurrentRequest` / `showPreviousRequests` / `showNewRequestButton` are all gated on `eitherPersonEligible` too — so only the `Your MHA` heading at `:146` survives.

**This PR didn't create the hole.** Pre-PR the branch rendered `<NameDisplay>` — the user's name and person number, not a table or an action or an explanation. The PR downgraded "your name and nothing else" to "nothing else." Recorded as pre-existing; it does not count toward the verdict.

Two ways it's reached: a staff member who filed an MHA request and later lost eligibility, and — more commonly — a transient window, since `MinisterHousingAllowanceContext.tsx:201-208` derives `userEligibleForMHA` from the HCM query while `:138-142` gates rendering only on the *requests* query's `loading`. Whenever requests resolve first, the panel is briefly blank.

Worth a follow-up: gate on the HCM query's loading state too, and render `EligibilityStatusTable` in the ineligible branch so the user learns *why* there's nothing to do.

Separately — with `NameDisplay` gone, a married couple's MHA page no longer shows whose MHA is displayed anywhere. That's consistent with the ticket's intent, but a `{names}` subheading under the heading would restore orientation without any person number.

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
```diff
@@ -61,6 +65,28 @@ const getTomorrow = (): DateTime => {
...
+  transferTo: string,
+  transferDate: DateTime | null,
+  t: TFunction,
+  locale: string,
+): string => {
+  if (!transferTo) {
+    return '';
+  }
+
+  const date = transferDate?.isValid
```
kegrimes, line 79: **[Suggestion]** The persisted `description` varies by the author's locale, and `.trim()` only works by accident.

Two related issues in `buildDefaultNote`, raised separately by UX, Standards, and Financial, and consolidated during debate:

1. **`.trim()` is order-dependent.** It cleans up the missing-date case only because English happens to end with `{{date}}`. A locale that reorders — `"{{date}} — transfert de {{lastName}} vers {{transferTo}}"` — leaves a leading or interior double space that `.trim()` can't reach.
2. **The value is persisted.** This string becomes the transfer `description` on the server. `dateFormatShort(transferDate, locale)` means a `fr` author bakes `15/02/2020` into a record an `en-US` reader later sees — and for any day ≤ 12, `2/15` and `15/2` are indistinguishable.

Financial found the cheapest fix, which closes three findings at once: **drop `{{date}}` from the default note.** The date is already carried as its own unambiguous column in all three consumers (the history grid, the print view, and `downloadTable.tsx:47-49`), so the note adds no information — and removing it also eliminates the entire date-change note-regeneration path.

```ts
return t('{{lastName}} transfer to {{transferTo}}', { lastName, transferTo });
```

If the date should stay for product reasons, use two separate keys (with/without date) rather than `.trim()`, and treat the locale-in-a-stored-record question as a deliberate decision.

src/components/HrTools/Shared/CalculationReports/NameDisplay/NameDisplay.test.tsx
```diff
@@ -54,69 +47,41 @@ describe('NameDisplay', () => {
...
           spouseHcmData: null,
         }}
       />,
     );
 
     expect(getByText('Doe, John')).toBeInTheDocument();
-    expect(getByText('Person Number: 000123456')).toBeInTheDocument();
   });
 
   it('renders correctly for a married person', () => {
```
kegrimes: **[Suggestion]** The married test no longer tests anything married-specific.

This case existed to cover the `personNumberCount > 1` pluralization branch (`'Person Numbers: …'`). With that branch deleted, its only remaining assertion is that a string prop renders — which the single-person test above already covers.

`NameDisplay` reads no context (the final file has no `useContext`), so `isMarried: true` and `spousePreferredName: 'Jane'` are inert. Telling detail: this test also quietly dropped `spouseHcmData` while the others kept it, which is itself evidence the object means nothing.

Relatedly (3.0): the whole `MinisterHousingAllowanceContext.Provider` wrapper in this file is now dead scaffolding. It forces an `as ContextType` cast and couples a `Shared/` component's test to an unrelated feature's context type — so an MHA refactor will break this file for no reason. Worth deleting the provider and the `contextValue` prop entirely.

Either delete this test or give it real married coverage through the one married-aware surface left, `spouseComponent`:

```tsx
it('renders both names alongside the spouse component', () => {
  const { getByText, getByTestId } = render(
    <TestComponent
      names="Doe, John and Jane"
      spouseComponent={<div data-testid="spouse-component">Spouse Info</div>}
    />,
  );

  expect(getByText('Doe, John and Jane')).toBeInTheDocument();
  expect(getByTestId('spouse-component')).toBeInTheDocument();
});
```

src/components/HrTools/MinisterHousingAllowance/Steps/StepThree/CalcComponents/PersonInfo.tsx
```diff
@@ -60,8 +60,8 @@ export const PersonInfo: React.FC = () => {
           }}
         >
           <Typography>
-            {t('Email: {{emailAddress}}', {
-              emailAddress: person?.emailAddress,
+            {t('Phone number: {{primaryPhoneNumber}}', {
```
kegrimes, line 63: **[Suggestion]** The relocated Phone cell splits the postal address in two on mobile.

After the shuffle, the grid is: container A = `[Name][Email]`, container B = `[street address][Phone number]`, container C = `[city, state zip]` alone. At `sm+` that reads fine in two columns. At `xs` every cell is full width, so the order becomes Name → Email → `100 Lake Hart Dr` → **Phone number** → `Orlando, FL 32832` — one address interrupted mid-block, which also affects screen-reader reading order.

Not a regression (the same slot previously held Email), but the PR left container C with a free half-cell, so it's now a two-line fix:

```tsx
<Grid container spacing={2}>
  <Grid size={{ xs: 12, sm: 6 }}>
    {person?.addressLine2 === null ? (
      <Typography>{person?.addressLine1}</Typography>
    ) : (
      <Typography>{person?.addressLine1}, {person?.addressLine2}</Typography>
    )}
    <Typography>
      {person?.city}, {person?.state} {person?.zipCode}
    </Typography>
  </Grid>
  <Grid size={{ xs: 12, sm: 6 }}>
    <Typography>
      {t('Phone number: {{primaryPhoneNumber}}', {
        primaryPhoneNumber: person?.primaryPhoneNumber,
      })}
    </Typography>
  </Grid>
</Grid>
```

(deleting the now-empty third `<Grid container>`)

The swap itself is correct — `emailAddress` and `primaryPhoneNumber` each render exactly once, nothing duplicated or dropped.

src/components/Shared/ConfirmUserGroupModal/Helper/getUserType.ts
```diff
@@ -7,12 +7,35 @@ export const getUserType = (
...
-      return t('Cru Global Staff');
+      return {
+        label: t('Cru Global Staff'),
+        sublabel: t(
+          'Users in this group receive (mostly) non-US donations and are paid through our Global NetSuite system.',
+        ),
+      };
     case UserTypeEnum.NonCru:
-      return t('Non Cru User');
+      return {
```
kegrimes, line 24: **[Suggestion]** The `NonCru` label is computed but can never render — and a test asserts it.

`ConfirmUserGroupModal.tsx:80-91` short-circuits `NonCru` with its own copy (*"We see you're not on staff with Cru."*) and only reaches `<Typography>{label}</Typography>` in the `else` branch. So `t('Non Cru User')` is unreachable, while `getUserType.test.ts:23-27` asserts `result.label === 'Non Cru User'` — green-test confidence in a string no user will ever see.

The knowledge of "what copy does each group get" is now split: the helper owns three of four labels, the component hard-codes the fourth. A translator updating "Non Cru User" would see a passing test and reasonably assume the UI changed.

Either move the NonCru copy into the helper so all four branches live in one place, or drop the dead assertion and add a comment noting the label is intentionally unused.

src/components/HrTools/Shared/AccountInfoBox/AccountInfoBox.test.tsx
```diff
@@ -5,51 +5,28 @@ import { AccountInfoBox } from './AccountInfoBox';
 describe('AccountInfoBox', () => {
   it('renders name, accountId, and overallBalance when provided', () => {
```
kegrimes: **[Suggestion]** Stale test title — flagged independently by three agents.

The title still says `accountId`, but the body no longer passes one and the prop no longer exists. The three sibling titles in this `describe` were all correctly updated in this diff, so this one is just a leftover.

```diff
-  it('renders name, accountId, and overallBalance when provided', () => {
+  it('renders name and overallBalance when provided', () => {
```

Minor: line 15's title says "renders empty strings when name and overallBalance are not provided", but it asserts `name` is `''` and `overall-balance` is *absent* — "omits the balance" would be more accurate.

Everything else in this file checked out: skeleton/component parity was verified exact in both `hasOverallBalance` states, so there's no load-time layout shift. Removing the `InfoOutlined` tooltip was also correct — it explained only the number that's now gone, and dropping its `tabIndex={0}` removes a non-interactive tab stop.

src/components/HrTools/Shared/CalculationReports/NameDisplay/NameDisplay.tsx
```diff
@@ -25,33 +22,19 @@ interface NameDisplayProps {
...
   titleTwo,
   amountOne,
   amountTwo,
   spouseComponent,
 }) => {
-  const { t } = useTranslation();
   const locale = useLocale();
   const theme = useTheme();
   const currency = 'USD';
   const title = (
```
kegrimes, line 35: **[Suggestion]** Header row is now lopsided without `alignItems`.

This flex Box has `justifyContent: 'space-between'` but no `alignItems`, so children default to `stretch`. The left child used to be a two-line column (`h6` + person-number `subtitle1`); it's now a single `<Typography variant="h6">`. The right child in the four AdditionalSalaryRequest call sites is `SpouseComponent`, which still renders two stacked blocks (icon+link row, then a `caption` line).

The name doesn't move — it stays top-left — but a ~32px title now sits beside a ~44px block, both top-aligned, where they previously matched heights.

```diff
-<Box sx={{ display: 'flex', justifyContent: 'space-between' }}>
+<Box sx={{ display: 'flex', justifyContent: 'space-between', alignItems: 'center' }}>
```

`NewSalaryCalculatorLanding` passes no `spouseComponent`, so it's unaffected either way.

kegrimes (commented):

**No blockers.** 5 specialized agents + 5 gap-review agents + dependency analysis, followed by a cross-examination debate and a rebuttal round.

The debate round did real work here, in both directions:

- It **lowered** the headline finding from 9.0 "BLOCKING" to 7.5. A round-1 probe claimed the auto-note feature destroys an *already-saved* transfer description. Three agents independently traced the open-path and disproved it — the probe's author retracted it.
- It **surfaced a new confirmed regression** the first pass missed entirely (the stranded Submit button), reproduced by four independent probes.
- It **collapsed three competing fixes into one** for the `lastName` finding, after two agents withdrew their own proposals.

**Tooling, run by agents against this branch:** `yarn lint:ts` exit 0 · eslint 0 errors on all 47 changed files · ~500 tests green · zero missed consumers for every prop removal and the `getUserType` return-type change.

**Risk 10/10 CRITICAL** — read this honestly: the score is driven by breadth (23 source files in the medium-risk tier), not danger. No auth, API, schema, or dependency changes.

---

### On your two `/dismiss` replies

**`useFormUserInfo.ts` (dead `email`, sev 3.0)** — accepted. Not re-raised.

**`TransfersPage.tsx:88` (sev 7.5)** — could not be auto-dismissed (the threshold is <7.0), but **your reasoning was verified correct.** Two agents independently confirmed HCM is genuinely unreachable in the SavingsFundTransfer tree. Your objection eliminated the wrong fix and the agents converged on a better one: `useGetUserQuery()` → `user.lastName`, which is already in flight app-wide via `UserPreferenceProvider` (`_app.page.tsx:65`), so it's a free cache read with no parsing at all.

---

Informational — these cannot be posted as line comments and do not count toward the verdict.

**[Medium] Orphaned `personNumber` GraphQL selections — 6.0/10**
- `NsoMpdQuestionnaire/Shared/NewStaffQuestionnaire.graphql:8`
- `NsGoalCalculator/GoalSettings/NewStaffGoalCalculation.graphql:55`
- `NsGoalCalculator/GoalSettings/goalSettingsFormValues.ts:29` + `GoalSettingsForm.tsx:98,107`

Their only readers were deleted by this PR, so the identifier still ships to the client and persists into `localStorage` via `persistCache` (`src/lib/apollo/client.ts:21-26`). Standards traced `goalSettingsApiMapping.ts` and confirmed deletion is safe — `personNumber` appears there only inside a JSDoc comment. Run `yarn gql` after.

Scope note: `MinisterHousingAllowance.graphql:27,41,60` and `AdditionalSalaryRequest.graphql:14` were verified **already orphaned before this PR** (their apparent readers read `Hcm.graphql:7`, a different operation) — out of scope. `Hcm.graphql:7` must stay; it still feeds `orientSalaryRequest`.

**[Suggestion] i18n keys — 3.5/10** — Standards resolved the repo convention: `crowdin.yml` designates `public/locales/en/translation.json` as the *source* file, so the bot writes only non-en targets — the en delta is the author's responsibility, and CI never fails on drift. **Do not run `yarn extract`** (main carries ~248 keys of pre-existing drift). Hand-edit 7 keys: add `"{{lastName}} transfer to {{transferTo}} {{date}}"`, `"Cru Hybrid Staff"`, and the hybrid sublabel; delete `"Person number information"`, `"Person Number: {{personNumbers}}"`, `"Person Numbers: {{personNumbers}}"`, `"Staff Account Number: {{personNumber}}"`. **Keep** the bare `"Person Number"` and `"Staff Account Number"` — still used by MpdSupervisorReport, matching your intentional carve-out.

**[Suggestion] Test-fixture rot — 2.0/10** — `personNumber` mock values left behind after every assertion reading them was deleted: `AboutForm.test.tsx:43,53`, `useFormUserInfo.test.ts:28`, `FormVersions/testUtils.tsx:21`, `NsoMpdQuestionnaireTestWrapper.tsx:22,25`.

**[Suggestion] `PendingSalaryCalculationLanding.test.tsx:21` — 3.0/10** `[Pre-existing]` — titled `'displays NameDisplay with staff info and balances'`, but this PR deleted that component from the file under test. The test now asserts only `SalaryInformationCard` output. The test file isn't in this PR.

---

**MHA blank panel — 4.5/10** — `!hasNoRequests && !eitherPersonEligible` renders only the `Your MHA` heading. Reachable (also transiently, whenever the requests query resolves before the HCM query). Pre-PR this state showed an empty name card rather than an empty panel, so the PR aggravated rather than caused it. Follow-up, not a merge gate.

---

### Working tree

Agents ran probe tests and one temporarily patched `TransferModal.tsx` to verify a fix. All were reverted — `git status --porcelain` confirmed empty.

> ⚠️ **2 Important findings (7.0–7.9).** Strongly recommended, don't block merge, and can't be `/dismiss`ed. Address here or file follow-ups.

src/components/HrTools/Shared/StaffInfoCard/StaffInfoCard.tsx
```diff
@@ -48,18 +46,9 @@ export const StaffInfoCard: React.FC<StaffInfoCardProps> = ({
...
-              </Typography>
-              {person.staffAccountId && (
-                <Typography color="text.secondary" lineHeight={1.2}>
-                  {t('Person Number: {{number}}', {
-                    number: person.staffAccountId,
-                  })}
-                </Typography>
-              )}
-            </Stack>
+            <Typography variant="h6" lineHeight={1.2}>
```
kegrimes: **[Suggestion]** Vestigial `lineHeight` override.

`lineHeight={1.2}` existed to tighten the gap between the two stacked lines inside the removed `<Stack>`. This `Typography` is now the only text node in a `<Stack direction="row" alignItems="center">` whose height is set by the `theme.spacing(5)` Avatar, so the override has no visual effect — it's a magic number where the theme token would otherwise apply.

```diff
-<Typography variant="h6" lineHeight={1.2}>
-  {person.name}
-</Typography>
+<Typography variant="h6">{person.name}</Typography>
```

The rest of this file is clean — `useTranslation` is correctly retained (still used at `:55` for `t('View {{name}}')`), and collapsing the Stack actually improves vertical centering against the avatar.

src/components/HrTools/NsGoalCalculator/GoalSettings/GoalSettingsHeader.tsx
kegrimes: /dismiss leaving this alone for now

src/components/HrTools/NsoMpdQuestionnaire/PersonalInformation/StaffInformation.tsx
kegrimes: /dismiss leaving this alone for now

src/components/HrTools/MinisterHousingAllowance/MinisterHousingAllowance.tsx
kegrimes: /dismiss

src/components/HrTools/SavingsFundTransfer/TransferModal/TransferModal.tsx
kegrimes: /dismiss this report is US staff only

src/components/HrTools/MinisterHousingAllowance/Steps/StepThree/CalcComponents/PersonInfo.tsx
kegrimes: /dismiss not worried about in this PR

wjames111 (approved):
I skimmed but looks good!

PR #1784, opened 2026-05-19
[MPDX-9561] - Submit button on PDS Goal Calculator final step

wjames111:
## Description

- Adds a "Finish & Apply Goal" submit button on the last step of the PDS Goal Calculator that saves the calculated overall total as the account's `monthlyGoal` via `updateAccountPreferences`.
- On success, surfaces a localized snackbar showing the formatted goal that was applied.
- Disables the submit action while the mutation is in flight and when required fields are incomplete.
- Extends `DirectionButtons` with a `disabledNextTooltip` prop so the final step can show "Complete all required fields to submit" instead of the generic "...to continue" copy.

Work for [MPDX-9561](https://jira.cru.org/browse/MPDX-9561).

## Testing

- Go to the PDS Goal Calculator (HR Tools)
- Step through each section, filling in required fields
- On any intermediate step, hover the disabled Next button with incomplete fields and confirm the tooltip reads "Complete all required fields to continue"
- Advance to the final step
- With required fields incomplete, hover the disabled submit button and confirm the tooltip reads "Complete all required fields to submit"
- Complete all required fields and confirm the button label reads "Finish & Apply Goal"
- Click "Finish & Apply Goal" and confirm:
  - The button is disabled while the request is in flight
  - A success snackbar appears with the formatted monthly goal (e.g. "Successfully updated your monthly goal to $X,XXX!")
  - The account's monthly goal is updated (verify on the dashboard / preferences)

## Checklist:

- [x] I have given my PR a title with the format "MPDX-(JIRA#) (summary sentence max 80 chars)"
- [x] I have applied the appropriate labels (_Add the label "Preview" to automatically create a preview environment_)
- [x] I have run the Claude Code `/pr-review` command locally and fixed any relevant suggestions
- [ ] I have requested a review from another person on the project
- [x] I have tested my changes in preview or in staging
- [ ] I have cleaned up my commit history

PR #1965, opened 2026-08-04
[MPDX-9909] Add loading spinners to slow request forms mutations

canac:
## Description

Add loading spinners to slow request forms mutations that load HCM data on the backend or call out to SAA on the backend. This improves the UX by giving the user feedback that their request is in progress.

MPDX-9909

## Testing

- Salary Calc
  - Go to HR Tools > Salary Calculator with no request in progress
  - Click "Calculate New Salary"
  - Check that the button shows a spinner and is disabled until the form opens, and that clicking it repeatedly does not create multiple requests
  - Fill out the form through step 4 and click "Submit"
  - Check that the confirm button ("Yes, Continue") shows a spinner and is disabled while submitting, that "GO BACK" is disabled, and that the modal cannot be dismissed by clicking the backdrop
  - Check that the "Submit" button behind the modal does not show a spinner
  - Start another calculation, click "Discard", then confirm
  - Check that the modal's confirm button spins and disables the same way
  - Go back to the Salary Calculator landing with a pending request
  - Click the trash icon, then confirm the cancel
  - Check that the modal's confirm button spins and disables (the trash icon itself stays unchanged)

- ASR
  - Go to HR Tools > Additional Salary Request with no open request
  - On "About this Form", click "Continue"
  - Check that the Continue button shows a spinner and is disabled while the request is created, and that the page no longer blanks out to a full-page loading spinner
  - Complete the form, click "Submit", and confirm
  - Check that the modal's confirm button spins and disables, and that the "Submit" button behind it does not
  - From the pending-request card, click "Cancel Request" and confirm
  - Check that the modal's confirm button spins and disables, and that "Cancel Request" behind it does not

- MHA
  - Go to HR Tools > MHA Calculator with no blocking request
  - Click "Request New MHA"
  - Check that the button shows a spinner and is disabled while the request is created, and that the page no longer blanks out to a full-page loading spinner
  - Complete the form, click "Submit", and confirm
  - Check that the modal's confirm button spins and disables
  - From the pending-request card, click "Cancel Request" and confirm
  - Check that the modal's confirm button spins and disables

## Checklist:

### DIFF
--- a/.claude/rules/code-review.md
+++ b/.claude/rules/code-review.md
@@ -221,11 +221,9 @@ Every item here is mandatory. The Standards agent must report compliance per ite
 **Code Quality**
 
 - [ ] Passes `yarn lint` and `yarn lint:ts`
-- [ ] No debug output (`console.log`, `console.debug`, `debugger`, `// TODO` without a Jira/MPDX ticket reference)
+- [ ] No `// TODO` without a Jira/MPDX ticket reference
 - [ ] No `new Date()` — use Luxon (`DateTime.now()`, `DateTime.local()`) per project convention
-- [ ] No unused imports or variables
 - [ ] No commented-out code blocks (delete, don't comment)
-- [ ] No empty `catch {}` blocks that swallow errors silently
 
 ## Security Focus Areas
 
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -67,6 +67,8 @@ module.exports = {
     curly: 'error',
     eqeqeq: 'error',
     'no-console': 'error',
+    'no-debugger': 'error',
+    'no-empty': 'error',
     '@typescript-eslint/no-loss-of-precision': 'warn',
     '@typescript-eslint/no-unused-vars': [
       'error',
--- a/src/components/HrTools/AdditionalSalaryRequest/RequestPage/RequestPage.test.tsx
+++ b/src/components/HrTools/AdditionalSalaryRequest/RequestPage/RequestPage.test.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { ThemeProvider } from '@mui/material/styles';
-import { render, waitFor } from '@testing-library/react';
+import { render, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { FormikProvider, useFormik } from 'formik';
 import { SnackbarProvider } from 'notistack';
@@ -675,17 +675,24 @@ describe('RequestPage', () => {
     ).not.toBeInTheDocument();
   });
 
-  it('shows loading indicator when creating a new request', async () => {
+  it('disables the continue button and shows a spinner while creating a new request', async () => {
     mockUseAdditionalSalaryRequest.mockReturnValue({
       ...defaultMockContextValue,
       requestData: { latestAdditionalSalaryRequest: null },
     } as unknown as ReturnType<typeof useAdditionalSalaryRequest>);
 
-    const { getByRole, findByTestId } = render(<TestWrapper />);
+    const { getByRole } = render(<TestWrapper />);
 
-    userEvent.click(getByRole('button', { name: /continue/i }));
+    const continueButton = getByRole('button', { name: /continue/i });
+    userEvent.click(continueButton);
 
-    expect(await findByTestId('Loading')).toBeInTheDocument();
+    await waitFor(() => {
+      expect(continueButton).toBeDisabled();
+      expect(continueButton).toHaveAttribute('aria-busy', 'true');
+      expect(
+        within(continueButton).getByRole('progressbar'),
+      ).toBeInTheDocument();
+    });
   });
 
   it('hides back href on About this Form step in New mode', () => {
--- a/src/components/HrTools/AdditionalSalaryRequest/RequestPage/RequestPage.tsx
+++ b/src/components/HrTools/AdditionalSalaryRequest/RequestPage/RequestPage.tsx
@@ -116,7 +116,7 @@ const MainContent: React.FC = () => {
     t,
   );
 
-  if (creatingRequest || (loading && !requestData)) {
+  if (loading && !requestData) {
     return <Loading loading />;
   }
 
@@ -175,6 +175,7 @@ const MainContent: React.FC = () => {
                   (exceedsCap && !!errors.additionalInfo)
                 }
                 overrideNext={isFirstFormPage ? handleContinue : undefined}
+                loadingNext={creatingRequest}
               />
             </Stack>
           )}
--- a/src/components/HrTools/SalaryCalculator/Landing/NewSalaryCalculationLanding/NewSalaryCalculatorLanding.tsx
+++ b/src/components/HrTools/SalaryCalculator/Landing/NewSalaryCalculationLanding/NewSalaryCalculatorLanding.tsx
@@ -1,7 +1,14 @@
 import Link from 'next/link';
 import { useRouter } from 'next/router';
 import React from 'react';
-import { Box, Button, Container, Typography, useTheme } from '@mui/material';
+import {
+  Box,
+  Button,
+  CircularProgress,
+  Container,
+  Typography,
+  useTheme,
+} from '@mui/material';
 import { styled } from '@mui/material/styles';
 import { Trans, useTranslation } from 'react-i18next';
 import { NameDisplay } from 'src/components/HrTools/Shared/CalculationReports/NameDisplay/NameDisplay';
@@ -133,6 +140,12 @@ export const NewSalaryCalculatorLanding: React.FC = () => {
                 color="primary"
                 onClick={handleStartCalculation}
                 disabled={creatingCalculation}
+                aria-busy={creatingCalculation}
+                startIcon={
+                  creatingCalculation ? (
+                    <CircularProgress size={20} color="inherit" />
+                  ) : undefined
+                }
               >
                 {t('Calculate New Salary')}
               </Button>
--- a/src/components/HrTools/SalaryCalculator/Landing/PendingSalaryCalculationLanding/components/PendingRequestActions.test.tsx
+++ b/src/components/HrTools/SalaryCalculator/Landing/PendingSalaryCalculationLanding/components/PendingRequestActions.test.tsx
@@ -1,4 +1,4 @@
-import { render, waitFor } from '@testing-library/react';
+import { render, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { SalaryRequestStatusEnum } from 'src/graphql/types.generated';
 import { LandingTestWrapper } from '../../NewSalaryCalculationLanding/LandingTestWrapper';
@@ -79,6 +79,27 @@ describe('PendingRequestActions', () => {
     userEvent.click(await findByRole('button', { name: 'Delete request' }));
 
     userEvent.click(await findByText('Yes, Cancel'));
+    await waitFor(() =>
+      expect(mutationSpy).toHaveGraphqlOperation('DeleteSalaryCalculation'),
+    );
+  });
+
+  it('disables the cancel buttons and shows spinners while cancelling', async () => {
+    const { findByRole, getByRole } = render(
+      <TestComponent calculation={mockCalculation} onCall={mutationSpy} />,
+    );
+
+    const deleteButton = await findByRole('button', { name: 'Delete request' });
+    userEvent.click(deleteButton);
+
+    const confirmButton = await findByRole('button', { name: /yes, cancel/i });
+    userEvent.click(confirmButton);
+
+    expect(confirmButton).toBeDisabled();
+    expect(confirmButton).toHaveAttribute('aria-busy', 'true');
+    expect(within(confirmButton).getByRole('progressbar')).toBeInTheDocument();
+    expect(getByRole('button', { name: 'GO BACK' })).toBeDisabled();
+
     await waitFor(() =>
       expect(mutationSpy).toHaveGraphqlOperation('DeleteSalaryCalculation'),
     );
--- a/src/components/HrTools/SalaryCalculator/Landing/PendingSalaryCalculationLanding/components/PendingRequestActions.tsx
+++ b/src/components/HrTools/SalaryCalculator/Landing/PendingSalaryCalculationLanding/components/PendingRequestActions.tsx
@@ -19,7 +19,7 @@ export const PendingRequestActions: React.FC<PendingRequestActionsProps> = ({
 }) => {
   const { t } = useTranslation();
   const accountListId = useAccountListId();
-  const { deleteSalaryCalculation } = useDeleteSalaryCalculation();
+  const { deleteSalaryCalculation, deleting } = useDeleteSalaryCalculation();
   const [removeDialogOpen, setRemoveDialogOpen] = useState(false);
 
   const handleDelete = async () => {
@@ -67,6 +67,7 @@ export const PendingRequestActions: React.FC<PendingRequestActionsProps> = ({
           handleClose={() => setRemoveDialogOpen(false)}
           handleConfirm={handleDelete}
           isCancel
+          submitting={deleting}
         />
       )}
       <IconButton
--- a/src/components/HrTools/SalaryCalculator/Shared/useDeleteSalaryCalculation.ts
+++ b/src/components/HrTools/SalaryCalculator/Shared/useDeleteSalaryCalculation.ts
@@ -3,11 +3,12 @@ import { useDeleteSalaryCalculationMutation } from './DeleteSalaryCalculation.ge
 
 interface UseDeleteSalaryCalculationReturn {
   deleteSalaryCalculation: (calculationId: string) => Promise<void>;
+  deleting: boolean;
 }
 
 export const useDeleteSalaryCalculation =
   (): UseDeleteSalaryCalculationReturn => {
-    const [deleteSalaryCalculationMutation] =
+    const [deleteSalaryCalculationMutation, { loading: deleting }] =
       useDeleteSalaryCalculationMutation();
 
     const deleteSalaryCalculation = async (
@@ -20,8 +21,9 @@ export const useDeleteSalaryCalculation =
           },
         },
         refetchQueries: [LandingSalaryCalculationsDocument],
+        awaitRefetchQueries: true,
       });
     };
 
-    return { deleteSalaryCalculation };
+    return { deleteSalaryCalculation, deleting };
   };
--- a/src/components/HrTools/SalaryCalculator/StepNavigation/StepNavigation.tsx
+++ b/src/components/HrTools/SalaryCalculator/StepNavigation/StepNavigation.tsx
@@ -18,7 +18,7 @@ export const DiscardButton: React.FC<ButtonProps> = (props) => {
   const { push } = useRouter();
   const accountListId = useAccountListId();
   const { calculation } = useSalaryCalculator();
-  const { deleteSalaryCalculation } = useDeleteSalaryCalculation();
+  const { deleteSalaryCalculation, deleting } = useDeleteSalaryCalculation();
   const [removeDialogOpen, setRemoveDialogOpen] = useState(false);
 
   const handleDiscard = async () => {
@@ -48,6 +48,7 @@ export const DiscardButton: React.FC<ButtonProps> = (props) => {
           handleClose={() => setRemoveDialogOpen(false)}
           handleConfirm={handleDiscard}
           isDiscard
+          submitting={deleting}
         />
       )}
     </>
@@ -92,7 +93,8 @@ export const ContinueButton: React.FC<ButtonProps> = (props) => {
 export const SubmitButton: React.FC<ButtonProps> = (props) => {
   const { t } = useTranslation();
   const { handleNextStep, calculation } = useSalaryCalculator();
-  const [submit] = useSubmitSalaryCalculationMutation();
+  const [submit, { loading: submitting }] =
+    useSubmitSalaryCalculationMutation();
   const [submitDialogOpen, setSubmitDialogOpen] = useState(false);
   const { title, content, subContent } = useSubmitDialogContent();
 
@@ -132,6 +134,7 @@ export const SubmitButton: React.FC<ButtonProps> = (props) => {
           overrideTitle={title}
           overrideContent={content}
           overrideSubContent={subContent}
+          submitting={submitting}
         />
       )}
     </>
--- a/src/components/HrTools/Shared/CalculationReports/DirectionButtons/DirectionButtons.tsx
+++ b/src/components/HrTools/Shared/CalculationReports/DirectionButtons/DirectionButtons.tsx
@@ -161,12 +161,15 @@ export const DirectionButtons: React.FC<DirectionButtonsProps> = ({
                 <Button
                   variant="contained"
                   color="primary"
-                  endIcon={loadingNext ? undefined : <ChevronRight />}
+                  endIcon={<ChevronRight />}
                   startIcon={
-                    loadingNext ? <CircularProgress size={20} /> : undefined
+                    loadingNext ? (
+                      <CircularProgress size={20} color="inherit" />
+                    ) : undefined
                   }
                   onClick={overrideNext ?? handleNextStep}
                   disabled={disableNext || loadingNext}
+                  aria-busy={loadingNext}
                 >
                   {loadingNext
                     ? (loadingNextTitle ?? buttonTitle ?? t('Continue'))
--- a/src/components/HrTools/Shared/CalculationReports/StatusCard/StatusCard.test.tsx
+++ b/src/components/HrTools/Shared/CalculationReports/StatusCard/StatusCard.test.tsx
@@ -2,7 +2,7 @@ import { ThemeProvider } from '@emotion/react';
 import { SvgIconProps } from '@mui/material/SvgIcon';
 import { LocalizationProvider } from '@mui/x-date-pickers';
 import { AdapterLuxon } from '@mui/x-date-pickers/AdapterLuxon';
-import { render } from '@testing-library/react';
+import { render, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { SnackbarProvider } from 'notistack';
 import TestRouter from '__tests__/util/TestRouter';
@@ -158,7 +158,7 @@ describe('CardSkeleton', () => {
 
     userEvent.click(getByRole('button', { name: /yes, cancel/i }));
 
-    expect(queryByRole('dialog')).not.toBeInTheDocument();
+    await waitFor(() => expect(queryByRole('dialog')).not.toBeInTheDocument());
   });
 
   it('hides print icon when hidePrint is true', () => {
@@ -201,7 +201,34 @@ describe('CardSkeleton', () => {
 
     expect(handleConfirmCancel).toHaveBeenCalled();
 
-    expect(queryByRole('dialog')).not.toBeInTheDocument();
+    await waitFor(() => expect(queryByRole('dialog')).not.toBeInTheDocument());
+  });
+
+  it('disables the modal buttons and shows a spinner while cancelling', async () => {
+    let resolveCancel: () => void = () => {};
+    handleConfirmCancel.mockImplementationOnce(
+      () =>
+        new Promise<void>((resolve) => {
+          resolveCancel = resolve;
+        }),
+    );
+
+    const { getByRole, findByRole, queryByRole } = render(
+      <TestComponent isRequest={true} />,
+    );
+
+    userEvent.click(await findByRole('button', { name: 'Cancel Request' }));
+
+    const confirmButton = await findByRole('button', { name: /yes, cancel/i });
+    userEvent.click(confirmButton);
+
+    await waitFor(() => expect(confirmButton).toBeDisabled());
+    expect(confirmButton).toHaveAttribute('aria-busy', 'true');
+    expect(within(confirmButton).getByRole('progressbar')).toBeInTheDocument();
+    expect(getByRole('button', { name: 'GO BACK' })).toBeDisabled();
+
+    resolveCancel();
+    await waitFor(() => expect(queryByRole('dialog')).not.toBeInTheDocument());
   });
 
   it('should hide second button when hideLinkTwoButton is true', () => {
--- a/src/components/HrTools/Shared/CalculationReports/StatusCard/StatusCard.tsx
+++ b/src/components/HrTools/Shared/CalculationReports/StatusCard/StatusCard.tsx
@@ -36,7 +36,7 @@ interface StatusCardProps {
   hideActions?: boolean;
   hideLinkTwoButton?: boolean;
   handlePrint?: () => void;
-  handleConfirmCancel: () => void;
+  handleConfirmCancel?: () => Promise<void>;
   handleLinkTwo?: () => void;
   styling?: SxProps<Theme>;
 }
@@ -66,10 +66,17 @@ export const StatusCard: React.FC<StatusCardProps> = ({
   const { t } = useTranslation();
 
   const [openCancel, setOpenCancel] = useState(false);
+  const [cancelling, setCancelling] = useState(false);
 
-  const handleCancel = () => {
-    setOpenCancel(false);
-    handleConfirmCancel();
+  const handleCancel = async () => {
+    setCancelling(true);
+    try {
+      await handleConfirmCancel?.();
+      setOpenCancel(false);
+    } catch {
+    } finally {
+      setCancelling(false);
+    }
   };
 
   return (
@@ -168,6 +175,7 @@ export const StatusCard: React.FC<StatusCardProps> = ({
               handleClose={() => setOpenCancel(false)}
               handleConfirm={handleCancel}
               isCancel={true}
+              submitting={cancelling}
             />
           )}
         </Box>
--- a/src/components/HrTools/Shared/CalculationReports/StatusCard/StatusCard.tsx
+++ b/src/components/HrTools/Shared/CalculationReports/StatusCard/StatusCard.tsx
@@ -74,6 +74,7 @@ export const StatusCard: React.FC<StatusCardProps> = ({
       await handleConfirmCancel?.();
       setOpenCancel(false);
     } catch {
+      // The Apollo error link already notifies the user of the failure
     } finally {
       setCancelling(false);
     }
--- a/src/components/HrTools/Shared/CalculationReports/SubmitModal/SubmitModal.tsx
+++ b/src/components/HrTools/Shared/CalculationReports/SubmitModal/SubmitModal.tsx
@@ -122,11 +122,11 @@ export const SubmitModal: React.FC<SubmitModalProps> = ({
             startIcon={
               submitting ? <CircularProgress size={20} color="inherit" /> : null
             }
+            endIcon={<ChevronRight />}
           >
             <b>
               {additionalApproval ? t('Submit For Approval') : cancelButtonText}
             </b>
-            {!submitting && <ChevronRight sx={{ ml: 1 }} />}
           </Button>
         )}
       </DialogActions>
--- a/src/components/HrTools/MinisterHousingAllowance/SharedComponents/CurrentBoardApproved.tsx
+++ b/src/components/HrTools/MinisterHousingAllowance/SharedComponents/CurrentBoardApproved.tsx
@@ -103,7 +103,6 @@ export const CurrentBoardApproved: React.FC<CurrentBoardApprovedProps> = ({
       hideLinkTwoButton={hasOpenRequest}
       isRequest={false}
       handlePrint={handlePrint}
-      handleConfirmCancel={() => {}}
       styling={{ p: 0 }}
     >
       <TableContainer sx={{ padding: 0 }}>
--- a/src/components/HrTools/MinisterHousingAllowance/Steps/StepTwo/RentOwn.tsx
+++ b/src/components/HrTools/MinisterHousingAllowance/Steps/StepTwo/RentOwn.tsx
@@ -71,7 +71,9 @@ export const RentOwn: React.FC = () => {
         variant: 'success',
       });
       setHasCalcValues(false);
-    } catch (error) {}
+    } catch {
+      // The Apollo error link already notifies the user of the failure
+    }
   };
 
   const [pendingValue, setPendingValue] = useState<MhaRentOrOwnEnum | null>(
--- a/src/components/HrTools/SalaryCalculator/Landing/NewSalaryCalculationLanding/NewSalaryCalculatorLanding.test.tsx
+++ b/src/components/HrTools/SalaryCalculator/Landing/NewSalaryCalculationLanding/NewSalaryCalculatorLanding.test.tsx
@@ -1,4 +1,4 @@
-import { render, waitFor } from '@testing-library/react';
+import { render, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import {
   LandingTestWrapper,
@@ -154,7 +154,7 @@ describe('NewSalaryCalculatorLanding', () => {
     });
   });
 
-  it('disables button while mutation is in progress', async () => {
+  it('disables button and shows a spinner while mutation is in progress', async () => {
     const { findByRole } = render(<TestComponent hasApprovedCalculation />);
 
     const button = await findByRole('button', {
@@ -164,6 +164,8 @@ describe('NewSalaryCalculatorLanding', () => {
     userEvent.click(button);
     await waitFor(() => {
       expect(button).toBeDisabled();
+      expect(button).toHaveAttribute('aria-busy', 'true');
+      expect(within(button).getByRole('progressbar')).toBeInTheDocument();
     });
   });
 });
--- a/src/components/HrTools/SalaryCalculator/StepNavigation/StepNavigation.test.tsx
+++ b/src/components/HrTools/SalaryCalculator/StepNavigation/StepNavigation.test.tsx
@@ -1,4 +1,4 @@
-import { render, waitFor } from '@testing-library/react';
+import { render, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { AutosaveForm } from 'src/components/Shared/Autosave/AutosaveForm';
 import {
@@ -88,6 +88,28 @@ describe('DiscardButton', () => {
       }),
     );
   });
+
+  it('disables the modal buttons and shows a spinner while deleting', async () => {
+    const { findByRole, getByRole } = render(
+      <SalaryCalculatorTestWrapper onCall={mutationSpy} editing={true}>
+        <DiscardButton />
+      </SalaryCalculatorTestWrapper>,
+    );
+
+    userEvent.click(await findByRole('button', { name: 'Discard' }));
+
+    const confirmButton = await findByRole('button', { name: /yes, discard/i });
+    userEvent.click(confirmButton);
+
+    expect(confirmButton).toBeDisabled();
+    expect(confirmButton).toHaveAttribute('aria-busy', 'true');
+    expect(within(confirmButton).getByRole('progressbar')).toBeInTheDocument();
+    expect(getByRole('button', { name: 'GO BACK' })).toBeDisabled();
+
+    await waitFor(() =>
+      expect(mutationSpy).toHaveGraphqlOperation('DeleteSalaryCalculation'),
+    );
+  });
 });
 
 describe('SubmitButton', () => {
@@ -105,6 +127,29 @@ describe('SubmitButton', () => {
       expect(mutationSpy).toHaveGraphqlOperation('SubmitSalaryCalculation', {
         input: { id: 'salary-request-1' },
       }),
+    );
+  });
+
+  it('disables the modal buttons and shows a spinner while submitting', async () => {
+    const { findByRole, getByRole } = render(
+      <SalaryCalculatorTestWrapper onCall={mutationSpy} editing={true}>
+        <SubmitButton />
+      </SalaryCalculatorTestWrapper>,
+    );
+
+    userEvent.click(await findByRole('button', { name: 'Submit' }));
+
+    const confirmButton = await findByRole('button', { name: 'Yes, Continue' });
+    userEvent.click(confirmButton);
+
+    expect(confirmButton).toBeDisabled();
+    expect(confirmButton).toHaveAttribute('aria-busy', 'true');
+    expect(within(confirmButton).getByRole('progressbar')).toBeInTheDocument();
+
+    expect(getByRole('button', { name: 'GO BACK' })).toBeDisabled();
+
+    await waitFor(() =>
+      expect(mutationSpy).toHaveGraphqlOperation('SubmitSalaryCalculation'),
     );
   });
 });
--- a/src/components/HrTools/Shared/CalculationReports/DirectionButtons/DirectionButtons.test.tsx
+++ b/src/components/HrTools/Shared/CalculationReports/DirectionButtons/DirectionButtons.test.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { ThemeProvider } from '@mui/material/styles';
-import { render, waitFor } from '@testing-library/react';
+import { render, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { Formik } from 'formik';
 import { SnackbarProvider } from 'notistack';
@@ -189,9 +189,8 @@ describe('DirectionButtons', () => {
 
     const savingButton = await findByRole('button', { name: 'Saving...' });
     expect(savingButton).toBeDisabled();
-    expect(
-      savingButton.querySelector('.MuiCircularProgress-root'),
-    ).toBeInTheDocument();
+    expect(savingButton).toHaveAttribute('aria-busy', 'true');
+    expect(within(savingButton).getByRole('progressbar')).toBeInTheDocument();
   });
 
   it('suppresses the disabledNextTooltip while loadingNext is true', async () => {


src/components/HrTools/Shared/CalculationReports/StatusCard/StatusCard.tsx
```diff
@@ -66,10 +66,17 @@ export const StatusCard: React.FC<StatusCardProps> = ({
...
 
-  const handleCancel = () => {
-    setOpenCancel(false);
-    handleConfirmCancel();
+  const handleCancel = async () => {
+    setCancelling(true);
+    try {
+      await handleConfirmCancel?.();
+      setOpenCancel(false);
+    } catch {
```
canac, line 76: **[Important]** Empty `catch {}` — flagged by **all 6 agents** (highest-consensus finding in the review).

Verbatim hit on the rules-file checklist: *"No empty `catch {}` blocks that swallow errors silently."* The Standards agent censused every bare `catch` in `src/` — 12 of them either have a body or an explanatory comment; the only comment-free empty one is `RentOwn.tsx:74`, a single instance, **below the 3-instance convention bar**. And `.eslintrc.js` extends no `eslint:recommended`, so core `no-empty` is off — CI will not catch this.

Two error classes, and they need different answers:

- **Apollo rejections** — already surfaced by the global error link (`src/lib/apollo/client.ts:57-65`), and both consumers reject rather than resolve. Swallowing is *correct* here; a manual toast would double-fire.
- **Everything else** — real, not hypothetical. Both consumers do post-success work **inside** the awaited chain (`AdditionalSalaryRequestContext.tsx:220-235` calls `resetSteps`/`setPageType`/`enqueueSnackbar`; MHA `CurrentRequest.tsx:74-78` enqueues a snackbar). A `TypeError` there is swallowed **after the record is already deleted server-side** — no toast, no console, nothing for Rollbar/DataDog.

What pushed this to 7.0: on retry the backend answers `"Approved requests cannot be deleted or canceled"` (`salary_request.rb:79-80` — `event :cancel` has no `from: :canceled`), which `client.ts:58` snackbars verbatim. A staff member who just cancelled a pending salary request reads that it was **approved**. It wasn't.

**Standards' ruling as checklist arbiter: a comment alone is NOT sufficient** — it satisfies the item for the Apollo class only. Compliant minimum:

```tsx
} catch (error) {
  // Keep the dialog open so the user can retry. GraphQL/network errors are
  // already surfaced by the global Apollo error link (src/lib/apollo/client.ts).
  if (!(error instanceof ApolloError)) {
    throw error;
  }
} finally {
```

The repo already has this comment idiom four lines from code this PR edits — `CurrentBoardApproved.tsx:79-80`.

Also add the rejecting-path test. It needs **zero plumbing** (`handleConfirmCancel` is a module-scoped `jest.fn()`, so `mockRejectedValueOnce` works today) and it's the only site where a stuck spinner is observable, because the dialog now deliberately stays open on failure:

```tsx
it('re-enables the modal buttons when cancelling fails', async () => {
  handleConfirmCancel.mockRejectedValueOnce(new Error('Server Error'));

  const { getByRole, findByRole } = render(<TestComponent isRequest={true} />);
  userEvent.click(await findByRole('button', { name: 'Cancel Request' }));

  const confirmButton = await findByRole('button', { name: /yes, cancel/i });
  userEvent.click(confirmButton);

  await waitFor(() => expect(confirmButton).not.toBeDisabled());
  expect(confirmButton).toHaveAttribute('aria-busy', 'false');
  // The dialog intentionally stays open so the user can retry.
  expect(getByRole('dialog')).toBeInTheDocument();
});
```

src/components/HrTools/MinisterHousingAllowance/MinisterHousingAllowance.test.tsx
```diff
@@ -282,10 +282,18 @@ describe('MinisterHousingAllowanceReport', () => {
...
+      name: 'Request New MHA',
+    });
     userEvent.click(requestButton);
 
-    expect(await findByTestId('Loading')).toBeInTheDocument();
+    await waitFor(() => {
+      expect(requestButton).toBeDisabled();
+      expect(requestButton).toHaveAttribute('aria-busy', 'true');
+      expect(
+        requestButton.querySelector('.MuiCircularProgress-root'),
```
canac: **[Important]** All 7 new spinner tests assert the busy state **appears**; none asserts it ever **goes away**.

Stuck spinner / permanently-disabled button is exactly this PR class's failure mode, and the repo's own reference implementation does test it — `CsvHeaders.test.tsx:422-427`, `CsvValues.test.tsx:286-289`, and `CsvPreview.test.tsx:317-320` each resolve a deferred save and assert `aria-busy` returns to `'false'`.

This file is the sharpest case. `MinisterHousingAllowance.tsx:75-103` uses hand-rolled `useState`, and the **only** path that resets `creatingRequest` is `onError` (line 93) — on success it deliberately stays `true` through navigation. Refactor `onError` away and the button locks forever with a green suite. This file currently has **zero** error-path tests.

The Financial agent's argument for promoting this to 7.0 is the one that convinced the panel: *the absence of this single assertion is the direct cause of three agents reaching three incompatible conclusions about the one window this PR exists to close.* It is also the cheapest finding in the review.

Two corrections from the debate, so nobody wastes a cycle:

1. It's **6 of 7**, not 7 of 7 — `StatusCard.test.tsx:204-231` does assert exit-by-unmount on success.
2. The `PendingRequestActions` site is **not testable as `aria-busy: false`** — its parent unmounts the spinner mid-refetch (`PendingSalaryCalculationLanding.tsx:22-24`). That one needs an unmount assertion instead.

This wrapper can't inject a mutation error yet — `TestComponentProps` exposes only `hcmMock`/`mhaRequestsMock` and the `CreateHousingAllowanceRequest` mock is hardcoded at :59-65. Add an `extraMocks` seam following the documented in-repo precedent at `PdsGoalCalculatorTestWrapper.tsx:147-158` (~6 lines), then:

```tsx
it('re-enables the request button when creation fails', async () => {
    <TestComponent
      hcmMock={singleMhaNoException}
      mhaRequestsMock={[{ ...mockMHARequest, status: MhaStatusEnum.HrApproved }]}
      extraMocks={{ CreateHousingAllowanceRequest: () => { throw new Error('Server Error'); } }}
    />,
  );

  const requestButton = await findByRole('button', { name: 'Request New MHA' });
  userEvent.click(requestButton);

  await waitFor(() => expect(requestButton).toBeDisabled());
  await waitFor(() => expect(requestButton).toBeEnabled());
  expect(requestButton).toHaveAttribute('aria-busy', 'false');
  expect(mockPush).not.toHaveBeenCalled();
});
```

src/components/HrTools/Shared/CalculationReports/StatusCard/StatusCard.tsx
```diff
@@ -36,7 +36,7 @@ interface StatusCardProps {
   hideActions?: boolean;
   hideLinkTwoButton?: boolean;
   handlePrint?: () => void;
-  handleConfirmCancel: () => void;
+  handleConfirmCancel?: () => Promise<void>;
```
canac, line 39: **[Medium]** Making `handleConfirmCancel` optional decouples it from `isRequest` — the flag that actually renders the cancel UI.

Nothing breaks today. All three consumers were enumerated and verified type-compatible (`ASR CurrentRequest.tsx:74` returns `Promise<void>` via `AdditionalSalaryRequestContext.tsx:103`; `MHA CurrentRequest.tsx:99` is `async`; `CurrentBoardApproved.tsx` passes `isRequest={false}` so the cancel branch never renders). And the removal from `CurrentBoardApproved` was **forced** — `() => {}` returns `void`, which isn't assignable to `() => Promise<void>`. That's the real reason the prop became optional.

The cost is a compile-time guarantee traded for a runtime silent failure, in a `Shared/` component. `await handleConfirmCancel?.()` **resolves** when the prop is absent, so `setOpenCancel(false)` runs on exactly the same path as a successful cancellation — a future `isRequest={true}` card that forgets the handler shows a Cancel button whose confirm spins, closes, and cancels nothing.

Compounding it (line 74): success is inferred from *promise resolution*, and Apollo `mutate()` **resolves instead of rejecting** the moment an `onError` callback is supplied. This feature already does that twice — `CurrentBoardApproved.tsx:80` and `MinisterHousingAllowance.tsx:92` — and the former carries a comment in the repo's own words explaining exactly that: `// Global Apollo error link surfaces the snackbar, this only prevents an unhandled promise rejection`. The first consumer to follow the neighbouring idiom gets a cancel dialog that closes cleanly on a **failed** mutation.

The panel produced three competing fixes and converged on one. **Standards' ruling: the one-line runtime gate**, because the TypeScript checklist explicitly says *"prefer explicit null checks"*, and a discriminated union has only **1** precedent in all of `src/` — below the convention bar, and it would be API churn on a shared component to prevent a bug no consumer has.

```tsx
// StatusCard.tsx:145 — the modal at :172 is gated on `openCancel`, which
// can only be set by this button, so gating the button is sufficient.
{isRequest && handleConfirmCancel && (
```

Plus a TSDoc line on the prop, since the type can't express the contract:

```ts
/**
 * Must REJECT on failure — the dialog only closes when this resolves.
 * Do not pass an Apollo `onError` callback in the mutation options: that
 * makes `mutate()` resolve on failure and the dialog would close as if the
 * cancel succeeded. The global error link already snackbars the message.
 */
handleConfirmCancel?: () => Promise<void>;
```

src/components/HrTools/SalaryCalculator/StepNavigation/StepNavigation.tsx
```diff
@@ -48,6 +48,7 @@ export const DiscardButton: React.FC<ButtonProps> = (props) => {
           handleClose={() => setRemoveDialogOpen(false)}
           handleConfirm={handleDiscard}
           isDiscard
+          submitting={deleting}
```
canac, line 51: **[Medium]** The new `submitting={deleting}` guard has a hole: confirm re-arms **after** the mutation resolves but **before** the route change.

This one flipped a verdict mid-review. The Financial agent originally concluded "Double-delete: NO", then reversed on the traced sequence; the Data Integrity agent then **raised** its own score 6.0 → 6.5 on evidence Financial surfaced.

The mechanism is a scope mismatch: `handleDiscard` (:24-29) is a two-phase operation — mutation, then navigation — but its busy flag covers phase one only. `deleting` is `useMutation`'s `loading`, so it flips `false` the instant `client.mutate()` resolves, strictly *before* `push()` is even called. Nothing in `handleDiscard` closes the dialog, and `removeDialogOpen` (:22) is the sole gate at :45. So React commits a render with `submitting={false}` and the dialog still mounted, at which point `SubmitModal.tsx:120` re-enables confirm, `:111` re-enables GO BACK, and `:87` restores backdrop/Escape.

**The window is a full network round trip, not a React tick.** The push target's page module exports `getServerSideProps = blockImpersonatingNonDevelopers` (`pagePropsHelpers.ts:48-51`), so Next must fetch `/_next/data/<buildId>/….json` before the transition commits. And the awaited refetch adds no masking latency, because `LandingSalaryCalculations` isn't mounted on this route.

Combined with the spinner vanishing while the dialog looks unchanged, a second click is a natural user response. On an `ActionRequired` calculation that second delete returns the affirmatively false `"Approved requests cannot be deleted or canceled"`.

Capped by the backend — `find(id)` raises `RecordNotFound` on a destroyed record, so no duplicate record is created. Note that the guard is server-side; the frontend has none.

One-line fix (the component is about to unmount, so there's nothing to leak):

```tsx
const handleDiscard = async () => {
  if (calculation) {
    await deleteSalaryCalculation(calculation.id);
    setRemoveDialogOpen(false);
    push(`/accountLists/${accountListId}/hrTools/salaryCalculator`);
  }
};
```

Worth adding a `try/catch` around the `await` here and in `PendingRequestActions.tsx:25-29` while you're at it — both currently leak unhandled rejections out of `SubmitModal.tsx:118`'s `onClick`. Financial called that single edit the highest-value code fix in the review. For contrast, `SubmitButton` is genuinely clean: `StepNavigation.tsx:149-151` returns `null` at the receipt step, so the modal unmounts in the same commit that flips `submitting`.

src/components/HrTools/SalaryCalculator/Shared/useDeleteSalaryCalculation.ts
```diff
@@ -20,8 +21,10 @@ export const useDeleteSalaryCalculation =
           },
         },
         refetchQueries: [LandingSalaryCalculationsDocument],
+        // Keep `deleting` true until the refetch finishes
+        awaitRefetchQueries: true,
```
canac, line 24: **[Medium]** The comment is accurate about Apollo and **inaccurate as documentation of either call site**.

This was the review's three-way dispute, and both extremes withdrew. Architecture retracted "load-bearing misinformation" (the Apollo mechanism *is* real — `@apollo/client` resolves to **3.7.1** via `yarn.lock`, and `markMutationResult` does `Promise.all(refetchPromises).then(...)` with `useMutation` clearing `loading` only in that `.then`). Standards retracted "the comment is simply CORRECT" — a comment can be mechanically true about Apollo and still misdescribe *this* code.

Settled position, per call site:

- **`DiscardButton` — inaccurate.** `refetchQueries` only refetches **already-registered** observables. All five `useLandingData` mounts were enumerated (`salaryCalculator/index.page.tsx:27`, `NewSalaryCalculatorLanding.tsx:40`, `PendingRequestCard.tsx:30`, `PendingSalaryCalculationLanding.tsx:20`, `SalaryInformationCard.tsx:41`), and `SalaryInformationCard` was specifically chased as the likely counterexample — it renders only from the two landing components. None is reachable from `[calculationId]/index.page.tsx` → `SalaryCalculator`. So the include array matches nothing, `Promise.all([])` resolves immediately, and "until the refetch finishes" means "immediately." Apollo also emits a dev warning (*"unknown query requested in refetchQueries options.include array"*) on **every discard**.
- **`PendingRequestActions` — literally true, pragmatically misleading.** The observable is registered, so `deleting` does stay true. But `client.ts:73-74` sets `notifyOnNetworkStatusChange: true` globally, `useLandingData.ts:262-266` ORs `calculationLoading` into its `loading`, and `PendingSalaryCalculationLanding.tsx:22-24` early-returns `<Loading loading />`. The refetch this flag awaits **unmounts the very button the extended `deleting` was meant to keep spinning** — the full-page `Loading` this PR set out to replace takes over mid-flight.

Also new from this line: `awaitRefetchQueries` folds **refetch failure into mutation failure**, so on the landing path a *successful* cancel whose refetch fails is reported to the caller as a *failed* cancel. Pre-flag, a successful delete always resolved.

The flag is worth keeping — on the landing route it guarantees the cache is fresh before `index.page.tsx:63` re-picks its branch. Just make the comment say that:

```ts
refetchQueries: [LandingSalaryCalculationsDocument],
// Ensures the landing list is fresh before it re-picks its
// pending-vs-new branch. Note this is a no-op when called from the
// calculator wizard, where LandingSalaryCalculations isn't mounted.
awaitRefetchQueries: true,
```

src/components/HrTools/SalaryCalculator/Landing/PendingSalaryCalculationLanding/components/PendingRequestActions.tsx
```diff
@@ -67,6 +67,7 @@ export const PendingRequestActions: React.FC<PendingRequestActionsProps> = ({
           handleClose={() => setRemoveDialogOpen(false)}
           handleConfirm={handleDelete}
           isCancel
+          submitting={deleting}
```
canac, line 70: **[Medium]** Dialog dismissal depends entirely on a parent re-render; `handleDelete` neither closes the dialog nor catches.

`handleDelete` (:25-29) works only because the awaited refetch nulls `latestCalculation` → `shouldShowPending` false (`useLandingData.ts:77-82`) → `index.page.tsx:63` swaps the component, unmounting the modal. Nothing in the handler itself dismisses it.

Credit where the debate landed: the *worse* version of this — "dialog stays open and re-armed on refetch failure" — was claimed, challenged by three agents, and **conceded as unreachable**. The parent's `loading` early-return (`PendingSalaryCalculationLanding.tsx:22-24`) unmounts the subtree at refetch **start**, destroying `removeDialogOpen`, so a remount comes back closed. No second-delete window exists here.

What remains is a real ambiguity: success and refetch-failure produce an **identical** dismissal. On a rejected refetch after a committed `cancel!`, the user is bounced back to the still-present pending card with a red toast implying the cancel failed — and a retry hits the false `"Approved requests cannot be deleted or canceled"`. `useLandingData.ts:77-82` restricts this component to `Pending || ActionRequired`, so the first delete *always* takes the `cancel!` branch and the record *always* ends `canceled`. That outcome is deterministic, not one possibility among several.

Plus an unhandled promise rejection escaping `SubmitModal.tsx:118`'s `onClick`.

```tsx
const handleDelete = async () => {
  if (!calculation) {
    return;
  }
  try {
    await deleteSalaryCalculation(calculation.id);
  } finally {
    // Don't rely on the refetch unmounting this component to dismiss the
    // dialog; on failure the global Apollo error link surfaces the message.
    setRemoveDialogOpen(false);
  }
};
```

src/components/HrTools/MinisterHousingAllowance/MinisterHousingAllowance.tsx
```diff
@@ -222,6 +224,13 @@ export const MinisterHousingAllowanceReport = () => {
                   color="primary"
                   sx={{ mt: 2 }}
                   onClick={onCreateMHARequest}
+                  disabled={creatingRequest}
+                  aria-busy={creatingRequest}
```
canac: **[Medium]** The new `disabled`/`aria-busy`/`startIcon` covers only this button, but removing the full-page `Loading` (the deleted `if (creatingRequest)` early return above) left the rest of the page interactive during a create whose `onCompleted` navigates **unconditionally**.

`onCreateMHARequest` (:74-102) clears `creatingRequest` only in `onError`; on success `router.push(requestLink)` fires regardless of what the user has done in the meantime. The old early return made a competing interaction structurally impossible.

Two agents disagreed on the reachable control set and the UX agent settled it with a full predicate evaluation over `MhaStatusEnum`:

- `showNewRequestButton` is true only for **`HrApproved`** and **`Cancelled`** — and `MinisterHousingAllowance.test.tsx:311` already enumerates exactly those two.
- `isCurrentRequestPending` (:108-113) short-circuits to `true` for every non-`BoardApproved` status, so `<CurrentRequest>` co-renders with this button.
- For **`HrApproved`**, Cancel is rendered **disabled** — `approved = [HrApproved, BoardApproved].includes(status)` → `disableCancel` (`CurrentRequest.tsx:100`) → `StatusCard.tsx:162` — and Edit is hidden (`canEdit` is false). Live set: the **View Request** `NextLink` and the **Print** `IconButton`, whose `handlePrint` does `await router.push(viewLink)` then `window.print()`.
- For **`Cancelled`**, `approved` is false, so **Cancel is enabled** — a full cancel mutation racing the create.

Either way: clicking Print or View mid-create races `onCompleted`'s push. The user lands on the view page while a new `InProgress` MHA request has been created behind their back — and since `hasBlockingRequest` (:116) then becomes true, they're **blocked from starting another** until they find and discard the one they never saw.

`StatusCard` already has the props for this — `hideActions` (:36/:123) and `hidePrint` (:34/:109). Thread them from `creatingRequest` through both `CurrentRequest` and `CurrentBoardApproved`; no new StatusCard surface needed.

src/components/HrTools/Shared/CalculationReports/DirectionButtons/DirectionButtons.tsx
```diff
@@ -161,7 +161,7 @@ export const DirectionButtons: React.FC<DirectionButtonsProps> = ({
                 <Button
                   variant="contained"
                   color="primary"
-                  endIcon={loadingNext ? undefined : <ChevronRight />}
+                  endIcon={<ChevronRight />}
```
canac, line 164: **[Medium]** Three separate defects on this button, all fixed by one three-line edit. Flagged by 5 agents.

**(a) Reverts a deliberate design.** `git log -L 158,168` confirms commit `d5c2533c0` ([MPDX-9561], #1784) introduced `endIcon={loadingNext ? undefined : <ChevronRight />}` **in the same hunk** that added the `startIcon` spinner — the spinner was meant to *replace* the chevron, not join it. With MUI v7 defaults (verified: `src/theme.ts` has no `MuiButton` or `palette.action` override) each icon slot is ~24px, so the loading button now grows **+24px** where it previously held width.

This started the review at 7.5 and was revised to 6.0 after five agents challenged the calibration — the UX agent re-verified its own git evidence (which held) and then found the sibling-displacement impact **unreachable at both live consumers**: on the ASR path `loadingNext` can only be true when `isFirstFormPage`, and both Back (`showBackButton={!isFirstFormPage}`) and Discard (`isFormPage`) are unmounted in exactly that window; on PdsGoalCalculator no `handleDiscard` is passed, so Back sits in a separate left-pinned flex group. So the residual is the button's own left edge moving while it's `disabled` — cosmetic, no mis-click hazard, latent for a future consumer that passes both props.

**(b) The only busy button in this PR without `aria-busy`.** The other three all set it (`MinisterHousingAllowance.tsx:228`, `NewSalaryCalculatorLanding.tsx:144`, `SubmitModal.tsx:121`) and their tests assert it. That's precisely why `RequestPage.test.tsx:689-694` is the one new test that *can't* — the asymmetry is visible in the test suite. `aria-busy` has 7 instances repo-wide, well past the convention bar.

**(c) `CircularProgress` missing `color="inherit"`** (pre-existing, from the same commit, but the same three lines). House pattern is 7 instances *with* it vs 2 without on a `contained` button. On `variant="contained"` + `disabled`, MUI paints the label and `endIcon` in `action.disabled` grey while a default-`primary` spinner stays palette-blue — so this diff's now-permanent chevron sits **grey beside a blue spinner** in the same button.

```tsx
<Button
  variant="contained"
  color="primary"
  aria-busy={loadingNext}
  endIcon={loadingNext ? undefined : <ChevronRight />}
  startIcon={
    loadingNext ? <CircularProgress size={20} color="inherit" /> : undefined
  }
```

Then `RequestPage.test.tsx:689-694` can assert `aria-busy` like its four siblings, and `DirectionButtons.test.tsx:181-195` should gain a chevron assertion — the icon behaviour change is currently untested.

src/components/HrTools/Shared/CalculationReports/SubmitModal/SubmitModal.tsx
```diff
@@ -122,11 +122,11 @@ export const SubmitModal: React.FC<SubmitModalProps> = ({
             startIcon={
               submitting ? <CircularProgress size={20} color="inherit" /> : null
             }
+            endIcon={<ChevronRight />}
```
canac, line 125: **[Medium]** Same unconditional-`endIcon` shift — and it made a pre-existing shift 3× worse and flipped its sign.

Arithmetic, re-derived against verified MUI defaults: **before**, idle was `<ChevronRight sx={{ml:1}}/>` as a direct child = 24px `SvgIcon` + 8px `ml` = 32px, and submitting was 24px spinner + label → delta **−8px** (shrank). **After**, `endIcon` = 8px `marginLeft` + 20px glyph + −4px `marginRight` = 24px idle, and submitting = 24 + label + 24 → delta **+24px**. `DialogActions` is `justifyContent: flex-end`, so GO BACK slides ~24px left. This is the confirm step for five flows: cancel MHA, cancel ASR, delete a salary calculation, submit a salary calculation, discard.

Revised down from 7.0 during debate on two mitigators: the dialog is `maxWidth="sm" fullWidth` so it doesn't resize, and GO BACK is `disabled={submitting}` (:111) at the exact instant it moves — the displacement lands on an inert control, so there's no mis-click hazard at all.

**Keep the restructure — it's a genuine improvement.** Moving the chevron out of the `<b>` label and into the real `endIcon` slot gains the correct 8px `marginLeft`, `marginRight: -4`, and a `fontSize: 20` override, replacing a mis-sized 24px glyph glued into the label text. Only the `submitting` gate needs to come back:

```tsx
endIcon={submitting ? undefined : <ChevronRight />}
```

One line, and the idle-state improvement survives intact. Also worth noting: `SubmitModal.test.tsx:291-336` never asserts the chevron, so this behaviour change is untested on both sides.

canac (commented):

**6 specialized agents + dependency analysis + cross-examination debate + rebuttal round.** No blockers: every agent independently concluded nothing reaches severity 8.0, and the Financial agent traced the strongest composite end-to-end to confirm it.

| | |
|---|---|
| Risk | **10/10 CRITICAL** — see calibration note below |
| Critical / High blockers | **0 / 0** |
| Important (7.0-7.9) | **2** |
| Medium (5.0-6.9) | **12** |
| Suggestions (<5.0) | 8 (informational) |
| Pre-existing | 6 (informational) |
| Unresolved | 1 (see below) |

> ⚠️ **2 Important findings (7.0-7.9)** — strongly recommended, don't block merge, and can't be `/dismiss`ed. Address in this PR or file follow-ups.

### Risk calibration note

The 10/10 is a **scoring-model artifact**, not a judgment on the diff: 10 Medium-tier files × +1 each saturates the cap, on 75 non-test lines of loading-state wiring. Consequence: this PR is ineligible for auto-approve (requires LOW/MEDIUM), so the auto-approve workflow was **not** triggered.

### The debate changed the outcome

Pre-debate the top two findings were layout-shift issues at 7.5 and 7.0. Five agents challenged the calibration. The UX agent re-verified its own git evidence (which held — `d5c2533c0` did introduce the chevron suppression in the same hunk as the spinner), then found its sibling-displacement impact claim **unreachable at both live consumers**, and revised itself down. Net movement:

- Four agents revised their own findings **down** under challenge.
- Two agents **reversed** their own conclusions on new evidence: Financial on "Double-delete: NO", Standards on the `awaitRefetchQueries` comment being "simply correct".
- Data Integrity **raised** one of its own findings (6.0 → 6.5) on evidence a challenger surfaced.
- Three findings were **dropped or withdrawn**: a `DocumentNode`-vs-string nit (noise); a test-flake concern (raised by two agents, withdrawn by both — settled empirically by `GoalsList.test.tsx:90-98`, a pre-existing green test doing the identical thing); and a "landing arrives warm" claim (withdrawn by its author).

### What capped the severity

Financial and Data Integrity were the only agents to read the backend (`mpdx_api`), and that is what kept every double-click finding non-catastrophic: `user.salary_requests.find(id)` raises `RecordNotFound` on a destroyed record, and AASM's `event :cancel` has no `from: :canceled` transition, so `SaaService.cancel_salary_request` is exactly-once. **The guard is server-side; the frontend has no equivalent.** Worth recording — had the server been permissive, the `StepNavigation.tsx:51` re-arm window alone would be an 8.5 blocker.

That same guard produces the sharpest user-facing risk here: a retry after an already-successful cancel returns **"Approved requests cannot be deleted or canceled"**, which the global error link snackbars verbatim. A staff member who just cancelled a pending salary request reads that it was *approved*.

### Recognized as done well

- Four full-page-`<Loading/>`-for-a-mutation anti-patterns replaced with in-place feedback — the right architectural direction.
- Four `SubmitModal` confirm buttons that were **always enabled during their mutation** are now disabled. The PR net-**narrows** double-submit exposure.
- `RequestPage.tsx:119` correctly narrowed to `loading && !requestData`, keeping the genuine initial-fetch gate.
- The `SubmitModal` chevron restructure (inline child → real `endIcon` slot) fixes a mis-sized 24px glyph. Keep it — just re-add the `submitting` gate.
- `CurrentBoardApproved.tsx`'s removed `handleConfirmCancel={() => {}}` was verified **dead** by three agents (`isRequest={false}` gates the whole cancel affordance). No coverage lost.
- The two `waitFor` wrappers added in `StatusCard.test.tsx` are exactly the right and complete set of adaptations to the now-async `handleCancel`.

---

### Findings on files or lines not in this PR's diff

Informational — these cannot be posted as line comments, do not count toward the verdict, and cannot be `/dismiss`ed.

**`useDeleteSalaryCalculation.test.tsx` — 6.0/10 — hook contract untested** (Testing, Data Integrity). The hook gained a public return value (`deleting`) and a new option; its 26-line single-test file was untouched. The comment `// Keep \`deleting\` true until the refetch finishes` is currently an unverified claim. The Testing agent wrote the settling test: assert the operation **order** is `[Landing, Delete, Landing, resolved]` — it fails loudly if the flag is removed or stops matching. A control variant (drop the landing query) proves the DiscardButton no-op.

**`DirectionButtons.tsx:196` — 5.5/10 — the shared discard modal is the one gap this PR missed** (Architecture, Data Integrity, UX, Dependency all converged here independently). `handleDiscardConfirm:96-101` fires `handleDiscard()` **without awaiting** and closes the dialog immediately, so on the ASR path a delete + `awaitRefetchQueries` refetch runs with *zero* user signal and, on failure, no dialog left to retry from. It is also the one call site where this PR's unconditional `endIcon` lands without a compensating loading signal — net-negative drift rather than merely incomplete. Reaches ASR, all three MHA steps, and PdsGoalCalculator. Standards: *"if one follow-up is picked from this review, that is it."*

**Fix-one-fix-all — 12 remaining mutation buttons** in `src/components/HrTools/**` with no loading state: `PartnerRemindersReport.tsx:237` (no `disabled` at all), `HouseholdExpensesHeader.tsx:181/193/198`, `HoursPerWeekGrid.tsx:440/462`, `GoalCalculatorGrid.tsx:533/593`, `RentOwn.tsx:190/200`, `CurrentBoardApproved.tsx:102`, `DirectionButtons.tsx:196`. Conformant precedents to copy: `GoalsList.tsx:66-68`, `CreateGoalDialog.tsx:163-166`, `GoalSettingsForm.tsx:189-196`.

**`[Pre-existing]` Missing `cache.evict` after delete** — reclassified from 6.5 to informational. Verified at the base commit: `refetchQueries` is line 22 of the pre-PR file and appears as a **context line** in this diff, and `DeleteSalaryCalculation.graphql`'s `{ id }`-only payload selection is untouched. Eviction behavior is byte-identical pre- and post-PR, so this is out of scope here. Still worth a follow-up: the normalized `SalaryRequest:<id>` survives the delete, `SalaryCalculatorContext.tsx:80` reads it back via `salaryRequest(id:)`, and `apollo3-cache-persist` persists it to `localStorage` in production. Against 30+ evict-after-delete precedents this hook is the outlier. **Correction to the obvious fix:** keep `refetchQueries` **and** add the evict — dropping the refetch would leave the root field pointing at an evicted entity, a worse cache state.

**`[Pre-existing]` `MinisterHousingAllowance.tsx:85,92-102`** — manual `enqueueSnackbar` in `onError` **plus** the global error link → double error toast; and a success toast against the "prefer no success snackbar" rule. Untouched by this diff, but removing the full-page unmount makes this error path newly visible.

**`[Pre-existing]` `StatusCard.test.tsx:182`** — `await userEvent.click(...)`, against "never await userEvent". Not a diff line; cheap while you're in the file.

**`[Pre-existing]` `ASR CurrentRequest.tsx:74`** — the Cancel affordance is wired to `handleDeleteRequest(id, false)`, so cancelling toasts "discarded successfully" and resets to a New request. Reads as a mislabeled financial action.

**`[Pre-existing]` Backend, 5.5/10 — `mpdx_api/app/graphql/mutations/salary_request/update_mutation.rb:12-14`** has no status guard, so `update!` will write to a **cancelled** request. Independent of this PR; deserves its own ticket.

---

### Convention drift (advisory — not findings, not scored, not dismissible)

The Standards agent re-verified three candidates against the this-diff-caused test and **dropped one**.

- ~~Mutation-in-flight button contract~~ — **DROPPED.** Verified against `HEAD~1` and `d5c2533c0`: both the contract and the doc gap pre-date this PR.
- **`StatusCard.tsx:74-79`** moves `setOpenCancel(false)` inside the `try`, so a confirmation dialog now stays **open on failure** — new in this diff. `.claude/rules/code-review.md` → UX Focus Areas → Dialog UX documents only "close on success" and is silent on the failure branch. Suggest a bullet, and requiring an in-dialog error surface (`SubmitModal.tsx:93` already renders an `<Alert>`).
- **Spinner assertions** — this diff takes `.MuiCircularProgress-root` from **2 → 9** instances, against **56** `*ByRole('progressbar')` call sites across 28 files. Testing Focus Areas prefers role queries in spirit but is silent on scoped-spinner assertions. Suggest sanctioning `within(button).findByRole('progressbar')` explicitly.

---

### Unresolved — needs a 5-minute experiment, not more argument

**Does Apollo 3.7 tear down the `LandingSalaryCalculations` `ObservableQuery` on soft navigation into `/salaryCalculator/[calculationId]`?** Raised by Architecture, escalated by Data Integrity, unresolvable in-repo (`node_modules` is outside the review's search boundary).

The key reframing: **the two branches are mutually exclusive.** If it's torn down, the stale-landing impact occurs but the discard-route refetch rejection cannot. If it survives, the refetch works so the stale-landing impact vanishes — but it has no observers, so nothing renders the result, and the rejection path opens. **Never score them additively.** Certain either way: for hard-navigation entries (bookmark, reload, direct URL) the observable was never created, so the no-op holds.

Suggestive but not probative: `useUpdateTasksQueries.ts:44-49` guards `if (!observableQuery.hasObservers())` against the same registry handing back observer-less observables.

**To settle it:** in `StepNavigation.test.tsx`, mount the landing, soft-navigate via `TestRouter`, discard, then assert the **call count** of `LandingSalaryCalculations` on `mutationSpy`. Two calls → the observable survived. One → teardown happened.

src/components/HrTools/Shared/CalculationReports/StatusCard/StatusCard.test.tsx
```diff
@@ -201,7 +201,36 @@ describe('CardSkeleton', () => {
...
+
+    userEvent.click(await findByRole('button', { name: 'Cancel Request' }));
+
+    const confirmButton = await findByRole('button', { name: /yes, cancel/i });
+    userEvent.click(confirmButton);
+
+    await waitFor(() => expect(confirmButton).toBeDisabled());
+    expect(confirmButton).toHaveAttribute('aria-busy', 'true');
+    expect(
+      confirmButton.querySelector('.MuiCircularProgress-root'),
```
canac: **[Medium]** `.querySelector('.MuiCircularProgress-root')` couples 7 new assertions to an MUI-internal class name.

The Standards agent re-grepped both sides so the numbers are trustworthy: this class appears **9 times** repo-wide and **7 of those are added by this PR**. Only 2 pre-date it (`PdsGoalCalculator.test.tsx:369`, `DirectionButtons.test.tsx:193`) — **below the 3-instance convention bar**, which means this PR is what turns it into the house idiom, not prior art.

The competing form is far more established than either agent first estimated: **56** `*ByRole('progressbar')` call sites across 28 test files. Including one in this PR's own dependency tree — `SubmitModal.test.tsx:300` already does exactly `within(submitButton).getByRole('progressbar')`.

The class selector breaks on any MUI class-prefix change, and `querySelector` returning `null` reads as an ordinary assertion failure rather than a query error. The role query also asserts something real about what a screen reader perceives.

`within` is confirmed available and idiomatic — imported from `@testing-library/react` in 71 test files. Apply to all 7 sites (`StatusCard.test.tsx:228`, `RequestPage.test.tsx:692`, `MinisterHousingAllowance.test.tsx:294`, `NewSalaryCalculatorLanding.test.tsx:170`, `PendingRequestActions.test.tsx:101`, `StepNavigation.test.tsx:107` and `:150`), adding `within` to the existing import:

```tsx
expect(within(confirmButton).getByRole('progressbar')).toBeInTheDocument();
```

src/components/HrTools/AdditionalSalaryRequest/RequestPage/RequestPage.test.tsx
```diff
@@ -675,17 +675,23 @@ describe('RequestPage', () => {
...
 
-    userEvent.click(getByRole('button', { name: /continue/i }));
+    const continueButton = getByRole('button', { name: /continue/i });
+    userEvent.click(continueButton);
 
-    expect(await findByTestId('Loading')).toBeInTheDocument();
+    await waitFor(() => {
+      expect(continueButton).toBeDisabled();
+      expect(
+        continueButton.querySelector('.MuiCircularProgress-root'),
```
canac: **[Medium]** Multi-assertion `waitFor` where the checklist requires `findBy*` for the presence half.

The rules file is explicit: *"Tests use `findBy*` for async assertions rather than `waitFor(() => getBy*)`."* The `toBeDisabled()` half legitimately has no `findBy*` equivalent — but the spinner-presence half does, so that assertion is exactly the pattern the checklist prohibits. Same shape at `MinisterHousingAllowance.test.tsx:290-296` and `NewSalaryCalculatorLanding.test.tsx:166-170`.

Beyond the letter of the rule: three assertions inside one `waitFor` means a genuinely-missing spinner produces a 1s timeout with the error attributed to whichever assertion happened to fail last. The repo's reference tests narrow the `waitFor` to the single racing condition and assert the rest synchronously (`CsvHeaders.test.tsx:418-420`), and `StatusCard.test.tsx:225-230` in this very PR already follows that shape.

```tsx
expect(await within(continueButton).findByRole('progressbar')).toBeInTheDocument();
expect(continueButton).toBeDisabled();
```

There's also a forward-looking reason to narrow it: the current form is safe only because nothing yields between the un-awaited `userEvent.click` and the `waitFor`. `creatingRequest` is a real Apollo `loading` value that *does* reset, so inserting an `await findBy*` in between would later make this order-dependent.

src/components/HrTools/SalaryCalculator/Landing/NewSalaryCalculationLanding/NewSalaryCalculatorLanding.tsx
```diff
@@ -134,6 +141,12 @@ export const NewSalaryCalculatorLanding: React.FC = () => {
                 color="primary"
                 onClick={handleStartCalculation}
                 disabled={creatingCalculation}
+                aria-busy={creatingCalculation}
+                startIcon={
+                  creatingCalculation ? (
+                    <CircularProgress size={20} color="inherit" />
```
canac, line 146: **[Medium]** During load, the spinner and the greyed fill are the **only** state indicators — shape and colour alone.

The rules file says *"Color should never be the only indicator of state (add icons or text)."* The visible label stays `Calculate New Salary` throughout, and `disabled` (rather than `aria-disabled`) removes the button from the tab order without announcing why. Same at `MinisterHousingAllowance.tsx:235`.

`DirectionButtons.tsx:171-173` already solves this correctly with `loadingNextTitle ?? buttonTitle ?? t('Continue')` — and `PdsGoalCalculator.tsx:123` uses it (`t('Saving...')`). Since `aria-busy` support for buttons is patchy across NVDA and JAWS, the text swap is the more reliable signal of the two, and this PR adds `aria-busy` without it.

```tsx
{creatingCalculation ? t('Starting Calculation…') : t('Calculate New Salary')}
```

New string, wrapped in `t()`, no interpolation, no dynamic key — `yarn extract` will pick it up.

src/components/HrTools/AdditionalSalaryRequest/RequestPage/RequestPage.tsx
```diff
@@ -175,6 +175,7 @@ const MainContent: React.FC = () => {
                   (exceedsCap && !!errors.additionalInfo)
                 }
                 overrideNext={isFirstFormPage ? handleContinue : undefined}
+                loadingNext={creatingRequest}
```
canac, line 178: **[Suggestion]** `loadingNext` passed without its paired `loadingNextTitle`.

`DirectionButtons.tsx:171-173` resolves the label as `loadingNextTitle ?? buttonTitle ?? t('Continue')`, so with neither passed the label is unchanged during load. The only other consumer pairs them — `PdsGoalCalculator.tsx:122-123` passes `loadingNextTitle={t('Saving...')}`.

Combined with the now-unconditional `endIcon`, the ASR loading button reads spinner + "Continue" + chevron: two icons, no textual state change. Screen-reader users get nothing (this is also the one busy button in the PR without `aria-busy` — see the `DirectionButtons.tsx:164` comment).

```tsx
loadingNext={creatingRequest}
loadingNextTitle={t('Creating request…')}
```

src/components/HrTools/SalaryCalculator/Landing/PendingSalaryCalculationLanding/components/PendingRequestActions.test.tsx
```diff
@@ -83,4 +83,27 @@ describe('PendingRequestActions', () => {
...
+    const deleteButton = await findByRole('button', { name: 'Delete request' });
+    userEvent.click(deleteButton);
+
+    const confirmButton = await findByRole('button', { name: /yes, cancel/i });
+    userEvent.click(confirmButton);
+
+    expect(confirmButton).toBeDisabled();
+    expect(confirmButton).toHaveAttribute('aria-busy', 'true');
+    expect(
+      confirmButton.querySelector('.MuiCircularProgress-root'),
```
canac: **[Suggestion]** Consistency only — **not a flake.** Recording the resolution so it isn't re-raised in a future review.

Two agents flagged these fully-synchronous assertions after a non-awaited `userEvent.click` as a possible race, and **both withdrew** after the Testing agent's analysis. It was then settled *empirically* rather than by argument: `GoalsList.test.tsx:90-98` is a pre-existing, green test doing the identical thing — `userEvent.click(button)` followed immediately by `expect(button).toBeDisabled()` on a button whose `disabled` comes from an Apollo `useMutation` `loading` inside an async handler.

Both edges are deterministic. **True edge:** user-event 13.5.0's `click` is synchronous, RTL 13.4's `eventWrapper` wraps dispatch in `act()`, Apollo's `execute` calls `setResult({loading: true})` synchronously before `client.mutate()`, and React 18 flushes on `act()` exit. **False edge:** no `await` and therefore no microtask checkpoint exists between the click and the assertion — and a hypothetical synchronous mock response would make these fail 100% of the time, not intermittently. So they either always pass or always fail.

What's left is style: `StatusCard.test.tsx:225` wraps the same assertion in `waitFor` while this file and `StepNavigation.test.tsx` don't. The repo's established shape (3 instances: `CsvHeaders.test.tsx:418`, `CsvValues.test.tsx:282`, `CsvPreview.test.tsx:313`) is `await waitFor(() => expect(btn).toBeDisabled())` then sync assertions — cheap insurance if Apollo's internals ever defer `loading`.

```tsx
await waitFor(() => expect(confirmButton).toBeDisabled());
expect(confirmButton).toHaveAttribute('aria-busy', 'true');
```

src/components/HrTools/Shared/CalculationReports/StatusCard/StatusCard.test.tsx
```diff
@@ -201,7 +201,36 @@ describe('CardSkeleton', () => {
...
-    expect(queryByRole('dialog')).not.toBeInTheDocument();
+    await waitFor(() => expect(queryByRole('dialog')).not.toBeInTheDocument());
+  });
+
+  it('disables the modal buttons and shows a spinner while cancelling', async () => {
+    let resolveCancel: () => void = () => {};
+    handleConfirmCancel.mockImplementationOnce(
+      () =>
+        new Promise<void>((resolve) => {
+          resolveCancel = resolve;
```
canac, line 212: **[Suggestion]** The deferred-promise fixture is sound; one latent ordering hazard.

Checked and confirmed fine: `resolveCancel` **is** guaranteed assigned before use — a `Promise` executor runs synchronously during construction, so the assignment completes before `mockImplementationOnce`'s factory returns (the `= () => {}` initializer is dead code satisfying TS definite-assignment). The dangling promise is safe too: Jest exits regardless, and if `resolveCancel()` is never reached because an assertion above threw, the test already fails for the right reason. The pattern matches `CsvHeaders.test.tsx`'s deferred-save fixture.

The one real hazard: `handleConfirmCancel` is module-scoped (:22) and `jest.config.js` sets `clearMocks: true`, which is `mockClear()` — it resets *calls* but does **not** drain a queued `mockImplementationOnce`. If an assertion between the mock setup and the click ever throws, the unconsumed once-implementation leaks into the next test in the file. Latent today (the next test never invokes the mock), but it couples test ordering.

Add `handleConfirmCancel.mockReset()` to a `beforeEach`, or scope the deferred mock with `mockImplementation` plus explicit teardown.

src/components/HrTools/Shared/CalculationReports/StatusCard/StatusCard.tsx
canac: /dismiss This component could be refactored a lot, but that's not in scope yet

.claude/rules/code-review.md
```diff
@@ -208,11 +208,9 @@ Every item here is mandatory. The Standards agent must report compliance per ite
 **Code Quality**
 
 - [ ] Passes `yarn lint` and `yarn lint:ts`
-- [ ] No debug output (`console.log`, `console.debug`, `debugger`, `// TODO` without a Jira/MPDX ticket reference)
+- [ ] No `// TODO` without a Jira/MPDX ticket reference
```
canac, line 224: **[Critical]** This edit removes two checks that CI does **not** enforce — flagged by **all 3 agents** (Architecture 9.5, Standards 8.5, Testing 7.0).

Commit `bfb04b9d9` is titled *"Remove code review rules covered by CI."* That's true for one of the three deleted items and false for two:

| Deleted checklist item | Enforced by lint? | Evidence |
|---|---|---|
| No unused imports or variables | **YES** | `.eslintrc.js:71-74` — `@typescript-eslint/no-unused-vars: 'error'` |
| `console.log` / `console.debug` | **YES** | `.eslintrc.js:69` — `no-console: 'error'` |
| `debugger` | **NO** | no `no-debugger` rule anywhere |
| Empty `catch {}` | **NO** | no `no-empty` rule anywhere |

`.eslintrc.js` never extends `eslint:recommended`, and `plugin:@typescript-eslint/recommended`'s `eslint-recommended` compat layer only turns rules *off* — it never turns them on. `@typescript-eslint/no-empty-function` (line 102) is a different rule, and it's `off` for tests. Two agents reproduced this in isolated sandboxes with the repo's exact pinned versions (`eslint@8.57.0`, `@typescript-eslint@7.x`): a file containing `catch {}` and `debugger;` produces **zero** output.

So after this change, `debugger` and empty `catch {}` pass both `yarn lint:ci` and the manual checklist undetected.

**Sequencing:** commit `9bdea3c5b` (13:41) added an empty `catch {}` in `StatusCard.tsx`; `bfb04b9d9` (15:04, same author, same branch) deleted the rule that catches it. That's the one defect on this branch the removed rule would have caught.

**Fix** — keep the unused-vars removal, restore the other two (or add the real ESLint rules if CI enforcement is the intent):

```diff
  - [ ] Passes `yarn lint` and `yarn lint:ts`
  - [ ] No `// TODO` without a Jira/MPDX ticket reference
+ - [ ] No `debugger` statements (no `no-debugger` rule in .eslintrc.js)
+ - [ ] No empty `catch {}` blocks that swallow errors silently (no `no-empty` rule in .eslintrc.js)
  - [ ] No `new Date()` — use Luxon (`DateTime.now()`, `DateTime.local()`) per project convention
  - [ ] No commented-out code blocks (delete, don't comment)
```

src/components/HrTools/Shared/CalculationReports/StatusCard/StatusCard.tsx
canac, line 76: **[High]** Bare empty `catch {}` — flagged by **all 3 agents** (Standards 8.5, Testing 7.0, Architecture 4.0).

Verified by grep: this is the **only** empty catch block in `src/` + `pages/`. There are 18 `catch {` blocks total; the other 17 all have a body — either remedial code (`enqueueSnackbar`, `setValue(defaultValue)`, `setEntries(prev...)`) or an explanatory comment (`// Ignore malformed JWTs`).

The convention already exists in a file **this same PR touches**. `CurrentBoardApproved.tsx:80` does the identical swallow and documents why:

```ts
onError: () => {}, // Global Apollo error link surfaces the snackbar, this only prevents an unhandled promise rejection
```

The path is reachable, not theoretical: `handleConfirmCancel` is wired to real Apollo mutations with no `errorPolicy` override (`MinisterHousingAllowance/SharedComponents/CurrentRequest.tsx:67-80`), so `mutate()` rejects on any GraphQL or network error and this catch fires for real.

**Fix:**

```diff
     } catch {
+      // Global Apollo error link surfaces the snackbar (src/lib/apollo/client.ts);
+      // swallow here only to keep the modal open and re-enabled so the user can retry.
     } finally {
```

src/components/HrTools/Shared/CalculationReports/StatusCard/StatusCard.test.tsx
```diff
@@ -201,7 +201,34 @@ describe('CardSkeleton', () => {
 
     expect(handleConfirmCancel).toHaveBeenCalled();
 
-    expect(queryByRole('dialog')).not.toBeInTheDocument();
+    await waitFor(() => expect(queryByRole('dialog')).not.toBeInTheDocument());
+  });
+
+  it('disables the modal buttons and shows a spinner while cancelling', async () => {
```
canac, line 207: **[High]** The rejection path — the entire reason the new `try`/`catch`/`finally` exists — is untested. (Consensus tier: **Important**, 2 of 3 agents, not counted as a blocker.)

Flagged by Testing (8.5) and Standards (WARN on the "error-path testing, not just happy paths" checklist item).

This new test drives only the resolve path via `resolveCancel()`. Nothing verifies that when `handleConfirmCancel` **rejects**, the button re-enables, the spinner clears, and the dialog stays open for a retry. If the `finally` block ever regresses, the Cancel button sticks disabled permanently after a failed cancel and no test catches it.

```tsx
it('clears the spinner and keeps the modal open when cancelling fails', async () => {
  handleConfirmCancel.mockImplementationOnce(() =>
    Promise.reject(new Error('network error')),
  );

  const { findByRole } = render(<TestComponent isRequest={true} />);
  userEvent.click(await findByRole('button', { name: 'Cancel Request' }));

  const confirmButton = await findByRole('button', { name: /yes, cancel/i });
  userEvent.click(confirmButton);

  await waitFor(() => expect(confirmButton).toBeEnabled());
  expect(await findByRole('dialog')).toBeInTheDocument();
});
```

Separately: the deferred-promise fixture in this test is sound — `resolveCancel` is guaranteed assigned before use because the `Promise` executor runs synchronously.

src/components/HrTools/AdditionalSalaryRequest/RequestPage/RequestPage.tsx
canac, line 178: **[Important]** Mutation-rejection paths are untested at all four new spinner sites.

None of these handlers catches a rejection, and no new test exercises one:

- `RequestPage.tsx:71-94` — `handleContinue`
- `PendingRequestActions.tsx:25-29` — `handleDelete`
- `StepNavigation.tsx:24-29` — `handleDiscard`
- `StepNavigation.tsx:101-112` — `handleSubmit`

Concretely here: `handleContinue` does `await trackMutation(createRequest(...))`, and `useTrackMutation` (`src/hooks/useTrackMutation.ts:6-14`) only attaches `.finally()` — it re-throws. If `createRequest` rejects, `handleNextStep()` never runs and the rejection escapes the `onClick` handler unhandled.

The existing test *"does calls handleNextStep when createNewRequest fails"* (`RequestPage.test.tsx:594`) mocks a **null GraphQL payload** (`createAdditionalSalaryRequest: null`), which *resolves* successfully. It does not exercise an actual rejection, so this path is genuinely untested end to end.

Apollo does reset `loading`, so the button recovers — but nothing asserts that, and the unhandled rejection is invisible to CI.

```tsx
it('re-enables the continue button when createNewRequest rejects', async () => {
  // ...mocks with CreateAdditionalSalaryRequest throwing
  const continueButton = getByRole('button', { name: /continue/i });
  userEvent.click(continueButton);
  await waitFor(() => expect(continueButton).toBeEnabled());
});
```

src/components/HrTools/SalaryCalculator/Shared/useDeleteSalaryCalculation.ts
```diff
@@ -20,8 +21,9 @@ export const useDeleteSalaryCalculation =
           },
         },
         refetchQueries: [LandingSalaryCalculationsDocument],
+        awaitRefetchQueries: true,
```
canac, line 24: **[Medium]** Hook's public API changed but its colocated test wasn't touched — flagged by Testing (6.5) and Architecture (3.5).

This hook gained `deleting: boolean` and `awaitRefetchQueries: true`, and it backs the spinner in **both** consumers (`PendingRequestActions`, `StepNavigation`'s `DiscardButton`). `useDeleteSalaryCalculation.test.tsx` has no corresponding change.

Commit `4cd033e27` deleted the comment `// Keep \`deleting\` true until the refetch finishes` alongside the `awaitRefetchQueries: true` it explained, without pinning that behavior in a test. The new UI tests only assert disabled/`aria-busy` immediately after the click — none isolates the refetch window, so nothing distinguishes "stays true through the mutation" from "stays true through the refetch."

Use the deferred-promise technique already in `StatusCard.test.tsx` to isolate it:

```tsx
it('reflects the mutation loading state through the refetch', async () => {
  const { result } = renderHook(() => useDeleteSalaryCalculation(), { wrapper });
  expect(result.current.deleting).toBe(false);

  let deletePromise: Promise<void>;
  act(() => {
    deletePromise = result.current.deleteSalaryCalculation('calculation-id');
  });
  await waitFor(() => expect(result.current.deleting).toBe(true));

  await deletePromise!;
  await waitFor(() => expect(result.current.deleting).toBe(false));
});
```

src/components/HrTools/Shared/CalculationReports/StatusCard/StatusCard.tsx
```diff
@@ -66,10 +66,17 @@ export const StatusCard: React.FC<StatusCardProps> = ({
...
+  const [cancelling, setCancelling] = useState(false);
 
-  const handleCancel = () => {
-    setOpenCancel(false);
-    handleConfirmCancel();
+  const handleCancel = async () => {
+    setCancelling(true);
+    try {
+      await handleConfirmCancel?.();
+      setOpenCancel(false);
```
canac, line 75: **[Medium]** New state-update-after-unmount hazard — flagged by Architecture.

`handleCancel` used to be synchronous: `setOpenCancel(false)` fired *immediately* on click, then `handleConfirmCancel()` ran fire-and-forget. Now both setters fire **after** the await.

Both `isRequest={true}` parents render `StatusCard` conditionally on the very record the mutation deletes — `{request && <CurrentRequest .../>}` (`AdditionalSalaryRequest.tsx:78`) and `showCurrentRequest && currentRequest` (`MinisterHousingAllowance.tsx:209-211`) — and both underlying mutations already use `awaitRefetchQueries: true` with an `onCompleted` that drives the parent re-render. So the subtree can unmount at essentially the same moment the awaited promise resolves, leaving `setOpenCancel(false)` / `setCancelling(false)` to fire on an unmounted component.

Likely harmless in practice (the whole subtree disappears either way), but note the two sibling confirm-then-mutate flows deliberately avoid this: `PendingRequestActions.handleDelete` and `StepNavigation.DiscardButton.handleDiscard` never call a local close-setter after the await — they let the cache-driven re-render or `router.push` dismiss the modal. `StatusCard` is now the only place combining "await first" with "manage your own dialog state after," so this isn't an established convention to defer to.

Options: guard the post-await setters with a mounted ref, or drop the success-path `setOpenCancel(false)` and rely on the parent unmount (matching the siblings), keeping local state only for the failure path where the component is guaranteed still mounted.

src/components/HrTools/Shared/CalculationReports/SubmitModal/SubmitModal.tsx
canac, line 125: **[Medium]** Behavior changed with no test change in either direction — flagged by Testing.

`ChevronRight` moved from a conditionally-rendered child (`{!submitting && <ChevronRight sx={{ ml: 1 }} />}`) to an always-rendered `endIcon`. The chevron used to **disappear** while submitting; now it renders alongside the spinner. `SubmitModal.test.tsx`'s existing `describe('submitting state')` block (lines 291-336) checks spinner / `aria-busy` / `disabled` but says nothing about the chevron.

This is a Shared component with **5 dependents**, and the change is visible at every call site that passes `submitting`: `DirectionButtons:183`, `StatusCard:173`, `PendingRequestActions:65`, `StepNavigation:46` and `:130`.

Low functional risk, but it's a silent, unverified UI change with wide blast radius — worth one assertion in that describe block. See also the still-open prior-review comments on the dual chevron+spinner look; no UX agent ran in this quick-mode pass.

canac (commented):
## Multi-Agent Code Review — PR #1965

**Agents:** 3 (Testing, Standards, Architecture) + dependency impact analysis
**Mode:** quick — no debate round, no coverage-gap pass, no UX/Security/Data-Integrity agent

---

### Verdict: BLOCKERS FOUND

1. **`.claude/rules/code-review.md:211`** (sev 9.5, all 3 agents) — the rules-file edit removes two checks that CI does **not** actually enforce, on the same branch that introduces a violation of one of them.
2. **`StatusCard.tsx:76`** (sev 8.5, all 3 agents) — bare empty `catch {}`, the only one in `src/` + `pages/`.

Both fixes are small: two restored checklist lines, and a three-line comment.

### Risk Assessment

**10/10 — CRITICAL** · Reviewer level: **SENIOR** · 16 files, +163 −34

Score is dominated by file-count tiering (9 Medium-tier `src/components/**` files). The substantive driver is the Critical-tier `.claude/rules/**` edit, not the spinner work.

### Dependency Impact

| File | Dependents | Level |
|---|---|---|
| `DirectionButtons.tsx` | 5 | HIGH |
| `SubmitModal.tsx` | 5 | HIGH |
| `StatusCard.tsx` | 3 | MEDIUM |
| `useDeleteSalaryCalculation.ts` | 2 | MEDIUM |

**Contract changes — all verified safe:**

- `handleConfirmCancel` required → optional: all 3 call sites checked. Both `isRequest={true}` sites pass a `Promise<void>`-returning handler; the site that dropped it (`CurrentBoardApproved.tsx:93`) has `isRequest={false}`, so no cancel UI renders. TypeScript would reject a sync `() => void`.
- `deleting` added to the hook's return: both consumers destructure it.
- Unconditional `ChevronRight` now renders *alongside* the spinner at 7 call sites that pass `submitting`/`loadingNext` — previously hidden while loading. Affected: `DirectionButtons:183`, `StatusCard:173`, `PendingRequestActions:65`, `StepNavigation:46` & `:130`, `PdsGoalCalculator:107`, `RequestPage:134`.

### What's good

The spinner work itself is well executed — a consistent `disabled` + `aria-busy` + `CircularProgress` contract across all six new call sites, none inventing a divergent pattern. Replacing `savingButton.querySelector('.MuiCircularProgress-root')` with `within(button).getByRole('progressbar')` across five test files is a real improvement, and `awaitRefetchQueries: true` correctly keeps the spinner up until the list actually refreshes.

---

### Out of Scope (informational — not counted toward the verdict)

**`DirectionButtons.tsx:198-206`** — the discard-confirmation `SubmitModal` receives no `submitting` prop, unlike its submit-modal sibling 15 lines above, and unlike `StepNavigation`'s `DiscardButton` which this PR *does* fix. ASR/MHA confirm-discard therefore still shows no spinner while a real delete mutation is in flight, and `handleDiscardConfirm` doesn't await `handleDiscard()`. No line in this diff touches it, so it is out of scope per the review scope rule — but it is a direct gap in this PR's stated goal and a good follow-up candidate.

### Coverage Caveat

Quick mode ran no UX, Security, or Data Integrity agent. All three agents that did run independently raised the same UX question none of them was scoped to answer: `SubmitModal.tsx:125` and `DirectionButtons.tsx:164` now render the trailing `ChevronRight` **and** the leading `CircularProgress` simultaneously while loading. The prior review scored this 6.0 / 5.5; those comments are still open and were **not** re-adjudicated in this pass. Re-run in standard mode for a UX verdict.

### Dismissed (acknowledged by developer)

**`StatusCard.tsx:39`** (sev 5.5) — `handleConfirmCancel` required → optional removes the compile-time guarantee that an `isRequest={true}` consumer supplies a handler. Dismissed by @canac: *"This component could be refactored a lot, but that's not in scope yet"*. Not counted toward the verdict. The dependency analysis independently confirms all three current call sites are correct.

---

**Summary:** 1 critical + 1 high-priority blocker · 2 important · 4 medium · 1 dismissed · 1 out-of-scope

src/components/HrTools/SalaryCalculator/Landing/PendingSalaryCalculationLanding/components/PendingRequestActions.test.tsx
```diff
@@ -83,4 +83,25 @@ describe('PendingRequestActions', () => {
...
+      <TestComponent calculation={mockCalculation} onCall={mutationSpy} />,
+    );
+
+    const deleteButton = await findByRole('button', { name: 'Delete request' });
+    userEvent.click(deleteButton);
+
+    const confirmButton = await findByRole('button', { name: /yes, cancel/i });
+    userEvent.click(confirmButton);
+
+    expect(confirmButton).toBeDisabled();
```
canac, line 98: **[Medium]** Synchronous assertions immediately after an un-awaited `userEvent.click`, with no `waitFor` — flagged by Testing.

Also at `StepNavigation.test.tsx:104-107` and `:145-149`.

Not a demonstrated flake — these pass repeatedly locally. But grep found this is the **only** place in the repo with the unwrapped form, and this same PR uses the `waitFor`-wrapped version of the identical assertion three times elsewhere (`NewSalaryCalculatorLanding.test.tsx:166-170`, `RequestPage.test.tsx:689-695`, `StatusCard.test.tsx:225`). The inconsistency is within a single PR.

```diff
     userEvent.click(confirmButton);

-    expect(confirmButton).toBeDisabled();
-    expect(confirmButton).toHaveAttribute('aria-busy', 'true');
-    expect(within(confirmButton).getByRole('progressbar')).toBeInTheDocument();
-    expect(getByRole('button', { name: 'GO BACK' })).toBeDisabled();
+    await waitFor(() => {
+      expect(confirmButton).toBeDisabled();
+      expect(confirmButton).toHaveAttribute('aria-busy', 'true');
+      expect(within(confirmButton).getByRole('progressbar')).toBeInTheDocument();
+      expect(getByRole('button', { name: 'GO BACK' })).toBeDisabled();
+    });
```

Note this supersedes the prior review's 3.5-severity note on the same lines, which recorded them as "consistency only".

.claude/rules/code-review.md
wjames111, line 224: Question: Was this intentionally included in this PR? Totally fine if so, just wanted to double check.

.eslintrc.js
```diff
@@ -67,6 +67,8 @@ module.exports = {
     curly: 'error',
     eqeqeq: 'error',
     'no-console': 'error',
+    'no-debugger': 'error',
+    'no-empty': 'error',
```
wjames111, line 71: Question: Also curious if this slipped in or was intenional.

wjames111 (approved):
Looks just fine to me!

.claude/rules/code-review.md
canac: Yeah, I removed some things that ESLint can determine from static analysis.

.eslintrc.js
canac: Here are the ESLint rules that replace the Claude Code review rules. ESLint is more reliable than Claude.

